PR #6, opened 2011-06-27
Added scribble file extension to Racket.

elibarzilay:
This is our documentation language, http://docs.racket-lang.org/scribble/,
and the files are still racket files even though the syntax looks completely
different.


linkicloud (approved):
Thx

PR #249, opened 2012-09-05
Ignore Qt project files.

ghost:
Qt's project files (generated with qmake) shouldn't be included.


thorade:
Now also discussed in #2186

PR #464, opened 2013-04-08
Add OpenGL shading language (GLSL)

julik:
This is the language used for writing OpenGL shaders, which are becoming much more mainstream lately.


larsbrinkhoff:
@julik Why did you put GLSL in the C group?

julik:
Because the syntax looks very C-like to me and I could not find a neat guide on how to pick a different (probably better) group

pchaigno:
Yep, it's not really clear what it is... (see #1548)
I'll open a pull request to add a comment on that.

larsbrinkhoff:
Thanks!

My view is that GLSL should not be in any group.  I have opened #2196 to see what the *Staff* thinks.

PR #995, opened 2014-03-17
Fix Fortran Capitalization

acidburn0zzz:
http://en.wikipedia.org/wiki/Fortran

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -414,7 +414,7 @@ F#:
   - .fsi
   - .fsx
 
-FORTRAN:
+Fortran:
   type: programming
   lexer: Fortran
   color: "#4d41b1"


anonymouse64:
@arfon Is there any update on the process of renaming languages?

arfon:
> @arfon Is there any update on the process of renaming languages?

I'm afraid I'm not at GitHub any longer so I don't know. / cc @TwP

lildude:
@anonymouse64 Renaming languages is now possible, and has happened a few times since this PR was closed. @arfon even had fun doing this in a big rename bonanza in https://github.com/github/linguist/pull/3278 😁

The process? Submit a PR like this one with the new name and provide an explanation as to why we need the rename.

In this case, it appears to be just a case change so the impact should be negligible as language searching is case-insensitive.

lildude:
So nothing to do. 🎉

Issue #1025, opened 2014-03-27
Some Objective-C files are being misidentified as M files.


ScrimpyCat:
I'm not sure when the change has occurred as I've only just noticed it now, though I know it wasn't originally like this. But certain Objective-C files in one of my projects are being detected as M files.

The files that are being detected are unit tests (using XCTest) that don't have an accompanying header (e.g. SomeTests.m), those that do have a header (e.g. SomeTests.m + SomeTests.h) don't seem to be affected. I mention that they're tests, because other projects of mine that only have a single Obj-C file, don't seem to be affected.

see: https://github.com/ScrimpyCat/Common/search?l=m&p=1

Update: I just now took a look at the project to see if it's something I could just offer a fix for, and see it's using heuristics (didn't know how it was originally going about its detection, so disregard some of the above comments).

If I add one of those tests from my project as a sample or the header it includes as a sample, it works fine. But I don't think that's a good solution. I tried creating a general samples to try address the problem.

```
#import <XCTest/XCTest.h>

@interface ExampleTests : XCTestCase
@end

@implementation ExampleTests

-(void) testSomething
{
    const char a = 'c';
    const short b = 0;
    const int c = 0;
    const long d = 0L;
    const long long e = 0LL;
    const float f = 0.0f;
    const double g = 0.0;
    const long double h = 0.0L;

    XCTFail(@"");
    XCTAssertNil(nil, @"");
    XCTAssertNotNil(nil, @"");
    XCTAssert(1, @"");
    XCTAssertTrue(1, @"");
    XCTAssertFalse(1, @"");
    XCTAssertEqualObjects(nil, nil, @"");
    XCTAssertNotEqualObjects(nil, nil, @"");
    XCTAssertEqual(1, 1, @"");
    XCTAssertNotEqual(1, 1, @"");
    XCTAssertEqualWithAccuracy(1.0f, 1.0f, 0.0f, @"");
    XCTAssertNotEqualWithAccuracy(1.0f, 1.0f, 0.0f, @"");
    XCTAssertThrows(nil, @"");
    XCTAssertThrowsSpecific(nil, NSException, @"");
    XCTAssertThrowsSpecificNamed(nil, NSException, NSGenericException, @"");
    XCTAssertNoThrow(nil, @"");
    XCTAssertNoThrowSpecific(nil, NSException, @"");
    XCTAssertNoThrowSpecificNamed(nil, NSException, NSGenericException, @"");
}

@end
```

[...]

PR #1027, opened 2014-03-29
Changed tokenizer number literals to be more encompassing

ScrimpyCat:
This is a solution to https://github.com/github/linguist/issues/1025

The new number literal matching regex will skip hexadecimal and C style literals. It is not perfect, but to keep it simple I didn't worry about it.

### DIFF
--- a/lib/linguist/tokenizer.rb
+++ b/lib/linguist/tokenizer.rb
@@ -94,7 +94,7 @@ module Linguist
           end
 
         # Skip number literals
-        elsif s.scan(/(0x)?\d(\d|\.)*/)
+        elsif s.scan(/(0x\h(\h|\.)*|\d(\d|\.)*)([uU][lL]{0,2}|([eE][-+]\d*)?[fFlL]*)/)
 
         # SGML style brackets
         elsif token = s.scan(/<[^\s<>][^<>]*>/)
--- a/test/test_tokenizer.rb
+++ b/test/test_tokenizer.rb
@@ -25,6 +25,10 @@ class TestTokenizer < Minitest::Test
     assert_equal %w(add \( \)), tokenize('add(123, 456)')
     assert_equal %w(|), tokenize('0x01 | 0x10')
     assert_equal %w(*), tokenize('500.42 * 1.0')
+    assert_equal %w(), tokenize('1.23e-04')
+    assert_equal %w(), tokenize('1.0f')
+    assert_equal %w(), tokenize('1234ULL')
+    assert_equal %w(G1 X55 Y5 F2000), tokenize('G1 X55 Y5 F2000')
   end
 
   def test_skip_comments


arfon:
@bkeepers - given we've done significant work to address https://github.com/github/linguist/issues/1025 elsewhere. Do you think we still need this PR?

bkeepers:
If I'm understanding the samples.json output, it looks like this might actually be a good improvement for several other languages as well. It might be worth still incorporating it.

ScrimpyCat:
If the samples.json is an improvement for other languages, that'll be because since this pr was made other samples have changed or possibly been removed. 

The main stuff was the number literal check change. So it looked for more number literal formats languages often support. The other thing was a new sample in Obj-C to get some weights on XCTAssert library so my issue was fixed, however I don't think this is really needed (better being solved another way). 

If you're interested in the number literal formatting change, I'll clean up the pr (get rid of the other stuff, and add some tests). Otherwise I don't think its a good idea to merge this as it is.

bkeepers:
The samples.json gets regenerated before we run the tests and when we do a new release now. 

Tests for the number literal formatting change would be great.

ScrimpyCat:
Was just adding some tests for it and I noticed with the current master some tests fail. These are tests depending on the samples however. Specifically: 

```
[ 13/112] TestBlob#test_language = 0.98 s                                                                
  1) Failure:
TestBlob#test_language [linguist/test/test_blob.rb:460]:
G-code/lm.g.
<"G-code"> expected but was
<"GAP">.

[ 28/112] TestClassifier#test_classify_ambiguous_languages = 0.48 s       
  2) Failure:
TestClassifier#test_classify_ambiguous_languages [linguist/test/test_classifier.rb:62]:
linguist/samples/G-code/lm.g
[["GAP", -4.917893242397641], ["G-code", -5.477509030333064]].
<"G-code"> expected but was
<"GAP">.
```

Looking at the `samples/G-code/lm.g` this file seems pretty problematic, as it's only numbers. So while the current skip number literal check would leave behind some characters ('e' and '-', e.g. from 5.51e-05) mine does not.

pchaigno:
@fingolfin added support for G-Code in #1530. He might be able to help us on that.
I though G-Code lines of code were always starting by a letter followed by a number (see [square.g](https://github.com/github/linguist/blob/f38e15790e80a72c9a3f35457a2bd65fd891b8b1/samples/G-code/square.g) for instance) :/

ScrimpyCat:
If it is valid G-Code, a heuristic will need to be added. Possibly something along the lines of checking if only number literals and whitespace are contained within the file (could probably simplified to just checking if an occurrence of something other than whitespace, number, '.', 'e', '+', '-', is contained within the file) then specify that it is G-Code.

fingolfin:
I am pretty sure it is valid G-code; it comes from here:
  https://github.com/BLLIP/bllip-parser/blob/master/first-stage/DATA/EN/lm.g

bkeepers:
How should we proceed with this?

arfon:

pchaigno:
#2484 should fix the build for this.

arfon:
@ScrimpyCat - would you mind brining this branch up to date with master?

arfon:
Superseded by #2524

PR #1143, opened 2014-04-28
NewLisp language added with some heuristics

pchaigno:
Updated version of #949 (and neater).

Pull request taken up from #919.
Some example NewLISP repositories on GitHub:
https://github.com/kanendosei/artful-newlisp
https://github.com/taoeffect/dragonfly-newlisp
https://github.com/newlisponrockets/newLISP-on-Rockets


arfon:
Sorry to go back and forth on this @pchaigno - do you think we should keep `defactor`or not here? It seems like we get better results with?

Regardless we should probably try and get this PR updated and merged in. Agree?

pchaigno:
As @larsbrinkhoff noted, `defactor` is not a frequent Common Lisp keyword. I only found it in one of the 4 projects I tested. I tried to make the perfect heuristic and it was in overfitting :/ That's the reason why I removed the `defactor` keyword. Does it make sense? Should we investigate more on this keyword?

arfon:
OK cool. Let's not overfit this - I think this PR is in pretty good shape as-is. Thanks! :zap:

larsbrinkhoff:
`defactor` is not a keyword, it's just a macro in a library:
https://github.com/naveensundarg/Common-Lisp-Actors/blob/53e6259cea16e32eec6bc0c717d425f41f377397/actors.lisp#L90

arfon:
:+1: thanks @pchaigno and @larsbrinkhoff

Issue #960, opened 2014-02-20
.sls files are also SaltStack states (YAML), not Scheme


stevage:
Plenty of sample material here if needed:

https://github.com/saltstack/salt-states

pchaigno:
@stevage @jandd This issue should be fixed.

@stevage The statistics of your repository are currently not up-to-date. They will update if you make a new commit. Alternatively you can contact support@github.com to ask them to update it.

stevage:
Thanks. (Whoa, that was a blast from the past...)

pchaigno:
@stevage Yep, sorry it took so long...

DavidJFelix:
This is still an issue.
Examples:

-  https://github.com/saltstack/salt-winrepo
-  https://github.com/saltstack/salt-jenkins/search?l=scheme

pchaigno:
@DavidJFelix These repositories are indeed not up-to-date. The new results are much better:
```
~/salt-jenkins$ linguist --breakdown
77.91%  ApacheConf
20.35%  Python
1.51%   HTML
0.23%   Shell

Python:
_grains/external_ip.py
_grains/python_executable.py
_modules/runtests.py
_states/runtests.py

ApacheConf:
apache/httpd.conf

Shell:
halite/master/files/install_node_npm.sh

HTML:
halite/settings.jinja
```

```
~/salt-winrepo$ linguist --breakdown
100.00% Scheme

Scheme:
gimp.sls
gpg4win-light.sls
npp.sls
openoffice.sls
truecrypt.sls
```

salt-winrepo still has a few mis-detected files. If the project was under an OSS license I could add these files to our samples but even so, results will never be perfect :/

DavidJFelix:
@pchaigno I think a big factor in the visibility of this bug is that salt is not represented in the percentage "color graph" but scheme is. Even if 90% of the the files are identified as salt and 5% are scheme, at least half of the bar will show up as scheme because the salt files are suppressed from the graph.

puppet has it's own color (slightly different from ruby's red). I think it might be appropriate to affiliate salt with python or jinja templates (which in many of the cases above are actually detected as django templates) and display it in the bar.

pchaigno:
Salt state files are not counted in the statistics because Linguist considers them as `data`. Only `programming` and `markup` languages are part of the language bar.

DavidJFelix:
[...]

PR #1304, opened 2014-06-22
Add support for .sls YAML files (SaltStack states)

pchaigno:
This PR adds supports for SaltStack states files (with a `.sls` extension) as asked in #960.

I tested on [saltstack/salt-states](https://github.com/saltstack/salt-states) and [coffeestats/coffeestats-django](https://github.com/coffeestats/coffeestats-django). The detection isn't perfect but it's better than before (those files were detected as Scheme).
@arfon I guess we'll have to wait for the test system to merge this :-)

EDIT: Also fixes the issue with this repo: [uggedal/states](https://github.com/uggedal/states).

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2602,12 +2602,13 @@ Sage:
 SaltStack:
   type: data
   group: YAML
-  tm_scope: source.yaml.salt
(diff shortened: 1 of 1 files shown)


aroben:

PR #1343, opened 2014-07-03
Add support for Drupal's inc, module and install extensions

jacobbednarz:
Drupal uses the .inc, .module and .install file extensions and it is currently
not recongised by GitHub as being a PHP based file.


arfon:
Sorry for the incredibly slow response rate here - over the past few weeks I've been systematically trying to get through old pull requests.

To answer your question @rivimey - the issue with `.inc` is that there are many different languages that use the extension. Tackling this is a fairly major task and so IMO its better for them to remain unclassified than have some of them correct and the majority correct. 

Digging into this pull request a little more, it turns out that [`.module`](https://github.com/search?utf8=%E2%9C%93&q=NOT+php+extension%3Amodule&type=Code&ref=searchresults) and [`.install`](https://github.com/search?utf8=%E2%9C%93&q=NOT+php+extension%3Ainstall&type=Code&ref=searchresults) are both also used outside of Drupal which means we have the same issue again.

The current state of the Linguist classifier is that it _has_ to return a Language result if the extension is 'known' and to be honest this [is rather dumb](https://github.com/github/linguist/issues/1571). 

We have a Google Summer of Code project this year to completely re-write the Linguist classifier to address problems just like this one. It definitely should be possible to make contributions like this to Linguist (that is, adding support for a single language use of an extension) but for now it's not and so I'm afraid this pull request can't be merged.

arfon:
I also meant to add that if this is causing you issues on your current repositories then we have a number of self-serve override solutions that you can use to force Linguist to classify these files as `PHP`: https://github.com/github/linguist#overrides

rivimey:
Hi @arfon, thanks for getting back to us re this issue.

I have looked at the linked issue and the linguist overrides, and the .gitattributes solution is a possible one, though sadly manual (I presume). Maybe the github website could offer an interface to set it?

In posting my comment I was thinking that while I fully understand that file extensions are widely duplicated, the solution was to implement language-specific tests, in the manner of Unix file (magic) database. For example, .info, .module and .install in a drupal context will, I think in all cases, be in the same directory, and the .info file must be present. Moreover, .module and .install will almost certainly have a <?php in the first few lines. I think that would make a fairly robust check on its own for the 'Drupal' meaning of those extensions.

.inc is trickier, in that it is as you suggest widely used. If I was trying to detect it I would start by saying there must be a <?php tag in the first few lines. That might not be sufficient, but it would be a very good start to say this is a 'php' file. To differentiate further may be helpful, but I'd need use-cases I think.

As for performance, the tests could be cached, surely, so you only need to check file types when a file is added (or changed?), not for every access.

pchaigno:
@rivimey Working on it at #2180 ;)

PR #1421, opened 2014-07-28
Adding Font Awesome to vendored files.

stevex86-old:
Added Font Awesome css files to vendor.yml and appropriate test.


james69miller:
G gg

Issue #1388, opened 2014-07-16
Does linguist still support extensions with multiple file segments?


cldwalker:
I'm specifically asking because `.cljs.hl` syntax highlighting for Hoplon [no longer works](https://github.com/tailrecursion/hoplon-demos/blob/master/todoFRP/src/index.cljs.hl). I saw this was merged in #854 but a quick look at languages.yml doesn't reveal other extensions with multiple file segments e.g. `.seg1.seg2`. Do I need to move Hoplon, which is a clojurescript framework, to its own entry? If not, pointers on how to fix this? Thanks!

PR #1438, opened 2014-08-04
Remove file extensions with multiple segments

pchaigno:
Since Linguist doesn't support file extensions with multiple segments (see #1388), this PR removes all of them.
They are replaced with the last segments. For example, `.cmake.in` becomes `.in`.


pchaigno:
@larsbrinkhoff There is already a pull request (#1944) to replace `.in` with `.cmake.in`.

PR #1511, opened 2014-09-06
Better detection for AGS Script, AsciiDoc, and Public Key .asc files

edmundito:
Went into the Explore languages feature within github to see the "AGS Script" projects and noticed a couple of things:

 1. .asc files were getting marked as AGS Script but formatted as AsciiDoc
 2. .asc files that were neither format were also getting marked as AsciiDoc.

So, I added a heuristics checker for .asc files. ~~Additionally, because AGS Script is a nameless language  designed specifically for the Adventure Game Studio tool (and mentioned as `'C'-style script` within the AGS source code, but no one calls it that), I renamed the script to Adventure Game Studio to avoid ambiguity (similar to how the ColdFusion's language is called ColdFusion), and brings more clarity to Adventure Game Studio game projects within github~~.

After making these changes, I cloned a selection of the projects that were incorrectly marked as AGS Script and ran a `exec rake linguist <clone_path> --breakdown` on them to ensure that things were working properly.


arfon:
Hi @edmundito - thanks for submitting this. I'm sorry it's taken so very long to get to review this. If you're still interested I'd be happy to work with you to get these changes merged. A couple of high-level comments:

1. A lot has changed in the [heuristics code](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb) since this pull request was submitted. Do you think you could bring this branch up to date with `master` using the new format heuristics?
2. Unfortunately we can't change language names currently on GitHub. This is a long-standing issue with our search infrastructure and so would you mind leaving the language name as it currently is?

edmundito:
Ok, no problem. I will into this. I can leave the language name as it was and work on ensuring that the syntax is not confused with AsciiDoc and vice versa.

arfon:
:+1: thanks @edmundito!

edmundito:
@arfon I was able to successfully merge the latest changes and revert the name change (with rake test passing). I updated the heuristics, but when I try to run linguist against progit2 as mentioned in #1638, it's not returning "AsciiDoc" as it should (it doesn't even classify the files). I added a print the ascii match statement at the top of the disambiguate call and it was indeed matching some of the .asc files. I will continue to look around, perhaps there is something else that it's making it exclude the files.

edmundito:
Oh, I got it! I studied a the linguist source further and because AsciiDoc is of type *prose* it will not be part of the breakdown:

https://github.com/github/linguist/blob/e87b89ab5b0e3dbbcf21bea8142e6cfe807baa95/lib/linguist/repository.rb#L163

Code should be ready for review. :+1:

lib/linguist/heuristics.rb
```diff
@@ -150,8 +150,17 @@ def call(data)
...
+    disambiguate "AsciiDoc", "AGS Script", "Public Key" do |data|
+      if /^[=-]+(\s|\n)|{{[A-Za-z]/.match(data)
+        Language["AsciiDoc"]
+      elsif /^(\/\/.+|((import|export)\s+)?(function|int|float|char)\s+((room|repeatedly|on|game)_)?([A-Za-z]+[A-Za-z_0-9]+)\s*[;\(])/.match(data)
+        Language["AGS Script"]
+      elsif /^-----BEGIN/.match(data)
+        Language["Public Key"]
+      else
+        # Plain text can also be AsciiDoc.
+        Language["AsciiDoc"]
```
arfon: @bkeepers - what do you think about having this as the fallback. Given the relative use of `"AsciiDoc", "AGS Script", "Public Key"` it seems like a reasonable assumption. That said, perhaps we should just return nothing and let the classifier do its thing.

arfon: @aroben - any thoughts on this comment?

aroben: Leaving it up to the classifier seems safer to me.

arfon: :+1: - great. I'll get this merged in :soon:

arfon:
:+1: thanks @edmundito. This change should be live on GitHub in the next couple of weeks.

PR #1527, opened 2014-09-15
Extend linguist command to work with multiple arguments

fingolfin:
This change modifies the command line interface of linguist to accept more than one argument. It simply loops over the arguments and treats them individually. In combination with e.g. `find` and `xargs`, this makes it very convenient to e.g. check all files matching a certain pattern inside a given repository. And much more efficient: Invoking `bundle exec linguist FILENAME` on my machine takes 2 (!) seconds (despite my SSD). Invoking it on a dozen files takes about the same time...

This is a very simple hack, and I won't be surprised if you'll want refinements before accepting it (if at all), but it was enough to scratch an itch of mine (namely testing only the .tst files in a big repo, see PR #1523), and I thought perhaps others might benefit from it.

I am not quite happy with it myself; e.g. I think --breakdown should only be accepted once, and as first parameter (but that change would have been incompatible with current behaviour). On the other hand, it should accept any previously valid invocation. 

[Review on Reviewable](https://reviewable.io/reviews/github/linguist/1527)

### DIFF
--- a/bin/linguist
+++ b/bin/linguist
@@ -8,17 +8,14 @@ require 'linguist/language'
 require 'linguist/repository'
 require 'rugged'
 
-path = ARGV[0] || Dir.pwd
-
-# special case if not given a directory but still given the --breakdown option
-if path == "--breakdown"
-  path = Dir.pwd
+if ARGV.include? "--breakdown"
   breakdown = true
+  ARGV.delete("--breakdown")
 end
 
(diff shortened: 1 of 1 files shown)


arfon:
:flags: flagging this as stale.

arfon:
Please feel free to open another PR if this is still something you're interested in working on.

PR #1541, opened 2014-09-17
Improve GAP support: set right lexer, add .tst extension (Alternate version of PR #1523)

fingolfin:
This is an alternative to my PR #1523 which avoids extending the heuristics API with the filename extension.

This solution is strictly inferior. but if the linguist team is opposed to giving the heuristics code more information, then this is better than nothing.


arfon:
Closing in favour of https://github.com/github/linguist/pull/2030

PR #1549, opened 2014-09-22
Treat Go dependency tree as generated and vendored code

kr:
The directory tree rooted at `Godeps` is analogous to `Gemfile.lock` and `node_modules`, so this patch treats it accordingly.

These dependencies are saved by https://github.com/tools/godep, for example in https://github.com/heroku/hk/tree/master/Godeps.

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -63,6 +63,7 @@ module Linguist
         generated_jni_header? ||
         composer_lock? ||
         node_modules? ||
+        godeps? ||
         vcr_cassette? ||
         generated_by_zephir?
     end
@@ -229,6 +230,14 @@ module Linguist
     # Returns true or false.
     def node_modules?
       !!name.match(/node_modules\//)
+    end
+
+    # Internal: Is the blob part of Godeps/,
+    # which are not meant for humans in pull requests.
+    #
+    # Returns true or false.
+    def godeps?
+      !!name.match(/Godeps\//)
     end
 
     # Internal: Is the blob a generated php composer lock file?
--- a/test/test_blob.rb
+++ b/test/test_blob.rb
@@ -262,6 +262,10 @@ class TestBlob < Test::Unit::TestCase
 
 
     assert Linguist::Generated.generated?("node_modules/grunt/lib/grunt.js", nil)
+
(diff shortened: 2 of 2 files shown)


dpiddy:
Any luck?

arfon:
> Any luck?

Sorry, for the slow reply. Yes, some progress. I think I understand the bug now - tl;dr the full pathname (e.g. `Godeps/_workspace/src/github.com/bgentry/heroku-go/account.go`) of the files in the diff view on GitHub isn't being passed properly which basically means the [regex](https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L239) doesn't ever match. I'm afraid I don't have a fix yet but I'm pretty sure we now understand the issue.

arfon:
Working on a fix here: https://github.com/github/linguist/pull/1976

kr:
Interesting, good to hear a fix is underway! :)

Out of curiosity, how did it ever work for node? The Go regex is identical in structure to the [node regex](https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L231).

arfon:
> Out of curiosity, how did it ever work for node? The Go regex is identical in structure to the node regex.

I have to assume that it didn't :-\

arfon:
OK this should be fixed in https://github.com/github/linguist/pull/1976

It will be another week or so before we cut a new Linguist gem (and therefore see the issue fixed on GitHub)

kr:
Any update on this? Looks like #1976 was merged, reverted in #2014, and then un-reverted in #2017, which is in progress. Is that right? Is #2017 the place to watch for status?

arfon:
@kr - yeah #2017 is the place to watch. This ended up causing a bunch of problems in production when deployed so I had to rollback. I still want to get this fixed up, it is just going to require some more work to get this behaving properly on GitHub.

titanous:
Any update on this?

arfon:
> Any update on this?

@titanous @kr - this should how be fixed. Sorry it took so long.

titanous:
Thanks!

dpiddy:
:clap:

Thanks to everyone involved!

kr:
Fantastic. Thank you so much!

PR #1601, opened 2014-10-16
Add dependency on Rugged to gemspec

calebhearth:
This is option 2. Opening a parallel pull request to not
`require 'linguist/repository'`.

This approach means that all 3rd party apps using Linguist but not needing to
work with repositories (including [html-pipeline](https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/syntax_highlight_filter.rb#L2))
also need to add a dependency on Rugged, which is unrelated to their goal.

The reason for this change is that a 3rd-party with "require 'linguist'" gets a
load error for Rugged being missing since it is required in linguist/repository.


BM5k:
The rugged dependency makes this difficult / impossible to deploy on heroku.

PR #1604, opened 2014-10-17
Comment styles; don't choke on `#!/usr/bin/env foo=bar`

geoff-codes:
- Add `--` and `"` style single-line comments.
  - Rearrange `elsif`s so `"` comments don't interfere with removing double-quoted strings.
- Don't fail to detect valid `#!/usr/bin/env` shebangs which actually "`env`" environmental variables.

**Edit**:

---
- Add tests to test_tokenizer.rb to match tokenizer.rb.
- Add samples for `"` and `--` comments and complicated `env` shebangs.
- Also, add `.tool` as shell extension (basically, exactly as #1622).


arfon:
@bkeepers - this looks :sparkles: to me. Whaddya think?

bkeepers:
:+1:

arfon:
Thanks @geoff-codes. Sorry it took so very long to get this merged.

This change should be live on GitHub in the next week or so.

geoff-codes:
No problem :wink:. I know you guys are busy.

PR #1627, opened 2014-10-23
Disambiguate C, C++, Objective-C

arfon:
Updated version of https://github.com/github/linguist/pull/1036


lib/linguist/heuristics.rb
```diff
@@ -61,6 +61,15 @@ def call(data)
       @heuristic.call(data)
     end
 
+    disambiguate "Objective-C", "C++", "C" do |data|
+      if (/@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/.match(data))
```
jtbandes: Oops — I think this should've been `synchronized`!

pchaigno: @jtbandes Could you open a pull request to fix it?

jtbandes: #2592

PR #1652, opened 2014-10-31
Disambiguate Matlab, Objective-C, Mathematica, M, and Mercury

arfon:
This is looking pretty good now.


pchaigno:
@hrothgar @sebgod Is there anything that could help distinguish Mercury and Matlab?

sebgod:
@pchaigno , for distinguishing triples (quadruples) of languages, what is the preferred/recommended way?
Like this:
```ruby
disambiguate "Mercury", "M", "Matlab" do |data| 
    if data.include?(":- module") 
        Language["Mercury"] 
    else if data.include?(" ??? ")
        Language["Matlab"]
    else if data.include?(" ??? ")
        Language["M"]
    end
end 
````
or each pair of languages should be distinguished separately?

**Comment by pchaigno** [comment]
@sebgod As you did should work fine (see [`disambiguate "Objective-C", "C++", "C"`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L72)). Do you think `:- module` could be used to distinguish Matlab and Mercury?

**Comment by sebgod** [comment]
@pchaigno , yes, I think only Mercury has the ":- module" directive, the other languages don't. Also the module statement should be always present.

**Comment by pchaigno** [comment]
@arfon From the ideas raised in this thread I made [a method to disambiguate the 5 languages](https://github.com/pchaigno/linguist/blob/m-heuristics/lib/linguist/heuristics.rb#L167-L180). If you can grab me a corpus of files from [this search](https://github.com/search?o=desc&q=extension%3Am+if&s=indexed&type=Code&utf8=%E2%9C%93), I could see how well it works and we could incorporate it in this pull request.

**Comment by whitten** [comment]
Why do you think that search ie: extension:m if
would work for M files? I would actually be surprised if you got very many
M files at all, since M allows IF to be abbreviated as just the single
letter I and M programmers tend to prefer terse code, so rarely type the
two letter form. Also, M programmers tend to use uppercase as that the
portability standards used to not support lowercase, so even if they did
type "IF" they wouldn't type "if" so I don't think you are going to find
much code in M with that search.

Dave Whitten
713-870-3834

**Comment by pchaigno** [comment]
@whitten Thanks for your help! What keyword other than `IF` could I use to match M files (single letters don't seem to work :/)?

**Comment by whitten** [comment]
M code is incredibly terse.  You probably could find IF commands looking
for a space followed by I followed by space " I ", Intrinsic Function calls
would be a dollar sign followed by a single letter followed by an open
parenthesis. like "$P(" or "$E("

Spaces ARE significant in M.

Dave Whitten
713-870-3834

**Comment by pchaigno** [comment]
@whitten Apparently we can't perform searches with special characters or spaces :disappointed:

**Comment by whitten** [comment]
If all you want is M code, these repositories have over 50000 ".m" files in
them:

https://github.com/whitten/VistA-FOIA

https://github.com/whitten/FOIA-RPMS

**Comment by pchaigno** [comment]
@whitten @arfon Ok then, let's go with [the initial search page I proposed](https://github.com/search?o=desc&q=extension%3Am+if&s=indexed&type=Code&utf8=%E2%9C%93) and I'll test against the two repositories @whitten proposed too.

@whitten Thanks again for pointing out the issue with my search results :-)

**Comment by halirutan** [comment]
I was directed from my [closed issue 2022](https://github.com/github/linguist/issues/2022) to this issue. The proposed heuristic to look for Mathematica's comment opener `(*` sounds good. If there is another heuristics needed, then one should take into account that Mathematica (and I think between these 4 languages Mathematica only) uses `[..]` to denote function calls. So even in the most basic files you find several pairs of them.

Question: Can I check somewhere when this issue is merged into master and included in the version that is running on GitHub? Currently, I have a header full of Mathematica comments and the file is still interpreted as MATLAB.

**Comment by arfon** [comment]
Ok sorry for the delay folks. I've brought this PR up to date with master and have done some work on the heuristics based upon suggestions earlier in this (and other) threads. For convenience here's the method as implemented in this PR:

```ruby
disambiguate "M", "Mathematica", "Matlab", "Mercury", "Objective-C" do |data|
  if /^[ \t]*@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/.match(data)
    Language["Objective-C"]
  elsif data.include?(":- module")
    Language["Mercury"]
  elsif /^\s*;/.match(data)
    Language["M"]
  elsif /^\s*\(\*/.match(data)
    Language["Mathematica"]
  elsif /^\s*%/.match(data)
    Language["Matlab"]
  end
end
```

@pchaigno - I've run the [search](https://github.com/search?o=desc&q=extension%3Am+if&s=indexed&type=Code&utf8=%E2%9C%93) you suggested and the results are:

**Current master (cc56ddb3543679cfec9aa2b50ceed6905e477093)**
```
49.65%  Matlab
28.90%  Objective-C
21.45%  M
```

**This PR as it stands (4d2b6ee99e3f3e0ecef623c92342d04c05704035)**

arfon:
Also, @halirutan I note that https://github.com/halirutan/HeyexImport is identified as 100% Mathematica with these changes - even after removing the overrides :smile:

lib/linguist/heuristics.rb
```diff
@@ -160,6 +160,20 @@ def call(data)
       end
     end
 
+    disambiguate "M", "Mathematica", "Matlab", "Mercury", "Objective-C" do |data|
+      if /^[ \t]*@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/.match(data)
```
bkeepers: I think `^[ \t]*` can be simplified to `^\s*`.

http://rubular.com/r/s2usmgMO3A

lib/linguist/heuristics.rb
```diff
@@ -160,6 +160,20 @@ def call(data)
...
+        Language["Objective-C"]
+      elsif data.include?(":- module")
```
bkeepers: Should this be anchored to the beginning of the line?

```ruby
/^\s*:- module/.match(data)
```

lib/linguist/heuristics.rb
arfon: Makes sense. I've just copied the regex from here: https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L81

arfon:
@halirutan - would you mind if we added https://github.com/halirutan/HeyexImport/blob/master/HeyexImport/HeyexImport.m as a sample here for Mathematica? If that's OK we could do with you adding a license to your repository (such as MIT, Apacge 2 etc).

lib/linguist/heuristics.rb
bkeepers: Gotcha. It might be worth pulling that out to a constant to reuse it.

lib/linguist/heuristics.rb
arfon: Whenever I've seen the sequence it's been at the start of a line but it's unusual enough that perhaps we don't need to be that strict?

halirutan:
@arfon No, I wouldn't mind. I added the MIT license to this repo.

arfon:
> @arfon No, I wouldn't mind. I added the MIT license to this repo.

Wonderful - thanks!

lib/linguist/heuristics.rb
sebgod: @arfon , @bkeepers ,
technically, the `:- module` statement doesn't have to be anchored to the begging of the line,
but I guess for all practical purposes it is indeed anchored.
Feel free to add the `^`
Can you also provide a break-down for
https://github.com/Mercury-Language/mercury ?

arfon:
At the [request of @sebgod ](https://github.com/github/linguist/pull/1652/files#r23502100) here's the breakdown for https://github.com/Mercury-Language/mercury

|   mercury |      master      |  This PR |
|----------|-------------|------|
| Mercury	| 75.33% | 	81.19% |
| C	| 14.38% | 	14.38% |
| M	| 5.40%  | 	0.04% |
| Shell	| 1.34%  | 	1.34% | 
| C++	| 0.95%  | 	0.95%  |
| TeX	| 0.74%  | 	0.74%  |
| Matlab	| 0.54%  | 	- |
| Game Maker Language	| 0.29%  | 	0.29% |
| Java	| 0.22%  | 	0.22% |
| Emacs Lisp	| 0.21%  | 	0.21%  |
| Makefile	| 0.20%  | 	0.20% |
| Perl	| 0.10%  | 	0.10%  |
| Prolog	| 0.06%  | 	0.06%  |
| XSLT	| 0.06%  | 	0.06% |
| Matlab | - |  0.04% |
| Awk	| 0.04%  | 	0.04% |
| Haskell	| 0.04%  | 0.04% | 
| Objective-C	| 0.03%  | 	0.02% |
| Assembly	| 0.03%  | 	0.03%  |
| VimL	| 0.02%  | 	0.02%  |
| Bison	| 0.02%  | 	0.02%  |
| Python	| 0.01%  | 	0.01%  |
| C#	| 0.00%  | 	0.00%  |
| Erlang	| 0.00%  | 	0.00%  |

arfon:
OK, I'd really like to get this merged if possible. It's close to a week since I pushed some updates on this. Unless there are any strong objections I'd like to merge this and cut a new gem tomorrow.

Thoughts?

halirutan:
@arfon What is your way of testing this? Is it possible to look at an arbitrary repository and see [how each file is classified](https://github.com/Mercury-Language/mercury/search?l=c) with this branch?

arfon:
> @arfon What is your way of testing this? Is it possible to look at an arbitrary repository and see how each file is classified with this branch?

Not really sorry. The way I've been testing is to clone the repositories and testing with a local version of Linguist based upon this branch. Are there any you'd like me to check?

arfon:
Thanks for your help on this everyone!

PR #1669, opened 2014-11-04
Support hack

arfon:
Adding support for Hack (replaces https://github.com/github/linguist/pull/1447)


staabm:
Shouldnt the `.hh` files been highlighted after this PR? Missing colors here.... https://github.com/raeesiqbal/QuickPress/blob/master/includes/main.hh

arfon:
@staabm - actually I think this is due to our recent switch to using Textmate grammars for syntax-highlighting.

We need to set [this line](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1223) to: `tm_scope: text.html.php`

I'll open up a pull request now with the change.

PR #1699, opened 2014-11-11
Added imatix/GSL to the list of known languages.

sappo:


Asmod4n:
@sappo @pchaigno close to 1000 .gsl files now https://github.com/search?utf8=%E2%9C%93&q=extension%3Agsl+if&type=Code&ref=searchresults

Would that be enough?

PR #1713, opened 2014-11-14
use syntax hilighting for uncompilable codeblocks

oli-obk:
the rust test-suite compiles all code examples in the documentation. Since some examples are not compilable, the docs often have {rust,ignore} as language for codeblocks. Syntax hilighting should still work for those docs


arfon:
Closing this in favour of a different approach (perhaps in the [html-pipeline](https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/syntax_highlight_filter.rb) as @bkeepers suggested)

PR #1731, opened 2014-11-17
Support for multiple file extension segments

bkeepers:
This is a first stab at supporting multiple file extension segments. `FileBlob#extensions` will return an array of the materialized extensions, with the most specific one first.

``` ruby
>> Linguist::FileBlob.new("app/views/things/index.html.erb").extensions
=> [".html.erb", ".erb"]
```

Right now it will return languages for the most specific extension with languages defined. In other words, if `.foo.bar` has languages defined, `.bar` will not match. Is that ok?

/cc https://github.com/github/linguist/pull/1729 @arfon @pchaigno

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -110,8 +110,8 @@ module Linguist
       # A bit of an elegant hack. If the file is executable but extensionless,
       # append a "magic" extension so it can be classified with other
       # languages that have shebang scripts.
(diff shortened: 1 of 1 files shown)


Mailaender:
This does not seem to work according to the failing tests at https://github.com/github/linguist/pull/1922.

aroben:

PR #1770, opened 2014-11-26
Add support for the Clarion language

fushnisoft:
Clarion is a Rapid Application Development tool (RAD) with a 4th Generation language. The language was create by Bruce Barrington back in the early 1980 and operated in DOS until Clarion for Windows came out in 1994.

Some examples of Clarion on GitHub:
https://github.com/devroadmaps/DevroadmapsClarionLibrary
https://github.com/fushnisoft/ClarionMetroWizard
https://github.com/CarlosGtrz/ClarionJson
https://github.com/MarkGoldberg/ClarionCommunity
https://github.com/MarkGoldberg/CwUnit
https://github.com/Rick-UpperPark/ClarionLiveDLLTutor
https://github.com/ClarionLive?tab=repositories


arfon:
> @arfon Is there a way on your end to get 1,000 (or more) `.inc` files randomly?

I'm not sure there is sorry. What are you looking for here? Just more variety? Perhaps I could sort by last updated which would add _some_ randomness to it.

pchaigno:
I'm looking for a representative sample of `.inc` files. Maybe this search could work: https://github.com/search?o=desc&q=if+OR+mov+OR+where+OR+end+OR+procedure+OR+html+extension%3Ainc&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93 (it should match most languages).

arfon:
@pchaigno I've just sent you a bunch of files for that search :zap:

larsbrinkhoff:
@pchaigno I've been using this trick: http://github.com/search?q=extension%3Ainc+NOT+wordwhichisnotfound&ref=searchresults&type=Code

pchaigno:
@larsbrinkhoff Thanks for the suggestion! I noticed this "hack" too. However, I'm not sure I understand how it works (meaning "if it really works"). For example you'll notice that with two different keywords you get two slightly different numbers of results...

I chose to not use this technique here because, whichever sorting option you select, the results are grouped by repositories. Thus, the samples would be less random, or at least less representative of the entire collection of `.inc` files.
Does this make sense?

larsbrinkhoff:
@pchaigno Yes, that makes sense.  It's too bad we can't get a truly random sample.

fushnisoft:
I only just noticed that there was a conflict on this pull request. Hopefully I managed to update everything ok. Thanks for sticking with this guys, if there is anything else I can provide to help let me know.

PR #1785, opened 2014-11-28
Add Fish Shell support.

ghost:
I thought Linguist was still using [Pygments](https://bitbucket.org/birkenfeld/pygments-main/), so I [went ahead and committed](https://bitbucket.org/birkenfeld/pygments-main/pull-request/422/add-fishshelllexer-class-to-support-the/diff) a `FishLexer` to the main repo @ BitBucket. Later found out Linguist uses Textmate grammars so I set out to write one for Fish. Fortunately, I found [this relic](https://github.com/l15n/fish-tmbundle), tested it with [LightShow](lightshow.githubapp.com/) and added the relevant references in this commit.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3202,6 +3202,7 @@ fish:
   extensions:
   - .fish
   tm_scope: source.fish
+  ace_mode: none
 
 mupad:
   extensions:


RSully:
@brj any idea how hard it might be to convert a Pygments syntax to something compatible with linguist? Both of the fish bundles references in this thread (l15n and toru-hamaguchi's) seem to have significant issues after the test of time, and neither are quite active projects at this point.

brj:
@RSully I don't know how hard it would be to convert it, but shouldn't be we looking just at https://github.com/davidcelis/language-fish-shell/blob/master/grammars/fish.cson?

RSully:
@brj if I was personally going to start working on a fish grammar I'd be starting from scratch, maybe using bash/zsh as a starting point and using the existing fish stuff as inspiration and for head scratchers; I simply do not have high confidence in either of the existing solutions, and I'd rather not spend time trying to hack on them to become correct. 

My interest in conversion is a question of having to maintain multiple versions once a proper grammar finally *is* written. Fish moves a lot faster than other shells, and I don't want to be manually duplicating changes.

PR #1786, opened 2014-11-28
Remove .script! hack

bkeepers:
It is no longer necessary.

/cc @arfon @pchaigno


arfon:
@bkeepers @pchaigno - I'm trying to clean up old PRs and this is _old_. What are our thoughts on this these days?

bkeepers:
For some reason this PR is showing a ton of other commits. First I tried just merging master. Then I tried rebasing and force pushing. But neither of those seem to fix it. I'm not sure what happened here, but I'll just open another PR.

/cc @mhagger

PR #1797, opened 2014-11-30
Cs smalltalk

arfon:
Replacement PR for https://github.com/github/linguist/pull/1446


khellang:
Why was this change made? I assume this is why GitHub gists now classifies .cs files as smalltalk?



Wouldn't you say it's a bit weird picking a language barely in the TIOBE top 100 list over the 4th most popular language, by default?

pchaigno:
@khellang GitHub should be selecting the first extension for Gists (`.st` for SmallTalk). The issue has already been reported to support@github.com (it can't be fixed here).

khellang:
Right, I reported it myself as well I was just wondering why the change was made :smile:

pchaigno:
See #1446 (tl;dr because `.cs` is also a Smalltalk extension ;)

arfon:
Yeah, the UI is somewhat confusing right now - a fix for this is on the way. What gist _says_ it is going to do is actually often not correct :-/ , that is, I believe this will actually render as C sharp.

PR #1806, opened 2014-12-02
Syntax highlighting for Nit language

R4PaSs:
Since you guys changed the way you handle highlighting, here's the updated version for the Nit language !

Samples were already included in #1320, I included a few samples however since it has some stuff that was not covered by the old samples (triple-quoted strings, FFI)


aroben:
I took care of this over in #1973. It should be live on github.com within a couple of weeks.

aroben:
And thanks!

PR #1818, opened 2014-12-03
SCSS isnt SASS

arfon:


hcatlin:
We consider the Sass language to encompass both "indented Sass" and it's more common flavour "SCSS". They are the same language, with two syntaxes that are supported.

PR #1823, opened 2014-12-04
add support for Isabelle ROOT file syntax

lsf37:
This pull request adds support for Isabelle ROOT files. The Isabelle theory language is already part of linguist, ROOT files are additional files with different syntax, somewhat like Makefiles with automatic dependency analysis.

Sample file is included. Usage on github is for every larger Isabelle project, for instance:
https://github.com/seL4/l4v/blob/master/proof/ROOT
https://github.com/seL4/l4v/blob/master/spec/ROOT
https://github.com/seL4/l4v/blob/master/tools/autocorres/ROOT
https://github.com/seL4/isabelle/blob/master/src/HOL/ROOT
https://github.com/seL4/isabelle/blob/master/src/FOL/ROOT
etc

Ambiguity: there seem to be plenty of lowercase "root" or "Root" files on github for different languages, but I couldn't find any all caps "ROOT" (the name is also not mentioned in other languages linguist knows about).

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1356,6 +1356,14 @@ Isabelle:
   tm_scope: source.isabelle.theory
   ace_mode: text
 
+Isabelle ROOT:
+  type: programming
+  group: Isabelle
+  filenames:
+  - ROOT
+  tm_scope: source.isabelle.root
+  ace_mode: text
+
 J:
   type: programming
   extensions:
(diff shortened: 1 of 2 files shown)


lsf37:
After increased search-foo: `session filename:ROOT` gives [~12,900 results](https://github.com/search?utf8=✓&q=session+filename%3AROOT&type=Code).

At around page 35, files that are not just `ROOT` but have extensions or are not all uppercase are starting to show up, so it's roughly 350 actual usages.

arfon:
Closing in favour of #2481. Sorry for the ridiculously slow turnaround on this :-\

lsf37:
No worries, and thanks!

PR #1825, opened 2014-12-04
SaltStack state files use the same extension as old scheme library source files

iggy:
Add support for detecting the difference between scheme library source files
and SaltStack state files (which are technically YAML+Jinja).

Add languages entry and a disambiguation entry that should work in 99% of cases
(at least I can't think of a better way).


iggy:
Superseded by 1929

PR #1830, opened 2014-12-05
Add .xml.fr and .html.fr

larsbrinkhoff:
In #1779, I tried to add samples for `.xml.fr` and `.html.fr`.  However, #1731 doesn't accommodate this, although that was probably the intent.

This tries to fix the bug.  New extensions and samples serve as test cases, but would also be nice to add.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1242,6 +1242,7 @@ HTML:
   extensions:
   - .html
   - .htm
+  - .html.fr
   - .html.hl
   - .st
   - .xht
@@ -3448,6 +3449,7 @@ XML:
   - .xliff
   - .xmi
   - .xml.dist
+  - .xml.fr
   - .xsd
   - .xul
   - .zcml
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
@arfon @bkeepers So now that the first part of this PR is merged, thumbs up or down for the second part?

larsbrinkhoff:
Example of misidentified files:
https://github.com/hlkzmy/windows-php-webserver/search?l=frege

bkeepers:
Sorry for not helping push this forward. I'm a little uneasy about adding `.html.rf` as an extension for HTML when there are a bunch of other extensions that we would also have to add.

I took a stab at fixing this in https://github.com/github/linguist/pull/2009 by using all parts of the extension to get possible languages. The idea being then we would leave it up to the classifier to properly detect the right one.

larsbrinkhoff:
Cool, I'll close this when your branch is merged.

bkeepers:
I'm going to work on getting #2009 ready. But in the mean time, I'm going to go ahead and close this, because I don't think adding extensions for locales is the route we should go.

PR #1832, opened 2014-12-05
Optimization of generated?

pchaigno:
This pull request reorganizes the methods called in `generated?` to only load the data (calls to `lines`) in last resort.


aroben:

PR #1837, opened 2014-12-05
Add Lisp and PicoLisp language.

Grahack:
No extension is specified.
Users must attach to these languages from their `.gitattributes` file.


Grahack:
Hi. That's horribly frustrating! (no anger here)
I thought I was not far from a merge. Is there a problem?

arfon:
> I thought I was not far from a merge. Is there a problem?

I think this is OK. @bkeepers - I would appreciate a quick :eyes: from you.

bkeepers:
I'm not familiar with the differences between Lisp and Common Lisp. Could you help me understand what this change accomplishes?

- Is the goal to make it so the language bar and explore shows `Lisp` instead of `Common Lisp`?
- Should `Common Lisp` be in the `Lisp` group?
- Since `Lisp` and `Common Lisp` use the same grammar, would it make sense to move any of the file extensions to `Lisp`?

Grahack:
> Is the goal to make it so the language bar and explore shows Lisp instead of Common Lisp?

No, Common Lisp files should still be treated the same.  
The goal here is to be able tell from the `.gitattributes` file that some files should be highlighted as Lisp source code, without cluttering the extensions space. Since the targeted language PicoLisp is *quite confidential*, pchaigno told [here](https://github.com/github/linguist/pull/1827#issuecomment-65716637) me that "... so I'm afraid we can't add support for it on Linguist :/", which I understand perfectly.

> Should Common Lisp be in the Lisp group?

Yes, historically speaking. But if this is not convenient, I really don't care at all since I don't use CL!

> Since Lisp and Common Lisp use the same grammar, would it make sense to move any of the file extensions to Lisp?

Why not, but I don't know all the implications. Lisp is a really big family, with what should be a really huge list of extensions. Doing this for the extensions listed in CL would be the beginning of a game with no end. Also, maybe your suggestion could have some unwanted implications on CL users.

Grahack:
Strange.
After having "vendored" some html files, my repo now says that I use Common Lisp files.
See here: <https://github.com/Microalg/microalg>.
Maybe I was not clear with the first "No" in my previous answer.
It's not perfect (I'd prefer to have Lisp instead of Common Lisp), but it's somehow better.

larsbrinkhoff:
Hello,

As you can see in #2449, I have been working to add PicoLisp support to GitHub's linguist.  All `.l` files in [Grahack/EmuLisp](https://github.com/Grahack/EmuLisp), [Grahack/minipicolisp](https://github.com/Grahack/minipicolisp), and [Microalg/microalg](https://github.com/Microalg/microalg) will be correctly identified.

It might still be a good idea to add `Lisp` as a generic name for various Lisp dialects.

arfon:
@larsbrinkhoff @Grahack - given that #2449 has merged can this be closed now?

larsbrinkhoff:
@arfon There is some merit to the idea of having a separate `Lisp` language without any file extension.  To be used with Linguist overrides.

But given that no users are screaming for this, I guess it's not needed at this point.

Grahack:
The PicoLisp community is very thankful for the detection of (one of) their language of choice, which is better for us than the workaround that was thought at first (*a separate Lisp language without any file extension*). Historically speaking, *a separate Lisp* would be nice as it is a very ancient and big family. But indeed I guess that nobody will be screaming for this.
Sadly, if it is implemented and deployed, the fact that it has to be explicitly set via .gitattributes will prevent  this feature to be often used even if there had been a real need.
Thanks to all !

arfon:
OK thanks @Grahack & @larsbrinkhoff

PR #1850, opened 2014-12-09
Add WebIDL support

othree:
WebIDL[[1]](http://www.w3.org/TR/WebIDL/) is very common in web standards.
Mozilla also use it as browser environment init configuration.
Lots of WebIDL file is already in the web world [[2]](https://github.com/mozilla/gecko-dev/tree/master/dom/webidl).
So add support for this file type.


aroben:

othree:
Great!

Issue #1626, opened 2014-10-23
Fixing C/C++/Objective-C classifications


arfon:
A fair fraction of the outstanding issues with Linguist mis-classifications are to do with C, C++, Objective-C and any other languages that use `.m` or `.h` as extension.

While it's possible we can see further improvements by increasing our samples for the Bayesian classifier I'm pretty convinced we need to craft a few reliable heuristics to shortcut the classifier.

We currently have the following in [`heuristics.rb`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L36-L41) but it's not being called. 

``` ruby
def self.disambiguate_c(data, languages)
  matches = []
  matches << Language["Objective-C"] if data.include?("@interface")
  matches << Language["C++"] if data.include?("#include <cstdint>")
  matches
end
```

@DX-MON had a good go at this in #1036 but at the time we didn't have a good way to benchmark the effects of these changes. With the new [benchmarks](https://github.com/github/linguist/blob/master/Rakefile#L36) we are now in a good place to tackle this problem.

The [current heuristic](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L36-L41) is deliberately limited in scope, that's because any heuristics we define should be accurate, simple to understand and fast.

Basically I'm asking for some help with this. I'm more than happy to grab a bunch (1000s) of files to test any potential changes on. Questions for you all:
1. How do we feel about #1036 - I'd love to see a case for `C` in that `disambiguate_c` method too if possible. 
2. #1036 is matching on a [bunch of keywords](https://github.com/github/linguist/pull/1036/files#diff-e56a8d50b1a75fc4212ffb3504faa007R43) - are all of these necessary? Are we potentially sacrificing accuracy for completeness?
3. Are there any good heuristics for pure C code?

arfon:
From https://github.com/github/linguist/issues/2035 both of these files are being incorrectly identified as C++:

- https://github.com/Vedaad-Shakib/MPS/blob/master/include/que.h
- https://github.com/Vedaad-Shakib/MPS/blob/master/src/que/que.h

I haven't dug into why yet but wanted to report this. 

We also have another new C/C++ issue here: https://github.com/github/linguist/issues/1993

phillipberndt:
I have another example for a C header incorrectly classified as C++:
https://github.com/phillipberndt/pqiv/blob/469fe63df4b27463f8accf8a45ae515a17c202e7/pqiv.h

[...]

PR #1856, opened 2014-12-10
Add a bunch of C header samples for the Bayesian classifier

lutoma:
This pull request is a result of the discussion in #1626. It contains a whole bunch of C header files taken from https://github.com/lutoma/xelix. As it's a kernel, the headers tend to be a bit unconventional. Also, they contain a lot of structs and typedefs which were previously very underrepresented in the samples.

I know its a lot of files, but it was the only set that managed to push the misdetections as C++ down to 3.15% (from the previous 12.8%). But I'd be glad to remove some of them if you think it's too much.

### DIFF
--- a/samples/C/bitmap.h
+++ b/samples/C/bitmap.h
@@ -1,0 +1,47 @@
+#pragma once
+
+/* Copyright © 2010 Christoph Sünderhauf
+ *
+ * This file is part of Xelix.
+ *
+ * Xelix is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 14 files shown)


lutoma:
@arfon Just wondering if this has been deployed in the meantime? I'm still seeing the old stats for my repositoy.

arfon:
Hi @lutoma - the language stats only update when you push to a repo. I'm able to force this update through our admin UI (which I've just done). These seems to be a little better now.

If you want a 'complete' solution then you might want to take a look at the [overrides here](https://github.com/github/linguist#overrides).

PR #1866, opened 2014-12-12
Better C++ regex

arfon:
Based upon this comment: https://github.com/github/linguist/issues/1626#issuecomment-66371622


arfon:
> @DX-MON - do we have any example files that would match these?

Per this comment a few weeks ago I'd love some example files here that this new regex addresses. @DX-MON do you have any we could use?

lib/linguist/heuristics.rb
```diff
@@ -73,7 +73,7 @@ def call(data)
       if (/@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/.match(data))
```
DX-MON: *sighs* I feared that might happen.. only thing we can do for the time being is anchor to the start of the line and add a skip for white-space only. this will still match under certain conditions even when it shouldn't but without teaching linguist how to scan comments to exclude them it's the best we can do.

DX-MON: ok @arfon, having given it some thought, the revised regex needs to be:

/^[ \t]*@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/

arfon:
Confirmed with the `Entity.h` sample (C++) file that the newest regexes are working as expected (i.e. not matching Objective-C on the `@class` in the comment _and_ correctly matching with the C++ regexes.

arfon:
Thanks for your help @DX-MON

vinz243:
Thank you very much!

bkeepers:
:clap:

aroben:

vinz243:
Thank you! It's working! https://github.com/whitedrop/whitedrop

PR #1871, opened 2014-12-12
Fish

arfon:


jakwings:
The regex for "comment" is wrong, it should work like the bash one with a look-backward `(?<!\\S)`, anyone can fix it?

PR #1872, opened 2014-12-13
Auto-detect, auto-convert character encodings to UTF-8, no "Charlock" needed.

geoff-codes:
This is an improved reimplementation of #855, something I was hacking on a year ago.

The long and short of it: replace @brianmario / [charlock_holmes](https://github.com/brianmario/charlock_holmes) with a fast, pure Ruby encoding detector for common encodings. It certainly doesn't do all that the ICU library can do — it won't detect legacy multi-byte encodings, Big5, EBCDIC. But it also doesn't have to `dlopen()` a ~25Mb C++ library/module when invoked. And if I had to take a guess, I'd say this probably covers ±99.9% of text files hosted on GitHub.

While I don't have exhaustive benchmarks like last time (@arfon), nor is the module really written as it should be (it currently just monkey patches String), I believe it's at least 50% faster than Charlock (as its used in linguist) and returns far fewer errors, where just about anything that's not UTF-8 gives a  `'split': invalid byte sequence in UTF-8 (ArgumentError)`.

I've put the code for this in a [small gem](https://github.com/geoff-codes/charlotte). The implementation is [here](https://github.com/geoff-codes/charlotte/blob/master/lib/charlotte.rb). Let me know if the comments/code are unclear; I'd be happy elaborate. The repo also has a couple quick and dirty test scripts to compare against charlock and rchardet.

This PR isn't really meant to be merged as is — I've just quickly replaced "charlock" with "charlotte" to demonstrate. It works as-is, but I think a major refactor of blob_helper.rb, file_blob.rb, etc. would be in order if this idea were to be adopted. So perhaps it's more perhaps an idea for a topic branch?

Also, it would probably be wise to bring `linguist` in line with [github/lib/github/encoding.rb](https://github.com/github/github/blob/master/lib/github/encoding.rb),  [gist/lib/gists/encoding.rb](https://github.com/github/gist/blob/master/lib/gists/encoding.rb),  [gitrpc/lib/gitrpc/encoding.rb](https://github.com/github/gitrpc/blob/master/lib/gitrpc/encoding.rb), and kin. Although I don't believe that code is presently open source.

So maybe this is something to talk about on or sometime after Monday, @tmm1? :wink:

:octocat: — Geoff

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.files = Dir['lib/**/*'] - ['lib/linguist/grammars.rb']
   s.executables << 'linguist'
 
-  s.add_dependency 'charlock_holmes', '~> 0.7.3'
+  s.add_dependency 'charlotte', '~> 0.1.4'
   s.add_dependency 'escape_utils',    '~> 1.0.1'
   s.add_dependency 'mime-types',      '>= 1.19'
   s.add_dependency 'rugged',          '~> 0.22.0b4'
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -1,5 +1,5 @@
 require 'linguist/generated'
-require 'charlock_holmes'
+require 'charlotte'
 require 'escape_utils'
 require 'mime/types'
 require 'yaml'
@@ -104,15 +104,11 @@ module Linguist
     end
 
     def encoding
-      if hash = detect_encoding
-        hash[:encoding]
-      end
+      data.encoding
     end
 
     def ruby_encoding
-      if hash = detect_encoding
-        hash[:ruby_encoding]
-      end
+      data.encoding
     end
 
     # Try to guess the encoding
@@ -121,7 +117,7 @@ module Linguist
     #          this will return nil if an error occurred during detection or
     #          no valid encoding could be found
     def detect_encoding
-      @detect_encoding ||= CharlockHolmes::EncodingDetector.new.detect(data) if data
+      @detect_encoding ||= "UTF-8"
     end
 
     # Public: Is the blob binary?
@@ -135,14 +131,9 @@ module Linguist
       # Treat blank files as text
       elsif data == ""
         false
-
-      # Charlock doesn't know what to think
-      elsif encoding.nil?
-        true
-
-      # If Charlock says its binary
-      else
-        detect_encoding[:type] == :binary
+  
+        else
+        data.encoding == "BINARY"
       end
     end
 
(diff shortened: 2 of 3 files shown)


brianmario:
> replace @brianmario / charlock_holmes with a fast, pure Ruby encoding detector for common encodings.

I'm all for using something faster for detecting encodings, but it has to be accurate as well. Though ICU is written in C++ and compiles it's character encoding tables into an optimized and architecture-specific database so I have a hard time imagining anything pure-ruby would be nearly as fast. At least with as many encodings as we need to support.

Currently, our mean detection time with `charlock_holmes` is somewhere ~1ms, max hovering ~100ms and a handful of outliers ~200ms. Those pathological cases are really rare though.

> But it also doesn't have to dlopen() a ~25Mb C++ library/module when invoked.

I may be misunderstanding how shared libraries work with Ruby, but the libicu lib should only be loaded once at boot and used in-memory from then on out, right? If that's not the case then this would be a concern of every C extension that depends on an external library as well?

> And if I had to take a guess, I'd say this probably covers ±99.9% of text files hosted on GitHub.

The vast majority of encodings we see are ASCII (we track that separately from UTF-8), UTF-8 and "binary". But ISO-8859-1, Big5, Shift_JIS and a few Russian encodings are popular as well, especially on Enterprise installs where the every repo may be in one of these "legacy" encodings.

> ...and returns far fewer errors, where just about anything that's not UTF-8 gives a 'split': invalid byte sequence in UTF-8 (ArgumentError)

Just wanted to note that this exception has nothing to do with `charlock_holmes` itself. The `split` method is being attempted on a string that is marked as one encoding, but either contains invalid bytes in that encoding (as the exception says) or is actually in a different encoding and is marked incorrectly in Ruby. That last part may be the doing of `charlock_holmes`, but in almost every case I've seen this happen in the last 3 years we've been using it in production:

* We didn't have enough data to hand over for detection. ICU suggests "hundreds of bytes" in order to make any kind of somewhat accurate detection.
* The data was in fact corrupt/invalid. I've seen cases where an entire file is actually UTF-8 (or some other encoding), except a single byte (usually something like a smart-quote, probably pasted from a word processor using some other encoding like windows-1252).
* ICU just straight up guessed wrong

But given the amount of data we correctly guess the encoding for, the above 3 cases are _extremely_ rare. We might see a support ticket once a month or so.

While detecting a few more standard encodings could potentially be made faster, the trade-off of not being able to detect even the top 5 "less popular" encodings used on our hosted service would be a deal breaker. That would mean entire companies in Japan or China who are still using those other encodings for legacy reasons (iirc almost all email clients in Japan still uses Shift_JIS by design because of that).

ICU is maintained by IBM, and ships with OSX to provide all of it's encoding functionality. We trust it and it's served us well at our scale.

Although, it's not without it's annoyances. We constantly have compile-time issues in development whenever we move to a newer version of libicu. Though we're working on a "fix" for that [here](https://github.com/brianmario/charlock_holmes/pull/82). I do wish we were able to provide less data and still get an accurate detection. But encoding detection is tricky business. One perfect example I saw recently was an Illustrator file that had about 256k of XML at the top, but the rest of the 1MB file was just binary. Had we not performed detection on the entire file, or used something naive like git by checking for a `NUL` byte in the first 8kb, we would have happily served the file to the user as text.

But all of that said, I definitely appreciate your research here. I don't think we'll be able to go with this library as it stands today. But if you're able to come up with something that's faster, at least as accurate and supports the majority of the 40 or so encodings that are still in somewhat public on our service I'd love to take a look :)

PR #1878, opened 2014-12-15
Add heuristic to distinguish between Component Pascal and C++ for .cp

pguyot:
.cp is quite commonly used for C++ rather than for Component Pascal.

https://github.com/search?utf8=✓&q=include+extension%3Acp&type=Code&ref=searchresults
currently yields 10442 results, mostly C++.

https://github.com/search?utf8=✓&q=MODULE+extension%3Acp&type=Code&ref=searchresults
currently yields 699 results (and some actually are C++ as the search is case insensitive).


arfon:
Hi @pguyot - I think https://github.com/github/linguist/pull/2055 should address this issue.

PR #1888, opened 2014-12-17
reorganized Sublime Text extensions, added some

MattDMo:
Moved extensions that were in JSON to JavaScript, as comments are allowed in the Sublime files, and were being incorrectly highlighted as errors (see [this file](https://github.com/titoBouzout/SideBarEnhancements/blob/st3/Side%20Bar.sublime-settings), for example). Added several more extensions to JavaScript that hadn't been included previously. Finally, added 2 extensions to XML.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1462,12 +1462,12 @@ JavaScript:
   - .sublime-macro
   - .sublime-menu
   - .sublime-mousemap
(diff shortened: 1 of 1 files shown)


MattDMo:
Update - the Travis build is now passing!

aroben:
Thanks @MattDMo!

PR #1892, opened 2014-12-18
Contribution guideline: avoid false positives.

larsbrinkhoff:
Suggestion: Contributions should avoid false positives.

It's easy to add a new language with a file extension in languages.yml (or add a new extension to an existing language), but that may result in a great number of files being misdetected.  This isn't doing new languages any service, since e.g. searches, trendling lists, etc will be cluttered with unrelated results.  Usually this can be avoided with a heuristic.


larsbrinkhoff:
@arfon Looks great!  Closing this.

arfon:
:+1: thanks!

PR #1894, opened 2014-12-18
Add Clojure support for .boot files

alandipert:
[Boot](http://boot-clj.com/) is a Clojure build tool that uses the `.boot` extension for Clojure code run in the Boot environment.  It would rock if `.boot` files were Clojure highlighted.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -492,6 +492,7 @@ Clojure:
   color: "#db5855"
   extensions:
   - .clj
+  - .boot
   - .cl2
   - .cljc
   - .cljs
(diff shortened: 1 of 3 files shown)


arfon:
Fixed in https://github.com/github/linguist/pull/2254 and https://github.com/github/linguist/pull/2300 https://github.com/github/linguist/pull/2300

peterromfeldhk:
cant you add a option to say in second line something like #[lang=clojure]
to override fileextension to use proper lang syntax? its common with lint...
+1

arfon:
Yep, you can do something very similar to that. Check out the overrides we support: https://github.com/github/linguist/blob/master/README.md

PR #1896, opened 2014-12-18
Move all grammars that use Git repos to submodules

aroben:
This makes it so we don't have to redownload all the grammars every time we build the grammars gem. We test that the list of submodules in the repo matches the list in grammars.yml. Eventually we'll be able to test that the list of scopes for each submodule is accurate as well, which was the original motivation for this change.

/cc @bkeepers @arfon @pchaigno

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ before_install:
   - git fetch origin test/attributes:test/attributes
   - git fetch origin test/master:test/master
   - sudo apt-get install libicu-dev -y
+  - git submodule init
+  - git submodule sync --quiet
   - script/fast-submodule-update
 rvm:
   - 1.9.3


bkeepers:
`git status` being delayed a second or two has been a (minor) annoyance.

@peff @mhagger This is probably an edge case, but I can't imagine we're the only one using submodules in this manner. Have either of you done any work on submodules? I'm curious if there are any optimizations to be had.

peff:
Neither of us really works on submodules.

It looks like the slowness is mostly from running `git status --porcelain` in each of the submodules. And it's not a CPU problem so much as a latency issue. Here are some timings on my box of a fresh `clone --recursive` of linguist:

```
$ time git status --ignore-submodules=none
real    0m0.270s
user    0m0.020s
sys     0m0.052s

$ time git status --ignore-submodules=all
real    0m0.011s
user    0m0.004s
sys     0m0.004s
```

which tells me:

  1. You should get a faster machine. It's only 270ms here.

  2. You can work around this by setting submodule.*.ignore (and/or diff.ignoreSubmodules). The second run is way faster.

  3. We're only spending 77ms of CPU time, but the wall-clock time is 270ms. Git could probably run all of the submodule processes in parallel and finish in 77ms on a single-core processor, or much less on a multi-processor.

I also ran the top command through `perf` on Linux. Here are the top time consumers:

```
    10.16%      git  ld-2.19.so          [.] do_lookup_x                            
     6.99%      git  ld-2.19.so          [.] _dl_relocate_object                    
     5.85%      git  git                 [.] git_diff_ui_config                     
     4.32%      git  git                 [.] git_status_config                      
     3.41%      git  ld-2.19.so          [.] _dl_lookup_symbol_x                    
     2.40%      git  [kernel.kallsyms]   [k] page_fault                             
     2.40%      git  ld-2.19.so          [.] strcmp                                 
     2.20%      git  git                 [.] git_default_config                     
     1.94%      git  [kernel.kallsyms]   [k] copy_page_rep                          
     1.82%      git  git                 [.] git_diff_basic_config                  
     1.72%      git  [kernel.kallsyms]   [k] __d_lookup_rcu                         
```

It looks like a big chunk of the CPU is going to process startup (all of the ld.so calls, and all of git's config parsing). You might be able to get some speedup by doing these `status` calls in-process. However, you may run into correctness problems (e.g., the status subprocess should be respecting the config in the submodule, and git doesn't have a clean way of pushing and popping context like this).

Doing it all in-process is probably a pretty huge project. Running the "is the submodule dirty" checks in parallel is probably a lot more do-able.

robrix:
> Git could probably run all of the submodule processes in parallel and finish in 77ms on a single-core processor, or much less on a multi-processor.

[Related issue on GHfM](https://github.com/github/Mac/issues/5681).

aroben:
On my Retina MacBook Pro running OS X 10.10:

```
$ time git status --ignore-submodules=none >/dev/null

real	0m0.835s
user	0m0.327s
sys	0m0.405s
$ time git status --ignore-submodules=all >/dev/null

real	0m0.071s
user	0m0.042s
sys	0m0.027s
```

So it seems there's a bigger performance difference on OS X than on Linux.

For me personally, the biggest annoyance is that `__git_ps1` (which adds git information to my shell prompt) is quite slow in this repo. I don't run `git status` all that often so having it be slow isn't a big deal. I wonder if you're in the same situation @arfon. There are ways to disable some of `__git_ps1`'s functionality in certain repositories to speed it up; I should probably look into doing that.

peff:
I'm not surprised there's some performance difference between Linux and OS X. `git status` is going to be quite heavy on `lstat` calls, and those are notoriously slower on OS X. Ignoring the submodules doesn't just save process startup cost, but it also avoids `lstat`ting all of the files in the submodules.

The discouraging thing is that your CPU time and wall-clock time for the slow case are quite similar.  Which means there's a lot less to gain from parallelizing (we still take advantage of multiple CPUs, but we're not improving the single-core utilization like the Linux case was). And probably the `lstat` load is much higher compared to the startup load.

I'd be interested to see a comparison of your first run above with a version of the repo in which all of the submodules are inserted directly. I.e., run:

```
git clone --recursive https://github.com/github/linguist
cd linguist

# hacky de-submodulize
rm -rf vendor/grammars/*/.git
git rm -r --cached vendor

# and then add back in the real files
git add vendor
git commit -m de-submodule

time git status
```

I get:

```
real    0m0.047s
user    0m0.032s
sys     0m0.032s
```

I'm curious what you see on OS X. This number should represent the _fastest_ we can get it by playing tricks with submodules (parallelizing calls and/or shoving the work into the parent process).

I also wonder what you would see (in both cases) with `git config --global core.preloadindex true`.

aroben:
Without de-submodulizing:

```
$ time git status >/dev/null

real	0m0.842s
user	0m0.341s
sys	0m0.404s
$ git config --global core.preloadindex true
$ time git status >/dev/null

real	0m0.810s
user	0m0.316s
sys	0m0.397s
```

With de-submodulizing:

```
$ time git status >/dev/null

real	0m0.240s
user	0m0.116s
sys	0m0.179s
$ git config --global core.preloadindex true
$ time git status >/dev/null

real	0m0.256s
user	0m0.121s
sys	0m0.185s
```

peff:
Interesting. So there is definitely some overhead to the sub-processes.

Unfortunately stuffing it all into one process (while using submodules) is probably going to be quite an invasive change. And I'm not sure how much parallelizing the external submodule calls is going to help, since the CPU seems pretty busy.

I'm not sure there's an easy solution, short of tweaking the config to avoid looking at the submodules at all. :cry:

peff:
And regarding `core.preloadindex`, I forgot that we made it the default in v2.1.0. If you're curious about its effect, you might try `time git -c core.preloadindex=false status`.

arfon:
> I wonder if you're in the same situation @arfon. 

Yeah, I have a similar setup to you I think @aroben. As for stats from my machine (MacBook Air, OS X 10.10)

```
time git status --ignore-submodules=none >/dev/null
real	0m1.020s
user	0m0.396s
sys	0m0.480s

time git status --ignore-submodules=all >/dev/null
real	0m0.088s
user	0m0.055s
sys	0m0.030s
```

Then...

```
git clone --recursive https://github.com/github/linguist
rm -rf vendor/grammars/*/.git
git rm -r --cached vendor
git commit -m de-submodule

time git status
real	0m0.302s
user	0m0.143s
sys	0m0.215s
```

PR #1899, opened 2014-12-18
Add HCL to languages.yml

sethvargo:
More information on HCL: https://github.com/hashicorp/hcl

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1153,6 +1153,14 @@ Groovy Server Pages:
   tm_scope: text.html.jsp
   ace_mode: jsp
 
+HCL:
+  type: programming
+  extensions:
+    - .hcl
+    - .tf
+  ace_mode: javascript
+  tm_scope: source.json
+
 HTML:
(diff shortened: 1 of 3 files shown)


arfon:
Looks like there are other languages using `.hcl` too: https://github.com/davidw/hecl - https://github.com/search?utf8=%E2%9C%93&q=source+extension%3Ahcl&type=Code&ref=searchresults

We'll need to work out a good way we can disambiguate these two before merging this.

arfon:

sethvargo:
Hi @arfon 

Thanks for the ping. I'm not sure of a good way to distinguish between the other HCL extensions. I'd welcome any ideas.

arfon:
The other files look to be `Hecl` which doesn't seem to have seen any updates since 2010. Also, a cursory search for `Hecl` files makes me think the level of usage is extremely low (tens of files).

I therefore think it's OK for us to merge this as-is.

Thanks for you patience @sethvargo - this should be live in the next release of Linguist.

PR #1900, opened 2014-12-19
Use source.smalltalk to hightlight STON file

dalehenrich:
While not perfect, it turns out that source.smalltalk is a better fit for highlighting STON files than source.json. When STON departs from pure JSON (often) the hightlighting is pretty bad.


dalehenrich:
Ping ... is there something else that needs to be done for this PR or is it just in the `merge queue`?

arfon:
:+1: this looks good @dalehenrich. It will be another week or so before this change is live on GitHub.

PR #1905, opened 2014-12-19
Add support for cql and ddl files

joaquincasares:
The Cassandra Query Language is pretty similar to that of SQL with a few trivial changes. Whenever I try .sql file extensions it maps closely to what I would expect.

https://cassandra.apache.org/doc/cql3/CQL.html

Here are the change for .cql and, more broadly, .ddl files.

Thanks!

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2611,12 +2611,12 @@ SQL:
   ace_mode: sql
   extensions:
   - .sql
+  - .cql
+  - .ddl
   - .prc
   - .tab
   - .udf
   - .viw
-  - .cql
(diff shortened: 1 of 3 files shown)


joaquincasares:
I've added a sample file with both the .cql and .ddl extensions. Thanks!

pchaigno:
Great! It looks good to me, you just have to wait for someone from GitHub to review and merge ;)

joaquincasares:
Awesome! Thanks for helping!

aroben:
Thanks @joaquincasares!

joaquincasares:
Thanks for merging!

PR #1909, opened 2014-12-22
Test that grammars.yml lists the right scopes for each submodule

aroben:
`convert-grammars` now supports a few flags that we can use to make it dump out the YAML just for the local grammar submodules. We can then compare this to the YAML that's actually in grammars.yml to check that they're the same. If they aren't, `grammars.yml` needs to be updated. This will help catch mistakes like using the wrong scope name.

/cc @bkeepers @arfon @pchaigno


pchaigno:
This test would give me some peace of mind when I add a new grammar :zap:

PR #1922, opened 2014-12-28
Added support for Mono .dll.config XML files

Mailaender:
Reference: http://www.mono-project.com/docs/advanced/pinvoke/dllmap/#example

Example: https://github.com/OpenRA/OpenRA/blob/release-20141029/thirdparty/SharpFont.dll.config

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3106,6 +3106,7 @@ XML:
   - .dita
   - .ditamap
   - .ditaval
+  - .dll.config
   - .filters
   - .fsproj
   - .glade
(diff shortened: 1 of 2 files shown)


arfon:
@Mailaender - would you mind adding this example file to the `samples` directory?

Mailaender:
Done.

bkeepers:
> @bkeepers - this PR means we support this kind of extension now right?

Theoretically, but we have not added one yet, so it looks like there are still some issues with it. I'll get it fixed up.

bkeepers:
Fixed in https://github.com/github/linguist/pull/1940. If you merge master the tests should be green.

Mailaender:
Rebased.

arfon:
:+1: thanks @Mailaender & @bkeepers

arfon:
@Mailaender it will likely be another couple of weeks before this change is live on GitHub as we'll need to cut a new Linguist gem.

aroben:

Mailaender:
Thanks!

PR #1923, opened 2014-12-28
lib/linguist/heuristics.rb: Disambiguate Mercury-M

sebgod:
In the [Mercury repository](http://github.com/Mercury-Language/mercury)
75 files have been classified as 'M', this patch tries to rectify that,
using the fact that Mercury modules must contain a ':- module'
directive.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -181,5 +181,13 @@ module Linguist
         Language["Text"]
       end
     end
+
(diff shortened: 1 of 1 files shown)


pchaigno:
Maybe this heuristic could be used to distinguish Mercury and Matlab as discussed in #1652?

arfon:
Closing in favour of #1652

PR #1930, opened 2014-12-31
Switch to Minitest::Test instead of Test::Unit::TestCase

aroben:
This gives us a consistent test framework across all Ruby versions which should help avoid errors that are only found when CI runs the tests on different Rubies. (And this fixes an immediate bug where there's no  `skip` method in the version of test-unit we're currently using only on Ruby 2.2.)

/cc @arfon @bkeepers

### DIFF
--- a/test/test_generated.rb
+++ b/test/test_generated.rb
@@ -1,6 +1,6 @@
 require_relative "./helper"
 
-class TestGenerated < Test::Unit::TestCase
+class TestGenerated < Minitest::Test
   include Linguist
 
   def samples_path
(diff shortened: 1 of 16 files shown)


test/test_generated.rb
```diff
@@ -11,14 +11,16 @@ class DataLoadedError < StandardError; end
 
   def generated_without_loading_data(name)
     blob = File.join(samples_path, name)
-    assert_nothing_raised(DataLoadedError, "Data was loaded when calling generated? on #{name}") do
+    begin
       Generated.generated?(blob, lambda { raise DataLoadedError.new })
+    rescue DataLoadedError
+      assert false, "Data was loaded when calling generated? on #{name}"
```
arfon: I think rescuing and reporting a failure is :+1:

arfon:
:+1: this looks great to me.

bkeepers:
:+1:

test/test_generated.rb
bkeepers: I've heard Ryan Davis (author of minitest) rant about this a few times: http://blog.zenspider.com/blog/2012/01/assert_nothing_tested.html

aroben: This made me realize we should be asserting that `Generated.generated?` returns true too.

PR #1931, opened 2014-12-31
Opt in to container-based Travis CI builds

aroben:
This speeds up our builds by providing lower latency, better caching, and more resources on the build machine. See http://docs.travis-ci.com/user/workers/container-based-infrastructure/.

The only major change here is that we can't install `libicu-dev` globally on the test machine. Instead we extract it to `vendor/debs` and point `charlock_holmes` at it.

Note that the build sometimes fails because apparently the `ace_modes.json` file sometimes can't be downloaded. This would normally cause us to skip the test that relies on that file, but `#skip` doesn't exist in the version of `test-unit` being used on Ruby 2.2. #1930 will fix the `#skip` problem. I doubt that the intermittent failure to download `ace_modes.json` is unique to container-based builds, though I suppose it's possible.

/cc @arfon @bkeepers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ before_install:
   - git fetch origin v2.0.0:v2.0.0
   - git fetch origin test/attributes:test/attributes
   - git fetch origin test/master:test/master
-  - sudo apt-get install libicu-dev -y
+  - script/vendor-deb libicu48 libicu-dev
+  - bundle config build.charlock_holmes --with-icu-include=$(pwd)/vendor/debs/include --with-icu-lib=$(pwd)/vendor/debs/lib
   - git submodule init
   - git submodule sync --quiet
   - script/fast-submodule-update
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,4 @@ notifications:
   disabled: true
 git:
   submodules: false
+cache: bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,5 @@
 sudo: false
-before_install:
-  - git fetch origin master:master
-  - git fetch origin v2.0.0:v2.0.0
-  - git fetch origin test/attributes:test/attributes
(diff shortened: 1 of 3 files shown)


arfon:
> This shaves about 1 minute off each test run (from 2:27 to 1:27).

This seems like a good direction to go it :smile:

aroben:
https://bugs.ruby-lang.org/issues/8074 looks like the fix for the Ruby 2.0 bug. It looks like specifying multiple build options to Bundler is broken in Ruby 2.1 and 2.2. The options end up getting surrounded by a single pair of quotes, which turns them into a single argument to `extconf.rb`.

aroben:
And https://bugs.ruby-lang.org/issues/6207 introduced the Ruby 2.0 bug.

aroben:
I haven't tracked down the Ruby 2.1/2.2 bundler bug yet. I also can't reproduce it locally.

aroben:
Wow, every time I run `bundle install --jobs=3 --path=vendor/gems`, the `BUNDLE_BUILD__CHARLOCK_HOLMES` setting in `.bundle/config` gets surrounded by an extra pair of quotes.

aroben:
https://github.com/bundler/bundler/issues/3261 might be the quote issue. Supposedly it's fixed in bundler 1.7.7 and newer.

aroben:
…but I'm using 1.7.11 locally and can still reproduce the bug. So I guess it isn't really fixed.

aroben:
I believe https://github.com/bundler/bundler/pull/3338 will fix the Bundler bug. Maybe we can use a patched bundler on Travis until that fix is merged.

aroben:
Or I guess we can use `--with-icu-dir` on everything but Ruby 2.0.

aroben:
We're green everywhere.

bkeepers:
:sparkles:

PR #1932, opened 2015-01-01
Update Elm support by adding its own syntax highlighting

deadfoxygrandpa:
Until now, Elm's been using Haskell syntax highlighting. The language has grown farther apart from Haskell in the past months and could really use its own highlighting at this point.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -526,3 +526,6 @@
 [submodule "vendor/grammars/Scalate.tmbundle"]
 	path = vendor/grammars/Scalate.tmbundle
 	url = https://github.com/scalate/Scalate.tmbundle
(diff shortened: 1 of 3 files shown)


arfon:
Looks good. Thanks @deadfoxygrandpa

deadfoxygrandpa:
Thanks for the quick merge.

arfon:
No problem. Btw it will be another couple of weeks before these changes are live as we'll need to cut a new gem

deadfoxygrandpa:
Sure, should be nice when it's live.

aroben:

deadfoxygrandpa:
Nice. Thanks.

PR #1933, opened 2015-01-01
Added CartoCSS to languages.yml

floscher:


arfon:
Thanks @floscher and happy new year! As per [the contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) could you link to some examples of this language in use on GitHub?

floscher:
Here are some examples I know of:
* [gravitystorm/openstreetmap-carto](https://github.com/gravitystorm/openstreetmap-carto) the main map-style of openstreetmap.org
* [mapbox/osm-bright](https://github.com/mapbox/osm-bright) an example-style by [Mapbox](https://mapbox.com)
* [hotosm/HDM-CartoCSS](https://github.com/hotosm/HDM-CartoCSS) style by the Humanitarian OpenStreetMap-team

[mapbox/carto](https://github.com/mapbox/carto) is the repo of the language itself, [TileMill](https://github.com/mapbox/tilemill) and [mapbox/mapbox-studio](https://github.com/mapbox/mapbox-studio) are the repos of editors for this language.

arfon:
There look to be a fair number of these files on GitHub: https://github.com/search?utf8=%E2%9C%93&q=feature+OR+zoom+OR+line+extension%3A.mss&type=Code&ref=searchresults

lib/linguist/languages.yml
```diff
@@ -449,6 +449,14 @@ Cap'n Proto:
   - .capnp
   ace_mode: text
 
+CartoCSS:
+  aliases:
+  - Carto
+  extensions:
+  - .mss
+  ace_mode: text
```
arfon: Would it make sense to use `CSS` here?

floscher: I've [tested highlighting of the sample file](https://raw.githubusercontent.com/floscher/openstreetmap-carto/master/addressing.mss) and it looks ok with `CSS`, except for some things:
* //-style comments are not recognized (these are not allowed in CSS but in CartoCSS)
* Strings are not recognized if these are contained in brackets []
* When nesting blocks (marked with curly braces {}), only the first inner block gets marked correctly (see example below)

In this example the word `zoom` and the first `width` are highlighted blue, the other `width` is not:
```
[zoom >= 10] {
  [width <= 2] {
    …
  }
  [width > 2]{
    …
  }
}
```

I can't judge, if these shortcomings are minor enough that setting this to `CSS` would be ok, or if it should be set to `text` until a TextMate-grammar is available.

arfon: OK thanks for checking @floscher. Let's wait until TextMate-grammar is available - do you know of any in development?

floscher: I only know about a [grammar for Atom](https://github.com/yohanboniface/carto-atom), that [looks pretty good](https://lightshow.githubapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fyohanboniface%2Fcarto-atom%2Fmaster%2Fgrammars%2Fcarto.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fgravitystorm%2Fopenstreetmap-carto%2Fmaster%2Flandcover.mss&code=).

arfon:
:+1: thanks @floscher

arfon:
It will be another couple of weeks before `CartoCSS` starts to show up on GitHub as we'll need to cut a new gem.

lib/linguist/languages.yml
arfon: Oh cool, we could use that. Can you take a look at the section on [adding a new grammar](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#syntax-highlighting-looks-wrong) and open a new pull request? We'll also need https://github.com/yohanboniface/carto-atom to add a license to their repo before we can add it.

arfon: ....update I've opened an issue here: https://github.com/yohanboniface/carto-atom/issues/1

aroben:

yohanboniface:
I'm being asked to [rename the repository](https://github.com/yohanboniface/carto-atom/issues/2), any risk that this will break anything here?

floscher:
[This line](https://github.com/github/linguist/blob/master/.gitmodules#L532) should be changed accordingly, apart from that, everything should be fine.

yohanboniface:
Ah, good to know, thanks :)
I guess github will redirect if the submodule is not changed while the repository has already the new name. So better to first rename the repo, and then make a PR for changing the submodule reference :)

floscher:
Yes, it seems that github [creates redirects](https://github.com/blog/1508-repository-redirects-are-here), so changing the name first would be better. But even if it didn't, renaming would only break things for people trying to clone this repo with submodules. Those with existing clones would not be bothered.

pchaigno:
@yohanboniface That said, @mnquintana only asked if you could rename the package, not the repository ;)

yohanboniface:
Oh, right, my question was biased! :s
Thanks for making that clear, @pchaigno! :)

@floscher so any issue if the **package** name change instead?

mnquintana:
Yup @yohanboniface, you definitely don't need to rename the repository (I just meant the package name).

floscher:
@yohanboniface renaming only the package should not influence this repository, as linguist simply [uses all .json or .cson files from the grammars/ directory](https://github.com/github/linguist/blob/a0d5a8338bacaa925c9e0215b4f8b2bbcd1115d1/script/convert-grammars#L47) and doesn't care about the rest of you repository :).
As long as the name of the name of repository and [scope](https://github.com/github/linguist/blob/master/grammars.yml#L194) stay the same, all should be well.

yohanboniface:
Great, thanks for your lights :)

PR #1935, opened 2015-01-02
Fix issue with github/linguist#1932

floscher:
#1932 introduced an error that prevented [download-grammars](https://github.com/github/linguist/blob/master/script/download-grammars) from working. The script issued an error message saying that `vendor/grammars/Elm.tmLanguage` is not a file (what is true because it is a directory).

The reason for this is:
Submodules must not end with neither `.tmlanguage` nor `.plist`. Otherwise the `download-grammars`-script will interpret the submodule as a `SingleFile` instead of a `DirectoryPackage`, which will lead to an error message.

The relevant line in `download-grammars` is https://github.com/github/linguist/blob/f2ab426d38dc391a6b0612cecf29cb67efa931b4/script/download-grammars#L153

### DIFF
--- a/script/download-grammars
+++ b/script/download-grammars
@@ -150,7 +150,7 @@ end
 
 def install_grammar(tmp_dir, source, all_scopes)
   is_url = source.start_with?("http:", "https:")
-  is_single_file = source.end_with?('.tmLanguage', '.plist')
+  is_single_file = source.downcase.end_with?('.tmlanguage', '.plist') && !File.directory?(source)
 
   p = if !is_url
         if is_single_file


arfon:
cc @aroben

aroben:
> Submodules must not end with neither `.tmlanguage` nor `.plist`. Otherwise the `download-grammars`-script will interpret the submodule as a `SingleFile` instead of a `DirectoryPackage`, which will lead to an error message.

Thanks for including this explanation. I think a better fix would be to ask the filesystem whether the local path is a file or directory rather than deciding based on the filename extension. A `File.directory?` check should be sufficient.

floscher:
Sure, that would be better. My solution was more like a quick and dirty solution to just make it work again. I'll look into it tomorrow and hopefully come up with a better solution.

Thanks for the tip. I'm not very familiar with Ruby, but let's learn by doing ;-).

floscher:
I've now changed the download-script instead of the name of the Elm-grammar.

But simply changing `source.end_with?('.tmLanguage', '.plist')` to `File.directory?(source)` was not sufficient, because e.g. `.tar.gz`-files produce `is_single_file=false`, so I simply combined both criteria.

script/download-grammars
```diff
@@ -150,7 +150,7 @@ end
 
 def install_grammar(tmp_dir, source, all_scopes)
   is_url = source.start_with?("http:", "https:")
-  is_single_file = source.end_with?('.tmLanguage', '.plist')
+  is_single_file = source.downcase.end_with?('.tmlanguage', '.plist') && !File.directory?(source)
```
aroben: I think it would be better to get rid of this variable entirely and split up the two checks. Something like:

```ruby
if !is_url
  if File.directory?(source)
    DirectoryPackage.new(source)
  else
    SingleFile.new(source)
  end
elsif source.downcase.end_with?('.tmlanguage', '.plist')
  SingleGrammar.new(source)
# etc
```

That way we never call `File.directory?` on a URL, and we don't care at all about the local path's filename extension. What do you think?

aroben:
I ran into and fixed this problem over in #1909. Thanks for your work here @floscher; I would have been really confused had you not already diagnosed the problem!

PR #1936, opened 2015-01-02
Rename Nimrod to Nim

adelq:
Nimrod has been renamed to Nim as of 4 days ago in the latest release.
Source: http://nim-lang.org/news.html#Z2014-12-29-version-0-10-2-released


arfon:
Thanks @adelq. It looks like the tests are failing as the Textmate grammar still uses the old name. I think just manually setting the `tm_scope` here should fix this.

pchaigno:
@arfon Will this be possible with Solr (it wasn't possible in #995 with Fortran)?

arfon:
> @arfon Will this be possible with Solr (it wasn't possible in #995 with Fortran)?

Oh yeah. Good call. I'm not sure what will happen here. I'll check in with the search team again for an update.

adelq:
Alright, I've added the scope so tests should pass now, and keep me updated on the feasibility of switching the name over.

aroben:
The scope is actually now called `source.nim`. If you rebase against master you shouldn't need to change `languages.yml` at all anymore.

pchaigno:
> Oh yeah. Good call. I'm not sure what will happen here. I'll check in with the search team again for an update.

@arfon Any news?

arfon:
> @arfon Any news?

Not yet sorry.

dom96:
what's the hold up on this PR?

arfon:
> what's the hold up on this PR?

Please see the reason here: https://github.com/github/linguist/pull/995#issuecomment-46665150

Basically we still don't have a good way of renaming languages without rebuilding our whole code search index.

arfon:
See comment [here](https://github.com/github/linguist/pull/2034#issuecomment-71346028) for background but basically we don't have a way to address this yet. We're working on a new code search index which should make this PR possible and so I've labelled this 'Language rename' for future discoverability. Sorry we don't have better solution right now :-\

dom96:
@arfon any news about this?

arfon:
I'm afraid not 😞. The best thing you can do right now is email support@github.com to ask for this.

PR #1937, opened 2015-01-02
Update languages.yml for Hack syntax-highlighting

arfon:
Setting correct tm_scope: text.html.php


aroben:

PR #1938, opened 2015-01-02
Add grammar for CartoCSS

floscher:
This pull request has to wait, until yohanboniface/carto-atom#1 is resolved. But I'm optimistic, that the owner will choose a suitable license. He [has noted in the package.json](https://github.com/yohanboniface/carto-atom/blob/master/package.json#L5) that he intended to release it under WTFPL.

In order to get this work, I had to change the `download-grammars`-script to accept files that end with `.json` instead of `.cson`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -531,3 +531,6 @@
 [submodule "vendor/grammars/sass-textmate-bundle"]
 	path = vendor/grammars/sass-textmate-bundle
 	url = https://github.com/nathos/sass-textmate-bundle
+[submodule "vendor/grammars/carto-atom"]
+	path = vendor/grammars/carto-atom
+	url = https://github.com/yohanboniface/carto-atom
--- a/grammars.yml
+++ b/grammars.yml
@@ -136,6 +136,8 @@ vendor/grammars/c.tmbundle:
 - source.c.platform
 vendor/grammars/capnproto.tmbundle:
 - source.capnp
+vendor/grammars/carto-atom:
+- source.css.mss
 vendor/grammars/ceylon-sublimetext:
 - module.ceylon
 - source.ceylon
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -450,12 +450,13 @@ Cap'n Proto:
   ace_mode: text
 
 CartoCSS:
+  type: programming
   aliases:
   - Carto
   extensions:
   - .mss
   ace_mode: text
-  tm_scope: none
+  tm_scope: source.css.mss
 
 Ceylon:
   type: programming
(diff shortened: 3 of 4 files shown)


arfon:
@aroben is [this](https://github.com/yohanboniface/carto-atom/blob/master/package.json#L5) good enough for our purposes?

arfon:
OK we're not waiting on the license any longer but I'd like @aroben to :+1: the changes to the download script.

script/download-grammars
```diff
@@ -139,7 +139,7 @@ def load_grammar(path)
   case File.extname(path.downcase)
   when '.plist', '.tmlanguage'
     Plist::parse_xml(path)
-  when '.cson'
+  when '.cson', '.json'
     cson = `"#{CSONC}" "#{path}"`
```
aroben: There's no need to pass `.json` files through `csonc`. It would be slightly nicer to just go directly to `JSON.parse` in this case.

aroben:
The changes look good but I have one small suggestion above.

script/download-grammars
floscher: Thanks for the advice. I'll look into it tomorrow.

arfon:
@floscher - would you mind rebasing against master too? It doesn't look like we can get a clean merge at this point.

floscher:
Done.
I've rebased against master and fixed the issue that json-files were passed through csonc.

arfon:
Looks good to me :+1:. Let's wait for a final :+1: from @aroben before merging.

script/download-grammars
```diff
@@ -143,6 +143,8 @@ def load_grammar(path)
...
     raise "Failed to convert CSON grammar '#{path}': #{$?.to_s}" unless $?.success?
     JSON.parse(cson)
+  when '.json'
+    JSON.parse(File.open(path, 'rb').read)
```
aroben: This could be simplified to `JSON.parse(File.read(path))` I believe.

aroben:
This looks great! One more teensy suggestion above and then I think we're good.

floscher:
Thank you for the suggestion. I've incorporated the change into the pull request and rebased again on 8430f694e5.

Please note that the script currently doesn't run through. But this is another issue for which I opened #1958.

floscher:
You're welcome.
Is there a way to rerun Travis on the above commits? The build should pass now, as #1959 is merged.

aroben:
http://docs.travis-ci.com/user/pull-requests/ makes it sound like the only way is to update the PR by pushing again.

floscher:
Thanks for the hint! I think I'll simply rebase.

aroben:
Awesome! Thanks so much @floscher.

aroben:

floscher:
> This is now live on github.com.

Wow, that was quick!
Thank you @arfon and @aroben for the guidance concerning this and the related pull requests and for merging those so fast.

PR #1940, opened 2015-01-02
Fix for extensions with multiple segments

bkeepers:
Fixes #1731 by cherry-picking commit from @larsbrinkhoff in https://github.com/github/linguist/pull/1830.


aroben:

Issue #1568, opened 2014-10-01
`.in` extension does not imply CMake syntax


roman-neuhauser:
it seems that https://github.com/github/linguist/commit/c8bc0a5c792b39b378a6bc743ad45949ea18f13b causes any file with a name ending in `.in` to be treated as CMake input.

`$whatever.in` is a traditional naming pattern for files meant to be preprocessed in some way.  half of my repos in github have files ending in `.spec.in`, *among others*.  i'd expect there to be a lot of `Makefile.in` files in github.  etc, etc.

some of my repositories with `.in` files:

* https://github.com/roman-neuhauser/bs-update
* https://github.com/roman-neuhauser/pastish
* https://github.com/roman-neuhauser/printscreen

arfon:
@pchaigno - do you know if this is this still an issue?

pchaigno:
Yes and no. We now support segmented file extensions and have a couple defined for `.in` (`.rs.in` and `.cmake.in`). We also improved the recognition of files based on their filename solely. However, we still don't recognize the `.spec.in` file extension.

> files ending in `.spec.in`, among others.

@roman-neuhauser What are these files? What other extensions are you thinking of?

pchaigno:
Flagging this as stale :flags:

tripleee:
Sorry for being late to reply. I don't know exactly what Roman was alluding to, but using `*.in` as a template for generating a build artefact is fairly universal practice, and this extension by itself should not be taken as more than possibly a weak hint for the file's language, type, or purpose. `Makefile.in` from Autotools is of course ubiquitous; quick googling also brought up the following;

* https://gist.github.com/telent/4118983 apparently builds an RPM `.spec` file from a template where the version etc is a hard-coded static token to be replaced dynamically at build time
* https://github.com/liip/mahararpmpkg/blob/master/Makefile does something fairly similar

roman-neuhauser:
exactly what @tripleee said.  *filename.in* is a ubiquitous pattern used to signal that the file is a main *in*put to generate *filename*, regardless of the actual syntax of the input, output, or the extent of processing.

you'll notice that `configure.in` (processed by `autoconf`), `Makefile.in` (processed by `configure`) have vastly different syntaxes, what they do have in common is that they're the main inputs used to produce `configure` and `Makefile` respectively.

@pchaigno wrote

> > files ending in `.spec.in`, among others.
>
> What are these files?

[...]

PR #1944, opened 2015-01-02
Revert #1438: add file extensions with multiple segments

pchaigno:
This pull request reverts #1438.
In #1438 I removed all file extensions with multiple segments.
Linguist now support them so we can add them back :zap: 

It should also solve #1568.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1198,7 +1198,6 @@ HTML+ERB:
   extensions:
   - .erb
   - .erb.deface
-  - .html.erb.deface
   ace_mode: html_ruby
 
 HTML+PHP:
@@ -1230,7 +1229,6 @@ Haml:
   extensions:
   - .haml
   - .haml.deface
-  - .html.haml.deface
   ace_mode: haml
 
 Handlebars:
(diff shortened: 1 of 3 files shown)


arfon:
Nice! For good housekeeping do you think we should add an example file for each extension @pchaigno ?

pchaigno:
I added the two samples (`.cmake.in` and `.html.hl`) that were missing (I already added some in #1438).

I noticed that `.html.hl` files can be a mix of HTML and Clojure. Should they really be recognized as HTML? @cldwalker may have more information on these files.

cldwalker:
@pchaigno They can be a mix of html and clojure. I chose html at the time because I didn't know if linguist supported syntax highlighting multiple languages for one file.

pchaigno:
@cldwalker I think it's possible if a grammar for `HTML+Clojure` exists. Do you know one?

cldwalker:
I'm not aware of a .html.hl [textmate grammar](https://github.com/github/linguist#syntax-highlighting). @alandipert Do you know of one? If not, it's currently set to render as html unless you'd like for it to stay as clojure

larsbrinkhoff:
Cherry-picked the CMake sample for #1978.

arfon:
This looks good to me. @bkeepers?

bkeepers:
:+1:

arfon:
Thanks @pchaigno

PR #1948, opened 2015-01-04
Add JSONiq & XQuery scope from Atom package

wcandillon:
I have built a XQuery and JSONiq Atom package: https://atom.io/packages/language-jsoniq
I would like the Javascript lexer to be used for XQuery and JSONiq syntax highlighting in github.

However, the highlighting is done programmatically, not by using a grammar file. Would that be a problem?
Please let me know what I can do to contribute better XQuery and JSONiq syntax highlighting to github.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1412,7 +1412,7 @@ JSONiq:
   ace_mode: jsoniq
   extensions:
   - .jq
-  tm_scope: source.xquery
+  tm_scope: source.jq
 
 Jade:
   group: HTML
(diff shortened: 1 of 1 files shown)


aroben:
Thanks for submitting this @wcandillon!

> However, the highlighting is done programmatically, not by using a grammar file. Would that be a problem?

This won't work with github.com's syntax highlighter. It only understands TextMate/Sublime/Atom language grammars. It can't execute your Atom module's code. So to get these languages highlighted on github.com you'll need to write grammars for them (or find existing ones).

Issue #1862, opened 2014-12-11
HTTP highlighting broken on github.com


abalmos:
It appears that HTTP highlighting is broken on github.com. Rather then creating an issue github support sent me here to report the bug.

For example:

**Request**
```http
GET /a/b/c HTTP/1.1
Host: cloud.com
Accept: application/json
```

**Response**
```http
HTTP/1.1 200 OK
Content-Type: application/json

{
  "a": "A",
  "b": {
    "c": "C"
  }
}
```

More examples of the issue can be found here: https://github.com/OADA/oada-docs/blob/master/rest-specs/REST-API-Examples.md. That page, and the others in that repo, used to be colorful.

languages.yml (https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1194) seem to indicate HTTP highlighting is still supported.

pchaigno:
@abalmos It looks good in Lightshow so I opened a pull request (#1952) to use this grammar.

arfon:
This will be addressed in https://github.com/github/linguist/pull/2011 which should go live very soon.

PR #1952, opened 2015-01-06
Grammar for HTTP

pchaigno:
This pull request adds [a grammar](https://github.com/httpspec/sublime-highlighting) for HTTP as discussed in #1862.


aroben:
Thanks! It'll be nice to have this working again.

Looks like the grammar has no license. Could you see about getting one added?

pchaigno:
Woops, completely forgot that in the two grammars I just added :/
Thanks for pointing it out @aroben!

Issue for the license opened at httpspec/sublime-highlighting#1

PR #1954, opened 2015-01-06
Test that all grammar submodules have an appropriate license

aroben:
We require all grammars to have a license that permits redistribution. A few existing grammars have been grandfathered in without a license (and we're working to remedy that), but no new grammars should be added without a license.

/cc @bkeepers @arfon @pchaigno @vmg

### DIFF
--- a/test/test_grammars.rb
+++ b/test/test_grammars.rb
@@ -14,12 +14,11 @@ class TestGrammars < Minitest::Test
   end
 
   def test_submodules_are_in_sync
-    submodules = `git config --list --file "#{File.join(ROOT, ".gitmodules")}"`.lines.grep(/\.path=/).map { |line| line.chomp.split("=", 2).last }
     # Strip off paths inside the submodule so that just the submodule path remains.
     listed_submodules = @grammars.keys.grep(/vendor\/grammars/).map { |source| source[%r{vendor/grammars/[^/]+}] }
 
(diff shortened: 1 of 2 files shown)


test/test_grammars.rb
```diff
@@ -49,4 +56,79 @@ def test_local_scopes_are_in_sync
...
+      "CC"
+    elsif content.include?("tidy-license.txt") || content.include?("If not otherwise specified (see below)")
+      "textmate"
+    elsif content =~ /^\s*[*-]\s+Redistribution/ || content.include?("Redistributions of source code")
+      "BSD"
+    elsif content.include?("Permission is hereby granted") || content =~ /\bMIT\b/
+      "MIT"
+    elsif content.include?("unlicense.org")
+      "unlicense"
+    end
```
bkeepers: Might be worth using https://github.com/benbalter/licensee for this.

aroben: Hm, that's a good idea. I'll experiment with it.

aroben: Looks like licensee doesn't look for licenses in READMEs, which is where many of our submodules put their license. https://github.com/benbalter/licensee#what-about-looking-to-see-if-the-author-said-something-in-the-readme discusses this a bit. For our purposes I think having a license in the README is fine, so I guess I'll just stick with the current logic.

PR #1955, opened 2015-01-06
Reclassify three samples files generated from Zephir code

pchaigno:
This pull request changes the language for three samples added in #1132.
I moved them from `Zephir/` to `C/` and `PHP/`.
The three files are generated from Zephir code but are actually C and PHP code.

I wonder if we could have written a test to detect this.
/cc @arfon @bkeepers


arfon:
> I wonder if we could have written a test to detect this. /cc @arfon @bkeepers

What about running all of the samples in the `filenames` directories through the classifier and warning if they don't match the parent directory language?

pchaigno:
Samples in the `filename` directories already [go through the detection process](https://github.com/github/linguist/blob/1f1ffcbfa678a3e362cb86436257c6066a4e2825/test/test_blob.rb#L463-L467) and through the classifier if they don't have a shebang and don't match any heuristic (as it was the case here). But since they are used to train the classifier, it almost always detect them as being written in the language they are classified.

bkeepers:
:+1: 

I can't think of a good way to test this.

PR #1953, opened 2015-01-06
Update all grammar submodules

aroben:
This was performed via `git submodule update --remote`.

```
* vendor/grammars/Agda.tmbundle 784f435...68a218c (1):
  > Merge pull request #1 from aroben/patch-1

* vendor/grammars/IDL-Syntax 9473b7f...3baeaea (1):
  > Merge pull request #3 from aroben/patch-1

* vendor/grammars/NimLime 7a2fb4e...9cef4b6 (4):
  > Updated command names
  > Updated generated documentation
  > renamed more from nimrod to nim
  > Renamed several files

* vendor/grammars/SCSS.tmbundle d6188e5...4147502 (1):
  > Merge pull request #181 from redgluten/master

* vendor/grammars/Sublime-VimL 6ab7e19...366fdc6 (1):
  > Merge pull request #2 from yous/fix-single-quoted-string

* vendor/grammars/factor 2dc5590...2453a78 (38):
  > json.writer: make sure we make hex values two digits zero padded.
  > json.writer: support escaping unicode > 0x10000.  Thanks @jonenst!
  > mason.git: fix status check.
  > compiler.cfg.*: new unit test vocabs
  > compiler.cfg.*: more docs
  > compiler.cfg.*: refactoring away the compute-global-sets word
  > compiler.cfg.*: docs and more tests
  > compiler.cfg.stacks.local: refactoring making stack-changes and height-changes take and return stuff instead of using variables and the make building
  > compiler.cfg.parallel-copy: docs
  > compiler.cfg.stacks.height: these words are unused
  > compiler.cfg.*: more compiler docs
  > compiler.cfg.stacks.finalize: initial docs
  > io.launcher: fix stack effects.
  > io.launcher: fix docs for with-process-reader and with-process-writer.
  > io.launcher: add versions of with-process that preserve process and status.
  > mason.git: fix use.
  > mason.git: fix for rename.
  > io.launcher: cleanup public interface, make some things private or internal.
  > gopher: set 1 minute timeout by default.
  > brainfuck: cleanup tests.
  > json.writer: don't escape spaces, thats weird.
  > unix: some using cleanups.
  > python: rename startup/shutdown hooks.
  > math.extras: adding the Möbius function.
  > alien.c-types: move definitions of stdint.h from unix.types.
  > gopher: use contents now that it works.
  > io.ports: Make buffered-port not have a length because of Linux virtual files and TCP sockets. Related to issues #1256 and #1259.
[...]

PR #1963, opened 2015-01-07
Recommend updating grammars when releasing

aroben:
This will help ensure we keep pulling in fixes from the grammar repos.

/cc @bkeepers https://github.com/github/linguist/pull/1953#issuecomment-69073698


arfon:
:+1:

PR #1965, opened 2015-01-07
Merge github-linguist-grammars into github-linguist

aroben:
Now that all our grammars are licensed (or grandfathered in), we can distribute them as part of the standard github-linguist gem. This makes it easier for projects to get up and running with Linguist.

/cc @bkeepers @arfon @vmg


bkeepers:
:+1:

arfon:
Awesome! :sparkles:

aroben:
I suggest we bump the version to 4.3.0 after this because the github-linguist gem will now include the `Linguist::Grammars.path` API (which was previously only present in github-linguist-grammars).

arfon:
> I suggest we bump the version to 4.3.0 

Agreed.

aroben:
I just added some text to `LICENSE` to point folks to the licenses for each grammar. Does it look OK to you guys?

talniv:
Agreed with @arfon: Maintaining the standard is important, and the license file is not currently where people would think to look for the info. Can we put it in a separate file and call it something short and indicative of the content? (If it's good enough maybe it will be the start of creating a new standard for that info).

aroben:
> the license file is not currently where people would think to look for the info.

I was following the convention set in files like https://github.com/atom/language-shellscript/blob/9839719721e3fb67c2df8461b2b296e6ff027e7f/LICENSE.md, which includes both the license itself and other license-related information. My thinking was that if someone is wondering "what is the license that covers these files?", they'd look in the LICENSE file included in our gem. But I'm happy to move the information somewhere else if you think that people are not likely to look there. Any suggestions?

arfon:
@talniv - any thoughts on @aroben's last comment?

arfon:
@mlinksva is this something you could help us with here?

mlinksva:
@arfon if the only issue is where to note licenses for grammars, an alternative which leaves LICENSE unmodified and detectable with licensee is to note in README.md, suggested in 5b72b4d

arfon:
> @arfon if the only issue is where to note licenses for grammars, an alternative which leaves LICENSE unmodified and detectable with licensee is to note in README.md, suggested in 5b72b4d

This is great. Thanks @mlinksva

aroben:
🎉

vmg:
Yey! One less gem in our lives!

PR #1969, opened 2015-01-08
Heuristic for CMake

### LINKED ISSUE #1961: Particular CMakeLists.txt file not detected as CMake
https://github.com/serge-v/ctests/blob/fbce433d9fa09465a2b883a1e320a5e2c242ba97/departures/CMakeLists.txt is getting detected as `Text`, rather than `CMake`:

```console
$ bundle exec linguist CMakeLists.txt 
CMakeLists.txt: 29 lines (20 sloc)
  type:      Text
  mime type: text/plain
  language:  Text
```

/cc @serge-v

larsbrinkhoff:
Fixes #1961.


aroben:
Thanks for getting this started @larsbrinkhoff.

I really wish we could just say:

```ruby
if name.downcase == "cmakelists.txt"
  Language["CMake"]
else
  Language["Text"]
end
```

It seems silly that we have to look at the blob data at all (either here or in the classifier) when the situation is so clear-cut. But I'm pretty new to Linguist's detection logic, so maybe this is a bad idea? /cc @bkeepers

larsbrinkhoff:
Agreed, but this is the best we can do with the current framework.  As far as I know.

On the other hand, there is a potential advantage to look at the file content and ignore the name.  Files are named with great creativity, i.e. are not always a reliable source of information.

PR #1970, opened 2015-01-08
Add missing ^ in regexp for Frege heuristic

larsbrinkhoff:
Forgot to match keywords only at the beginning of a line.  This caused some false positives.


aroben:
Is it possible to add tests that show the false-positives are gone?

larsbrinkhoff:
@aroben Yes, I added a new line to an existing sample file.  Before my change, it's detected as Frege, after as Text.

aroben:
Great, thanks!

PR #1976, opened 2015-01-08
Use path for Generated?

arfon:
/ cc @aroben

### DIFF
--- a/lib/linguist/lazy_blob.rb
+++ b/lib/linguist/lazy_blob.rb
@@ -1,7 +1,6 @@
 require 'linguist/blob_helper'
 require 'linguist/language'
 require 'rugged'
-require 'pry'
 
 module Linguist
   class LazyBlob
--- a/lib/linguist/lazy_blob.rb
+++ b/lib/linguist/lazy_blob.rb
@@ -24,6 +24,10 @@ module Linguist
       @mode = mode
     end
 
+    def path
+      name
+    end
+
     def git_attributes
       @git_attributes ||= repository.fetch_attributes(
         name, GIT_ATTR, GIT_ATTR_FLAGS)
--- a/lib/linguist/lazy_blob.rb
+++ b/lib/linguist/lazy_blob.rb
@@ -1,6 +1,7 @@
 require 'linguist/blob_helper'
 require 'linguist/language'
 require 'rugged'
+require 'pry'
 
 module Linguist
   class LazyBlob
@@ -14,18 +15,16 @@ module Linguist
 
     attr_reader :repository
     attr_reader :oid
-    attr_reader :name
+    attr_reader :path
     attr_reader :mode
 
-    def initialize(repo, oid, name, mode = nil)
+    alias :name :path
+
+    def initialize(repo, oid, path, mode = nil)
       @repository = repo
       @oid = oid
-      @name = name
+      @path = path
       @mode = mode
-    end
-
-    def path
-      name
     end
 
     def git_attributes
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -301,7 +301,7 @@ module Linguist
     #
     # Return true or false
     def generated?
-      @_generated ||= Generated.generated?(name, lambda { data })
+      @_generated ||= Generated.generated?(path, lambda { data })
     end
 
     # Public: Detects the Language of the blob.
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -99,7 +99,7 @@ module Linguist
       elsif name.nil?
         "attachment"
       else
-        "attachment; filename=#{EscapeUtils.escape_url(File.basename(name))}"
+        "attachment; filename=#{EscapeUtils.escape_url(name)}"
       end
     end
 
@@ -233,7 +233,7 @@ module Linguist
     #
     # Return true or false
     def vendored?
-      name =~ VendoredRegexp ? true : false
+      path =~ VendoredRegexp ? true : false
     end
 
     # Public: Get each line of data
--- a/lib/linguist/file_blob.rb
+++ b/lib/linguist/file_blob.rb
@@ -32,6 +32,17 @@ module Linguist
     # Returns a String
     attr_reader :name
 
+    # Public: Filename with Path
+    #
+    # Examples
+    #
+    #   FileBlob.new("/path/to/linguist/lib/linguist.rb").path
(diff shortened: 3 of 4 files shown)


lib/linguist/lazy_blob.rb
```diff
@@ -24,6 +24,10 @@ def initialize(repo, oid, name, mode = nil)
       @mode = mode
     end
 
+    def path
+      name
+    end
```
aroben: This seems OK, but I wonder if there's a way to get the full path here?

aroben:
Should be possible to write a test for this I think?

lib/linguist/lazy_blob.rb
arfon: Good point. Will check.

arfon: Actually it looks like `LazyBlob` returns the filename with the path already:

```
#<Linguist::LazyBlob:0x007f9f0e4962e0
 @git_attributes={"linguist-language"=>nil, "linguist-vendored"=>nil},
 @mode="100644",
 @name="contrib/hk-zsh-completion.sh",
 @oid="1c1d218d1c6f176e1f6b1874d183b5ca191a72d3",
 @repository=#<Rugged::Repository:70160548389160 {path: "/Users/arfon/hk/.git/"}>>
```

aroben: Hm. I wonder if we should rename `@name` to `@path` then, and if we should make `#name` just return the baseman.

aroben: I don't know where LazyBlob is used so I don't know what effect this would have.

arfon:
I've renamed `name` to `path` on `LazyBlob` and just left one aliased to the other for now.

> Should be possible to write a test for this I think?

Turns out, most of the tests we write here already pass the path (which wasn't what we were doing on .com): https://github.com/github/linguist/blob/master/test/test_generated.rb#L44-L45

The changes to `FileBlob` are just adding an `attr_reader` to an already instantiated ivar.

I'd like to test this but I'm wondered what exactly to test! Perhaps how `name` and `path` can differ depending on whether the base path is passed?

aroben:
I guess this is only testable if `#name` and `#path` return different things. Right now both FileBlob and LazyBlob return the same string from each. Should we change that (which would match dotcom)?

arfon:
> Should we change that (which would match dotcom)?

Perhaps but that makes me somewhat nervous. I tried modifying them both to return different results for `name` and `path` but ended up in a right mess with all the existing [vendor](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml) patterns that are defined where the regexes anchor to the start of the line.

aroben:
Looks like the vendor patterns are meant to be tested against the full path, so can we change `BlobHelper#vendored?` to use `path` instead of `name`?

arfon:
> Looks like the vendor patterns are meant to be tested against the full path, so can we change BlobHelper#vendored? to use `path` instead of `name`?

We can do that, but it will mean modifying any regexes [that start](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml#L20-L21) with `^` to `(^|/)`. What do you think?

aroben:
I assume rules that start with `^` are only meant to match paths at the root of the repository. I don't see why we'd need to change them. Am I misunderstanding?

arfon:
> I assume rules that start with `^` are only meant to match paths at the root of the repository. I don't see why we'd need to change them. Am I misunderstanding?

It's just that if we switch to using the path for the vendor matches then `path` is returning the full path to the file (e.g. /Users/arfon/github/linguist...).

aroben:
> path is returning the full path to the file

Oh, that's a problem. I believe `path` is only supposed to return the path relative to the repository root. That's certainly what the behavior is in github.com's codebase.

arfon:
> That's certainly what the behavior is in github.com's codebase.

Yeah, that makes sense. Rails presumably helps us with that right? (Everything being relative to `Rails.root`)

Do you know if there's a convention for addressing this kind of thing in a Gem? / cc @bkeepers incase he has any insights.

aroben:
> Yeah, that makes sense. Rails presumably helps us with that right? (Everything being relative to Rails.root)

In github.com we're dealing with files in a git repository (and usually we're not even dealing with files that exist uncompressed on disk), so we're just feeding the relative paths from git straight into Linguist. Rails doesn't really come into play in this case.

> Do you know if there's a convention for addressing this kind of thing in a Gem?

Well you just have to decide what path you want to be relative to, then use [`Pathname#relative_path_from`](http://www.rubydoc.info/stdlib/pathname/Pathname#relative_path_from-instance_method) to convert the absolute path to a relative one.

I'm not sure any of that is required for this PR, though:

Until now, `FileBlob#name` has returned an absolute path. In this PR you're essentially renaming `#name` to `#path`, and adding a new `#name` method that returns just the file's basename. Places that really want a path (like the `vendored?` and `generated?` code) will be changed to use `#path`, while those that want just a basename can use `#name` (do we have any code that wants just a basename?).

So assuming you update all the appropriate places to use `#path` instead of `#name`, this PR won't introduce any change in behavior from this gem's perspective. The only behavior change will be for classes that include `BlobHelper` and had a `#name` method that was not returning a path (like in github.com's codebase).

I agree it would be nice to make `FileBlob` smarter about using relative paths. Maybe it could be enhanced to look for an enclosing git repository and make the path relative to that. But I don't think it must be done as part of this PR.

arfon:
@aroben - When you get a chance I'd love a review of this.

aroben:
Would the tests break if we always did `enforce_relative=true`?

arfon:
> Would the tests break if we always did `enforce_relative=true`?

Yes, tests these ones fail: https://github.com/github/linguist/blob/path-for-fileblob/test/test_blob.rb#L85-L86

aroben:
Gotcha. Looks like the problem is that `FileBlob` uses the relative path to perform filesystem access, which of course only works if the path is relative to the current working directory.

Here's one solution: https://github.com/github/linguist/compare/f4c1cc576b1bf0a057213ca0bb333fa6de3afdf0...96c7bc30d86d99af5f29c23dc84f3ecc0d5e7841

What do you think about that approach? It introduces yet another API change, but since we're already changing API in this PR maybe that's OK.

arfon:
> What do you think about that approach? It introduces yet another API change, but since we're already changing API in this PR maybe that's OK.

Yeah, that's much nicer. I wasn't very happy with all those additions to the specs.

aroben:
This looks good to me!

PR #1951, opened 2015-01-06
Grammar for AutoHotkey

pchaigno:
This pull request adds [a grammar](https://github.com/robertcollier4/AutoHotkey) for AutoHotkey as discussed in #1898.

PR #1977, opened 2015-01-08
Auto

arfon:
Replaces https://github.com/github/linguist/pull/1951

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -532,3 +532,6 @@
 [submodule "vendor/grammars/sublime-bsv"]
 	path = vendor/grammars/sublime-bsv
 	url = https://github.com/thotypous/sublime-bsv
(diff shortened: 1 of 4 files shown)


arfon:
@aroben - [this line](https://github.com/github/linguist/blob/master/test/test_grammars.rb#L63) is failing to find the `UNLICENSE.txt` file in https://github.com/robertcollier4/AutoHotkey even though we [later check](https://github.com/github/linguist/blob/master/test/test_grammars.rb#L130) to identify the contents of this kind of license file.

Shall we just adapt the regex to include this type of filename?

```ruby 
license = files.find { |path| File.basename(path) =~ /\b(un)?licen[cs]e\b/i } || files.find { |path| File.basename(path) =~ /\bcopying\b/i }
```

aroben:
That sounds good to me.

PR #1978, opened 2015-01-09
More CMake samples

### LINKED ISSUE #1961: Particular CMakeLists.txt file not detected as CMake

larsbrinkhoff:
Fixes #1961 by adding six more CMake samples.


larsbrinkhoff:
Cherry-picked @pchaigno's sample from #1944.

aroben:
Should any of the new samples have a `.txt` extension? (Does that even matter?)

larsbrinkhoff:
I don't think it matters to the Bayesian classifier.

If it were needed, we also need to add `.txt` as a CMake extension.  But that would seem wrong.  I believe only the `CMakeLists.txt` file use the `.txt` extension, and the other CMake files conventionally use `.cmake`.

bkeepers:
Sweet. Did you confirm that this fixes it?

larsbrinkhoff:
@bkeepers Yes.

bkeepers:
Awesome. Looks good to me! :+1:

PR #1979, opened 2015-01-09
Create xojo.JSON-tmLanguage

ArturViader:
Xojo is very similar to vbnet. Codehighlight works with ```vbnet. But when you add a WHILE WEND bucle on the code highlight disappears.
This is for try fix this.

### DIFF
--- a/vendor/grammars/xojo.JSON-tmLanguage
+++ b/vendor/grammars/xojo.JSON-tmLanguage
@@ -1,0 +1,217 @@
+: "XOJO",
+  "scopeName": "source.xojo",
+  "fileTypes": ["vb"],
+  "patterns": [
+  	{ "include": "#singleLineComment" },
+  	{ "include": "#region" },
+	{ "include": "#quotedString" },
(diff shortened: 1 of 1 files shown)


pchaigno:
Grammars are usually hosted in their own repositories. I don't know if it's okay to keep it directly in Linguist...

/cc @bkeepers @arfon

bkeepers:
@ArturViader Thanks for the contribution! As @pchaigno suggets, we'd like to keep the grammars in their own repositories so that people can reuse them in their editors.

Would you be interested in pushing the grammar to a new repository, and then following the [docs on adding a grammar](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#i-want-to-add-support-for-the-x-programming-language)?  Thanks!

arfon:
:flags: as stale. @ArturViader - do you think you can do as @bkeepers asked above?

arfon:
Please feel free to open another PR if this is still something you're interested in working on.

PR #1980, opened 2015-01-10
Better PAWN language detection

eider-i128:
Some PAWN language files are wrongly detected as OpenEdge ABL. This commit fixes it.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2094,6 +2094,8 @@ PAWN:
   color: "#dbb284"
   extensions:
   - .pwn
+  - .p
+  - .inc
   tm_scope: source.c++
   ace_mode: c_cpp
 
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2094,8 +2094,8 @@ PAWN:
   color: "#dbb284"
   extensions:
   - .pwn
-  - .p
   - .inc
+  - .p
   tm_scope: source.c++
   ace_mode: c_cpp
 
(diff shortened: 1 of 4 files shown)


pchaigno:
Is there a keyword or keywords that could be used to search for PAWN files on GitHub? (To see how many there are for each extension, as per [the contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#i-want-to-add-support-for-the-x-programming-language).)

eider-i128:
I think you can search it by "sa-mp" but you'll have to filter plugins etc. Please note that PAWN language was already supported by Linguist, this commit just improves it.

pchaigno:
The requirement that a language must have some usage on GitHub before it's added is also used for extensions generally. Though it's more of an indication.

I was able to find [many `.inc` PAWN files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ainc+sa-mp&type=Code&ref=searchresults) with your indication but [only 2 for `.p`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ap+sa-mp&type=Code&ref=searchresults). Should I use another keyword for `.p`?

eider-i128:
It's not common to use .p extension in SA-MP community, but as you can see from this official screen from CompuPhase - .p is default extension for PAWN files - http://www.compuphase.com/pawn/quincy.png

More informatiosn available here: http://www.compuphase.com/pawn/pawn.htm

bkeepers:
Since there isn't a lot of usage with `.p`, and `.inc` is such a common extension, I'm :-1: on adding these extensions right now.  We can definitely reconsider in the future.

PR #1995, opened 2015-01-14
Support for web ontology language (OWL)

phuicy:
The Web Ontology Language (OWL) is a family of knowledge representation languages for authoring ontologies, created by W3. 
The languages' syntax is simply "xml". 
The languages' file extension is ".owl". 

With this in mind the addition was trivial.

Thanks
Guy

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3084,18 +3084,20 @@ Volt:
   tm_scope: source.d
   ace_mode: d
 
+Web Ontology Language:
+  type: markup
+  color: "#3994bc"
+  extensions:
+  - .owl
+  tm_scope: text.xml
+  ace_mode: xml
+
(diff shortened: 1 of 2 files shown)


pchaigno:
Just out of curiosity, if it is highlighted as XML why do you want to add it as a language? why not add it as a new extension for XML?

phuicy:
It is a widely used language, and it would interesting to be able search and gather stats about its use.
Also it is distinct from xml, it just uses for the xml syntax for interchange.

pchaigno:
Okay, thanks for answering my question :-)

arfon:
This looks good. Thanks @phuicy.

arfon:
It'll be another week or so before these changes are live on GitHub.

phuicy:
Brilliant thanks

PR #1996, opened 2015-01-15
Add FXML to languages.yml

aoetk:
FXML is an XML-based language that provides the structure for building a user interface of a JavaFX application.
http://docs.oracle.com/javase/8/javafx/api/javafx/fxml/doc-files/introduction_to_fxml.html


arfon:
:+1: thanks @aoetk.

PR #1998, opened 2015-01-15
Added interpreter "pike" for Pike.

Frigolit:
Fixes detection of Pike files without file extensions but with interpreter specified.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2230,6 +2230,8 @@ Pike:
   extensions:
   - .pike
   - .pmod
+  interpreters:
+  - pike
   ace_mode: text
 
 Pod:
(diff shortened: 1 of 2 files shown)


pchaigno:
Hey @Frigolit! Could you add a sample file in `samples/Pike/` with this shebang?

Frigolit:
There we go~ "/usr/bin/env pike" is preferred as distros tend to place it in either /usr/bin or /bin, and custom builds by default end up in /usr/local/bin.

pchaigno:
Thanks. I'm not from GitHub so I can't say for sure but pull requests pretty straightforward like this one are usually merged quickly. So you just need to wait now :-)

Frigolit:
Alrighty, thank you :)

arfon:
Looks good. Thanks @Frigolit and @pchaigno. It will be another couple of weeks before this change is live on GitHub.

PR #1999, opened 2015-01-15
More relative paths

arfon:


arfon:
Merging these into https://github.com/github/linguist/pull/1976

PR #2000, opened 2015-01-16
Added `htmlbars` as an alias for Handlebars

jayphelps:
See [tildeio/htmlbars](https://github.com/tildeio/htmlbars) for reference.


arfon:
:+1: thans @jayphelps. It'll be another couple of weeks before this is live on Github.

PR #2001, opened 2015-01-16
added Rusthon to languages list for syntax highlighting in md.

hartsantler:
fixes syntax highlighting for rusthon code in markdown on github.
```python
a = [1,2,3]
```
should have same syntax colors as above.
```rusthon
a = [1,2,3]
```

rusthon project page:
https://github.com/rusthon/Rusthon

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2386,6 +2386,11 @@ Python traceback:
   tm_scope: text.python.traceback
   ace_mode: text
 
+Rusthon:
+  type: programming
+  ace_mode: python
+  color: "#3581ba"
+
 QML:
   type: markup
   color: "#44a51c"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2376,6 +2376,8 @@ Python:
   - python
   - python2
   - python3
+  aliases:
+  - rusthon
 
 Python traceback:
   type: data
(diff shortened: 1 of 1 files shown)


pchaigno:
Hi @hartsantler! I'm not from GitHub but I can help you build your pull request ;)

First, to have Rusthon code highlighted in Markdown files you will need to find a grammar for it. It can be a Sublime Text package or a TextMate or Atom bundle. If you find one you can test it with [Lightshow](https://lightshow.githubapp.com/). Also, do you have a longer example of Rusthon code to test any grammar we could find?

Apart from that, is Rusthon associated to any file extension?

hartsantler:
Hi @pchaigno test grammar and literate code page is here: https://github.com/rusthon/Rusthon/blob/master/examples/example-project.md
Can the normal Python grammar be used? The syntax is almost the same it looks fine, and uses `.py` as file extension.

pchaigno:
If the syntax is the same as Python and you want to be able to highlight code snippets in Markdown documents, you could just add `rusthom` as an alias for Python. It would be easier to build a pull request for that. What do you think?

hartsantler:
adding `rusthon` as an alias for Python will work much better.
Where do i add that?

pchaigno:
You can update this pull request to revert your changes and just add the following for the Python entry in `languages.yml`:
```yaml
aliases:
  - rusthon
```

Don't hesitate to tell me if you want more details.

arfon:
/ cc @bkeepers

hartsantler:
Here is another example page where the rusthon fenced code block is not syntax highlighted.
https://github.com/rusthon/Rusthon/blob/master/examples/hello_threejs.md

arfon:
:+1: thanks @hartsantler. This should go live when we next cut a linguist gem (hopefully today).

PR #2002, opened 2015-01-16
Rearrange docs

bkeepers:
The goal of this is to move some of the content around so the `README.md` is focused mostly on GitHub.com users troubleshooting issues, and `CONTRIBUTING.md` is focused on how the library actually works and ways to make it better.

For this first pass, I just moved all the content from `README.md` and `Contributing.md` into a temp file and started moving sections back into each file in a way that made some sense. Now I want to go through and figure out what's missing and make it all more cohesive.

/cc @arfon @pchaigno


arfon:
:+1: to this direction.

pchaigno:
> The goal of this is to move some of the content around so the README.md is focused mostly on GitHub.com users troubleshooting issues, and CONTRIBUTING.md is focused on how the library actually works and ways to make it better.

Totally agree that this will be a significant improvement :+1:

Many users come here to simply add a new extension. They often tend to skip some of the steps in `Adding a language` because they think it doesn't apply to their case. Maybe `CONTRIBUTING.md` could include a separate section for `Adding a new extension` (maybe even an other one for `Adding a grammar`).

Having several sections for common contributions will surely mean having duplicated instructions. I still thing it would be a good thing because it would make it easier to follow the guidelines (and we're all lazy when it comes to reading documentations :stuck_out_tongue: ).

README.md
```diff
@@ -143,73 +49,50 @@ Alternatively, you can use Vim and Emacs style modelines to set the language for
...
-Any of the extensions defined are valid but the first in this array should be the most popular.
 
-### Testing
+These stats are also printed out by the `linguist` executable. You can use the
+`--breakdown` flag, and the binary will also output the breakdown of files by language.
+
+You can try running `linguist` on the root directory in this repository itself:
 
-Sometimes getting the tests running can be too much work, especially if you don't have much Ruby experience. It's okay: be lazy and let our build bot [Travis](http://travis-ci.org/#!/github/linguist) run the tests for you. Just open a pull request and the bot will start cranking away.
+$ bundle exec linguist --breakdown
```
arfon: I think we need to wrap this bit in back-ticks right?

bkeepers:
Finally took some time to wrap this up. It's not perfect, but I think this is a good start.

arfon:
:+1: thanks for this.

Issue #1907, opened 2014-12-21
C# code is recognized as SmallTalk


DerKnerd:
For an example check https://github.com/nant/nant

JoshKeegan:
Having the same problem on [this project](https://github.com/JoshKeegan/PiSearch)

JoshKeegan:
And also on [this project](https://github.com/JoshKeegan/hpop)

pchaigno:
@arfon I'm trying to write an heuristic for this issue. Could you send me corpus of files for the following searches?
- https://github.com/search?o=asc&q=extension%3Acs+public&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93 - should be mostly C# files
- https://github.com/search?o=desc&q=extension%3Acs+self+iftrue&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93 - should be mostly Smalltalk files

Thanks :-)

arfon:
@pchaigno - I've just sent you some samples for those searches.

franklinchou:
You will also notice that for (many) of these projects, if you click on the colorized bar that displays the ratio of the project's various languages, clicking on "Smalltalk" will actually point nowhere on the page. This is a phenomenon that I find occurs when Linguist incorrectly recognizes the language.

larsbrinkhoff:
Right, the language statistics bar can be inconsistent with search results from the repository.  Why is that?

pchaigno:
I opened a pull request at #2003 with heuristic rules to solve these issues.

@arfon The samples from the second search were all from one unique repository so I tested the PR against the repository instead of the sample files you send me (the repository contains more than 43,000 Smalltalk files).

arfon:
Thanks @pchaigno. https://github.com/github/linguist/pull/2003 is now merged but it will be another week or so before the changes are live on GitHub.

ChayimFriedman2:
I've this problem too: see https://github.com/ChayimFriedman2/SimpleNETting/.

rmunn:
[...]

PR #2003, opened 2015-01-17
Heuristics for .cs files: C# and Smalltalk

pchaigno:
This pull request adds heuristic rules to distinguish C# and Smalltalk on the `.cs` extension.
It solves the issues reported in #1907.

Here are the results of the detection when running Linguist on a few repositories (the 7 last are from #1907):

      | Actual language | C# | Smalltalk
------|-----------------------|-------------|--------
[1,000 first results from C# search](https://github.com/search?o=desc&q=extension%3Acs+public&ref=searchresults&s=&type=Code&utf8=%E2%9C%93) | C# | 100% | 0.00%
[theseion/pharogenesis](https://github.com/theseion/pharogenesis) | Smalltalk | **7.26%** | 92.74%
[nant/nant](https://github.com/nant/nant) | C# | 100.00 | 0.00%
[LouisTakePILLz/ExtensionLib](https://github.com/LouisTakePILLz/ExtensionLib) | C# | 100.00 | 0.00%
[Robmaister/SharpFont](https://github.com/Robmaister/SharpFont) | C# | 100.00 | 0.00%
[ShareX/ShareX](https://github.com/ShareX/ShareX) | C# | 100.00 | 0.00%
[sillsdev/phonology-assistant](https://github.com/sillsdev/phonology-assistant) | C# | 100.00 | 0.00%
[JoshKeegan/PiSearch](https://github.com/JoshKeegan/PiSearch) | C# | 100.00 | 0.00%
[JoshKeegan/hpop](https://github.com/JoshKeegan/hpop) | C# | 100.00 | 0.00%

The only miss-detection is for [theseion/pharogenesis](https://github.com/theseion/pharogenesis) which is a really big collection of Smalltalk files (~43,800 of them). The results from Linguist don't change before and after this pull request (7.23% C# before). I looked at the content of a few of the miss-detected files and they were all really small files (no more than 10 lines).

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -72,7 +72,7 @@ module Linguist
     disambiguate "C#", "Smalltalk" do |data|
       if /![\w\s]+methodsFor: /.match(data)
         Language["Smalltalk"]
-      elsif /^\s*namespace\s*[\w\.]+\s*{/.match(data) || /^\s*using\s*[\w\.]+\s*;/.match(data)
+      elsif /^\s*namespace\s*[\w\.]+\s*{/.match(data) || /^\s*\/\//.match(data)
         Language["C#"]
       end
     end
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -69,6 +69,14 @@ module Linguist
       end
     end
 
+    disambiguate "C#", "Smalltalk" do |data|
+      if /![\w\s]+methodsFor: /.match(data)
+        Language["Smalltalk"]
+      elsif /^\s*namespace\s*[\w\.]+\s*{/.match(data) || /^\s*using\s*[\w\.]+\s*;/.match(data)
+        Language["C#"]
+      end
+    end
+
     disambiguate "Objective-C", "C++", "C" do |data|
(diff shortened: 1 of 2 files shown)


arfon:
Awesome work. Thanks @pchaigno

xPaw:
I have two files in my repo that are still identified as smalltalk, even though the `namespace` regex should be catching it.

See https://github.com/SteamDatabase/ValveResourceFormat/search?l=smalltalk

arfon:
@xPaw - I looked into this and it's not matching as there are [Byte Order Marks - BOM](https://en.wikipedia.org/wiki/Byte_order_mark) `\xEF\xBB\xBF` before the `namespace` declaration which appear to be tripping up the regex. This means that the heuristic for `C#` isn't matching and so we're falling back to the Bayesian classifier which is (incorrectly) reporting `Smalltalk` as the answer.

I'm afraid I'm not very up-to-speed on text-encoding and BOMs (and whether you can strip them) but this is what's causing the issue for these two files.

xPaw:
@arfon Yeah, I was thinking that was the problem. For a fix in my repo, I can just throw in a `using System;` in these files.

arfon:
> @arfon Yeah, I was thinking that was the problem. For a fix in my repo, I can just throw in a using System; in these files.

OK, sounds good.

PR #2009, opened 2015-01-19
Use all extension segments

bkeepers:
A filename like `index.html.fr` would previously return `.html.fr` and
`.fr` as possible extensions. This makes it so it will also return
`.html`.

The tests are currently failing because now `bootstrap.css.map` is being detected as CSS when it should be nothing. I think this change has the potential to cause a lot of issues like that, so before this can merge we may want to investigate filtering out any results that don’t cross some confidence threshold in the classifier.

/cc @arfon

### DIFF
--- a/lib/linguist/file_blob.rb
+++ b/lib/linguist/file_blob.rb
@@ -76,9 +76,13 @@ module Linguist
     def extensions
       basename, *segments = name.split(".")
 
-      segments.map.with_index do |segment, index|
-        "." + segments[index..-1].join(".")
+      segments.map! { |segment| "." + segment }
+
+      combined = segments.map.with_index do |segment, index|
+        segments[index..-1].join
       end
+
+      (combined + segments).uniq
     end
   end
 end
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -173,12 +173,11 @@ module Linguist
     def self.find_by_filename(filename)
       basename = File.basename(filename)
 
-      # find the first extension with language definitions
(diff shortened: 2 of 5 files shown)


bkeepers:
> …we may want to investigate filtering out any results that don’t cross some confidence threshold in the classifier.

I actually don't understand the results that the classifier are returning. I'm going to need to dig a little more.

```
irb(main):004:0> Linguist::Classifier.classify(Linguist::Samples.cache, 
  Linguist::FileBlob.new("test/fixtures/Data/bootstrap.css.map").data, ["CSS"])
=> [["CSS", -33.31861787250029]]

irb(main):034:0> Linguist::Classifier.classify(Linguist::Samples.cache,
  Linguist::FileBlob.new("samples/css/bootstrap.css").data, ["CSS"])
=> [["CSS", -86769.19708126006]]


irb(main):027:0> Linguist::Classifier.classify(Linguist::Samples.cache, 
  Linguist::FileBlob.new("samples/Ruby/sinatra.rb").data, ["Ruby"])
=> [["Ruby", -4279.144827467248]]

irb(main):028:0> Linguist::Classifier.classify(Linguist::Samples.cache, 
  Linguist::FileBlob.new("samples/Ruby/grit.rb").data, ["Ruby"])
=> [["Ruby", -21.560011329550157]]
```

larsbrinkhoff:
:+1:

arfon:
I'm going to close this as it's pretty stale. Feel free to re-open if you're wanting to pick this up @bkeepers

PR #2024, opened 2015-01-23
New Mathematica (aka Wolfram Language) extensions added

halirutan:
Added Wolfram Language extensions .wl, .wlt and .mt to the Mathematica section.
Checked whether other languages have the same extensions I introduced which is not the case.
Added a sample .wlt file for a Wolfram Unit Test


arfon:
Hi @halirutan - thanks for this. The build is failing because the extensions need to be in the correct order:

> TestPedantic#test_extensions_are_sorted [/home/travis/build/github/linguist/test/test_pedantic.rb:15]:
.wl should come after .mt

Also, would you mind adding a sample to for each new extension you've listed here please?

halirutan:
@arfon Done. Note that the new extensions don't mean anything new. `*.wl` and `*.wlt` were only introduced due to the rigorous renaming of *the language Mathematica* to *Wolfram Language*. Therefore, `*.m` is similar to `*.wlt` and `*.mt` is similar to `*.wlt`. This is the reason I only copied the example package and the existing unit test I had included to files with the new extension.

pchaigno:
There seem to be [many `.mt` files on GitHub](https://github.com/search?o=desc&q=extension%3Amt+NOT+dnhfjdbgfjd&ref=searchresults&s=&type=Code&utf8=%E2%9C%93) but [not all of them look like Mathematica](https://github.com/search?o=desc&q=extension%3Amt+if&ref=searchresults&s=indexed&type=Code&utf8=✓)...

halirutan:
@pchaigno Hmm, that's unfortunate. I have only searched the linguist's `languages.yml` and ensured that I don't introduce extensions that are already in there.

arfon:
> If all those other .mt files would be categorized as Mathematica we should probably remove this extension until a heuristic for the majority of other languages exist.

:+1: - I agree. Let's do this until we can figure out what to do with the rest.

halirutan:
OK, I checked GitHub for `.wl` and for `.wlt` but that seems only very few entries. `.wl` is mostly used for an theoretical *while language* that is used for exercises in homework. For `.wlt` I find only two entries. So I think we can keep these two extensions. Let me make another pull request where I remove `.mt`.

arfon:
Sorry it's taken a while to get to this @halirutan.

PR #2025, opened 2015-01-23
Specify ace_mode for OpenSCAD

hyperair:


arfon:
:+1: thanks @hyperair

PR #2030, opened 2015-01-23
GAP and Scilab

arfon:
Replaces #1541 and #1523


arfon:
Apologies for taking so very long to get to this @fingolfin.

fingolfin:
Better late than never :).

Of course now the work I did on pygments is again worthless, as in the meantime it seems Linguist switched to another way of doing syntax highlighting... It feels a bit like chasing tails... ah well *sigh*, I guess that means one of these days I'll need to find out how you do this now, and see if I can write something for GAP again  :-).

Issue #2027, opened 2015-01-23
Puppet recognized as Pascal if comments contain too many minus characters


tOnI0:
I found out that if my source code (for instance this Puppet manifest https://github.com/tOnI0/aco-tomcat/blob/master/manifests/init.pp) contains too many minus characters placed in comments, it gets recognized as Pascal.

Converting '-' to '.' made the language being properly recognized.

tOnI0:
Note: changing all these long lines of '--------' in the class parameters to something shorter like just 3 minus signs ('---') or simply deleting some of the lines also helped. Makes me think the actual number of minus chars is actually involved.

pchaigno:
This is because [one of the three Pascal samples for the Bayesian classifier contains many dashes in comments](https://github.com/github/linguist/blob/master/samples/Pascal/gtkextra.pp). Usually Linguist removes the comments before feeding the samples to the classifier for training but, since here braces are used for multiline comments, it is difficult (removing lines between braces would destroy samples in many languages :/).

I opened a pull request to remove this sample and add new ones at #2031. It fixes the issue with your file.

arfon:
Fixed in #2031. @tOnI0 - it'll be another week or so before this change is live on GitHub.

PR #2031, opened 2015-01-24
Better samples for Pascal

pchaigno:
This pull request replaces one of the Pascal samples with three new ones.
There was an issue, reported in #2027, with the comments in the old one. A great number of dashes in Pascal multi-line comments were interpreted by the classifier during training. This had the effect to detect potential Pascal files (for example Puppet `.pp` files) with many dashes as Pascal files.


tOnI0:
Sounds legit, and it now works again with your changes. Thanks!

arfon:
Excellent. Thanks @pchaigno

Issue #1251, opened 2014-06-04
Ragel in C++ Host is incorrectly identified as 'Ragel in Ruby Host'


gsauthof:
This search returns some false positives:

https://github.com/gsauthof/imapdl/search?l=ragel%20in%20ruby%20host

There is no Ruby code at all - the correct host language would be C++.

PR #2034, opened 2015-01-24
Fix description of Ragel files

gsauthof:
Source:

Download http://www.colm.net/files/ragel/ragel-6.9.tar.gz and look unter
`examples`. There are a bunch of `.rl` files that all contain Ragel
specifications with some C or C++ in action snippets.  Not one example
including ruby.

Ragel (the state machine description compiler/language) also supports
languages besides C/C++, but a .rl file usually contains just small
snippets of the target language (C/C++ ...) in actions of a state machine.

Thus, the correct classification of `.rl` files is just 'Ragel' and
NOT 'Ragel in Ruby Host'.

See also my comment in the prematurely closed issue:

https://github.com/github/linguist/issues/1251#issuecomment-67895082


arfon:
Hi @gsauthof - I'm afraid renaming languages like this is currently not possible because of a dependency in our search infrastructure. I realise this far from optimal but this is a long-standing issue for us and my understanding is that we don't have a fix for this right now.

arfon:
Closing this as we don't have a way to address this yet. We're working on a new code search index which should make this PR possible and so I've labelled this 'Language rename' for future discoverability. Sorry we don't have better solution right now :-\

PR #2006, opened 2015-01-25
Use filenames as a definitive answer

pchaigno:
I think we should separate the detection by filenames from the detection by file extensions. I'm opening this issue to discuss the pros and cons of this approach.

So I think we should have a new Strategy that checks if a file match one of the filenames defined in `languages.yml`. If it does, then it can return directly the language associated. If none is found, the process can continue as usual.

This new strategy would fix many _miss-detections_ currently happening:
- [`CMakeLists.txt` detected as Text instead of CMake](https://github.com/search?l=text&o=desc&q=filename%3ACMakeLists.txt+cmake_minimum_required&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93)
- [`pom.xml` detected as XML instead of Maven POM](https://github.com/search?l=xml&q=filename%3Apom.xml+xml&ref=searchresults&type=Code&utf8=%E2%9C%93) (all `.xml` filenames have miss-detections)
- Executable files: see [Makefile](https://github.com/search?l=perl&q=filename%3AMakefile+all&ref=searchresults&type=Code&utf8=%E2%9C%93), [Dockerfile](https://github.com/search?l=bash&o=desc&q=filename%3ADockerfile+env&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93) or [Cakefile](https://github.com/search?l=javascript&q=filename%3ACakefile+if&ref=searchresults&type=Code&utf8=%E2%9C%93).

The only drawback I see so far is that there might be some cases where developers use filenames defined in `languages.yml` for other language than what there are usually associated to. I think it's an acceptable drawback because the filenames defined in `languages.yml` are sufficiently specific and if it does happen, they can always use the Overrides.

/cc @arfon @bkeepers


pchaigno:
I just made a pull request out of this issue.
It adds a new strategy and 5 files in `test/fixtures/` to test it.

 | Before | Now
-------|--------|-------
[Maven POM/pom.xml](https://github.com/pchaigno/linguist/blob/filename-strategy/test/fixtures/Maven%20POM/pom.xml) | XML | Maven POM
[Makefile/Makefile](https://github.com/pchaigno/linguist/blob/filename-strategy/test/fixtures/Makefile/Makefile) | Perl | Makefile
[Dockerfile/Dockerfile](https://github.com/pchaigno/linguist/blob/filename-strategy/test/fixtures/Dockerfile/Dockerfile) | Shell | Dockerfile
[CoffeeScript/Cakefile](https://github.com/pchaigno/linguist/blob/filename-strategy/test/fixtures/CoffeeScript/Cakefile) | JavaScript | CoffeeScript
[CMake/CMakeLists.txt](https://github.com/pchaigno/linguist/blob/filename-strategy/test/fixtures/CMake/CMakeLists.txt) | Text | CMake

arfon:
This looks solid to me. @bkeepers - what do you think?

lib/linguist/language.rb
```diff
@@ -166,36 +168,31 @@ def self.find_by_alias(name)
...
     def self.find_by_filename(filename)
       basename = File.basename(filename)
-
-      # find the first extension with language definitions
-      extname = FileBlob.new(filename).extensions.detect do |e|
-        !@extension_index[e].empty?
-      end
-
-      (@filename_index[basename] + @extension_index[extname]).compact.uniq
+      @filename_index[basename]
```
bkeepers: This is a breaking API change, so we'll want to bump the major version and doc breaking changes.

bkeepers:
I think this is a good next step.

Ultimately, I think we really need to figure out some kind of confidence scoring. We should be able to calculate a probability that a filename is a given language based on it being listed in the `filenames`, the number of extensions it matches for another language, etc.

lib/linguist/language.rb
pchaigno: Is there anything I need to do in this pull request for that? Add a warning somewhere or a note in comments maybe?

pchaigno:
I updated this PR with the last changes on modelines from `master`.

arfon:
Just looping back on this. @bkeepers you note that there's a breaking change to the API here so you're suggesting we go to `v5.0.0` here?

@pchaigno - would you mind documenting this change as @bkeepers suggests?

lib/linguist/language.rb
```diff
@@ -91,6 +92,8 @@ def self.create(attributes = {})
       Linguist::Strategy::Modeline,
       Linguist::Shebang,
       Linguist::Strategy::Filename,
+      Linguist::Strategy::Extension,
+      Linguist::Strategy::Filename,
```
arfon: @pchaigno - looks like we've got this strategy in here twice. I'm assuming the second time is unintentional?

pchaigno: > I'm assuming the second time is unintentional? 

Yep. It's fixed now.

larsbrinkhoff:
Hey @pchaigno @arfon @bkeepers We should probably not let this go stale.

arfon:
> Hey @pchaigno @arfon @bkeepers We should probably not let this go stale.

Agreed. I still want to see this land in `v5.0`

pchaigno:
@arfon Should we fix a date for the `v5.0` release? I just updated this PR.

arfon:
> @arfon Should we fix a date for the `v5.0` release? I just updated this PR.

Yeah we totally should. Realistically this would need to be sometime the week beginning 11th July as I'm out on holiday next week. The biggest thing here will be testing this properly on GitHub and making sure we've documented the changes well.

brandonblack:
@pchaigno @bkeepers @alhadis --

I'm interested in resurrecting this PR from the pile and working towards a v5.0 release before EOY. 

I can do the validation on GitHub's side of things, but my question for the group is whether or not there are any other breaking changes in the PRs in our backlog that we'd like to squeeze into this release?

I'd like to clear or close some of these stale PRs and since I'm a new maintainer for the project I'd love to get some feedback from you all.

pchaigno:
> I'm interested in resurrecting this PR from the pile and working towards a v5.0 release before EOY.

:tada: 

I went through the open PRs and couldn't find any other with breaking API changes. Could be a good occasion to merge #3278 if it's ready though. There are also a number of methods that have been deprecated for a while, which we could remove:
- [`find_by_shebang`](https://github.com/github/linguist/blob/71032cd2528d599d6563640c60f13ad865f17364/lib/linguist/language.rb#L180)
- [`Linguist::Language.detect`](https://github.com/github/linguist/blob/71032cd2528d599d6563640c60f13ad865f17364/lib/linguist/language.rb#L100)
- [`Linguist::Language.ace_modes`](https://github.com/github/linguist/blob/71032cd2528d599d6563640c60f13ad865f17364/lib/linguist/language.rb#L268)
- [`Linguist::Language.primary_extension`](https://github.com/github/linguist/blob/71032cd2528d599d6563640c60f13ad865f17364/lib/linguist/language.rb#L472)
- [`search_term` field in languages](https://github.com/github/linguist/blob/0e9109c3fcdacb17b27581710c7b435086b70485/lib/linguist/languages.yml#L17)
- [lazy loading in `Generated.generated?`](https://github.com/github/linguist/blob/d46a529b6a0f4e9994bb513e39f6dc672529ee19/lib/linguist/generated.rb#L7)
- [`BlobHelper`](https://github.com/github/linguist/blob/8dc4a1308f7d309ebdf5566251eca3f5562d8901/lib/linguist/blob_helper.rb#L8)?

#3359 addresses the first 5.

arfon:
🚀

brandonblack:
@pchaigno digging into a bit and it looks like GitHub internally is not ready to deprecate BlobHelper just yet. The effort to get it there is non-trivial. I've created github/linguist#3387 to flag that effort and we'll catch it in the next major release.

Issue #781, opened 2013-11-15
support Emacs mode strings


holtzermann17:
Here's an example file, written in Elisp, with the string `-*- Emacs-Lisp -*-` at the top.
https://github.com/holtzermann17/test-driven-development/blob/master/post-commit
For some reason, Github (i.e. Linguist) thinks that this is written in Perl.  It doesn't have a file type extension.

Emacs has a way of specifying major modes in files without file extensions (or to override the type that would be selected based on file extension).  This typically corresponds to syntax highlighting for the major mode.   For example, `-*- Emacs-Lisp -*-`, `-*- C -*-`, or `-*- mode: Diff; -*-`.  The syntax of these mode strings is described here: http://www.gnu.org/software/emacs/manual/html_node/emacs/Choosing-Modes.html

The value of the Emacs variable `auto-mode-alist` gives the associations that Emacs makes between common file extensions (specified as emacs regular expressions) and mode names, for instance, `.el` is associated with `emacs-lisp-mode`.   To get a fully populated `auto-mode-alist` one would want to load a lot of contributed modes as well as the "stock" examples.  There's a list of known modes here: http://www.emacswiki.org/emacs/CategoryModes

Note that the mode string is a case insensitive, so `-*- emaCs-LiSp -*-` also works as a way to select Emacs Lisp Mode.

I think it would be great if Linguist could support this additional way of specifying what highlighting syntax to use.  I'd be happy to help build a "definitive" list and e.g. translate the Emacs lisp regular expressions for matching file extensions into Ruby regular expressions or whatever else would be helpful, if this is a new feature that you would in principle like to support in Linguist.

Issue #1792, opened 2014-11-29
Support language highlighting on GitHub.com based upon Linguist overrides


arfon:
Currently the Linguist [overrides](https://github.com/github/linguist#overrides) only affect the language statistics and not the syntax highlighting. It would be good to allow these overrides to change the syntax highlighting behaviour also.

Emacs/Vim mode strings support would also be nice.

Related: https://github.com/github/linguist/issues/781 https://github.com/github/linguist/issues/875

arfon:
> Emacs/Vim mode strings support would also be nice.

PR in progress here: https://github.com/github/linguist/pull/2041

arfon:
OK #2041 is in master now. We plan to cut a gem tomorrow.

pchaigno:
@arfon Does this mean that the idea of overrides impacting highlighting is abandoned?

arfon:
Yikes. Closed this by mistake. We'd still like to support the Linguist-style overrides.

bgriffith:
Just to add my support for this feature or allowing Linguist to set the syntax highlighting (especially when working with custom extensions) :+1:

deathbeam:
Yes please, this would be awesome.

bryanhelmig:
:+1:

eugeneia:
:+1:

mintsoft:
:+1:

bendoh:
👍

oakfang:
:+1:

jordwalke:
I think this is important, and good for github because it allows github to be used as a place to build and share technologies that are not yet mainstream (like JSX before it became popular). Can you update on the plan for this?

zenparsing:
:+1: Sometimes people invent weird extensions for normal things (like "jpp" for a certain functional classification of "jsp").

DerekTBrown:
+1

NyproTheGeek:
Is this issue still being reviewed?

itzo:
Also interested in this.

scottbedard:
This feature would be super useful, I'd love to see it implemented.

sunjay:
This would be really useful to have as adding vim modelines to every file is extremely inconvenient with a lot of files with a non-standard file extension.

lildude:
Good news peeps, this is now a thing that was shipped to all users just over a week ago.

Your [language overrides](https://github.com/github/linguist#overrides) should now be reflected in syntax highlighting of blobs and diffs as well as the language statistics.

[...]

Issue #1210, opened 2014-05-21
Add support for Vim modeline filetype overrides


thefotios:
It would be very useful if you guys supported Vim modelines for overriding filetype detection. For example, I have one project that uses Perl Mason heavily (I know... I cry on the inside every day). The problem is that our Mason files use `.md` as the extension, which gets incorrectly identified as Markdown. Vim doesn't detect this either, but a simple modeline at the bottom of the file fixes this:

```
<% # vim: set filetype=mason : %>
```

This would allow developers an easy way to use their preferred filetype/extension without you guys having to detect everything under the sun. It would also provide an escape hatch for the weird corner cases or overlapping extensions.

Modelines are documented [here](http://vimdoc.sourceforge.net/htmldoc/options.html#modeline) and have a very simple grammar.

Issue #875, opened 2013-12-30
Language identifier


ghost:
> In your fenced block, add an optional language identifier and
> we'll run it through syntax highlighting.

[&sect; Syntax highlighting](http://help.github.com/articles/github-flavored-markdown#syntax-highlighting)

So, with fenced code blocks we can explicitly state the desired language for highlighting. I feel that this should be applied to files as well. For files with no/unknown extension, no shebang, you could override them if they are misidentified

``` rb
# language: ruby
# the previous line will set the highlighting
def foo
  puts 'foo'
end
```

This would be a boon to all users at GitHub; instead of waiting for a fix in the detection code, you can just add the identifier, then remove it if the auto detect code is fixed.

svnpenn:
For dummies, this can now be done with:

    # -*- mode: sh -*-

PR #2041, opened 2015-01-26
Emacs vim modelines

arfon:
Starting work on overrides based upon Vim and Emacs modelines.

Fixes https://github.com/github/linguist/issues/781 https://github.com/github/linguist/issues/875 https://github.com/github/linguist/issues/1792 https://github.com/github/linguist/issues/1210 https://github.com/github/linguist/pull/894


arfon:
@bkeepers would love your review on this.

arfon:
/ cc @holtzermann17 @svnpenn @thefotios

arfon:
Also, cc @pchaigno because of [this comment](https://github.com/github/linguist/issues/1991#issuecomment-70161091). I had expected this reordering to cause issues but I think we just need to make sure we're calling `Linguist::Strategy::Filename` before those strategies that _need_ the list of possible languages (e.g. `Linguist::Heuristics`)

lib/linguist/strategy/modeline.rb
```diff
@@ -0,0 +1,31 @@
...
+        else
+          return []
+        end
+      end
+
+      # Public: Get the modeline from the first n-lines of the file
+      #
+      # Returns a String or nil
+      def self.modeline(data)
+        data.lines.first(5).any? { |l| l.match(/\W(?:filetype=|ft=|-\*- mode:|-\*-)\s*(\w+)/) }
```
arfon: Designed to match the following:

```
Vim
/* vim: set filetype=perl: */
/* vim: set ft=prolog: */

Emacs
; -*- mode: php;-*-
; -*-php-*-

```

lib/linguist/strategy/modeline.rb
```diff
@@ -0,0 +1,31 @@
...
+      # Public: Detects language based on Vim and Emacs modelines
+      #
+      # blob               - An object that quacks like a blob.
+      #
+      # Examples
+      #
+      #   Modeline.call(FileBlob.new("path/to/file"))
+      #
+      # Returns an Array with one Language if the blob has a shebang with a valid
+      # interpreter, or empty if there is no shebang.
```
bkeepers: References to "shebang" need updated here.

arfon: :+1: - good catch.

lib/linguist/strategy/modeline.rb
bkeepers: 1. Would it be worth also anchoring to the `vim:` or `; -*-` sequences?
2. I feel like I've seen a lot of vim modelines at the end of the file. Is there a prevailing convention for where to put these?

test/test_modelines.rb
```diff
@@ -0,0 +1,34 @@
...
+  include Linguist
+
+  def fixtures_path
+    File.expand_path("../fixtures", __FILE__)
+  end
+
+  def fixture_blob(name)
+    name = File.join(fixtures_path, name) unless name =~ /^\//
+    FileBlob.new(name, fixtures_path)
+  end
```
bkeepers: Should we copy these two helpers to `test_helper.rb`?

arfon: :+1:

lib/linguist/strategy/modeline.rb
arfon: 1. Yes I think it would. Do you think we need to be tolerant of varying amounts of whitespace in these strings? `vim: set filetype`
2. I have no idea to be honest :-\

arfon: OK, how is this @bkeepers:

```ruby
# ; -*-php-*-
/(?:\;\s*-\*-)\s*(\w+)/

# ; -*- mode: php;-*-
/(?:\;\s*-\*-\s*mode:)\s*(\w+)/

# /* vim: set filetype=perl: */
/(?:vim\:\s*set\s*filetype=)\s*(\w+)/

# /* vim: set ft=perl: */
/(?:vim\:\s*set\s*ft=)\s*(\w+)/

# Which results in (hold onto your hats): 
/(?:\;\s*-\*-|\;\s*-\*-\s*mode:|vim\:\s*set\s*filetype=|vim\:\s*set\s*ft=)\s*(\w+)/
```

larsbrinkhoff:
The `;` isn't mandatory for Emacs.
> the simplest is to put the mode name in the first nonblank line, preceded and followed by ‘-*-’.

See http://www.gnu.org/software/emacs/manual/html_node/emacs/Choosing-Modes.html

svnpenn:
I agree with @larsbrinkhoff , the semicolon in this context merely signifies the start of a comment. It should be removed from the regex, or the regex should be made more robust (include `#`, etc)

lib/linguist/strategy/modeline.rb
bkeepers: Ready for more insanity? This enforces that the vim syntax is closed with vim syntax, and emacs is closed with emacs syntax. It also uses the extended regular expression syntax (`/x`), which ignores whitespace in the regex so you can split it across multiple lines.

```ruby
/
  (?:
    (-\*- \s* (?:mode:)? \s*) |                  # $1: Emacs
    (\/\* \s* vim: \s* set \s* (?:ft|filetype)=) # $2: Vim
  )
  (\w+)                                          # $3: language
  (?:
    (?(1)                                        # If $1 matched...
      ;?\s* -\*- |                               # then close Emacs syntax
      : \s* \*\/                                 # otherwise close Vim syntax
    )
  )
/x
```

arfon: ciivn0g

arfon: Actually, that's pretty neat :smile:

arfon:
Thanks @bkeepers!

larsbrinkhoff:
The semicolon is still there:
          EmacsModeline = /-\*-\s*(?:mode:)?\s*(\w+);?\s*-\*-/
It's not required.

arfon:
@larsbrinkhoff - the semi-colon is optional: http://rubular.com/r/4XEbgCVaUA

larsbrinkhoff:
Yes, sorry.  I deleted my comment as soon as I saw that.

arfon:
> Yes, sorry. I deleted my comment as soon as I saw that.

Cool, just wanted to check we're all on the same page :+1:

brucemiller:
I may just be confused: Is this not yet active on github?  Or is perhaps a modeline like:
```
# -*- CPERL -*-
```
not recognized as Perl? (CPerl mode is a popular variant of Perl mode that treats Perl a bit more C like).  Is there an alias mechanism in the language map that could be used (I'm sure there are other cases where mode neq language)?

arfon:
Hi @brucemiller. I'm afraid we need slightly more explicit modelines than this such as:

```
# -*- mode: CPERL -*-
```
See the [regex we're using](https://github.com/github/linguist/blob/master/lib/linguist/strategy/modeline.rb#L4) tested on a couple of strings here: http://rubular.com/r/Zer8M0oVDB

That said, even if you had `# -*- mode: CPERL -*-` in your files right now, we wouldn't know what to do as Linguist doesn't know about `CPERL`. You can teach it though by adding `CPERL` as [an alias](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L4) for Perl (if that's the correct thing to do).

brucemiller:
Yes, I can see why you'd prefer the more restrictive modelines for your application; that's fine.  And, in fact, my emacsen are configured to automatically use cperl mode for perl language files, so was perhaps not quite correct to be specifying cperl in the modeline, anyway. 

I've updated all my files to use the form
```
# -*- mode: Perl -*-
```
and everything seems to work as desired.

Thanks for having already fixed my issue! :>

arfon:
@brucemiller glad we could help!

tesch1:
I saw this after I submitted an issue #2230 for it... sorry, should have searched better before submitting the issue.

Mode names can have - and + in them too, not just \w (ie c++)

Maybe needs to be a map between emacs major mode names and linguist language names (or maybe the missing +- in the regex will fix that though, currently c++ is broken, not sure what it's matched against(?)

Here's a regex that matches better the syntax of actual emacs "modelines":

`-\*-\s*((?!mode)[\w-]+\s*:\s*([\w+-]+)\s*;?\s*)?(mode\s*:)?\s*([\w+-]+)\s*;?\s*((?!mode)[\w-]+\s*:\s*([\w+-]+)\s*;?\s*)?-\*-`

tested here: http://rubular.com/r/TKZ6Tq5Vnj

hth

Issue #1903, opened 2014-12-19
Syntax Highlighting not working anymore for Liquid


spplante:
Hi there,

My github has a project with Liquid syntax that was working well, until you guys migrated to TextMate. If I am not mistaken, Liquid has been created by the company "Shopify", they are pretty big here in Montreal. 

By looking at their GitHub, I found pretty much what you guys need in order to support Liquid syntax highlighting just like before :

https://github.com/Shopify/liquid-tmbundle

This is their TextMate bundle for liquid. If you guys could add this to GitHub's TextMate library it would be super cool and my "read me" file would start working back :)

Thanks!

pchaigno:
Hi @spplante!
I opened an issue on the repository to add a license: Shopify/liquid-tmbundle#3.
Once this is done, I'll open a pull request to add it as a grammar for Liquid.
[...]

PR #2044, opened 2015-01-27
Grammar for Liquid

pchaigno:
This PR adds a grammar for Liquid from a [TextMate bundle](https://github.com/bastilian/validcode-textmate-bundles) as discussed in #1903.

/cc @arfon @spplante


arfon:
:+1: thanks @pchaigno

PR #2045, opened 2015-01-27
Language detection of raw file data without file on disk

assimovt:
Linguist::FileBlob operates on the path to a file. In some cases the
file is not available on disk. Hence, the FileContents mimics FileBlob
and enables language detection based on the raw file data and
the name of the file.


bkeepers:
Thanks for the pull request! Since there isn't anywhere in linguist where we require the content to be a `FileBlob`, you can pass in an object that "quacks like a blob". We actually use this technique on GitHub.com for all of our blob content.

For example, you could use an `OpenStruct`:

```ruby
require "ostruct"
blob = OpenStruct.new(:name => "foo.rb", :data => "puts 'hello world'")
Linguist::Language.detect(blob)
#=> #<Linguist::Language name=Ruby>

blob.extend(Linguist::BlobHelper)
blob.language
#=> #<Linguist::Language name=Ruby>
```

assimovt:
Alright, we can live with ```OpenStruct```. Thanks for the hint! :+1:

PR #2048, opened 2015-01-28
Update Swift grammar

infininight:
Corrected an issue with Swift comments as reported at textmate/swift.tmbundle#11, updated submodule.


arfon:
:+1: - we'll pull these changes when we next cut the Linguist gem.

PR #2063, opened 2015-02-01
'Text' shouldn't qualify as a valid modeline language.

larsbrinkhoff:
A number of `.f` files previously correctly identified as Forth, are now identified as Text. This is because they contain the Emacs modeline `-*- text -*-`.

Perhaps it would be appropriate to exclude text from the modeline strategy?

http://github.com/search?q=extension%3Af+dup+swap+text&type=Code

### DIFF
--- a/lib/linguist/strategy/modeline.rb
+++ b/lib/linguist/strategy/modeline.rb
@@ -1,7 +1,7 @@
 module Linguist
   module Strategy
     class Modeline
-      EmacsModeline = /-\*-\s*(?:mode:)?\s*(\w+);?\s*-\*-/
+      EmacsModeline = /-\*-\s*mode:\s*(\w+);?\s*-\*-/
       VimModeline = /\/\*\s*vim:\s*set\s*(?:ft|filetype)=(\w+):\s*\*\//
 
       # Public: Detects language based on Vim and Emacs modelines
--- a/test/fixtures/Data/Modelines/example_smalltalk.md
+++ b/test/fixtures/Data/Modelines/example_smalltalk.md
@@ -1,1 +1,1 @@
-; -*-Smalltalk-*-
+; -*-mode:Smalltalk-*-


arfon:
That seems sane to me. @bkeepers?

bkeepers:
It doesn't feel right to me to ignore the modelines in certain situations, but I'm not sure I understand the use case. Do you know why the forth files use the text modeline?

arfon:
> It doesn't feel right to me to ignore the modelines in certain situations, but I'm not sure I understand the use case. Do you know why the forth files use the text modeline?

Is it defining the content-type or something similar?

larsbrinkhoff:
@bkeepers I guess the author(s) of these files wanted to use the Emacs text mode for editing.  Emacs doesn't come bundled with a Forth mode, although there is one floating around online.

pchaigno:
Couldn't it be because `.f` is associated with Fortran by default in Emacs? (I'm not a user of Emacs so I wouldn't know.)

larsbrinkhoff:
@pchaigno That may well be the primary reason!

bkeepers:
> Couldn't it be because `.f` is associated with Fortran by default in Emacs? (I'm not a user of Emacs so I wouldn't know.)

Wouldn't the modeline be for `forth` then? What does emacs do with unknown modelines?

pchaigno:
@bkeepers If `.f` is indeed associated to Fortran, then I guess that Emacs users use the Text modeline to "disable" highlighting.

bkeepers:
I would expect if the files used a modeline of `-*-forth-*-` and Emacs doesn't support `forth`, it would behave the same as if it used a modeline of `-*-text-*-`. Does that make sense?

I think we need more information before adding exceptions for modelines.

larsbrinkhoff:
@bkeepers If Emacs doesn't support `forth` (or otherwise doesn't recognise the modeline), it falls back on using the file extension.  And if that fails, it goes into `fundamental` mode.  `text`mode is for authoring natural-language documents.

I don't necessarily want an exception for `-*- text -*-`.  Basically I'd like those Forth files to be correctly identified.

larsbrinkhoff:
One might argue, perhaps, that `-*- text -*-` is a declaration that "I want to *edit* this file as text", and not primarily a statement about the *file contents*.

bkeepers:
> I don't necessarily want an exception for `-*- text -*-`.  Basically I'd like those Forth files to be correctly identified.

So what would that look like? If a file has a mode line, then we should take that into consideration, but add some other logic that tries to guess if the mode line is to be trusted? How do we determine that?

To me, it seems like the whole point of mode lines is they are an explicit declaration by the author. They're saying "I don't care what you think this file is, I'm telling you it's X". Mode lines should be authoritative.

larsbrinkhoff:
> To me, it seems like the whole point of mode lines is they are an explicit declaration by the author.

I'm an avid Emacs user, and to me that seems strange.  To me, the mode line is just a hint to the editor.  Not a declaration about the file contents.  Maybe that doesn't make sense, and I'm not explaining it right.

I believe the `.f` file extension are also an explicit declaration by the author, which then would contradict the mode line.

To me, it's quite obvious that those files are Forth, regardless of the mode line.  I guess I could contact the original author, Richard WM Jones, and ask him about his intentions.  But then, it's not just about these particular files.  Although there might well be few other files with the same problem.

> So what would that look like?

Possibly multiple strategies could each have their say about the final result.

bkeepers:
I talked with some other GitHubbers about this, and most of them were sympathetic to @larsbrinkhoff's view. Apparently many Emacs users even ignore them.

What about this as a solution to this:

1. Require `mode` in the [emacs mode line](http://www.gnu.org/software/emacs/manual/html_node/emacs/Specifying-File-Variables.html#Specifying-File-Variables) (suggested by @sbryant)
2. Detect mode lines after detecting by filename, so they are only used if there is ambiguity between languages with the same extension.

This makes them much less useful than we originally hoped, but it sounds like that is how Emacs works. I'm not sure about vim.

arfon:
1. sounds by far the best option to me.

larsbrinkhoff:
I reworked my PR to implement option 1.

bkeepers:
:+1: 

I was actually suggesting we do both of those things, not *just* 1. We could try 1 and then only do 2 if it's still a problem for people. What do you all think?

arfon:
> I was actually suggesting we do both of those things, not just 1. We could try 1 and then only do 2 if it's still a problem for people. What do you all think?

:+1: - this PR looks to be in pretty good shape I think too.

bkeepers:
@larsbrinkhoff thanks for working through this with us.

larsbrinkhoff:
@bkeepers No problem.  Good idea to solicit advice from others.

arfon:
@larsbrinkhoff - this is is production now.

larsbrinkhoff:
@arfon That was quick; thanks!

PR #2066, opened 2015-02-02
Add the markup language of TeXmacs

iblech:
TeXmacs is a scientific wysiwyg word processor inspired by TeX and Emacs, but sharing no code with those programs.

There are somewhat more than 5000 TeXmacs files on GitHub:
https://github.com/search?q=TeXmacs+extension%3Atm&ref=searchresults&type=Code&utf8=%E2%9C%93
https://github.com/search?q=TeXmacs+extension%3Ats&ref=searchresults&type=Code&utf8=%E2%9C%93

This pull request records TeXmacs in languages.yml and adds two sample documents. Note that TeXmacs files have the extensions .tm and .ts, which clashes with Tcl's .tm. In the case that the Bayesian classifier does not work, we could use the simple heuristic that TeXmacs files start with "<TeXmacs|". I'll gladly be of assistance.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2969,6 +2969,15 @@ TeX:
   - .sty
   - .toc
 
+TeXmacs:
+  type: markup
+  ace_mode: text
+  wrap: true
+  extensions:
+  - .tm
+  - .ts
+  tm_scope: none
+
 Tea:
   type: markup
   extensions:
--- a/samples/TeXmacs/new-presentation.ts
+++ b/samples/TeXmacs/new-presentation.ts
@@ -1,0 +1,94 @@
+<TeXmacs|1.0.6.14>
+
+<style|source>
+
+<\body>
+  <active*|<\src-title>
+    <src-package-dtd|presentation|1.0|presentation|1.0>
+
+    <\src-purpose>
+      European-style numbering.
+    </src-purpose>
+
+    <src-copyright|1998--2004|Joris van der Hoeven>
+
+    <\src-license>
+      This <TeXmacs> style package falls under the <hlink|GNU general public
+      license|$TEXMACS_PATH/LICENSE> and comes WITHOUT ANY WARRANTY
+      WHATSOEVER. If you do not have a copy of the license, then write to the
+      Free Software Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+      02111-1307, USA.
+    </src-license>
+  </src-title>>
+
+  <use-package|varsession|alt-colors>
+
+  <\active*>
+    <\src-comment>
+      Customizations
+    </src-comment>
+  </active*>
+
+  <assign|font-family|ss>
+
+  <assign|name|<macro|x|<with|font-family|rm|font-shape|small-caps|<arg|x>>>>
+
+  <assign|bg-color|<pattern|paper-ridged-light.png|*3/5|*3/5|#f4eee8>>
+
+  <assign|monochrome-bg-color|#f4eee8>
+
+  <assign|ornament-border|2ln>
+
+  <assign|ornament-color|<pattern|pine.png|*3/5|*3/5|#e0b050>>
+
+  <assign|ornament-sunny-color|#f0e0c0>
+
+  <assign|ornament-shadow-color|#c07055>
+
+  <assign|ornament-hpadding|1spc>
+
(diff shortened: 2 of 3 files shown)


pchaigno:
I'm not from GitHub but I can help to prepare this pull request for review.

First, we'll need to show that the Bayesian classifier can handle the conflict on `.tm`.
@arfon Could you send me corpus of files for https://github.com/search?q=extension%3Atm+texmacs&ref=reposearch&type=Code&utf8=%E2%9C%93 and https://github.com/search?q=extension%3Atm+namespace+OR+package&ref=reposearch&type=Code&utf8=%E2%9C%93?

Also, @iblech, Do you know of a grammar for TeXmacs? It can be a TextMate bundle or a Sublime Text package.

iblech:
Awesome, thanks!

I don't think there are syntax highlighting rules for TeXmacs, as TeXmacs files are not intended to be edited manually. However, I'll cc Miguel, a TeXmacs developer, who can give a definitive statement.

/cc @mdbenito

pchaigno:
Thanks @arfon for the samples!

So currently the results are:

 | TeXmacs | Tcl
------- | ------ | -----
1,000 files from [Tcl search](https://github.com/search?q=extension%3Atm+namespace+OR+package&ref=reposearch&type=Code&utf8=%E2%9C%93) | 0,00% | 100,00%
1,000 files from [TeXmacs search](https://github.com/search?q=extension%3Atm+texmacs&ref=reposearch&type=Code&utf8=%E2%9C%93) | 87.02% | **12,98%**

Maybe more samples for TeXmacs could solve that... @iblech Could you add a few more sample files?

mdbenito:
There are indeed no syntax highlighting rules for TeXmacs, but its syntax is basically xml-ish with `<\tag>content</tag>` and `<self-closing-tag|with|args>` as the basic constructs and `<TeXmacs|version>` as the first item. I can provide a list of standard tags if required. I can also provide hundreds of samples (all of texmacs.org and the manual are in the repo as TeXmacs files) if they are still needed.

iblech:
Yes, I think more sample files are needed. We can talk tomorrow and amend the pull request then!

iblech:
@pchaigno Miguel kindly provided lots of TeXmacs files. I'm a bit hesistant to add them all to this pull request, since they total more than 2 MiB. (Maybe I'm really old-fashioned?) Maybe you can download them and check how many files are really needed? If it's easier for you, I can also just go ahead and add the files.

https://dl.dropboxusercontent.com/u/15831115/mantm.zip https://dl.dropboxusercontent.com/u/15831115/webtm.zip

pchaigno:
> I'm a bit hesistant to add them all to this pull request, since they total more than 2 MiB.

Good thinking ;) The files themself shouldn't be a problem but the increase in size of the `samples.json` file (1.85MB => 2.8MB) could be (as @bkeepers noted in #2117). I think we should wait for a response to this issue in #2117, if that's okay with you.

At the moment, I only checked that adding all the sample files solves the mis-classifications. And it does! (see below) If you want to merge this quickly, I can test with less sample files.

 | TeXmacs | Tcl
------- | ------ | -----
1,000 files from [Tcl search](https://github.com/search?q=extension%3Atm+namespace+OR+package&ref=reposearch&type=Code&utf8=%E2%9C%93) | **0.51%** | 99.49%
1,000 files from [TeXmacs search](https://github.com/search?q=extension%3Atm+texmacs&ref=reposearch&type=Code&utf8=%E2%9C%93) | 100.00% | 0.00%

iblech:
Great that it works! I'm in no particular hurry and can wait for #2117 to be solved. I'll change the links to the TeXmacs files above to a more stable location.

To reduce that 0.51 % to 0 %, we could employ the heuristic that a TeXmacs file always has "TeXmacs" in its first line.

Alhadis:
@iblech You'll probably need to update this PR with a heuristic for TypeScript as well, since the `.ts` extension is now [widely associated with it](https://github.com/search?q=extension%3Ats+NOT+nothack&type=Code) (105,416,031 results currently indexed).

pchaigno:
To summarize this, since Stalebot closed it: we can't add enough sample files because of the increase of working memory, so on the short term, if someone wants to fix it, your best bet is to write [a heuristic rule](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml).

PR #2072, opened 2015-02-03
Add .4TH Forth extension.

larsbrinkhoff:
Forth already has the `.4th` extension.  Apparently, there are also Forth files with the upper-case `.4TH` extension:
http://github.com/OS2World/DEV-FORTH-Forth-2

Don't know if there are more, as I don't think it's possible to search specifically for upper-case file extensions.

Alternatively, would there be some downside to do case-insensitive comparisons for file extensions?  The only upper-case-only extensions I found in `languages.yml` were `.E` and `.ML`.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -918,6 +918,7 @@ Forth:
   color: "#341708"
   extensions:
   - .fth
+  - .4TH
   - .4th
   - .F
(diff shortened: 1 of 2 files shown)


arfon:
:+1:

larsbrinkhoff:
THX :100:

PR #2073, opened 2015-02-03
Add support for the AMPL modeling and script language

vitaut:
The proposed PR adds initial support for [AMPL](http://en.wikipedia.org/wiki/AMPL), a popular modeling and scripting language for mathematical optimization, including its subset called GNU MathProg. AMPL is used in a substantial number of GitHub repositories, so I think it is a useful addition. Here are a few usage examples:

https://github.com/search?p=1&q=extension%3Amod+var&ref=searchresults&type=Code&utf8=%E2%9C%93
https://github.com/search?utf8=%E2%9C%93&q=extension%3Amod+ampl&type=Code&ref=searchresults


vitaut:
Anyone?

pchaigno:
I'm not from GitHub but I can help you build the pull request proposal to get it merged.

First you'll need to show examples of AMPL on GitHub for each of the 4 extensions you're adding.

The last three (`dat`, `mod` and `run`) are very common extensions so you'll have to make sure that no other language is using them. If other languages are using them, we'll need to add those to Linguist.

Finally, do you know of a grammar that could be used to highlight AMPL code? It can be a TextMate bundle or a Sublime Text package.

vitaut:
Thanks, @pchaigno.

Here are examples of AMPL files with different extensions on GitHub:

* `.ampl`: https://github.com/search?utf8=%E2%9C%93&q=extension%3Aampl+var&type=Code&ref=searchresults

* `.mod`: https://github.com/search?utf8=%E2%9C%93&q=extension%3Amod+subject&type=Code&ref=searchresults

* `.dat`: https://github.com/search?utf8=%E2%9C%93&q=extension%3Adat+param&type=Code&ref=searchresults

* `.run`: https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+solve&type=Code&ref=searchresults

I used AMPL-specific keywords in the queries above, so the result give very few false positives. If necessary I can implement detection based on file content to avoid incorrect detection.

There is a Sublime Text package for AMPL available in the [sublime-ampl](https://github.com/JackDunnNZ/sublime-ampl) repository.

vitaut:
@pchaigno I've added the AMPL grammar as described in the [CONTRIBUTING](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) document and checked for other uses of `.mod`, `.dat` and `.run` extensions. `.mod` seems to mostly conflict with Modula-2 while `.dat` and `.run` are used for pretty much anything. Will it help if I implement a heuristic for AMPL detection depending on file content (if it is possible in Linguist)? It should be pretty easy to do since AMPL's statements start with fairly unique keywords like `var`, `param`, `minimize` or `subject`.

pchaigno:
Sorry, I didn't find the time to look into this lately :/

Great for the grammar!

For the `.mod` extension, there are around [5,400 AMPL files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Amod+subject&type=Code&ref=searchresults) as you noted before. In total there is more than 147,700 `.mod` files. Among these, I found [XML files](https://github.com/vogellacompany/vogella/blob/1464ea7e488c304589f09a91996c55082439416e/de.vogella.docbook.first/docbook-xml-4.5/dbcentx.mod), [C++ files](https://github.com/lecheel/fte-fork/blob/1ae31ae829ecb112d7574d260fe3c8d874926dc6/src/e_cmds.cpp.mod) and a few other languages that I'm no able to identify. So I think it's used for pretty much anything, like `.dat` and `.run`.

The process to add these extensions is a bit complicated because we have to add them to all languages using them. This first requires to identify all languages that use the extension commonly (i.e. several hundreds of files with that extension for the language). I'm currently doing this for `.inc` for instance.

If that's okay with you, I think we should only add `.ampl` for AMPL in this pull request and I'll open three other issues to track the process of adding `.run`, `.mod` and `.dat`.

vitaut:
> If that's okay with you, I think we should only add .ampl for AMPL in this pull request and I'll open three other issues to track the process of adding .run, .mod and .dat.

I think it's a good idea. Shall I update the PR removing `.mod`, `.dat` and `.run` entries then?

pchaigno:
> Shall I update the PR removing .mod, .dat and .run entries then?

Yep, I'll open the issues tomorrow.
Can you also update your branch? (it can't be automatically merged currently)

vitaut:
OK, I've rebased my branch and removed `.dat`, `.mod` and `.run` for now. Hope this can be merged now. Thanks for your help, @pchaigno!

lib/linguist/languages.yml
```diff
@@ -41,6 +41,14 @@ AGS Script:
   tm_scope: source.c++
   ace_mode: c_cpp
 
+AMPL:
+  type: programming
+  color: "#00008B"
+  extensions:
+  - .ampl
+  tm_scope: none
```
pchaigno: You'll need to change this line to `tm_scope: source.ampl` to use the grammar ;)

vitaut: Done in https://github.com/ampl/linguist/commit/51af1bd162eccaddf3f5d224dd4cbbbb3385d28e, thanks!

pchaigno:
@vitaut Should be good to go now! I opened issues for the three other extensions at #2168, #2169 and #2170.

@arfon This should be ready for you to review when you get the time :-)

arfon:
Thanks @pchaigno and @vitaut

arfon:
It will likely be another week or so before this is live on GitHub.

vitaut:
Cool, thanks @arfon and @pchaigno.

PR #2075, opened 2015-02-04
Add cpplint.py to vendor.yml

philix:
`cpplint.py` is Google's Python script used for linting C++ files.
I have a small C++ project with `cpplint.py` included mistakenly making
Python the main language of my project.


arfon:
:+1: makes sense. Thanks @philix.

It'll be another week or so before this change is live on GitHub.

Issue #2039, opened 2015-01-26
Prolog files are identified as Perl6 and Perl code


mrkkrp:
I've noticed that heavily commented Prolog code is often detected as Perl6 or just Perl. See [this repository](https://github.com/mrkkrp/lpnes) for example. If there is quite a bit of normal Prolog code, recognition works OK. As far as I know some Perl code is sometimes recognized as Prolog too.

Misrecognized files are easy to find because comments in them are not properly highlighted.

bkeepers:
There are currently some [very basic heuristics for Perl and Prolog](https://github.com/github/linguist/blob/0a5b5eadeb88a5d8ff9c8aeaad30011aa4a43479/lib/linguist/heuristics.rb#L89-L97), but looks like most of those files are all/mostly comments.

vim modeline support would also be helpful here (/cc #1210)

/cc @arfon

mrkkrp:
@bkeepers, well, if Linguist could understand Emacs (and Vi) mode lines, it would solve the problem (at least, for me and those who use the mode lines).

arfon:
PR in progress here: https://github.com/github/linguist/pull/2041

arfon:
@mrkkrp - Linguist does now indeed support Emacs and Vim-style modelines.

mrkkrp:
@arfon, I'm sorry, it doesn't seem to work. In [this repository](https://github.com/mrkkrp/lpnes) every file has the following (Emacs) mode line:
```
%% -*- Mode: Prolog; -*-
```
Emacs does recognize it, and as far as I know it's perfectly valid mode line. However, as you can see some files are still recognized by Linguist as Perl files. May be I'm missing something?

arfon:
Hi @mrkkrp - looks like we're requiring `mode` and not `Mode`: https://github.com/github/linguist/blob/6d770ab68f7aa9b4ea8a31c260b20cf962971daf/lib/linguist/strategy/modeline.rb#L4

@bkeepers @larsbrinkhoff perhaps we should change the regex to:

```
/-\*-\s*[mM]ode:\s*(\w+);?\s*-\*-/
```

mrkkrp:
It's case insensitive, so
```
%% -*- mOde: proLog; -*-
```
is also a valid mode line. So I guess `/-\*-\s*mode:\s*(\w+);?\s*-\*-/i` is good enough.

arfon:
[...]

PR #2083, opened 2015-02-06
Making modelines case-insensitive

arfon:
Addresses #2039 

/ cc @mrkkrp @bkeepers


bkeepers:
:+1:

aroben:

Issue #2081, opened 2015-02-06
xxx.init.d bash script detected as 'D' language


SjonHortensius:
As can be seen on https://github.com/Graylog2/graylog2-server/search?l=d&q=provides

the graylog2-server.init.d file (which has a correct shebang) is detected as being in the 'D' language, which is incorrect

arfon:
Hi @SjonHortensius - thanks for reporting this. The issue is that Linguist only knows about one language with that extension `D`: https://github.com/github/linguist/blob/6d770ab68f7aa9b4ea8a31c260b20cf962971daf/lib/linguist/languages.yml#L676 and so will automatically define it as `D`. If there were other languages defining the `.d` extension then we'd use the [shebang](https://github.com/github/linguist/blob/master/lib/linguist/shebang.rb) classification strategy (and would likely get the correct answer).

One possible fix here would be to define `.d` as `Shell` extension [here](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L2765-L2788).

As I side note, I was under the impression that `init.d` paths were usually directories in Unix/Linux systems but that might just be gross ignorance on my part :smile:

pchaigno:
This issue could be solved if we move the Shebang strategy to the top of the strategies to try (I suggested this before in #1991) .

larsbrinkhoff:
A cursory sweep suggests that out of 3000000 `.d` files, about 10% are actually in D.  I'm tempted to do something about that.

arfon:
@larsbrinkhoff - we'll so be doing [Shebang detection first](https://github.com/github/linguist/pull/2085) which will presumably help at some level here?

larsbrinkhoff:
@arfon It'll help if those init.d shell script do have shebang lines.  I just checked, and they seem to do -- great!

However,m a quick random sampling of search results seem to indicate that many of the .d files are Makefile dependency include files, and they don't contain any shebang.

arfon:
> However, a quick random sampling of search results seems to indicate that many of the .d files are Makefile dependency include files, and they don't contain any shebang.

OK, it might well be worth doing something about those then.

Issue #1991, opened 2015-01-12
Heavily commented shell script detected as groff


oldlaptop:
This (comment-heavy, but also explicitly shebanged as #! /bin/sh) shell script in one of my repos seems to be detected as groff: https://github.com/oldlaptop/uqm-crossbuilder/blob/master/config/includes.chroot/usr/lib/crossbuild/crossbuild_liblua5.1

This doesn't seem to be specific to Linguist, my editor (kate) also misidentifies it, as "Troff Mandoc".

arfon:
Perhaps I'm missing something but I can't seen any reference to Groff in the repo statistics here: https://github.com/oldlaptop/uqm-crossbuilder

![screen shot 2015-01-13 at 4 11 48 pm](https://cloud.githubusercontent.com/assets/4483/5730328/cab07428-9b3e-11e4-9171-9b3924a4aed2.png)

oldlaptop:
It's one 60-line shell script out of about a thousand lines and 25 or so scripts, that's apparently below the status bar's noise level. The file is listed as groff in the "detailed" search view (https://github.com/oldlaptop/uqm-crossbuilder/search?l=groff) and obviously is highlighted as groff.

arfon:
> It's one 60-line shell script out of about a thousand lines and 25 or so scripts, that's apparently below the stat bar's noise level. The file is listed as groff in the "detailed" search view (https://github.com/oldlaptop/uqm-crossbuilder/search?l=groff) and obviously is highlighted as groff.

Sorry I missed that. We currently support [overrides](https://github.com/github/linguist#overrides) for language reporting but these aren't used for [syntax highlighting](https://github.com/github/linguist/issues/1792) or search results yet I'm afraid.

pchaigno:
Has there ever been a case where the shebang strategy failed? This issue makes me think that maybe [the shebang strategy should be the first one to try](https://github.com/github/linguist/blob/master/lib/linguist/language.rb#L96-L100). It would require a bit of work since currently the first strategy is also generating the list of candidates but it could surely help solve this type of issue.

/cc @bkeepers

arfon:
Looks like this was fixed in https://github.com/github/linguist/issues/2085

PR #2085, opened 2015-02-07
Shebang first

arfon:
This PR moves the `Shebang` strategy above `Filename`:

```ruby
STRATEGIES = [
  Linguist::Strategy::Modeline,
  Linguist::Shebang,
  Linguist::Strategy::Filename,
  Linguist::Heuristics,
  Linguist::Classifier
]
```

IMO `Modeline` should still be before `Shebang` as it's an explicit statement of content by the user. This PR fixes https://github.com/github/linguist/issues/2081 and https://github.com/github/linguist/issues/1991

@pchaigno said in https://github.com/github/linguist/issues/1991:

> It would require a bit of work since currently the first strategy is also generating the list of candidates but it could surely help solve this type of issue.

I don't believe this is an issue as neither the `Modeline` or `Shebang` strategies rely upon 'incoming' languages like the [`Heuristics` does](https://github.com/github/linguist/blob/f5bc9735afee25446f537f6c4f895b80438c53cc/lib/linguist/heuristics.rb#L16). As the `Filename` strategy still precedes the `Heuristics` we should be OK.

/ cc @pchaigno @bkeepers

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -96,8 +96,8 @@ module Linguist
 
     STRATEGIES = [
       Linguist::Strategy::Modeline,
-      Linguist::Strategy::Filename,
       Linguist::Shebang,
+      Linguist::Strategy::Filename,
       Linguist::Heuristics,
       Linguist::Classifier
     ]
--- a/test/fixtures/Shell/crossbuild_liblua5.1
+++ b/test/fixtures/Shell/crossbuild_liblua5.1
@@ -1,0 +1,57 @@
+#! /bin/sh
+#
+# Builds and installs liblua5.1 for the cross toolchain.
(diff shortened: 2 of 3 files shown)


arfon:
Also, https://github.com/github/linguist/commit/88e79cd3a8bc7749783693a35ea4409a4709303a adds some fixtures that failed previously to this.

pchaigno:
> I don't believe this is an issue as neither the Modeline or Shebang strategies rely upon 'incoming' languages like the Heuristics does. As the Filename strategy still precedes the Heuristics we should be OK.

I forgot we don't need candidates for the Shebang strategy when I wrote that.
:+1: to this PR!

aroben:

PR #2087, opened 2015-02-07
Detection by extension made case-insensitive

pchaigno:
This is the implementation of an idea @larsbrinkhoff suggested in #2072.
With this pull request the search by file extensions is case-insensitive.
I wanted to see if it creates any issue with the detection.

It actually solves a few incorrect detections and I think it globally makes sense:
For instance, all `.e` files in the [Rosetta project](https://github.com/acmeism/RosettaCodeData) are detected as Eiffel but most of them are actually E code.
With this pull request the detection results get [a lot better](https://gist.github.com/pchaigno/7159e16ac9801507d60f).
In general, it looks like lower-case and upper-case extensions are regularly mixed. See #2072, #308, #268 or #1035 for examples.

### DIFF
--- a/lib/linguist/file_blob.rb
+++ b/lib/linguist/file_blob.rb
@@ -67,7 +67,7 @@ module Linguist
     #
     # Returns an Array
     def extensions
-      basename, *segments = File.basename(name).split(".")
+      basename, *segments = File.basename(name).downcase.split(".")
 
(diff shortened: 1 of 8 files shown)


arfon:
I've always thought this was weird. @bkeepers this seems like a good idea to me?

bkeepers:
:+1:

PR #2091, opened 2015-02-09
Add ColdC, the language from ColdMUD.

waywardmonkeys:
ColdC code is now landing in some repos in https://github.com/the-cold-dark/
among other possible places.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -531,6 +531,14 @@ CoffeeScript:
   interpreters:
   - coffee
 
+ColdC:
+  type: programming
+  color: "#3ebc27"
+  extensions:
+  - .cdc
+  ace_mode: text
+  tm_scope: none
+
 ColdFusion:
   type: programming
   group: ColdFusion
--- a/samples/ColdC/located.cdc
+++ b/samples/ColdC/located.cdc
@@ -1,0 +1,180 @@
+
+new object $located: $physical;
+
+var $described prose = [];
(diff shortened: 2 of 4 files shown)


waywardmonkeys:
https://github.com/search?utf8=%E2%9C%93&q=extension%3Acdc+object&type=Code&ref=searchresults

larsbrinkhoff:
I'm not a GitHub staff member, but I'll note that there is about 1600 files with the `.cdc`extension.  Of those, slightly more than half seem to be ColdC code.

pchaigno:
@waywardmonkeys I agree with @larsbrinkhoff (but I'm not from GitHub either).
I also want to point out that all the examples to which you pointed seem to come from the same repository ([the-cold-dark/ColdCore](https://github.com/the-cold-dark/ColdCore)). That will probably not be enough (see [the contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md));
> We try only to add languages once they have some usage on GitHub. In most cases we prefer that languages be in use in hundreds of repositories before supporting them in Linguist.

arfon:
:+1: @pchaigno, the vast majority of these files are [from a single repository](https://github.com/search?l=&q=extension%3Acdc+object+user%3Athe-cold-dark&ref=advsearch&type=Code&utf8=%E2%9C%93) and so I'm afraid `ColdC` doesn't have the level of adoption we require to add a new language to Linguist.

@waywardmonkeys - please feel free to open a new pull request in the future should `ColdC` adoption on GitHub increase.

Issue #2074, opened 2015-02-03
Plain perl mistaken for Perl6


kentfredric:
I was forwarded here by github relating to a failure to highlight this document: https://gist.github.com/kentfredric/04b64a80c5d90eea2924

Repeated attempt to indicate what language it was using the file extension always resulted in it thinking I was using Perl6.

Looking into your heuristics ( as per https://github.com/github/linguist/issues/2042 )  I find this: https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L54-L57

```ruby
disambiguate "Perl", "Perl6", "Prolog" do |data|
  if data.include?("use v6")
    Language["Perl6"]
  elsif data.include?("use strict")
    Language["Perl"]
  elsif data.include?(":-")
    Language["Prolog"]
  end
end
```

That perl heuristic is incredibly narrow and will exclude anything that is relying on strict to be injected by another pragma/module.

`use Moose` and `use Moo` both inject strict.
`use 5.xxx` injects strict where "xxx" can be converted to an integer that is 12 or greater.

Though it seems to me, a good heuristic to use in addition to `use strict` would be :

```
use\s+v?5\.
```

Which should catch any perl5 document that is using an explicit perl version requirement.

As would:
```
^#![^ ]*perl($|[^6])
```

( Because "perl" may never assume to mean "perl6", and may only imply perl5 )

bkeepers:
Thanks for reporting the issue!

> Though it seems to me, a good heuristic to use in addition to `use strict` would be…

Cool, I added that in https://github.com/github/linguist/pull/2092

> `^#![^ ]*perl($|[^6])`

The [shebang strategy](https://github.com/github/linguist/blob/master/lib/linguist/shebang.rb), which runs before the heuristics, should properly detect the perl version.

arfon:
This was fixed in https://github.com/github/linguist/pull/2092

PR #2092, opened 2015-02-09
Add  "use 5.xxx" to Perl Heuristic

bkeepers:
This updates the Perl heuristic to detect "use 5.xxx" as plain ol' Perl.

Fixes: https://github.com/github/linguist/issues/2074

/cc @kentfredric


arfon:
:+1:

aroben:

PR #2093, opened 2015-02-09
Call out fact that language updates are post-push

arfon:


arfon:
/ cc @bkeepers

bkeepers:
:+1: for adding more info about when/how we update the stats. This feels like it's not quite in the right spot though. Does this help them fix the misclassification?

Also, curious if this really warrants *emphasis*? Our old readme had a lot of random bold and emphasized phrases, which made none of them stand out.

Thoughts?

arfon:
> Does this help them fix the misclassification?

No but it's related... TBH I couldn't quite figure out where this should sit.

> Also, curious if this really warrants emphasis? Our old readme had a lot of random bold and emphasized phrases, which made none of them stand out.

Fair point. I'm happy to remove the emphasis (and also listen to suggestions about better places for this to go)

bkeepers:
In https://github.com/github/markup/pull/431 I added a section in the README on [how the library is integrated into GitHub.com](https://github.com/github/markup/pull/431/files#diff-04c6e90faac2675aa89e2176d2eec7d8L4). I wonder if it would make sense to add something similar here.

Issue #1020, opened 2014-03-26
SVG


jcubic:
I have project with [SVG icons](https://github.com/jcubic/Clarity) and it show that 100% of it is in shell.

arfon:
SVG is now detected by Linguist (https://github.com/github/linguist/pull/2097) but we've stuck with `data` for now which means it won't contribute to the language statistics. Please take a look at the discussion in https://github.com/github/linguist/pull/2097 for background.

jcubic:
Is it possible to add a way to set the file as language in .gitattributes?

jcubic:
@arfon Just found that one of my repo is in HTML. Is it possible to set data file as markup? Why HTML is detectable language but SVG is not?

Repo that have HTML: https://github.com/jcubic/sysend.js (include in stats)
Repo that have SVG: https://github.com/jcubic/clarity (not included in stats)

jcubic:
It seems it's possible, but I think that SVG is not marked even as data. You need to use:

```
*.svg linguist-detectable linguist-language=SVG
```

Without language it doesn't work, which probably mean that it's not detectable language.

Alhadis:
> Why HTML is detectable language but SVG is not?

HTML is designed to be hand-written, whereas SVGs are not. Moreover, SVG files tend to be quite large: it'd only take one or two detailed graphic files for a web project to be classified as an "SVG project" instead of, say, HTML or JavaScript. Needless to say, this would annoy a _lot_ of users…

> Without language it doesn't work, which probably mean that it's not detectable language.

You don't need the `linguist-language=SVG` attribute. The reason why jcubic/Clarity@dc9d3bf95ac8e0f95d66842e5717b7d9a13419fc didn't seem to work is because [repository stats are only refreshed when a change is pushed](https://github.com/github/linguist/blob/master/docs/how-linguist-works.md#how-linguist-works-on-githubcom). Hence why your follow-up commit, jcubic/Clarity@f17813def872a7595ed918fba005855c87bfb06c, appeared to "fix" the issue.

PR #2097, opened 2015-02-10
Detect all markup languages when computing language statistics

aroben:
Originally, only `programming` languages were included in repository language statistics. In 33ebee0f6a1e097550df0c7a69c5ffe57223f558 we started detecting a few selected `markup` languages as well. We didn't include all `markup` languages because at the time formats like Markdown and AsciiDoc were labeled as `markup` languages, and we thought that including those prose (i.e., non-code) languages in repository statistics on github.com was misleading for repositories that are largely about code but also contain a lot of documentation (e.g., https://github.com/rails/rails).

This hand-picked set of whitelisted `markup` languages can cause strange categorization for some repositories. For example, it includes CSS (and some variants) but not HTML. This results in repositories that contain the source code for a static website being classified as either a JavaScript (`programming`) or CSS (`markup`) repository, with no mention of HTML anywhere.

Fast-forward to today, and prose languages are no longer `markup` languages; they're now `prose` languages. So now we can include all `markup` languages in repository language statistics without worrying about undesirable effects for documentation-heavy repositories.

/cc @tnm @gjtorikian @bkeepers @arfon

### DIFF
--- a/lib/linguist/repository.rb
+++ b/lib/linguist/repository.rb
@@ -8,6 +8,8 @@ module Linguist
   # Its primary purpose is for gathering language statistics across
   # the entire project.
   class Repository
+    DETECTABLE_TYPES = [:programming, :markup].freeze
+
     attr_reader :repository
 
     # Public: Create a new Repository based on the stats of
@@ -159,8 +161,7 @@ module Linguist
           # Skip vendored or generated blobs
           next if blob.vendored? || blob.generated? || blob.language.nil?
 
-          # Only include programming languages and acceptable markup languages
-          if blob.language.type == :programming || Language.detectable_markup.include?(blob.language.name)
+          if DETECTABLE_TYPES.include?(blob.language.type)
             file_map[new] = [blob.language.group.name, blob.size]
           end
         end
--- a/lib/linguist/repository.rb
+++ b/lib/linguist/repository.rb
@@ -159,7 +159,7 @@ module Linguist
           blob = Linguist::LazyBlob.new(repository, delta.new_file[:oid], new, mode.to_s(8))
 
           # Skip vendored or generated blobs
-          next if blob.vendored? || blob.generated? || blob.language.nil?
+          next if blob.vendored? || blob.documentation? || blob.generated? || blob.language.nil?
 
           if DETECTABLE_TYPES.include?(blob.language.type)
             file_map[new] = [blob.language.group.name, blob.size]
--- a/lib/linguist/repository.rb
+++ b/lib/linguist/repository.rb
@@ -8,8 +8,6 @@ module Linguist
   # Its primary purpose is for gathering language statistics across
   # the entire project.
   class Repository
-    DETECTABLE_TYPES = [:programming, :markup].freeze
-
     attr_reader :repository
 
     # Public: Create a new Repository based on the stats of
@@ -158,12 +156,8 @@ module Linguist
 
           blob = Linguist::LazyBlob.new(repository, delta.new_file[:oid], new, mode.to_s(8))
 
-          # Skip vendored or generated blobs
-          next if blob.vendored? || blob.documentation? || blob.generated? || blob.language.nil?
-
-          if DETECTABLE_TYPES.include?(blob.language.type)
-            file_map[new] = [blob.language.group.name, blob.size]
-          end
+          next unless blob.include_in_language_stats?
+          file_map[new] = [blob.language.group.name, blob.size]
         end
       end
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
-Checking code you didn't write, such as JavaScript libraries, into your git repo is a common practice, but this often inflates your project's language stats and may even cause your project to be labeled as another language. By default, Linguist treats all of the paths defined in [lib/linguist/vendor.yml](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml) as vendored and therefore doesn't include them in the language statistics for a repository.
+Checking code you didn't write, such as JavaScript libraries, into your git repo is a common practice, but this often inflates your project's language stats and may even cause your project to be labeled as another language. By default, Linguist treats all of the paths defined in [lib/linguist/vendor.yml](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml) as vendored and therefore doesn't include them in the language statistics for a repository. Vendored files are also hidden by default in diffs on github.com.
 
 Use the `linguist-vendored` attribute to vendor or un-vendor paths.
 
@@ -43,7 +43,7 @@ special-vendored-path/* linguist-vendored
 jquery.js linguist-vendored=false
 ```
 
-Similar to vendored files, Linguist excludes documentation files from your project's language stats. [lib/linguist/documentation.yml](lib/linguist/documentation.yml) lists common documentation paths and excludes them from the language statistics for your repository.
+Similar to vendored files, Linguist excludes documentation files from your project's language stats. (Unlike vendored files, documentation files are displayed in diffs on github.com.) [lib/linguist/documentation.yml](lib/linguist/documentation.yml) lists common documentation paths and excludes them from the language statistics for your repository.
 
 Use the `linguist-documentation` attribute to mark or unmark paths as documentation.
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Linguist supports a number of different custom overrides strategies for language
 
 ### Using gitattributes
 
-Add a `.gitattributes` file to your project and use standard git-style path matchers for the files you want to override to set `linguist-language` and `linguist-vendored`.
+Add a `.gitattributes` file to your project and use standard git-style path matchers for the files you want to override to set `linguist-documentation`, `linguist-language`, and `linguist-vendored`.
 
 ```
 $ cat .gitattributes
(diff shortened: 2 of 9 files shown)


aroben:
I'd like to collect some data to see how this will change the categorization of various repositories, but I don't know how to do so. @tnm I think you've done this in the past; any tips?

arfon:
> I'd like to collect some data to see how this will change the categorization of various repositories, but I don't know how to do so. @tnm I think you've done this in the past; any tips?

@aroben - if you have a large corpus of repos you can use these rake tasks to quantify the changes: https://github.com/github/linguist/blob/master/Rakefile#L55-L114

aroben:
I now realize I can just use `bin/linguist` on a variety of repositories to get a sense of the effect.

aroben:
Hm, I don't understand this. Here's rails/rails before this change:

```
98.54%  Ruby
1.37%   CSS
0.08%   JavaScript
0.01%   Bison
```

And after:

```
99.01%  Ruby
0.90%   HTML
0.08%   JavaScript
0.01%   Bison
```

Why doesn't CSS show up anymore?

aroben:
Oh, CSS isn't a `markup` language. That seems like a mistake.

aroben:
OK, new "after" stats for rails/rails:

```
97.67%  Ruby
1.35%   CSS
0.89%   HTML
0.08%   JavaScript
0.01%   Bison
```

aroben:
Big (expected) improvement for https://github.com/diveintomark/diveintohtml5:

Before:

```
75.29%  JavaScript
16.23%  CSS
5.57%   Python
2.92%   Makefile
```

After:

```
83.88%  HTML
12.09%  JavaScript
2.61%   CSS
0.89%   Python
0.47%   Makefile
0.06%   ApacheConf
```

aroben:
Here's an interesting one: https://github.com/twbs/bootstrap. Before:

```
50.15%  CSS
48.44%  JavaScript
1.11%   Python
0.23%   Ruby
0.07%   Shell
```

After:

```
51.78%  HTML
18.89%  XML
14.71%  CSS
14.21%  JavaScript
0.33%   Python
0.07%   Ruby
0.02%   Shell
```

I don't think most people would say that twbs/bootstrap is an "HTML" or "XML" repository. The HTML/XML files are almost all in the `docs/` directory, so this is basically running into the problem we were trying to avoid by leaving `prose` languages out of the stats. (Some of the CSS/JS files are in the `docs/` directory as well, so they're already skewing the stats somewhat even without this PR.)

This probably indicates that we should be excluding documentation from repository stats, similar to how we exclude generated/vendored files today. If we were to do that we could probably include `prose` languages in the stats as well.

aroben:
Just to be concrete, here are the languages that this PR adds into the repository stats:

```ruby
irb(main):009:0> puts (Linguist::Language.by_type(:markup).map(&:name) - Linguist::Language.detectable_markup).join("\n")
ApacheConf
Eagle
HTML
HTML+Django
HTML+ERB
HTML+PHP
Haml
Handlebars
Jade
Kit
Latte
Liquid
MTML
Mask
Nginx
PostScript
Protocol Buffer
QML
RHTML
Scaml
Slim
Tea
Twig
Web Ontology Language
XML
```

aroben:
twbs/bootstrap probably wants `docs/`, `dist/`, and possibly `fonts/` excluded from its stats. This is achievable today using `.gitattributes`, but I wonder if there are some widely-applicable defaults that could be pulled into Linguist's logic.

aroben:
The `XML` files in Bootstrap are SVG files containing icon fonts. It feels kind of weird for them to be included in the stats since they're basically images (even though they happen to be represented in a textual format). I wonder if SVG should become its own language, classified as `data`, similar to Graph Modeling Language and Graphviz?

arfon:
There's an issue or PR to add SVG somewhere here. (I'm on my phone right now). I think we probably should add SVG and classify it as data.

aroben:
I suppose classifying SVG as data would be bad for repositories that are primarily SVG, like https://github.com/github/octicons. Though it's no worse than today, where SVG doesn't get classified at all.

aroben:
https://github.com/github/linguist/issues/1020 points to another primarily SVG repo where the author would like SVG to be included in the stats.

arfon:
> #1020 points to another primarily SVG repo where the author would like SVG to be included in the stats.

That's the one. This wouldn't address that issue then (although we would be adding SVG as a detectable language) - correct?

aroben:
Right. If we split SVG out from XML and mark it as `data`, SVG will continue not to be included in the language stats.

bkeepers:
:metal: I love this approach. Our special handling of markup/css has always bothered me.

We could also extend our `.gitattributes` support to do something like this:

```
*.svg linguist-type=data
```

aroben:
> ```
*.svg linguist-type=data
```

The only effect this would have would be to exclude these files from the language stats, right? In that case it's functionally the same as `linguist-vendored` I believe. Maybe it's still a nice thing to have because it gives Linguist more information to work with and perhaps we'd take advantage of it in the future.

**Comment by aroben** [comment]
If we classify SVG as `data` then twbs/bootstrap becomes:

```
63.80%  HTML
18.13%  CSS
17.51%  JavaScript
0.40%   Python
0.08%   Ruby
0.05%   XML
0.02%   Shell
```

I do think it would be interesting to try excluding `docs/` directories from repository language stats. That would change the stats to this:

```
41.93%  JavaScript
37.28%  CSS
19.65%  HTML
1.08%   Python
0.07%   Shell
```

What do others think? Is `docs/` a common enough pattern to warrant building it into Linguist? Is it really something we want to exclude?

**Comment by bkeepers** [comment]
> The only effect this would have would be to exclude these files from the language stats, right? 

Or to count something towards the language stats.

For example, if I had a repo that was a book and I wanted markdown to show up:

```
*.md linguist-type=markup
```

Although, once we get into other examples, this style doesn't make as much sense. For example, if I wanted my repo to show up as CSV because it's just an archive of data:

```
*.csv linguist-type=???
```

Maybe it would make more sense to say if we wanted to ignore something:

```
*.svg linguist-ignore
*.md linguist-ignore=false
*.csv linguist-ignore=false
```

**Comment by aroben** [comment]
I took a stab at adding `BlobHelper#documentation?`, similar to the existing `#vendored?` method. Common documentation paths are listed in `lib/linguist/documentation.yml`, but the defaults can be overridden with a new `linguist-documentation` Git attribute. Documentation files are excluded from repository language statistics.

This PR now classifies twbs/bootstrap like this:

```
41.93%  JavaScript
37.28%  CSS
19.65%  HTML
1.08%   Python
0.07%   Shell
```

(I suspect this actually might let us include `prose` languages in language statistics someday, once our documentation detection is comprehensive enough. This would let us categorize repositories like https://github.com/progit/progit2 correctly (well, once our AsciiDoc vs. AGS Script disambiguation is working). But let's just stick to `markup` languages for now.)

**Comment by aroben** [comment]
@arfon @bkeepers @pchaigno Any thoughts on the `documentation?` stuff I described above?

File: lib/linguist/repository.rb [hunk 1]
```diff
@@ -157,10 +159,9 @@ def compute_stats(old_commit_oid, cache = nil)
           blob = Linguist::LazyBlob.new(repository, delta.new_file[:oid], new, mode.to_s(8))
 
           # Skip vendored or generated blobs
-          next if blob.vendored? || blob.generated? || blob.language.nil?
+          next if blob.vendored? || blob.documentation? || blob.generated? || blob.language.nil?
 
-          # Only include programming languages and acceptable markup languages
-          if blob.language.type == :programming || Language.detectable_markup.include?(blob.language.name)
+          if DETECTABLE_TYPES.include?(blob.language.type)
```
bkeepers: Would it be worth wrapping this and the conditionals above into `BlobHelper#include_in_languages_stats?` or something like that?

```ruby
module BlobHelper
  def include_in_language_stats?
    !vendored? &&
    !documentation? &&
    !generated? &&
    language && DETECTABLE_TYPES.include?(language.type)
  end
end
```

aroben: Yeah that seems worthwhile. I'll add that.

aroben: Done.

bkeepers:
:+1: looks good to me.

aroben:
This should be a nice improvement for GitHub Pages sites like https://github.com/project-open-data/project-open-data.github.io:

Before:

```
81.50%  CSS
17.32%  Ruby
1.18%   Shell
```

After:

```
58.21%  HTML
20.53%  CSS
16.60%  XML
4.36%   Ruby
0.30%   Shell
```

pchaigno:
It's probably because I read the thread quickly but I don't see the difference between vendored files and documentation...?

aroben:
Vendored files:

* Are third-party content
* Have their diffs hidden by default on github.com

Documentation files:

* Are first-party content
* Have their diffs shown on github.com

pchaigno:
Okay, I see. It makes sense :-)
Should the `README.md` explain that?

aroben:
@pchaigno I just added some text to the readme to explain the difference with respect to diffs.

pchaigno:
Looks good!
:+1: to this PR; it will definitly make things more coherent.

arfon:
:zap:

aroben:

larsbrinkhoff:
I now see that source code repositories with plenty of documentation can be classified as e.g. HTML, and will no longer be found in:
- `language:` searches
- trending lists

larsbrinkhoff:
Not a solution, but maybe a new `programminglanguage:` search term would alleviate the problem.

aroben:
> I now see that source code repositories with plenty of documentation can be classified as e.g. HTML

Part of the work I did in this PR was to exclude documentation files from the language statistics for exactly this reason. I'm not surprised at all that there are documentation files that are sneaking through, though. Could you open an issue pointing to the repositories that are having trouble so we can take a look and fix it?

larsbrinkhoff:
@aroben Happy to: #2138

jcubic:
Is it possible to mark SVG as language in .gitattributes file?

arfon:
> Is it possible to mark SVG as language in .gitattributes file?

Unfortunately not.

Issue #2067, opened 2015-02-02
syntax highlighting doesn't work properly (readme with AS3 code snippets)


honzabrecka:
Hi,

I have a [readme file](https://github.com/honzabrecka/Screenshot/blob/master/README.md) that contains some code snippets in AS3 (each snippet has `as3` language identifier) but properly highlighted is only [the last one](https://github.com/honzabrecka/Screenshot/blob/master/README.md#custom-assertion).

Can someone take a look please?

egonSchiele:
My docs (generated with gh-pages) have ruby code that isn't getting highlighted either. Related issue?

Link: http://egonschiele.github.io/contracts.ruby/

pchaigno:
@honzabrecka The incorrect highlighting comes from [the grammar GitHub is using to highlight ActionScript](https://github.com/simongregory/actionscript3-tmbundle). I was able to reproduce the issue in [Lightshow](https://lightshow.githubapp.com/?utf8=%E2%9C%93&scope=source.actionscript.3&grammar_url=&grammar_text=&code_source=from-text&code_url=&code=%5BTest%28async%2C+ui%29%5D%0D%0Apublic+function+defaultColor%28%29%3Avoid%0D%0A{%0D%0A%09var+component%3ASquare+%3D+new+Square%28%29%3B%0D%0A%09Async.proceedOnEvent%28this%2C+component%2C+Event.ADDED%29%3B%0D%0A++++UIImpersonator.addChild%28component%29%3B%0D%0A%09%0D%0A%09%2F%2F+this+is+the+line!%0D%0A%09assertTrue%28Screenshot.compare%28%22SquareTest.defaultColor%22%2C+component%29%29%3B%0D%0A}).

[...]

PR #2100, opened 2015-02-10
change submodule url for proper AS3 syntax highlighting

honzabrecka:
Current grammar doesn't work well with. It properly color neither code written in Flash Professional nor well identified code snippets in README (discussed in #2067). It works only if the complete class is defined, which can be useful for text editor, but not for GitHub's purposes. I think that it should work as expected, otherwise it's useless feature.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -249,7 +249,7 @@
 	url = https://github.com/shellderp/sublime-robot-plugin
 [submodule "vendor/grammars/actionscript3-tmbundle"]
 	path = vendor/grammars/actionscript3-tmbundle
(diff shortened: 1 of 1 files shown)


arfon:
:+1:

aroben:

honzabrecka:
Do you plan to update this submodule to the latest revision, so the change will take effect?

aroben:
Hm, I updated all submodules as part of https://github.com/github/linguist/issues/2114. I don't know why this one didn't get updated. I just opened https://github.com/github/linguist/pull/2133 which updates all the submodules again, including this one.

PR #2102, opened 2015-02-11
Added a sample of a C header file that is currently recognized as C++

phillipberndt:
As per @DX-MON's request, see https://github.com/github/linguist/issues/1626#issuecomment-73870081

Taken from https://github.com/phillipberndt/pqiv/blob/469fe63df4b27463f8accf8a45ae515a17c202e7/pqiv.h
I am the file's original author.


arfon:
:+1:

PR #2112, opened 2015-02-13
Mark XS files as "type: programming"

nwellnhof:
I'd actually prefer to treat Perl XS files as C code, but this fix shouldn't be controversial.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3257,6 +3257,7 @@ XQuery:
   ace_mode: xquery
 
 XS:
+  type: programming
   extensions:
   - .xs
   tm_scope: source.c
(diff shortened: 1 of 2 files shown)


pchaigno:
Hey @nwellnhof - could you also add a few sample files for XS in `samples/XS/`?

nwellnhof:
@pchaigno Sure. Are there any restrictions regarding the licenses of the sample files?

pchaigno:
Yep, they need to have an [OSS license](http://choosealicense.com/).

nwellnhof:
@pchaigno I added an XS file written by me. I'm wary to add other XS files because I'm not 100% sure about licensing implications. XS files are typically dual-licensed under the GPL and Artistic License. It seems to me that the linguist repo should contain at least the full text of these licenses if files under these licenses are added.

arfon:
:+1:

Issue #1572, opened 2014-10-02
Bayesian classifier sample size is far too small


fingolfin:
Linguist is using a bayesian classifier to detect languages. However, the sample data provided to it is far too small. According to `wc -l samples/*/* there are just 309505 lines of sample code for about 220 "languages", making for an average of 1400 lines of sample code for each language.

I am surprised that despite this, linguist manages to work reasonably well -- my guess is that this is mostly due to the fact that data is pre-binned based on file extensions (fair enough). But as soon as more than one language uses the same extension, detection rates break down, as the tons of reports on Objective C vs C++ vs C show.

Which is not surprising -- the current approach linguist takes is akin to showing a spam filter half a dozen spam mails, no regular mails, and then expecting it to work well.

So, if linguist wants to keep using a Bayesian filter, it seems to me that the sample size should be enlarged substantially.

An alternative would be to use more hand-crafted heuristics, as has been suggested by many people on many issue reports and pull requests, but that has its drawbacks, too -- on the one hand, heuristics are hard to maintain, and on the other, they are surprisingly hard to get right, too -- most people are pretty bad at doing it. This is indeed something were Bayesian classifiers usually outperform humans. But only if fed with enough sample data...

arfon:
We're planning on working on this extensively over the coming months https://github.com/github/gsoc#linguist so closing this out. Thanks for your thoughts @fingolfin.

PR #2117, opened 2015-02-15
Samples from Rosetta Code project

pchaigno:
This pull request adds more than 11,000 sample files from [the Rosetta Code project](http://rosettacode.org). I wrote a [small Python script](https://gist.github.com/pchaigno/3905c8cecd5951f982c5) to copy the samples from [RosettaCodeData](https://github.com/acmeism/RosettaCodeData).

~~It also adds new extensions:~~
- ~~[`.ats`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aats+NOT+djhfdkjfhjbffjhdejdbnsjd&type=Code&ref=searchresults) for ATS~~
- ~~[`.idl`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aidl+NOT+djhfdkjfhjbffjhdejdbnsjd&type=Code&ref=searchresults) for IDL~~
- ~~[`.pascal`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Apascal+NOT+djhfdkjfhjbffjhdejdbnsjd&type=Code&ref=searchresults) for Pascal~~
- ~~[`.psh`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Apsh+NOT+djhfdkjfhjbffjhdejdbnsjd&type=Code&ref=searchresults) for PowerShell~~

~~and new interpreters:~~
- ~~[`dart`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+dart+extension%3Adart&type=Code&ref=searchresults) for Dart~~
- ~~[`racket`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+racket+extension%3Arkt&type=Code&ref=searchresults) for Racket~~
- ~~[`io`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+io+extension%3Aio&type=Code&ref=searchresults) for Io~~
- ~~[`factor`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+factor+extension%3Afactor&type=Code&ref=searchresults) for Factor~~
- ~~[`ocaml`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+ocaml+extension%3Aml&type=Code&ref=searchresults) for OCaml~~
- ~~[`rune`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+rune+extension%3Ae&type=Code&ref=searchresults) for E~~
- ~~[`csi`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+csi+extension%3Ascm&type=Code&ref=searchresults) for Scheme~~
- ~~[`runhaskell`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+runhaskell+extension%3Ahs&type=Code&ref=searchresults) for Haskell~~
- ~~[`instantfpc`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+instantfpc+extension%3Apas&type=Code&ref=searchresults) for Pascal~~
- ~~[`gosh`](https://github.com/search?utf8=%E2%9C%93&q=usr+bin+env+gosh+extension%3Ascm&type=Code&ref=searchresults) for Scheme~~
- ~~[`r6rs`](https://github.com/search?utf8=%E2%9C%93&q=r6rs+extension%3Ascm&type=Code&ref=searchresults) for Scheme~~

When the new samples had extensions or interpreters that were not widely used I changed them or removed the files.


bkeepers:
Brilliant! I love the idea of using known samples like this.

I think I remember a conversation that our system team had concerning the memory usage of the linguist classifier (since it loads `samples.json` into memory when linguist is required). In a quick test on my machine, a fresh ruby process that has only required `linguist` takes up ~52MB in master and ~68MB with this branch.

@vmg @dbussink Is this a concern?

Since the classifier always runs against a known set of languages, we could store the sample data in separate files per language and lazily load it. Repositories typically have a lot of the same type of data, so we could use some kind of LRU cache to avoid churn.

The other option would be to implement (or find an existing implementation of) a classifier in C, keeping memory management out of Ruby land.

/cc #1572

lazywei:
How is this PR going?

> The other option would be to implement (or find an existing implementation of) a classifier in C, keeping memory management out of Ruby land.

I think one possible solution is using the well-known liblinear / libsvm. The reasons:

(0) As mentioned, it's a good practice to generate features in advance and save the features into separate files. And using libsvm / liblinear can help with this: we can save the features into libsvm format, which is really useful for sparse data (e.g., bag-of-word). The loading and saving efforts will be reduced.
(1) liblinear / libsvm separates the training and predicting stages. That is, we can pre-train a SVM (or Logistic Regression) classifier and save its model file as we save the features in (0). Unless the user want to train their model for their own, they can just use our pre-trained classifier. The model file of libsvm / liblinear is very small (only saving information such as support vectors).
(2) The predicting stage of SVM or LogReg is very simple, which means we can even predict by ourself without using libsvm / liblinear as long as we have model file in (1). This implies that we can avoid extra external dependencies if the user is satisfied with our pre-trained classifier.

How do you think?

larsbrinkhoff:
I guess the Forth heuristics doesn't match all of the samples?

pchaigno:
@larsbrinkhoff Yes, that is why I reduced the scope in 7d6650a.

pchaigno:
@arfon Is the memory usage of Linguist still an issue? If it is, maybe we could open an issue to start discussing possible solutions to overcome it...?

arfon:
> @arfon Is the memory usage of Linguist still an issue? If it is, maybe we could open an issue to start discussing possible solutions to overcome it...?

I'll defer to @vmg on this one. @vmg - this PR is attractive as it adds a large, reliable source of samples data...but by doing this we're likely to increase the memory usage of Linguist (see @bkeepers [comment](https://github.com/github/linguist/pull/2117#issuecomment-74776022) above). 

I know that we had significant memory pressure in the past which made this a no-go but is this still the case?

larsbrinkhoff:
@pchaigno I don't know about adding zillions of samples, but this PR has so many other goodies.  Please don't forget about it.

Alhadis:
I'd love to know the size difference of converting spaces to tabs (the way nature intended).

`/optionated prick`.

brandonblack:
> Since the classifier always runs against a known set of languages, we could store the sample data in separate files per language and lazily load it.

This sounds like a really sensible approach to me. @pchaigno what do you think about re-working this effort to lazy load in separate language files?

PR #2118, opened 2015-02-15
Made assembly color more noticeable

xbony2:
It's bothered me for a while; the Java and Assembly color look almost exactly alike. [Here's](https://imgur.com/XayyV0L) a picture for a comparison.

Anyway, this pull request changes the Assembly color from #a67219 to [#6E4C13](http://www.colorpicker.com/6e4c13), which makes it look darker and more distinctive compared to Java.


bkeepers:
:metal: thanks!

PR #2121, opened 2015-02-16
Rename Nimrod to Nim

def-:
The language renamed about 2-3 months ago: http://nim-lang.org/


pchaigno:
Duplicate of #1936 ;)

Issue #2042, opened 2015-01-26
Errors when searching for the correct heuristic rule to use


pchaigno:
I found an issue with the heuristics and the way it looks for the right heuristic to use.
In the [Heuristic strategy](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L54-L57), these lines checks if a certain heuristic matches the candidate languages:
```ruby
# Internal: Check if this heuristic matches the candidate languages.
def matches?(candidates)
  candidates.any? && candidates.all? { |l| @languages.include?(l.name) }
end
```
It basically says that all candidate languages must be in the list of language this heuristic is trying to distinguish.

0. So if one of the candidate languages is not in the list, the heuristic will not be used.
0. It also means that, if the heuristic was written to distinguish more languages than we have candidates, it will still be used.

At the time we write the heuristic we generally make sure that the first case doesn't happen. For example when we wrote the heuristic for AsciiDoc and AGS Script (`.asc files`) they were the only two languages using the `.asc` extension. But since, Public Key has been added with the `.asc` extension. Thus the heuristic for AsciiDoc and AGS Script is not used anymore when trying to distinguish `.asc` files. This is why the issue #1638 is still opened.
We also wrote a heuristic rule for Prolog and IDL on `.pro` files. But now this file extension is also defined for [QMake](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L2417), explaining why [so many `.pro` files are still not correctly detected](https://github.com/search?l=idl&q=extension%3Apro+is&ref=searchresults&type=Code&utf8=%E2%9C%93).

The second point explains the issue with [these Perl files](https://github.com/rakudo/rakudo/search?l=prolog) (reported by @hoelzro in #1680). The `.pm` extension causes a conflict between Perl 5 and 6. Then Linguist uses the heuristic for Perl, Perl6 and Prolog because it includes all the candidate languages it's trying to distinguish. Unfortunately, ([because our heuristic doesn't check for comments](https://github.com/rakudo/rakudo/blob/421176c14605dcca7d339aec2d61aaee329b9f6d/src/core/io_operators.pm#L162)) Prolog is chosen as the correct language. Even if it wasn't a candidate for `.pm` files...

I'm not really sure on how this should be fixed. I also don't think I have enough experience with Ruby to write a test that detects these cases happening (I tried quickly but hit some specificities to Ruby's OOP).

/cc @bkeepers @arfon

arfon:
[...]

PR #2123, opened 2015-02-17
Some .for files are text.

larsbrinkhoff:
As of late, many `.for` files have been created which are neither FORTRAN nor Forth:
https://github.com/search?q=extension%3Afor+weekly&type=Code

The contents of these files are identical.  I added it as a `Text`sample.


pchaigno:
There isn't currently a heuristic rule to distinguish Text from FORTRAN and Forth. Because of #2042 if we add `.for` for Text the heuristic for FORTRAN and Forth won't be used anymore. It seems risky for only 107 files :/

larsbrinkhoff:
@pchaigno Sorry, I added Text to the heuristic rule for `.for`, but lost it before I made this pull request.

I'll update.  Do you still object?

pchaigno:
I'm not objecting, I just think we should be careful :-)
And we probably should fix (or at least clarify) the situation with #2042 before. Currently, if you add Text to the Forth and FORTRAN heuristic, `.F` files may be recognized as Text :/

larsbrinkhoff:
> `.F` files may be recognized as Text :/

I don't think that can happen (at this moment), because the heuristics never return Text.

But you are right that we should be careful.  Unless this PR is shot down, I'll request 1000 samples and check the results before and after.

pchaigno:
> I don't think that can happen (at this moment), because the heuristics never return Text.

Oh, you're right; I misread you :/
If it works with the classifier, perfect! You'll have to test that the new Text sample doesn't break other Text detections though.

larsbrinkhoff:
I'm all for checking thouroughly! :-)  Just awaiting the words of the elders.

bkeepers:
I'm curious what these `.for` files actually are. It looks like some data format. I wonder if we should be adding another definition for it.

larsbrinkhoff:
Me too.  I believe the original file came from here: http://www.cpc.ncep.noaa.gov/data/indices/

I emailed the contact at that page, but haven't got any reply yet.

larsbrinkhoff:
Tested with 1000 `Text` samples: no change in classification.

Tested with 1000 `.for`samples:
| | Before | After
|---|---|---
|FORTH|577|577
|Text|0|415
|Forth|423|8

I checked that those Forth files were indeed misidentified.

larsbrinkhoff:
Received word from the original author:

> ".for" is short for "formatted".

larsbrinkhoff:
According to http://www.weather.gov/disclaimer, the sample file is in the public domain.

larsbrinkhoff:
Another repository with plenty of `.for` data files:
https://github.com/OpenSceneryX/Library/search?q=extension%3Afor

larsbrinkhoff:
Ping @bkeepers @arfon.  Any thoughts?

arfon:
> Ping @bkeepers @arfon. Any thoughts?

I think we should probably merge this as it's a growing issue (178 files now). I'd really like to get https://github.com/github/linguist/issues/2042 resolved soon too.

larsbrinkhoff:
The CI build is erroring now, because of

> apt-get --assume-yes download libicu48 libicu-dev
> Err Downloading libicu48 4.8.1.1-3ubuntu0.1
> 404 Not Found [IP: 91.189.91.15 80]
> Err Downloading libicu-dev 4.8.1.1-3ubuntu0.1
> 404 Not Found [IP: 91.189.91.15 80]
> ar p vendor/apt/*.deb data.tar.gz
> tar -vzxC vendor/debs --strip-components=2
> ar: vendor/apt/*.deb: No such file or directory

arfon:
> The Travis CI build is erroring now, because of

@aroben - I had a look at fixing this earlier and didn't find an obvious solution. It looks like we need an `apt-get update` before the install but we can't do [the recommended](http://docs.travis-ci.com/user/common-build-problems/#Linux%3A-apt-fails-to-install-package-with-404-error.) as we can't use sudo on container builds.

Perhaps we should add the [`libicu48` deb](http://packages.ubuntu.com/precise/libicu48) to a local directory and install from there?

larsbrinkhoff:
> I think we should probably merge this as it's a growing issue

So, anything else I can do to facilitate this?

bkeepers:
I can't wait until we come up with a way to return "unknown" for files that don't have a high score from the classifier.

I don't love that we are classifying this as `Text`. Should we add a `Formatted` entry? Is there a spec for the file format anywhere?

larsbrinkhoff:
> I can't wait until we come up with a way to return "unknown"

Me too!

> I don't love that we are classifying this as Text. Should we add a Formatted entry? Is there a spec for the file format anywhere?

Agreed.  From looking at various `.for` files, I'm quite sure there is no spec.  Or maybe the spec would be "numbers formatted into whatever format looks pretty, plus some text that looks nice".

But anyway, I was thinking maybe something like this:

    Formatted Data:
      type: data
      group: Text # maybe?
      extensions:
      - .for
      # maybe .f too, sigh

larsbrinkhoff:
@arfon @bkeepers Added a commit to introduce a new `Formatted` language.  Does this look ok?

arfon:
This looks good to me. @bkeepers do you think we should add `group: Text`

bkeepers:
Looks good to me :+1: 

> do you think we should add `group: Text`?

No, it doesn't seem like this should be text to me. It's data, right?

arfon:
>  It's data, right?

Data encoded in plain text :-\

arfon:
Seriously though, I don't care that much. I could be convinced either way.

larsbrinkhoff:
I merged the two commits into one.  Leaving out `group: Text`.

arfon:
@larsbrinkhoff - probably won't cut a new gem until early next week.

larsbrinkhoff:
That's fine.  Thanks!

PR #2143, opened 2015-02-21
Classify Brainfuck as a programming language and include samples

CodingAnarchy:
The search results of usage on GitHub:

https://github.com/search?q=language%3A+brainfuck&ref=searchresults&type=Repositories&utf8=%E2%9C%93


pchaigno:
@CodingAnarchy The Travis build is failing because of [this test](https://github.com/github/linguist/blob/master/test/test_language.rb#L159).

@josh Is there a reason for Brainfuck to be classified as `nil`?

CodingAnarchy:
I understand that Brainfuck isn't a popular language, as esoteric as it is, but to have the files just be ignored in the repositories they are a part of is frustrating, at least in my perspective.

gjtorikian:
I've brought your commits into https://github.com/github/linguist/pull/2153, which I opened around the same as this PR. #2153 is, like, a superset of this PR, because it turns out many languages are lacking types.

CodingAnarchy:
Works for me.  Thanks!

PR #2145, opened 2015-02-21
Add MUF - Multi-user Forth

larsbrinkhoff:
This adds MUF - Multi-user Forth, which uses both `.m` and `.muf`.

Searches find about 1000 files:
https://github.com/search?q=extension%3Am+dup+swap
https://github.com/search?q=extension%3Amuf+dup+swap

The `.muf` extension seems almost exclusively used by MUF.  However, `.m` may be the most overloaded extension of all.  Will check carefully.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -178,11 +178,13 @@ module Linguist
       end
     end
 
-    disambiguate "M", "Mathematica", "Matlab", "Mercury", "Objective-C" do |data|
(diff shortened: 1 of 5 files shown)


larsbrinkhoff:
Verified with 1000 `.m` samples.  All files were classified the same before and after my changes.  (There were no MUF files.)

arfon:
:+1: thanks @larsbrinkhoff

Issue #2148, opened 2015-02-21
Recognize `.pb.go` files as generated


bdarnell:
Currently files generated by protocol buffers are only recognized in C++, Java, and Python; I'd like to have (at least) Go added to that list. I think that the naming convention of `*.pb.go` is sufficient to match these files, although if you want to match a comment within the file (as is done for the existing three languages), note that there are a couple of different variations of the initial comment (`protoc-gen-go` vs `protoc-gen-gogo`), neither of which is the same as in the existing languages:

```
// Code generated by protoc-gen-gogo.
// source: internal.proto
// DO NOT EDIT!
```
or
```
// Code generated by protoc-gen-go.
// source: google/protobuf/compiler/plugin.proto
// DO NOT EDIT!
```

Sample file: https://github.com/cockroachdb/cockroach/blob/4e789d0259c8336556d1b4bf883927efd2edd963/proto/api.pb.go

pchaigno:
#2152 should fix this.

bdarnell:
Thanks!

arfon:
Thanks @pchaigno. @bdarnell - this change should be live on GitHub in the next couple of weeks.

PR #2152, opened 2015-02-22
Detect Go files generated by Protocol Buffers

pchaigno:
This pull request adds detection for Go files generated by Protocol Buffers as asked for in #2148.


arfon:
:+1:

PR #2153, opened 2015-02-22
Test that languages have a type

gjtorikian:
This adds a test to enforce that every language has a `type`. It also adds types for languages missing them.


pchaigno:
This could probably replace #2143 (except for the sample files that could still be interesting to add).

gjtorikian:
I think that's a good call. I've plucked the commits adding the sample into this PR.

Sorry for the short OP. We talked about this internally and think that enforcing type is the way to go. I also noticed the Brainfuck issue about a month ago.

pchaigno:
> and think that enforcing type is the way to go.

Completely agree!

test/test_language.rb
```diff
@@ -358,6 +354,15 @@ def test_all_languages_have_grammars
     assert missing.empty?, message
   end
 
+  def test_all_languages_have_type
+    missing = Language.all.select { |language| language.type.nil? }
+    message = "The following languages' types are not listed in grammars.yml. Please add types for all new languages.\n"
+
+    width = missing.map { |language| language.name.length }.max
+    message << missing.map { |language| sprintf("%-#{width}s %s", language.name, language.tm_scope) }.sort.join("\n")
```
bkeepers: Is the `tm_scope` needed for this message?

Issue #2116, opened 2015-02-14
Says KiCad Project file is Prolog


HSchmale16:
Your language recognition software says that a kicad project file is a prolog file, yet it's closer to MS-INI. This should be fixed as a lot of software development tools use MS-INI format to describe projects.

pchaigno:
Hi @HSchmale16 - Do you have a link to the repository where this issue is occurring? What extension are KiCad project files using?

WyldeOscar:
hey





Wylde Oscar Enterprises
Oscars Adult Toys & Lingerie <http://oscarsadulttoys.com>
Edmonton,AB
Canada

HSchmale16:
https://github.com/HSchmale16/CustomComputerChair
The extension is `.pro`, There's a string right at the begining of the file that can be used to id kicad project files.

pchaigno:
I think the solution here will be to add `.pro` as a INI extension. Then, KiCad project files will be recognized and highlighted as INI. Anything against that?

HSchmale16:
Wouldn't that affect prolog detection too.

pchaigno:
We already have a [heuristic rule for the `.pro` extension](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L110-L116). I will extend it to classify KiCad files based on the line that you mentioned, at the beginning of the file.

HSchmale16:
Yeah, that would work. I was just reading the project file, and you're going to want to search for `[eeschema]` or `pcbnew` in a set of angle brackets.

pchaigno:
#2154 should fix this.

arfon:
Hi @HSchmale16 - https://github.com/github/linguist/pull/2154 has just been merged and a new version of Linguist should be live on GitHub in the next week or so.

PR #2154, opened 2015-02-22
Support for .pro INI files (KiCad project files)

pchaigno:
This pull request adds support for `.pro` INI files (used as KiCad project files). This was asked for in #2116.

I also updated the heuristic for `.pro` to include both INI and QMake files. As discussed in #2042, currently this heuristic rule isn't used at all (because QMake is missing). This pull request fixes it.


arfon:
:+1: thanks @pchaigno

PR #2160, opened 2015-02-25
Use correct VCL syntax highlighting

brandonwamboldt:
GitHub current uses Perl as it was "close enough" but I think we should use proper syntax highlighting :)

I've added my own repo which is a Sublime Text plugin for Syntax highlighting (MIT licensed). My Nginx one is currently used so I assume the VCL one should work the same.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -630,3 +630,6 @@
 [submodule "vendor/grammars/Lean.tmbundle"]
 	path = vendor/grammars/Lean.tmbundle
 	url = https://github.com/leanprover/Lean.tmbundle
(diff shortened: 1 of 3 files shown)


brandonwamboldt:
Travis CI doesn't appear to be reporting the actual error that occurred?

brandonwamboldt:
FYI this works on lightshow

brandonwamboldt:
Finally fixed all the issues (was using an old commit before I added the license and had ordering wrong). Should be good to go :smile:

arfon:
:+1: thanks @brandonwamboldt. This should be live on GitHub in the next week or so.

Issue #2155, opened 2015-02-22
nginx syntax highlighting error


plentz:
There's a problem with nginx conf files where comments sometimes isn't proper highlighted. If ou look here https://gist.github.com/plentz/6737338#file-nginx-conf-L6-L9 some comments aren't displayed in grey. But looking at the same file inside sublime text using nginx syntax, it looks okey.

![nginx](https://cloud.githubusercontent.com/assets/4867/6321025/d4a1bc54-bacd-11e4-948a-10cc8ae32be0.png)

maybe related with #2125

arfon:
Hi @plentz - we use [this grammar](https://github.com/brandonwamboldt/sublime-nginx) to highlight files on GitHub. Would you mind opening an issue over there? Once the bug is fixed in the grammar we can pull the changes in here.

pchaigno:
@arfon Except this time, the grammar works fine in Sublime Text :/ (I checked with the last dev version). This issue is very close to #2125 so it's probably best to follow it there.

plentz:
@pchaigno thanks :)

brandonwamboldt:
Turns out I had an error in my regex and sublime is a bit more forgiving (see my fix: https://github.com/brandonwamboldt/sublime-nginx/commit/872afddd2d61acae85560fa009658257f4030244). I will submit a PR to update the submodule here :)

arfon:
Ok thanks for the follow up. Does that mean this issue can be closed?

brandonwamboldt:
@arfon Once my PR gets created and merged :)

PR #2161, opened 2015-02-25
Update sublime-nginx submodule to fix highlighting bug

### LINKED ISSUE #2125: nginx highlighting of access_log vs error_log
It seems `access_log` is not highlighted whereas `error_log` does get highlighted. Not sure if this is intentional or has to do with them being defined slightly different in the vendor grammar file.

Grammar definitions:
* [access_log](https://github.com/brandonwamboldt/sublime-nginx/blob/master/Syntaxes/nginx.tmLanguage#L678-L700)
* [error_log](https://github.com/brandonwamboldt/sublime-nginx/blob/master/Syntaxes/nginx.tmLanguage#L241-L263)

brandonwamboldt:
Updating my sublime-nginx plugin as I fixed a bug with the syntax highlighting (https://github.com/brandonwamboldt/sublime-nginx/commit/872afddd2d61acae85560fa009658257f4030244).

Fixes #2125 and #2155.


arfon:
This will automatically be pulled in with the next Linguist update (so no need for the PR)

Issue #1799, opened 2014-12-01
Please use a heuristic to determine if ".inc" files are php files


ejc3:
There is a large body of code in my organization's repository that uses .inc to represent php code that should not be executed directly by the php yapache module. 

Could you add a heuristic that will look through .inc files for <?php and trigger php mode for them? Thanks!

arfon:
Thanks for reporting this. We're actually working on this issue over in https://github.com/github/linguist/pull/1770

PR #2163, opened 2015-02-25
Use PHP language detection on .inc files

brandonwamboldt:
It's pretty common for PHP projects to have `.inc` files which are PHP, so we should apply heuristic detection against .inc files for PHP detection.

Fixes https://github.com/github/linguist/issues/1799

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2171,6 +2171,7 @@ PHP:
   - .aw
   - .ctp
   - .fcgi
+  - .inc
   - .php3
   - .php4
   - .php5


pchaigno:
I'm working on adding this extension for all languages that use it (see #1770).
I'll open a pull request as soon as I have some time to finish to sort things out ;)

arfon:
Yep, we want to fix this @brandonwamboldt but unfortunately there are lots of `.inc` files that aren't PHP.

@pchaigno - anything I can do to help?

brandonwamboldt:
@arfon sorry, I thought any duplicate file extensions were passed through a language heuristic system to determine the actual contents.

arfon:
In theory you're absolutely correct. The issue here is that we don't have the `.inc` extension specified for all the languages that need it so with this PR a bunch of stuff that isn't PHP would be incorrectly classified.

pchaigno:
> @pchaigno - anything I can do to help?

You already sent me sample files :-)
I'll try to finish it this week.

PR #2164, opened 2015-02-25
Exclude `examples` folder

alvarotrigo:
I would say its even more common than the Samples folder (see Reveal.js, Three.js, Express, Meteor, Backbone...)

I was forced to use a `.gitattributes` to exclude the "examples" folder from the language stats, as my whole project was marked as HTML instead of Javascript.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -210,6 +210,9 @@
 # Samples folders
 - ^[Ss]amples/
 
+# Examples folders
+- ^[Ee]xamples/
+
 # LICENSE, README, git config files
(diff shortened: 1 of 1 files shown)


arfon:
@aroben - would this be a good candidate for inclusion in the [`documentation.yml`](https://github.com/github/linguist/blob/master/lib/linguist/documentation.yml)?

aroben:
Yes, I think so. We should probably move the existing "samples" rule there, and get rid of the duplication of rules for README/LICENSE etc.

arfon:
Closing in favour of #2541

Issue #2140, opened 2015-02-20
Specifying a language in .gitattributes doesn't affect syntax highlighting


deviantintegral:
Most of my Github projects are Drupal projects, so we have a ton of custom extensions containing PHP code. There's quite a few issues for adding support, but most of them are blocked on conflicts over the .inc extension.

According to the Linguist readme, we should be able to manually specify file masks to mark language types. While Linguist on my local properly picks up the .module and .inc extensions, Github is still not detecting the files as PHP code in search or in syntax highlighting.

* [An example repository](https://github.com/deviantintegral/linguist-module)
* [Linguist output](https://github.com/deviantintegral/linguist-module/blob/master/README.md)
* [Search results that are missing .inc and .module files](https://github.com/deviantintegral/linguist-module/search?l=php)
* [The .gitattributes file](https://github.com/deviantintegral/linguist-module/blob/master/.gitattributes)

Is this simply not supported as a feature, a bug, or am I not setting up the .gitattributes file correctly?

geerlingguy:
If, in Drupal 8, all the PHP files are made to end in `.php`, this will be a non-issue. Still an issue for older Drupal codebases... and would be an issue if that Drupal issue _doesn't_ make it in time for Drupal 8's release, but so far it looks like that might just happen :)

arfon:
Hi @deviantintegral - unfortunately this currently isn't supported (we're tracking the issue here https://github.com/github/linguist/issues/1792). This isn't quite the same fix but it's possible to override on a per-file basis using Vim or Emacs overrides: https://github.com/github/linguist#using-emacs-and-vim-modelines

deviantintegral:
Thanks @arfon - I filed a docs update to clarify this. It doesn't really help for upstream code where we don't want to modify it, but it would certainly work for custom code for a specific project.

PR #2165, opened 2015-02-25
Issue #2140: Note that only modelines control syntax highlighting.

deviantintegral:
A docs update to clarify that `.gitattributes` is only for language stats. See #2140 for the discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Linguist supports a number of different custom overrides strategies for language
 
 ### Using gitattributes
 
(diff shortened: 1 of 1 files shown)


arfon:
Sorry about this @deviantintegral - we definitely used to say this (quoting from [here](https://github.com/github/linguist/blob/67fae52b32f7e57749dceea1d039324b8425ddfc/README.md#using-gitattributes)):

> Please note that the overrides currently only affect the language statistics for a repository and not the syntax-highlighting of files.

I think this most have got lost in the re-shuffle of our README and contributing guidelines.

PR #2173, opened 2015-02-26
Instrument all calls and pass the blob, strategy and language candidates in the payload.

arfon:
/ cc @charliesome @vmg @bkeepers

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -105,7 +105,7 @@ module Linguist
       # Bail early if the blob is binary or empty.
       return nil if blob.likely_binary? || blob.binary? || blob.empty?
 
-      Linguist.instrument("linguist.detection") do
+      Linguist.instrument("linguist.detection", :blob => blob) do
         # Call each strategy until one candidate is returned.
         languages = []
         strategy = nil
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -107,18 +107,27 @@ module Linguist
 
       Linguist.instrument("linguist.detection") do
         # Call each strategy until one candidate is returned.
-        STRATEGIES.reduce([]) do |languages, strategy|
-          candidates = strategy.call(blob, languages)
+        languages = []
+        strategy = nil
+
+        STRATEGIES.each do |strategy|
+          candidates = Linguist.instrument("linguist.strategy", :blob => blob, :strategy => strategy, :candidates => languages) do
+            strategy.call(blob, languages)
+          end
           if candidates.size == 1
-            return candidates.first
+            languages = candidates
+            break
           elsif candidates.size > 1
             # More than one candidate was found, pass them to the next strategy.
-            candidates
+            languages = candidates
           else
-            # No candiates were found, pass on languages from the previous strategy.
-            languages
+            # No candidates, try the next strategy
           end
-        end.first
+        end
+
+        Linguist.instrument("linguist.detected", :blob => blob, :strategy => strategy, :language => languages.first) do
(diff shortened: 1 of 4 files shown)


bkeepers:
:cool:

It'd be nice to add some tests for this too. I'll send you an example.

arfon:
> It'd be nice to add some tests for this too. I'll send you an example.

Yep, thanks.

lib/linguist/language.rb
```diff
@@ -107,18 +107,27 @@ def self.detect(blob)
...
+            languages = candidates
           else
-            # No candiates were found, pass on languages from the previous strategy.
-            languages
+            # No candidates, try the next strategy
           end
-        end.first
+        end
+
+        Linguist.instrument("linguist.detected", :blob => blob, :strategy => strategy, :language => languages.first) do
```
charliesome: I'm guessing we care about this instrumentation for the data we're passing into it, rather than the timing of the block? Can we get away with not passing a block at all?

arfon: :+1: yep, that's true. Can you point me to an example of how that would look and I'll modify it?

charliesome: I'm just spitballing here but could we try something like:

```ruby
Linguist.instrument("linguist.detected", :blob => blob, :strategy => strategy, :language => languages.first)

languages.first
```

just to make it clear that we don't care about the block timing? No idea if that'll actually work, but it might be worth trying.

bkeepers: I know I've used it without a block. You can actually call `AS::Notifications.publish` (instead of `#instrument`), which is intended to be for just publishing events and not capturing timing information. But last time I tried to add code that didn't pass the timing information, it confused some people since the arguments passed to `#subscribe` are different.

charliesome:
:+1:

lib/linguist/language.rb
arfon: OK so just to confirm - are we good sticking with the block here for consistency?

charliesome: I don't mind either way, go for it

arfon:
Fiddling around with this a little more it seems that it's not possible to instrument within an existing instrumentation block (as we're currently doing). Is that right?

bkeepers:
> Fiddling around with this a little more it seems that it's not possible to instrument within an existing instrumentation block (as we're currently doing). Is that right?

Hmmm, I would expect it's pretty common to have nested instrumentation.

arfon:
> Hmmm, I would expect it's pretty common to have nested instrumentation.

Yeah, I agree. It's possible the `LocalInstrumenter` I'm using here just isn't behaving as I expect. I tried cribbing the solution you recommended here: https://github.com/github/failbot/issues/67

test/test_instrumentation.rb
```diff
@@ -0,0 +1,44 @@
...
+    # Shouldn't instrument this (as it's binary)
+    assert_equal 0, Linguist.instrumenter.events.size
+  end
+
+  def test_modeline_instrumentation
+    blob = fixture_blob("Data/Modelines/ruby")
+    Language.detect(blob)
+
+    assert_equal 3, Linguist.instrumenter.events.size
+    assert_equal "linguist.detected", Linguist.instrumenter.events.last.name
```
bkeepers: should we assert that it passed the `:language`, `:strategy` and `:blob` here?

arfon: :+1:

bkeepers:
This is looking good. It will be really great to have stats for timing and how each language is detected in GitHub.com.

PR #2082, opened 2015-02-06
update to support PL/SQL PLpgSQL and SQLPL

Shoelace:
update of old PL/SQL PLpgSQL and SQLPL patch based on current version
see add support for oracle PLSQL (#1003)

I have tested this on my own repo. plus a number of publicly available ones..
angoca/log4db2
okbob/plpgsql_lint
antzucaro/xonstatdb

and it is correct as far as i can tell..  would appreciate some other people to check it..
we may need to add more/better samples.

I think i have everything right with the new style now..

FYI @chezwicker @pchaigno and @arfon

PR #2175, opened 2015-02-27
Sqlllll

arfon:
Replaces https://github.com/github/linguist/pull/2082


arfon:
I think this is looking in pretty good shape now. / cc @bkeepers for a review.

bkeepers:
There are a lot of submodule changes that look unrelated. Was that intentional?

Everything else looks good to me. :+1:

arfon:
Looks like I accidentally merged in the updates grammars here. Any objection to merging as-is?

Shoelace:
great.. cant wait to see this live on github.. :)

PR #2178, opened 2015-02-28
Standardized color hexcode length to 6

munrocape:
C was the only language to have a hex code length of 3. 
#555555 == #555 == rgb(85, 85, 85)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -356,7 +356,7 @@ Bro:
 
 C:
   type: programming
-  color: "#555"
+  color: "#555555"
   extensions:
   - .c
   - .cats


pchaigno:
Why do you think this is an issue?

munrocape:
While there is nothing wrong for Linguist, I've noticed that this file gets referenced by a number of other repos.

Parsing this assuming a hexcode length of 6 will cause unexpected behaviour when "#555" is parsed using a bespoke function.

pchaigno:
But then isn't the issue in the repos using this file?

By the way, do you have a link to these repositories?

bkeepers:
No harm in changing this. I can't guarantee we won't add a color with a length of 3 in the future without a test.

munrocape:
@pchaigno I looked just by searching for references to this file on GitHub. Currently, it doesn't break any repos (by the looks of things)

I agree though, it should be the responsibility of any piece of code to verify the input it is processing. It wasn't immediately clear to me that there would exist 3 character hex codes which is why I researched the actual list.

Issue #1685, opened 2014-11-07
Ignore symlinks


quasado:
Currently searching files via github code search (i.e. searching for files with extension *.svg) results in a lot of symlinks found rendering the search useless. Symlinks should be ignored as binary files are.

arfon:
Closing as stale.

PR #2179, opened 2015-02-28
Ignore symbolic links

pchaigno:
I am opening this pull request to get some help on it.
I am trying to implement the change suggested in #1685 to ignore symbolic links.

@bkeepers suggested to change [these lines in `repository.rb`](https://github.com/github/linguist/blob/master/lib/linguist/repository.rb#L153-L155). Is that going to affect the search results as well? (as @quasado wanted)

I made the changes which should ignore symlinks based on the mode attribute. I tested on the two files that I added in `test/fixtures/SVG/`. It doesn't work, the mode doesn't seem to change for a symbolic link at all... :/

### DIFF
--- a/lib/linguist/repository.rb
+++ b/lib/linguist/repository.rb
@@ -150,9 +150,9 @@ module Linguist
         next if delta.binary
 
         if [:added, :modified].include? delta.status
-          # Skip submodules
+          # Skip submodules and symlinks
           mode = delta.new_file[:mode]
-          next if (mode & 040000) != 0
+          next if (mode & 0040000) != 0 || (mode & 0120000) != 0
 
           blob = Linguist::LazyBlob.new(repository, delta.new_file[:oid], new, mode.to_s(8))
 
--- a/lib/linguist/repository.rb
+++ b/lib/linguist/repository.rb
@@ -152,7 +152,8 @@ module Linguist
         if [:added, :modified].include? delta.status
           # Skip submodules and symlinks
           mode = delta.new_file[:mode]
-          next if (mode & 0040000) != 0 || (mode & 0120000) != 0
+          mode_format = (mode & 0170000)
+          next if mode_format == 0120000 || mode_format == 040000 || mode_format == 0160000
 
           blob = Linguist::LazyBlob.new(repository, delta.new_file[:oid], new, mode.to_s(8))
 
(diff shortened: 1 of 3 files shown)


bkeepers:
Weird. It looks to me like this should work. What's the mode of the symlink getting reported as?

pchaigno:
Okay, so I found my error. I didn't copy the symbolic link correctly. The modes are now different for the two files.

But now all files are detected as symbolic links. Maybe I didn't used the right value for the mode (`120000`)...

arfon:
Not sure what's going on here:

```
git cat-file -p e9626a40d232cc7d87730887ded46a91b6647cc4
120000 blob 2f2cb2e43b3a957526c6239362f8c36b0a263fdd	alg_schema.link.svg
100644 blob cd7099148865b1d766d287506c9e07dc0ec78788	alg_schema.svg
```

I tried setting the mode values to `120000` and  `040000` (slightly different from yours @pchaigno) but still no success.

Perhaps someone from @github/git could assist here?

peff:
`0120000` is the right mode for symlinks. For submodules, it is `0160000`; `040000` is the mode of a subtree.

But notice that a regular file is `0100644`. So AND-ing that with a symlink will yield the high bit set, and fail your `!= 0` test. In general with Unix mode bits, the format section of the mode bits is not itself a bitmask; the correct way to check a mode `m` for format `f` is: `(m & S_IFMT) == f`, where `S_IFMT` masks out all but the bits of the feormat value. The magic value for `S_IFMT` is `0170000`. This is usually hidden behind macros like `S_ISLNK` (provided by the OS, but git provides its own `S_ISGITLINK`).

In git trees, this is a little bit simplified, because we don't actually store permission bits at all for submodules, symlinks, or trees. So the non-format bits are always zeroes anyway, you could get by with just checking for equality with `0120000`. That breaks down for regular files, where both `0100644` and `0100755` are regular files (and there are even a few more exotic combinations, as very early versions of git respected group permissions, so you could see `100664` or similar).

Even though it probably wouldn't matter either way, given that it's not much more work to do it correctly (`(mode & 0170000) == 0120000` versus `mode == 0120000`), I think it doesn't hurt to do so.

pchaigno:
Thanks @peff!

This should be okay now :-)

arfon:
Great. Thanks so much for the pointers @peff :zap:

PR #2181, opened 2015-03-01
More specific heuristic for Prolog

pchaigno:
This pull requests improves the heuristic for Prolog to avoid false positives. For instance, Perl files are sometimes detected as Prolog because they contain [a smiley in a comment](https://github.com/textmate/perl.tmbundle/blob/3396190be167310600d00da6ff9af7807cea4b12/Support/exception_handler.pm#L32).


arfon:
:+1:

PR #2184, opened 2015-03-01
Add designated color of eC language for display in language bar.

redj:
I'm hoping this very trivial change will give eC it's designated color in the language bar of the repository page.
Many thanks!


arfon:
:+1: this change will be live on GitHub later this week.

Issue #1152, opened 2014-04-30
QML not detected at all


strahlex:
My repository: https://github.com/strahlex/QtQuickVcp is marked as mainly C++, however, it is mainly QML. QML files seem to be ignored at all. Would be easy to identify them by the file extension .qml.

arfon:
> This problem already occurred with SQL when it was set to programming.
Maybe we could set QML to programming and see if anyone comes here to complain.
@arfon Any thoughts about this?

@pchaigno - I have no serious objections to this. @bkeepers - what do you think ?

arfon:
:flags: flagging this as stale.

roschern:
I was under the impression this would be marked as a code from now on. Did that not happen?

pchaigno:
@arfon I think @roschern is right :). Should I open a PR?

arfon:
> @arfon I think @roschern is right :). Should I open a PR?

Yeah, let's try it out.

pchaigno:
Done.

PR #2189, opened 2015-03-02
QML as a programming language

pchaigno:
This pull request sets QML as a programming language as discussed in #1152.


arfon:
:+1: let's do it.

Issue #2113, opened 2015-02-13
.nl files are incorrectly reported as NewLisp


vitaut:
[.nl files](http://en.wikipedia.org/wiki/Nl_%28format%29) are incorrectly reported as NewLisp in the description of [AMPL/MP](https://github.com/ampl/mp) project (and probably in other places):

![screenshot from 2015-02-13 12 58 06](https://cloud.githubusercontent.com/assets/576385/6195165/59271ac8-b380-11e4-8410-b32c2857f333.png)

Here's an example of such file: https://github.com/ampl/mp/blob/master/test/data/assign0.nl

NL files are very easy to detect: they all start with letter `b` or `g` immediately followed by a number.

arfon:
@vitaut - what language should these be detected as?

pchaigno:
@arfon [As the NL file format apparently](http://en.wikipedia.org/wiki/Nl_%28format%29). I found [more than 1,200 NL files on GitHub](https://github.com/search?o=desc&p=100&q=extension%3Anl+g0&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93).

@vitaut Do you know a Sublime Text package or TextMate bundle to highlight these files?

vitaut:
Yes they should be ideally reported as something like "NL Format". This is a low-level format designed for machine rather than human input so I don't think there is a grammar for it.

pchaigno:
@vitaut Does that mean that they are generated from a higher language?

vitaut:
@pchaigno Yes, they are usually generated from [AMPL](https://en.wikipedia.org/wiki/AMPL) or other mathematical programming system (such as [Pyomo](https://software.sandia.gov/trac/pyomo) or [OpenSolver](http://opensolver.org/)). So it's not really a language, more like a serialization format for optimization problems. It just doesn't seem right that it is detected as NewLisp.

arfon:
So it sound like these should probably be added (name TBD) and then set as `data`?

pchaigno:
Would NL be okay for the name?

arfon:
> Would NL be okay for the name?

Yeah, I think so. http://en.wikipedia.org/wiki/Nl_%28format%29 . Looks like there's `SOL` too: http://en.wikipedia.org/wiki/Sol_(format)

vitaut:
"NL" is perfectly fine.

pchaigno:
@arfon Do we need to add SOL too??

arfon:
[...]

PR #2193, opened 2015-03-03
Support for NL file format

pchaigno:
This pull request adds the NL file format as discussed in #2113.
I tested it on [ampl/mp](https://github.com/ampl/mp) with 0 incorrect detections.


arfon:
:+1:

PR #2196, opened 2015-03-04
GLSL should not be in the C group.

larsbrinkhoff:
I believe it was a mistake to put GLSL in the C group.

E.g. this repository is reported as 100% C in the language bar:
https://github.com/larsbrinkhoff/glsl-sphere-tracing


larsbrinkhoff:
https://github.com/larsbrinkhoff/linguist/commit/a93e9493e209ef333984e5936492c4a0a7148cfc#commitcomment-10036159

Issue #2194, opened 2015-03-03
.jbuilder extensions are ruby


wyaeld:
Rails includes rails/jbuilder by default, and views with name.json.jbuilder extensions are becoming more common.

This is still just ruby though, much like erb.  https://github.com/facelessuser/ApplySyntax just delegates to the ruby bundle under sublime

pchaigno:
Hi @wyaeld!

Is the extension always `.json.jbuilder` or can it be just `.jbuilder`?

wyaeld:
usage is just convention I think.  Rails convention seems to be
.json.jbuilder, but the actual handler registration with sprockets is based
only on .jbuilder, so thats probably more reliable.  I've seen gists from
sinatra projects where they only have .jbuilder on the template

pchaigno:
#2202 should fix this ;)

wyaeld:
excellent, thankyou

PR #2202, opened 2015-03-06
.jbuilder as a Ruby extension

pchaigno:
This pull request adds `.jbuilder` as Ruby extension as discussed in #2194.


arfon:
:+1:

arfon:
It'll be another couple of weeks before this change is live on GitHub

PR #2204, opened 2015-03-06
Exclude gettext catalogues from statistics

aquileia:
Gettext catalogues are used for translations and are thus essentially prose, but were classified as "programming" in 507d191d7d.

In large projects like e.g. wesnoth/wesnoth, gettext can dominate the language statistics with about 95% although the actual code is C++.


arfon:
:+1:

arfon:
It'll be another couple of weeks before this change shows up on GitHub.com

dkavanagh:
thanks!

PR #2213, opened 2015-03-08
Detect generated source maps

ntkme:
This PR marks source maps as generated files.


lib/linguist/generated.rb
```diff
@@ -95,6 +96,20 @@ def minified_files?
...
+    # Internal: Is the blob a generated source map?
+    #
+    # Source Maps usually have .css.map or .js.map extensions. In case they
+    # are not following the name convention, detect them based on the content.
+    #
+    # Returns true or false.
+    def source_map?
+      return false unless extname.downcase == '.map'
+
+      name.downcase =~ /(\.css|\.js)\.map$/ ||         # Name convention
```
bkeepers: I don't know if it matters much, but a case-insensitive regex might be better than a `.downcase`

```ruby
name =~ /(\.css|\.js)\.map$/i
```

bkeepers:
Seems reasonable to me. Merge master and the tests will pass.

@arfon, @aroben: what do you think?

aroben:
Yup, seems fine. @pchaigno also knows a bit about our generated file detection so might have thoughts.

lib/linguist/generated.rb
```diff
@@ -95,6 +96,20 @@ def minified_files?
...
     end
 
+    # Internal: Is the blob a generated source map?
+    #
+    # Source Maps usually have .css.map or .js.map extensions. In case they
+    # are not following the name convention, detect them based on the content.
+    #
+    # Returns true or false.
+    def source_map?
+      return false unless extname.downcase == '.map'
```
pchaigno: Is this needed with the following line of the name convention?

pchaigno:
Looks good to me too :-)

@ntkme Could you add a test for this with a fixture file (`test/fixtures/`)?

ntkme:
@pchaigno Test cases added, as well as a little refactoring to make it nicer.

lib/linguist/generated.rb
```diff
@@ -96,6 +97,20 @@ def minified_files?
...
+
+      name =~ /(\.css|\.js)\.map$/i ||                 # Name convention
+      lines[0] =~ /^{"version":\d+,/ ||                # Revision 2 and later begin with the version number
```
arfon: We should check that the file isn't empty here, e.g. something similar to this: https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L284

ntkme: @arfon That is to avoid calling `.include?` on `nil`. I think no need for here like this one:
https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L106

arfon: Oh yes, you're right.

test/test_generated.rb
```diff
@@ -14,38 +13,58 @@ def generated_without_loading_data(name)
     end
   end
 
-  def generated_loading_data(name)
-    blob = File.join(samples_path, name)
+  def generated_loading_data(blob)
     assert_raises(DataLoadedError, "Data wasn't loaded when calling generated? on #{name}") do
```
arfon: `name` isn't set here any longer, correct?

ntkme: Good catch!

arfon:
@pchaigno this looks good to me. What do you think?

pchaigno:
@arfon :+1:

arfon:
Thanks @ntkme - this should be live on GitHub in the next week or so.

PR #2215, opened 2015-03-08
apt-get update before downloading deps

bkeepers:
Apparently [it is advised to run `sudo apt-get update`](http://docs.travis-ci.com/user/common-build-problems/#Linux%3A-apt-fails-to-install-package-with-404-error) before using any `apt` commands on travis.

This adds it to our `before_install` script, which also requires that we remove the `sudo: false` flag. We'll see how that impacts the duration of the builds.

/cc @arfon @aroben

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 before_install: script/travis/before_install
 rvm:
   - 1.9.3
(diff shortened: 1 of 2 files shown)


arfon:
:zap:

aroben:
Thanks for fixing this. If we find builds are slower to start or slower to run we could ping some Travis CI folks for recommendations.

PR #2218, opened 2015-03-10
Detect *.sma files as SourcePawn

xPaw:
This is a slightly older file format used by [AMX Mod X](https://github.com/alliedmodders/amxmodx/tree/master/plugins) which is very similiar to SourcePawn (they're both Pawn), so there shouldn't be any problems with syntax highlighting. There are also plans to move AMXX to use same SourcePawn version.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2936,6 +2936,7 @@ SourcePawn:
   - sourcemod
   extensions:
   - .sp
+  - .sma
   tm_scope: source.sp
   ace_mode: text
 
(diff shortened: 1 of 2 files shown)


pchaigno:
@xPaw See the [contribution guidelines to add a language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language) (They also apply to add an extension) ;)

xPaw:
Added sample file. Search for `.sma`: https://github.com/search?utf8=%E2%9C%93&q=extension%3Asma+NOT+githubsearchisgreat&type=Code&ref=searchresults

arfon:
:+1: thanks for this. This change should be live on GitHub in the next couple of weeks.

PR #2222, opened 2015-03-11
languages.yml: don't assume .conf is Apache

chriskuehl:
The assumption that `.conf` files are Apache is causing many projects to be detected incorrectly as being primarily "ApacheConf".

The `.conf` extension is widely used by software; Apache accounts for only a very tiny proportion of its overall use. The addition of `.conf` for ApacheConf has resulted in projects which contain none (or almost no) Apache config being marked as primarily containing it.

The problem was introduced by 18a3ef9e5e80889d3daf88dab9338a65c97a7310

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -148,7 +148,6 @@ ApacheConf:
   - apache
   extensions:
   - .apacheconf
-  - .conf
   tm_scope: source.apache-config
   ace_mode: apache_conf
 


chriskuehl:
This is a trivial change; is signing a CLA necessary for it to be considered? If so, I'd prefer to disregard the PR.

Phasesaber:
Yes, but it's not that much. Just hit the "I Agree" button and you're good to go. If not, you can close the pull request.

arfon:
> This is a trivial change; is signing a CLA necessary for it to be considered?

I'm afraid agreeing to the CLA is now compulsory for contributions to Linguist.

Issue #2223, opened 2015-03-12
Command-line `linguist` throws NoMethodError “undefined method `instrument`”


roryokane:
I just installed Linguist and tried to run `linguist` on the directory of a software project cloned [from GitHub](https://github.com/evalEmpire/perl5i). Straight away, it throws a `NoMethodError`:

```
undefined method `instrument' for Linguist:Module (NoMethodError)
```

A summarized transcript, including a stack trace:

```
$ git clone git@github.com:evalEmpire/perl5i.git
Cloning into 'perl5i'...
…
$ rbenv version
2.1.4 (set by /Users/roryokane/.ruby-version)
$ gem install github-linguist
…
Fetching: github-linguist-4.4.3.gem (100%)
Successfully installed github-linguist-4.4.3
…
4 gems installed
$ linguist perl5i/
/Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/language.rb:108:in `detect': undefined method `instrument' for Linguist:Module (NoMethodError)
	from /Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/blob_helper.rb:328:in `language'
	from /Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/lazy_blob.rb:54:in `language'
	from /Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/blob_helper.rb:343:in `include_in_language_stats?'
	from /Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/repository.rb:159:in `block in compute_stats'
	from /Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/repository.rb:145:in `each_delta'
	from /Users/roryokane/.rbenv/versions/2.1.4/lib/ruby/gems/2.1.0/gems/github-linguist-4.4.3/lib/linguist/repository.rb:145:in `compute_stats'
[...]

PR #2224, opened 2015-03-12
Simplifying requires

arfon:
This fixes https://github.com/github/linguist/issues/2223

Not sure why we weren't just doing this originally / cc @bkeepers


bkeepers:
We used to require each file individually, but don't do that anymore.  :+1:

Issue #1548, opened 2014-09-21
What is the group field in languages.yml?


pchaigno:
[languages.yml](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L218) contains an undocumented `group` field.
It causes some strange behavior. For example, Batch is referenced as Shell in the language statistics. In the languages.yml file, Batch is in the group Shell.
What is the purpose of this field?

This is related to #1483.

PR #2231, opened 2015-03-14
Description of group attribute

pchaigno:
This pull request adds a description for the `group` attribute in `languages.yml`.
See #1548.


arfon:
:+1:

PR #2233, opened 2015-03-14
fix emacs modeline parsing

tesch1:
This change to the emacs mode parser should correctly detect the emacs format file-type specification.  I've tested it against my current version of emacs (24.4.1), but I'd guess this interface has been stable for the last 20 years or so.

There are two major changes here:
- regex that selects the "mode:" portion of the per-file variable specification (which is a more accurate description of the line in the file that holds a variable specification for 'mode', rather than "modeline" which is a part of the GUI to emacs speakers.)
- added an alias to the Text language for 'fundamental' because that's the emacs major mode name for pure text.

Caveat: the TravisCI tests "fail" now because one of the DTrace test files has an emacs mode line in it that says it should be formatted as a C++ file.  What's the right thing to do about that?


arfon:
/ cc @bkeepers for some :eyes: on this regex.

lib/linguist/strategy/modeline.rb
```diff
@@ -22,8 +22,13 @@ def self.call(blob, _ = nil)
...
       def self.modeline(data)
-        match = data.match(EmacsModeline) || data.match(VimModeline)
-        match[1] if match
+        match = data.match(EmacsModeline)
+        if match
+          match[4]
+        else
+          match = data.match(VimModeline)
+          match[1] if match
+        end
```
pchaigno: I think if you use `?:` for the other capturing groups you wouldn't have to change this part.

bkeepers:
At first glance it looks like the regex could be simplified a bit, but I don't think I will have a chance to dig into it this week. Maybe switching to the `/x` extended syntax would help make it more clear?

larsbrinkhoff:
My #2253 overlaps https://github.com/tesch1/linguist/commit/5f647f2236ef87a9dc0bbbeec11269e33f3eb4c3

arfon:
:+1: thanks @tesch1. This should be live on GitHub in the next few days.

PR #2239, opened 2015-03-18
Handling misconecption of Ocaml and Standard ML in issue #2208

pawandeep2155:
I added code in heuristics.rb of lib folder.
There are few unique keywords which are present uniquely in both the languages.
Ocaml :-     /->|assert|class|external|functor|match|mutable|struct|try|inherit|module|virtual/
Standard ML :-    /=>|abstype|datatype|handle|infixr|nonfix|withtype|local|andalso|orelese/
I created the regular expression and matched it with the data and if keywords are present then return the language which it matched.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -71,6 +71,20 @@ module Linguist
         Language["BitBake"]
       end
     end
+    
+    # disambiguity support for OCaml and Standard ML 
+    # -> is very common in Ocaml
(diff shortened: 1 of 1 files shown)


pchaigno:
A pull request has already been opened at #2227. It should be merged soon.

arfon:
> A pull request has already been opened at #2227. It should be merged soon.

Yep, thanks for your effort on this @pawandeep2155. This has been resolved in #2227

PR #2250, opened 2015-03-21
Add Neovim config file names to VimL language

shaneog:
[Neovim](http://neovim.org/) uses configuration files named `.nvimrc` rather than `.vimrc`

See this PR for details neovim/neovim#330

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3203,12 +3203,15 @@ VimL:
   search_term: vim
   aliases:
   - vim
+  - nvim
   extensions:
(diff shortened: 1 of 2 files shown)


pchaigno:
The Travis build is failing because filenames need to be sorted in `languages.yml`.
Could you also add samples files for these two extensions in `samples/VimL/filenames/`?

shaneog:
Updated, with sample added

arfon:
:+1: thanks @shaneog. This change should be live on GitHub in the next week or so.

PR #2251, opened 2015-03-21
Allow hyphens in Emacs mode

hollingberry:
Hi there!

I was wondering why [my Spacemacs config](https://github.com/hollingberry/dotfiles/blob/master/emacs/.spacemacs) didn't have syntax highlighting... I think I found the problem.

I modifed the EmacsModeline regexp to allow hyphens in the mode name. This means (I hope) that files with

    ;; -*- mode: emacs-lisp -*-

at the top are now correctly recognized as written in Emacs Lisp.

This doesn't solve the problem of more complex modelines like

    ;; -*- mode: emacs-lisp; coding: utf-8 -*-

because the regexp expects the `-*-` terminator after the first semicolon. If you'd like, I can make the regexp support this sort of thing as well.

### DIFF
--- a/lib/linguist/strategy/modeline.rb
+++ b/lib/linguist/strategy/modeline.rb
@@ -1,7 +1,7 @@
 module Linguist
   module Strategy
     class Modeline
-      EmacsModeline = /-\*-\s*mode:\s*(\w+);?\s*-\*-/i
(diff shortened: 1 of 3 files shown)


pchaigno:
I think #2233 will already solve your issue.

hollingberry:
Oh sweet! That looks good then

PR #2253, opened 2015-03-21
Fix Emacs modeline in DTrace sample.

larsbrinkhoff:
One of the DTrace sample files specifies the Emacs C++ mode.  The original author probably had a good reason for doing this, but for now I think it's better to modify the sample.  I believe this will fix the build failure in #2233

Apparently, the DTrace mode for Emacs is called dtrace-script:
https://github.com/dotemacs/dtrace-script-mode

For this reason, I added `dtrace-script` as a Linguist alias for `DTrace`.  I'm not at all sure if this is how the Emacs ↦ Linguist language mapping should be done.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -711,6 +711,8 @@ DM:
 
 DTrace:
   type: programming
+  aliases:
+  - dtrace-script
   extensions:
   - .d
   interpreters:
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
@bkeepers Another example of why the Emacs modeline can't be entirely trusted.

arfon:
:+1:

PR #2254, opened 2015-03-21
add .boot to clojure extensions

shaunlebron:
[Boot](http://boot-clj.com/) is a build tool for clojure.  The `.boot` files are just clojure files though.  Would be nice to have those highlighted accordingly on github.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -514,6 +514,7 @@ Clojure:
   ace_mode: clojure
   color: "#db5855"
   extensions:
+  - .boot
   - .clj
   - .cl2
   - .cljc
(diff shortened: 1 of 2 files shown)


pchaigno:
The Travis build is failing because the first extension corresponds to the primary extension. Only the following extensions should be sorted. So, `.boot` should come after `.clj`.

shaunlebron:
thanks @pchaigno :)

pchaigno:
Could you also add a sample for this extension in `samples/Clojure/`?

shaunlebron:
added a sample

arfon:
:+1: thanks @shaunlebron and @pchaigno. This should be live on GitHub next week sometime.

shaunlebron:
cool, thanks @arfon

PR #2266, opened 2015-03-24
Add .ruby to the list of ruby file extensions

Dorian:
Can be found in Rails partial, eg `.html.ruby`

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2713,6 +2713,7 @@ Ruby:
   - .rbw
   - .rbx
   - .ru
+  - .ruby
   - .thor
   - .watchr
   interpreters:


arfon:
> Can be found in Rails partial, eg .html.ruby

I've not see this usage before. Could you point me to some examples of usage?

bkeepers:
There are [~2000 results on github with the `.ruby` extension](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aruby+NOT+foobar&type=Code&ref=searchresults). I have never seen that extension used or even encouraged in the ruby world.

arfon:
Nevertheless people definitely seem to be using it so...

Dorian:
Thanks, it's in the list of default rails template file extensions, eg like: `.erb`, `.slim`, `.haml` (and `.ruby`). I used it in a view that needed a lot of ruby logic

PR #2269, opened 2015-03-24
Use #000080 for Lua color

stuartpb:
This is the color of the official Lua logo, as seen at http://www.lua.org/images/


arfon:
Hi @stuartpb - we'll need you to sign the CLA before accepting this pull request. :smile:

stuartpb:
Signed.

arfon:
Thanks @stuartpb. It'll probably be another week or so before this change is live as we just cut a new gem this afternoon.

PR #2270, opened 2015-03-25
Add interpreters for OCaml.

larsbrinkhoff:
Some OCaml programs use a shebang.


arfon:
:+1: thanks @larsbrinkhoff

PR #2274, opened 2015-03-25
intro.wisp: Fixed typographical error ('rather then' -> 'rather than').

csimons:


gjtorikian:
:cake:

samples/wisp/intro.wisp
```diff
@@ -258,7 +258,7 @@ My name is wisp!"
 ;; evaluating the reference to the value of the corresponding variable.
```
reyes76: fffgh

PR #2275, opened 2015-03-25
Bump escape-utils to 1.1.0

ptoomey3:
EscapeUtils recently bumped to version 1.1.0 after merging https://github.com/brianmario/escape_utils/pull/59. I'd like to pull this version into a few other projects and the the `~> 1.0.1` is preventing Bundler from being able to find a compatible version. This bump *shouldn't* have any side effects on Linguist, as the only two uses I can find are https://github.com/github/linguist/blob/a0d5a8338bacaa925c9e0215b4f8b2bbcd1115d1/lib/linguist/blob_helper.rb#L102 and https://github.com/github/linguist/blob/a0d5a8338bacaa925c9e0215b4f8b2bbcd1115d1/lib/linguist/language.rb#L459, both of which use `escape_url`, which should not have changed.

P.S. That new verison of EscapeUtils adds support for a new method `escape_uri_component` that does URL escaping but with space encoded as `%20` instead of `+` (the way `escape_url` does). So, https://github.com/github/linguist/blob/a0d5a8338bacaa925c9e0215b4f8b2bbcd1115d1/lib/linguist/language.rb#L459 could be updated to not require the gsub.

/cc @brianmario

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.executables << 'linguist'
 
   s.add_dependency 'charlock_holmes', '~> 0.7.3'
-  s.add_dependency 'escape_utils',    '~> 1.0.1'
+  s.add_dependency 'escape_utils',    '~> 1.1.0'
   s.add_dependency 'mime-types',      '>= 1.19'
   s.add_dependency 'rugged',          '~> 0.23.0b1'
 


brianmario:
Awesome :+1:

ptoomey3:
@arfon - Just wanted to get a :+1:  since you seem to manage most of the releases lately.

arfon:
:+1: this should be fine @ptoomey3. Is there any immediate rush to get this deployed? I wasn't planning on cutting a new gem until next week sometime.

ptoomey3:
@arfon - no rush at all.

PR #2276, opened 2015-03-26
Fix ada aliases

MakeNowJust:
tiny tiny fix.


arfon:
:+1: thanks @MakeNowJust

Issue #2238, opened 2015-03-17
Add .odd extension


joewiz:
There are [nearly 500 ODD files](https://github.com/search?q=extension%3Aodd+NOT+djfhdirijdskqdhd&type=Code&utf8=%E2%9C%93) in GitHub.  [ODD](http://en.wikipedia.org/wiki/Text_Encoding_Initiative#ODD) is an XML-based vocabulary for defining schemas for Text Encoding Initiative (TEI) projects, used widely in the humanities.  ODD should be rendered using XML syntax but is currently displayed as plain text (see [this example](https://github.com/wolfgangmm/tei-simple-pm/blob/master/odd/teisimple.odd)).  

I would be happy to submit a pull request, but I'm not sure about whether to define .odd as an entry within the [existing XML languages](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3263) or whether I should submit this as a new language, along the lines of the entry for [XProc](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3345).  I'm inclined toward the former, but I'd appreciate guidance before I submit the wrong type of pull request.

Thanks in advance.

Phase:
I think you could add it to XML language if they're the same thing.

pchaigno:
You could define it as a new language in Linguist if you wanted it to be counted as a different language (from XML) in the statistics or if you wanted to use a different grammar for highlighting.
XML isn't counted in the statistics anyway, so, in your case, I think you should define it as a new extension.

Though 500 files might not be enough to add it to linguist (@arfon ?).

arfon:
> You could define it as a new language in Linguist if you wanted it to be counted as a different language (from XML) in the statistics or if you wanted to use a different grammar for highlighting. XML isn't counted in the statistics anyway, so, in your case, I think you should define it as a new extension.

Yes, I think we should add this as a new extension to `XML`. @phase - would you mind doing this? You can read our [contributing guidelines here](https://github.com/github/linguist/blob/master/CONTRIBUTING.md)

joewiz:
Sure, will do.  Thanks for your guidance, everyone!

PR #2279, opened 2015-03-26
Add .odd to XML extensions

joewiz:
See discussion at #2238. 

[ODD](http://en.wikipedia.org/wiki/Text_Encoding_Initiative#ODD) is an XML-based vocabulary for defining schemas for Text Encoding Initiative (TEI) projects, used widely in the humanities.  ODD should be rendered using XML syntax but is currently displayed as plain text (see [this example](https://github.com/wolfgangmm/tei-simple-pm/blob/master/odd/teisimple.odd)).  There are [nearly 500 ODD files](https://github.com/search?q=extension%3Aodd+NOT+djfhdirijdskqdhd&type=Code&utf8=%E2%9C%93) in GitHub, and this number is projected to grow.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3326,6 +3326,7 @@ XML:
   - .mxml
   - .nproj
   - .nuspec
+  - .odd
   - .osm
   - .plist
   - .pluginspec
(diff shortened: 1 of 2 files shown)


arfon:
Thanks @joewiz! For good housekeeping would you mind adding an example `.odd` file to the XML samples?

joewiz:
Thanks, @arfon.  I've added a sample ODD file.

arfon:
Excellent. Thanks @joewiz. This should be live on GitHub next week sometime.

joewiz:
Great!  Thanks so much for your help!

PR #2289, opened 2015-03-31
add tests/fixtures to vendor.yml

lepture:
My repo https://github.com/lepture/mistune is detected as HTML.

I've moved those HTML files to fixtures directory(https://github.com/lepture/mistune/commit/54197e12e2e4547e24ae5508f09f5516ebb8a8bc), but it won't work. I've checked this repo, it seems that only `test/fixtures/` is ignored.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -224,7 +224,7 @@
 - ^readme$
 
 # Test fixtures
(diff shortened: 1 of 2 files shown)


arfon:
:+1: thanks @lepture.

It'll be another few days before this change is live on GitHub.

lepture:
Thanks man.

PR #2296, opened 2015-04-01
Add colors to semi-popular languages

phase:
These may not be the most popular, but they are still used a lot here on GitHub, and they deserve a color.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -28,6 +28,7 @@
 
 ABAP:
   type: programming
+  color: "#E8274B"
   extensions:
   - .abap
   ace_mode: abap
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -337,6 +337,7 @@ Boo:
 
 Brainfuck:
   type: programming
+  color: "#2F2530"
   extensions:
   - .b
   - .bf
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -644,6 +644,7 @@ Creole:
 
 Crystal:
   type: programming
+  color: "#76EEF5"
   extensions:
   - .cr
   ace_mode: ruby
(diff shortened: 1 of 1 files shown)


phase:
`The command "script/travis/before_install" failed and exited with 1 during .` oh noes

pchaigno:
For Travis, it's probably just a temp. failure. It happens sometimes. You can try to re-launch the build by closing/opening the issue.

Have you checked that the new colors don't conflict with [existing ones](http://github.github.io/linguist/)?

phase:
@pchaigno I highly doubt that of the thousands of combinations out there, I made one that is the same as another. I took into account the name & what's it's used for. (Like Crystal is a light blue)

pchaigno:
> I highly doubt that of the thousands of combinations out there, I made one that is the same as another.

Maybe not the exact same but something too close to be easily distinguish on the tiny statistic bar. It happened before so I think it's worth a quick check. It's especially important that it doesn't happen for two languages that are often found together (Makefile and C for instance).

arfon:
Thanks for this @phase. Please do take a look at http://github.github.io/linguist/ to see if these colours are close to existing ones.

phase:
I believe that none are similar to these, but I might have missed one.

pchaigno:
I tested with @gjtorikian's new test from #2298 and it says it's okay too :zap:

arfon:
> I tested with @gjtorikian's new test from #2298 and it says it's okay too :zap:

:+1:

phase:
How long does it take for it to get updated? I was looking at https://github.com/phase/minekraft and the Kotlin still looks grey. When I did this to Diff in #2236, it looked like it was instant.

arfon:
We have to cut a new gem so you probably just got lucky with timing last time :-) 

So to answer your question - this will probably be updated in the next week or so.

phase:
Since this was over 3 years ago, I have no idea! I think it was due to there already being some languages that use black. At the time of this PR, black [wasn't used on the website](http://archive.is/nSbpn) that much.

PR #2298, opened 2015-04-01
Add proximity test for colors

gjtorikian:
https://github.com/github/linguist/pull/2296 got me thinking. It's a bit harsh to have to go through the languages list and determine whether new colors match or not.

I put together a simple gem called [color-proximity](https://github.com/gjtorikian/color-proximity) to determine the closeness of two colors with each other. It uses [a Euclidean distance algorithm](http://www.compuphase.com/cmetric.htm) to determine likeliness.

This PR adds a test to Linguist to ensure that no two colors in the languages list are too alike. The threshold value of 10 was set by trial and error. If the proximity of any color falls below this value, the test will fail.

For an example of what is considered failure for a threshold of even 20, these two existing colors wouldn't pass:

screen shot 2015-04-01 at 10 14 26 am
screen shot 2015-04-01 at 10 14 30 am

To math, that may be close, but to my human eyes, they seem far enough apart to be fine. 

/cc @github/languages @pchaigno @phase


arfon:
YES YES YES YES YES YES YES YES

phase:
Did those two example colors pass? They look different enough.

arfon:
```
[[#<Linguist::Language name=Ada>, ["#ff8877"]], [#<Linguist::Language name=Perl>, ["#199c4b"]], [#<Linguist::Language name=Perl6>, ["#199c4b"]], [#<Linguist::Language name=VCL>, ["#199c4b"]]]
```

Looks like Perl and Perl6 are using the came colour.

phase:
@arfon But since they're so similar, why not allow them to have the same color?

arfon:
Yeah, Perl and Perl6 should probably have the same colour?

aroben:
I think only group colors are displayed on github.com but I can't check right now to be sure. If that's true then we should perhaps only assign colors to groups instead of to individual languages.

-Adam

phase:
I haven't seen anything about group colors before, if it's not a thing I'd love to implement it.

larsbrinkhoff:
>  I think only group colors are displayed on github.com

That was indeed the case when I checked recently.  Only language groups are displayed in the language statistics bar, hence only group colors.

Are language colors displayed anywhere else than the language bar?

pchaigno:
@gjtorikian Could this be extended to assign colors automatically to all languages which don't have one? Are you able to tell approximately how many different colors exist according to this metric (assuming we can use all colors)?
<br/>

> Yeah, Perl and Perl6 should probably have the same colour?

Perl 5 and Perl 6 are often considered as two different languages so I don't think they should be in the same group or have the same color.
<br/>

> That was indeed the case when I checked recently. Only language groups are displayed in the language statistics bar, hence only group colors.

:+1:

gjtorikian:
> Could this be extended to assign colors automatically to all languages which don't have one? 

Sure; we could even add a test to always assume a color, if that's helpful (and it'd randomly assign you one otherwise). If "every language needs a color" is something we're interested in pursuing, it should probably be done in another PR, as groups would also need to be taken into consideration.

> Are you able to tell approximately how many different colors exist according to this metric (assuming we can use all colors)?

I mean, currently, based on a search of `color:`, we have 175 colored languages (vs 349 total). Is that what you meant? 

> Perl 5 and Perl 6 are often considered as two different languages so I don't think they should be in the same group or have the same color.

I completely agree here. The "6" is part of the name, not a direct reference to a semantic upgrade. From [Wikipedia](https://en.wikipedia.org/wiki/Perl_6):

> Compatibility with Perl 5 is not a goal, though a compatibility mode is part of the specification.

> I haven't seen anything about group colors before

If you check the *languages.yml* file, you'll notice that the `group` key exists every now and then.

pchaigno:
>  If "every language needs a color" is something we're interested in pursuing, it should probably be done in another PR, as groups would also need to be taken into consideration.

:+1: 
<br/>

> I mean, currently, based on a search of color:, we have 175 colored languages (vs 349 total). Is that what you meant? 

Sorry, I wasn't clear. I'm wondering how many colored languages we could have without this test failing. For example with a simple euclidean distance function on the RGB model we could have approximately 256^3 / 10^3 colored languages (size of RGB space divided by size occupied by one color). There are probably more "available" colors (i.e. colors considered different with this test) than we will ever need, though.
<br/>

I'm a bit surprised that Isabelle (#fdcd00) and Parrot (#fdd000) don't fail this test...
fdcd00  fdd000

gjtorikian:
> I'm a bit surprised that Isabelle (#fdcd00) and Parrot (#fdd000) don't fail this test...

Hmm, I see Parrot as `f3ca0a`. Where did you get that hex from?

gjtorikian:
> I'm wondering how many colored languages we could have without this test failing. 

I'm going to embarrass myself with some napkin math here.:stuck_out_tongue_closed_eyes:

First question: what's the minimum hex distance between two points before a pass occurs?

```
irb(main):005:0> cp = ColorProximity.new(10, color)
=> #<ColorProximity:0x007faaec0b9c28 @threshold=10, @collection=["000000"]>
irb(main):006:0> cp.within_threshold?("000000")
=> [false, ["000000"]]
irb(main):007:0> cp.within_threshold?("000001")
=> [false, ["000000"]]
irb(main):008:0> cp.within_threshold?("000002")
=> [false, ["000000"]]
irb(main):009:0> cp.within_threshold?("000003")
=> [false, ["000000"]]
irb(main):010:0> cp.within_threshold?("000004")
=> [false, ["000000"]]
irb(main):011:0> cp.within_threshold?("000005")
=> [false, ["000000"]]
irb(main):012:0> cp.within_threshold?("000006")
=> [true]
```

Okay, looks like it's six.

Second question: what's the maximum number of colors possible in the CSS spectrum? `ffffff` resolves to  16,777,215.

If every sixth color is "safe" from failing the test, I would guess then, that 16,777,215 / 6 = ~2,796,202 possibly different colors.

pchaigno:
> Hmm, I see Parrot as f3ca0a. Where did you get that hex from?

[From Google](https://www.google.fr/search?client=ubuntu&channel=fs&q=%23f3ca0a&ie=utf-8&oe=utf-8&gfe_rd=cr&ei=hyAeVaS1D8jI8gfG1oHgBA) (I shouldn't trust it so much). (Searching for `#f3ca0a` redirects you to other colors :/)

I created [a repository to see the two colors side by side](https://github.com/pchaigno/TestColors). It's not as easy to see the difference on the statistics bar as it is with blocks of color.

phase:
You also have to take into account how common the languages are in use together. If Java & Scala were both orange, it'd suck, but if Objective-C & Crystal were both blue, no one would care.

gjtorikian:
> I created a repository to see the two colors side by side. It's not as easy to see the difference on the statistics bar as it is with blocks of color.

Well that's weird. The test should fail. I see the problem now; in your original quote you listed

>  I'm a bit surprised that Isabelle (#fdcd00) and Parrot (#fdd000) don't fail this test...

But *languages.yml* [lists Parrot as `f3ca0a`](https://github.com/github/linguist/blob/add-proximity-test/lib/linguist/languages.yml#L2299). In any event I agree those are crazy similar. According to the code:

```
irb(main):005:0> cp = ColorProximity.new(10, ['fdcd00'])
=> #<ColorProximity:0x007ffb5a061758 @threshold=10, @collection=["fdcd00"]>
irb(main):006:0> cp.within_threshold?('f3ca0a')
=> [true]
irb(main):007:0> cp.thresholds('f3ca0a')
=> [23.108440016582687]
```

These two colors have a distance of 23.10. 

I will revise the algorithm in the `color-proximity` gem [by reading up a bit more](http://stackoverflow.com/questions/492211/is-there-an-easy-way-to-compare-how-close-two-colors-are-to-each-other). Thanks for keeping this honest! If you know if any other existing similarities, do let me know as it'll help the test corpus.

gjtorikian:
@pchaigno Thanks again so much for keeping this honest. I switched to a different standard algorithm and it feels much better. It's now reporting, for example, that Frege's [`00cafe`](http://www.color-hex.com/color/00cafe) is too close to Flux's [`33ccff`](http://www.color-hex.com/color/33ccff), which is absolutely right.

I'll go through and address these errors before merging in. Do let me know if something else seems suspicious.

pchaigno:
@gjtorikian It do look better! Though, I still find Objective-C and Objective-C++ a bit difficult to differentiate on the language bar. On an other level, there is also Python and Processing. I updated [pchaigno/TestColors](https://github.com/pchaigno/TestColors) to reflect that.

I can see the difference between these languages, it's just a bit difficult. So, maybe increasing the threshold would be sufficient...?

gjtorikian:
@pchaigno Done. A bump of the threshold suggested many other changes that I also thought were a bit too close.

BTW, in case anyone from The Future&trade; comes along and wonders how the colors changed, here's what I usually do:

* I always head to http://www.color-hex.com/ to make sure that the colors are truly the same
* If they are, I try to do a Wikipedia search for the language. If it has a logo or symbol, I try to pluck that color.
* If there's no logo, I try to pluck a color from an O'Reilly book on the subject.
* If there's no book, I just make it up by playing with the numbers on the ColorHex website.
* I often change unpopular languages over popular ones. That is to say, if there's a conflict between a Lisp variant and HTML, I'll change the Lisp variant in a heartbeat.
  * This is not always true, however. For example, Python was listed as [`3581ba`](http://www.color-hex.com/color/3581ba) and conflicted with [Clean](https://en.wikipedia.org/wiki/Clean_%28programming_language%29)'s [`3a81ad`](http://www.color-hex.com/color/3a81ad). I changed both: Python to match one of its snake's blues and Clean to match its turquoise logo.

As you can imagine, the process is a bit tedious, so I'm going to call it complete for now.

arfon:
Thanks @gjtorikian!

pchaigno:
@gjtorikian Thanks!

joelerego:
> @pchaigno Done. A bump of the threshold suggested many other changes that I also thought were a bit too close.
> 
> BTW, in case anyone from The Future™ comes along and wonders how the colors changed, here's what I usually do:
> 
> * I always head to http://www.color-hex.com/ to make sure that the colors are truly the same
> * If they are, I try to do a Wikipedia search for the language. If it has a logo or symbol, I try to pluck that color.
> * If there's no logo, I try to pluck a color from an O'Reilly book on the subject.
> * If there's no book, I just make it up by playing with the numbers on the ColorHex website.
> * I often change unpopular languages over popular ones. That is to say, if there's a conflict between a Lisp variant and HTML, I'll change the Lisp variant in a heartbeat.
>   
>   * This is not always true, however. For example, Python was listed as [`3581ba`](http://www.color-hex.com/color/3581ba) and conflicted with [Clean](https://en.wikipedia.org/wiki/Clean_%28programming_language%29)'s [`3a81ad`](http://www.color-hex.com/color/3a81ad). I changed both: Python to match one of its snake's blues and Clean to match its turquoise logo.
> 
> As you can imagine, the process is a bit tedious, so I'm going to call it complete for now.

Hi. I was wondering how the C++ language color (Light salmon pink) was chosen. Standard C++ (isocpp.org) has a dark blue logo, which is available on Wikipedia. The C++ Programming Language textbook written by Bjarne Stroustrup is also of the same color. I am curious to know how the current salmon pink color was chosen. Was it an arbitrary selection? Or does it have any link with the C++ language?

lildude:
@joelerego from @Alhadis's comment at https://github.com/github/linguist/pull/4818#issuecomment-597676741 and the [colour guidelines](https://github.com/github/linguist/blob/b545f1c6befe87c2074e5e608f47ca0f6334d355/CONTRIBUTING.md#changing-the-color-associated-with-a-language):

> Many of the colours associated with the languages within Linguist have been in place for a very long time. The colours were often chosen based on the colours used by the language at the time and since then users will have become familiar with those colours as they appear on GitHub.com.

And mine at https://github.com/github/linguist/pull/4818#issuecomment-597721922:

>> Was it an arbitrary selection?
>
> Probably and we may never know as the C++ was one of the languages that Linguist knew about when colours were first assigned to the languages over 8 years ago in https://github.com/github/linguist/commit/c3d6fc5af8cf9d67afa572bba363bf0db256a900 .  This predates pull requests (and GitHub even having a formal design team) so we can't even read the substantiation within a PR, and as all the colours were added in one commit, it's impossible to tell why each colour was chosen. 
> 
> I can't remember why I made a choice for a line of code, random or otherwise, a last year, let alone 8 years ago, so asking the commitor, who may not have been the original author, is probably futile too. They don't work at GitHub anymore either.

Alhadis:
Since we're on the topic... has the colour-proximity requirement been abjured yet, @lildude? 😉

lildude:
Nope. Still waiting on Product Design 😞

Alhadis:
Hire me as Product Design lead for 10 minutes so I can make the decision for them, and then I'll fire myself afterward. 😁

(kidding, sorta)

PR #2299, opened 2015-04-02
Fixing up shebang detection to match new tokenizer behaviour

arfon:
Since https://github.com/github/linguist/pull/1604 the build has been broken :crying_cat_face: 

The change to the tokenizer behaviour [here](https://github.com/github/linguist/blob/edadca9085ba69f907b44e142fa7b45520367f96/lib/linguist/tokenizer.rb#L145) and the new sample file with a shebang that looked like this:

```
#! /usr/bin/env A=003 B=149 C=150 D=xzd E=base64 F=tar G=gz H=head I=tail sh
```

meant that the shebang extraction in [shebang.rb] (https://github.com/github/linguist/blob/master/lib/linguist/shebang.rb#L20-L51) was returning `A=003` as the interpreter (when it should be `sh`).

This pull request brings the shebang detection in `tokenizer.rb` and `shebang.rb` into parity to fix the current build failure. Given how similar these methods are they should basically be the same. I'm saving that for another day.

### DIFF
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -23,23 +23,29 @@ module Linguist
       # First line must start with #!
       return unless shebang && shebang.start_with?("#!")
 
-      # Get the parts of the shebang without the #!
-      tokens = shebang.sub(/^#!\s*/, '').strip.split(' ')
+      s = StringScanner.new(shebang)
 
       # There was nothing after the #!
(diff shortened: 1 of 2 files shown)


arfon:
/ cc @bkeepers for :eyes:

bkeepers:
TIL `StringScanner`

:+1: if you add an explicit test.

PR #2300, opened 2015-04-02
Adding in a Makefile.boot filename for .boot detection

arfon:
Looks like we added support for `.boot` in https://github.com/github/linguist/pull/2254 without considering the discussion we had in https://github.com/github/linguist/pull/1894

This pull request adds `Makefile.boot` which seems to be the dominant file type using the `.boot` extension.


arfon:
Fix for failure in https://github.com/github/linguist/pull/2299

PR #2302, opened 2015-04-03
Add Libraries & Libs to vendor.yml

phase:
This removes the following from being checked by linguist:
* libs
* Libs
* lib
* libs
* Libraries
* libraries

These are usually only used for libraries for a project, and not the project itself.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -15,6 +15,10 @@
 # Dependencies
 - ^[Dd]ependencies/
 
+# Libraries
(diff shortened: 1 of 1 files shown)


arfon:
Hi @phase. I can't speak for other languages but `lib` is pretty important in Ruby-land. Take a look at the Linguist gem itself - `lib` is basically where all of the Ruby code _for this project_ lives.

phase:
Yeah, I was looking at some other repos and found loads that use lib(s)?/ for their code. :stuck_out_tongue_closed_eyes:

Issue #2292, opened 2015-03-31
Limbo files incorrectly recognized as M, Objective-C and Brainfuck


powerman:
[Limbo](http://inferno-os.info/inferno/limbo.html) is main programming language for [OS Inferno](http://inferno-os.info/inferno/index.html) (sorta OS Plan9 with VM). It uses `.b` files for sources (like `.c`) and `.m` files for public interfaces (like `.h`).

There are actually a lot of repo with Limbo code on GitHub, for example search for [module PATH con dis](https://github.com/search?l=m&q=module+PATH+con+dis&ref=searchresults&type=Code&utf8=%E2%9C%93) - all files you see when selecting one of languages M, Objective-C or Brainfuck are actually Limbo code. Example repos:
https://github.com/powerman/inferno-contrib-watchdog
https://github.com/powerman/inferno-contrib-tap

powerman:
I'm not sure is this unique enough detection rule (using PCRE), but all `.m` Limbo files begins with (after optional empty or comment lines) `\w+\s*:\s*module\s*{` and all `.b` Limbo files begins with (after optional empty or comment lines) `implement\s+\w+;`. Comment lines begins with `#`.

phase:
It's also being recognized as Mercury (if that helps at all :bowtie:).

pchaigno:
Hi @powerman - I'm not from GitHub but I can help you build a pull request to add this language if you'd like.

To add this language with its extensions, we first need to show that they are widely used on GitHub (as per the [contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language)). I was able to find:
- [1,176 `.b` Limbo files](https://github.com/search?utf8=%E2%9C%93&q=module+path+con+dis+extension%3Ab&type=Code&ref=searchresults)
- [4,270 `.m` Limbo files](https://github.com/search?utf8=%E2%9C%93&q=module+path+con+dis+extension%3Am&type=Code&ref=searchresults)

I wasn't able to build search queries for `.c` and `.h` Limbo files. For example [this search](https://github.com/search?utf8=%E2%9C%93&q=module+path+con+dis+extension%3Ac&type=Code&ref=searchresults) returns many files that don't look like Limbo.
@powerman Do you have an idea for a better search query? (the search engine is case-insensitive)

powerman:
[...]

PR #2306, opened 2015-04-04
add Limbo: language, samples

powerman:
I've no idea how to "Test the performance of the Bayesian classifier".
I don't know Ruby, but I've already provided PCRE for detecting at https://github.com/github/linguist/issues/2292#issuecomment-88268369 and may try to update heuristics.rb if someone tell me how to run and test it.


powerman:
How exactly I'm supposed to fix conflict with Brainfuck's samples? Just copy all samples/Brainfuck/*.bf files to samples/Brainfuck/*.b?

pchaigno:
To test your changes locally:
```
$ bundle install
$ bundle exec rake samples # trains the classifier
$ bundle exec rake # tests
```

For the `.b` Brainfuck sample, I think you can use [this file](https://github.com/windoze/brainfuck/blob/9ef0f6eef684da2fa27173f31d4622e56c83ff88/test/factor.b). It seems to be under GPL license.

powerman:
Looks like this doesn't work:
```
$ bundle install
# To unbundle, run this file
echo install
sed 's/.//' >install <<'//GO.SYSIN DD install'
sed: can't read install: No such file or directory
//GO.SYSIN DD install
$ bundle exec rake samples
# To unbundle, run this file
echo exec
sed 's/.//' >exec <<'//GO.SYSIN DD exec'
sed: can't read exec: No such file or directory
//GO.SYSIN DD exec
echo rake
sed 's/.//' >rake <<'//GO.SYSIN DD rake'
sed: can't read rake: No such file or directory
//GO.SYSIN DD rake
echo samples
sed 's/.//' >samples <<'//GO.SYSIN DD samples'
sed: read error on samples: Is a directory
//GO.SYSIN DD samples
```

powerman:
Added in last commit samples/Brainfuck/factor.b is “Submitted on behalf of a third-party or third parties: GitHub user windoze”.

pchaigno:
> Looks like this doesn't work

Never seen these errors before :/
Maybe @arfon can help?

powerman:
Ok, I've switched ruby from 1.9 to 2.0 (Gentoo provide such a feature), repeat all steps and now it works:
```
$ bundle exec rake
...
833 runs, 10233 assertions, 0 failures, 0 errors, 0 skips
```
Is this mean everything is fine and there is no needs in manually adding my regexps to heuristics.rb?

pchaigno:
These are only the unit tests, they use the samples. Since these samples were used to train the classifier, we can be almost sure that the Bayesian classifier will recognize them correctly :/ 
To test your pull request completely you should test it against some Limbo repositories:
`bundle exec linguist ../path/to/repository --breakdown`.

powerman:
Cool, I've checked several of my repos with Limbo code - looks like it works! :)

arfon:
@pchaigno - adding another language with `.m` will mean that the heuristic will stop being used because of [this issue](https://github.com/github/linguist/issues/2042) right?

pchaigno:
@arfon Ouch, you're right. Thanks! I had completely forgotten that :/

@powerman Could you add the heuristic you mentioned in #2292 for `.m` (`\w+\s*:\s*module\s*{`) at [`heuristics.rb#L187-L201`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L187-L201)?

powerman:
Travis fails on before_install script while updating subrepos for some ruby versions and pass for others, so I suppose this RP is ok.

arfon:
:+1: thanks @powerman. This change should be live on GitHub in the next week or so.

Issue #2187, opened 2015-03-02
*.sch is both Eagle and KiCad schematics file


Miceuz:
Now *.sch files are classified as Eagle files.

Miceuz:
Repo in question - https://github.com/Miceuz/ultimate-temperature-controller

Miceuz:
Kicad sch file is not XML, it starts like this:

`EESchema Schematic File Version 2`

pchaigno:
@Miceuz I see that both your files have a lot of lines starting with `LIBS`. Is that common in the KiCad format or is it just in your files? (I'm trying to see if we could use your files to train the classifier to recognize `.sch` files as KiCad.)

Miceuz:
@pchaigno In fact it depends on how much libs are used int particular project, might be a single library containing a resistor. But the file has a unique header - `EESchema Schematic File Version 2` on another hand `$Comp` and `$EndComp` is a unique feature.

pchaigno:
Okay, thanks for the clarification! I'll see what I can do.
The header could surely be used for [a heuristic rule](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb) but we try to only use those if the Bayesian classifier can't do the job.

mithat:
@pchaigno  I've got a few KiCad designs that you can use for training the classifier if it's helpful:
* https://github.com/mithat/Volume-AlpsRK16814MG
* https://github.com/mithat/Arduino_Minimal_PTH
* https://github.com/mithat/Arduino_Pro_Mini_PTH

pchaigno:
@Miceuz @mithat Do you know a Textmate bundle or Sublime Text package that could be used to highlight KiCad code?

mithat:
@pchaigno I don't know of such a bundle. But I found file format documentation for KiCad at <http://bazaar.launchpad.net/~kicad-developers/kicad/doc/files/head:/doc/help/file_formats/docs_src/> if that's helpful.

Also, I remembered that the `*.sch` extension is used by another FLOSS EDA suite, the [gEDA project](http://www.geda-project.org/). Its file format documentation is at <http://wiki.geda-project.org/geda:file_format_spec>.

Alhadis:
> @Miceuz @mithat Do you know a Textmate bundle or Sublime Text package that could be used to highlight KiCad code?

I wrote one [for all three languages](https://github.com/Alhadis/language-pcb#pcb-languages):

[...]

PR #2309, opened 2015-04-05
KiCad language with .sch extension

pchaigno:
This pull requests adds KiCad with its `.sch` file extension.
The extension collides with Eagle's extension but the two language are very different; the Bayesian seems to be able to distinguish easily.
[Examples of KiCad files on GitHub.](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asch+EESchema+Schematic+File+Version&type=Code&ref=searchresults)

This fixes the issues reported in #2187.


arfon:
:+1:

PR #2311, opened 2015-04-06
New samples for SaltStack

pchaigno:
This pull request adds some new SaltStack samples from [saltstack/salt-winrepo](https://github.com/saltstack/salt-winrepo). This should solve an issue recently reported at #960.


arfon:
:+1:

PR #2314, opened 2015-04-07
Tighten up OCaml heuristic

larsbrinkhoff:
Only match `module` at the beginning of a line.

Matching anywhere yields [false positives](https://github.com/search?q=language%3Asml+module+-extension%3Asig+-extension%3Acache&type=Code).


arfon:
:+1:

PR #2319, opened 2015-04-09
Fixing incorrect detection of the NewLisp language

Ignotus:
The proposed sample file that contains tokenized Dutch words has been wrongly recognized as a NewLisp code in my repository.

This patch is a try to fix this ambiguity.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -222,6 +222,15 @@ module Linguist
       end
     end
 
+    disambiguate "Text", "NewLisp" do |data|
+      if /^\s*\(define /.match(data)
+        Language["NewLisp"]
+      else
+        Language["Text"]
+      end
+    end
+
+
     disambiguate "TypeScript", "XML" do |data|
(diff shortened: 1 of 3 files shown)


pchaigno:
Isn't the classifier sufficient to solve this issue? I.e., does adding new samples solve it?
Otherwise, have you tested that this new heuristic doesn't break recognition of NewLisp files?

Ignotus:
Yeah, you can try to check
```
bundle exec linguist samples/Text/corpus_1000.nl
```
No, I only tested on unit tests.

pchaigno:
> Yeah, you can try to check
```
bundle exec linguist samples/Text/corpus_1000.nl
```

The files used for training should always be recognized correctly (otherwise there's something really wrong with the classifier). Do you have some other files wrongly detected as NewLisp?

Ignotus:
No, unfortunately. I have fixed in my project by adding .gitattributes. Probably I will create another pull request in the case if I will have enough data to test in the future.

PR #2322, opened 2015-04-10
Make Common Lisp heuristic case insensitive.

larsbrinkhoff:
Some Common Lisp files are misidentified as NewLisp.  This is mostly because the heuristics don't recognize `(DEFUN` in Common Lisp files:

https://github.com/search?q=language%3Anewlisp+%22defun%22+-extension%3Anl&type=Code

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -129,7 +129,7 @@ module Linguist
     end
 
     disambiguate "Common Lisp", "OpenCL", "Cool" do |data|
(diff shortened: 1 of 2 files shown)


bkeepers:
Good catch. Are there any tests we could add to prove this?

larsbrinkhoff:
@bkeepers Added a sample which fails the tests without the updated heuristics.

arfon:
:+1: thanks @larsbrinkhoff

Issue #1483, opened 2014-08-28
Add Windows Batch Language


Kwimbee:
Please add the windows "Batch" Language.

It's still used by some people and tagged as Shell by Github.

Thanks

arfon:
:flags: - flagging this as stale.

arfon:
> @bkeepers In light of #1548, should we just remove Batch from the Shell group?

Crazy slow response here but @pchaigno I think this is the correct way to proceed.

bkeepers:
doh, sorry for dropping the ball.

PR #2324, opened 2015-04-10
Remove Batch from Shell group

pchaigno:
This pull request removes Batch from the Shell group as discussed in #1483.


bkeepers:
Got a test failure from that.

arfon:
:+1:

PR #2325, opened 2015-04-10
Adding support for Clarion

arfon:
Replaces https://github.com/github/linguist/pull/1770


arfon:
@fushnisoft sorry it's taken a while to get to this. We can definitely add support for Clarion but I'm afraid we can't include `.inc` until we've resolved https://github.com/github/linguist/issues/2180

As Linguist will now support Clarion, a short term fix would be to set `.inc` files as Clarion manually using the Linguist overrides: https://github.com/github/linguist#overrides

fushnisoft:
@arfon many thanks!

PR #2330, opened 2015-04-13
Add 'Linker Script' language

larsbrinkhoff:
Add `Linker Script` language.  This is the command language for GNU Link and other Unix linkers.

[300k `.ld` files](https://github.com/search?q=extension%3Ald+NOT+slartibartfast)
[400k `.lds` files](https://github.com/search?q=extension%3Alds+NOT+slartibartfast)
["many" `ld.script` files](https://github.com/search?q=filename%3Ald.script) (is there a way to get exact matches with `filename:`?)

I didn't find any significant other use of `.ld` or `.lds`.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1703,6 +1703,16 @@ Limbo:
   tm_scope: none
   ace_mode: text
 
+Linker Script:
+  type: data
+  extensions:
+  - .ld
+  - .lds
+  filenames:
+  - ld.script
+  tm_scope: none
+  ace_mode: text
+
 Liquid:
   type: markup
   extensions:
--- a/samples/Linker Script/filenames/ld.script
+++ b/samples/Linker Script/filenames/ld.script
@@ -1,0 +1,50 @@
+/*
+ * ld.script for compressed kernel support of MIPS
+ *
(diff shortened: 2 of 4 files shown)


larsbrinkhoff:
Historical note from [http://airs.com/ian/](Ian Lance Taylor):

> I believe [the GNU linker script language] is somewhat loosely based on the script language used by the SVR3 linker, plus significant additions from the MRI linker scripting language.

lib/linguist/languages.yml
```diff
@@ -1703,6 +1703,16 @@ Limbo:
   tm_scope: none
   ace_mode: text
 
+Linker Script:
+  type: data
```
pchaigno: Why not `programming`?

larsbrinkhoff: Because, even though it's called a script, the language doesn't provide for programming constructs.  I.e. no conditionals, variables, loops, procedures, etc.

pchaigno: Hum, I see...

pchaigno:
:+1: thanks @larsbrinkhoff!

larsbrinkhoff:
Ping @arfon @bkeepers

arfon:
Looks good. Thanks @larsbrinkhoff

Issue #2318, opened 2015-04-08
Puppet incorrectly classified as Pascal


3flex:
This file is Puppet code, not Pascal: https://github.com/jfryman/puppet-nginx/blob/master/manifests/init.pp

It might help if a new sample is added which contains class parameters, none of the samples include this:

```puppet
    # A class with parameters
    class apache ($version = 'latest') {
      package {'httpd':
        ensure => $version, # Using the class parameter from above
        before => File['/etc/httpd.conf'],
      }
      file {'/etc/httpd.conf':
        ensure  => file,
        owner   => 'httpd',
        content => template('apache/httpd.conf.erb'), # Template from a module
      }
      service {'httpd':
        ensure    => running,
        enable    => true,
        subscribe => File['/etc/httpd.conf'],
      }
    }
```

Example is from https://docs.puppetlabs.com/puppet/latest/reference/lang_classes.html#syntax

Also missing is an example of inheritance which the problem file also uses.

```puppet
class example (
  $my_param = $example::params::myparam
) inherits example::params {

  notice('This is a test notice')
}
```

Modified from https://docs.puppetlabs.com/puppet/latest/reference/lang_classes.html#aside-when-to-inherit

pchaigno:
New samples are always welcome!
Do you want to open a pull request? I'm not from GitHub so I won't be able to merge your PR (though it should be merged quickly) but if there's any issue I can help you.

3flex:
So I just drop these into /samples/Puppet, then they'll do their magic?

It would probably be best to have a disambiguation in heuristics.rb but I don't know Pascal nor am I any good with regular expressions.

3flex:
Now I think about it, if someone could help come up with something that detects the first keyword in the file is either `class`, `define` or `node` (ignoring any comments in the preceding lines [i.e. lines that start with `#`]) then we can say that's Puppet. It would be good enough for most cases.

pchaigno:
We usually prefer to add heuristic rules if the issue can't be solved only with the Bayesian classifier. We could first try to add new samples and if that doesn't solve it (we'll need incorrectly classified repositories to test it), we can build a new heuristic.

arfon:
> We usually prefer to add heuristic rules if the issue can't be solved only with the Bayesian classifier. 

Yes, please add some samples to `/samples/Puppet`. If this doesn't fix the issue then let's look at adding a heuristic.

3flex:
[...]

PR #2333, opened 2015-04-15
Add additional Puppet sample

3flex:
Will hopefully address #2318 

This file was copied directly from https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/init.pp

It's Apache 2.0 licensed - I'm not sure what's required here in terms of carrying that license over, should I  the license to the top of the file as a comment?

### DIFF
--- a/samples/Puppet/apacheinit.pp
+++ b/samples/Puppet/apacheinit.pp
@@ -1,0 +1,378 @@
+# Class: apache
+#
+# This class installs Apache
+#
+# Parameters:
+#
+# Actions:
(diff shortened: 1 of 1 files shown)


bkeepers:
Hi @3flex, thanks for the pull request!

> It's Apache 2.0 licensed - I'm not sure what's required here in terms of carrying that license over, should I the license to the top of the file as a comment?

We haven't required licenses for samples in the past, but I have an inquiry with our legal department to see what we really should be doing here.

PR #2335, opened 2015-04-16
Change Salt to programming per #2307

DavidJFelix:
- Add a color
- Remove group

PR #2337, opened 2015-04-16
More Language Colors

phase:
More colors. Don't merge yet, as I want to get some more added.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -164,6 +164,7 @@ Apex:
 
 AppleScript:
   type: programming
+  color: "#E0224B"
   aliases:
   - osascript
   extensions:
@@ -268,6 +269,7 @@ Awk:
 Batchfile:
   type: programming
   search_term: bat
+  color: "#382428"
   aliases:
   - bat
   - batch
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -157,6 +157,7 @@ ApacheConf:
 
 Apex:
   type: programming
+  color: "#EB711A"
   extensions:
   - .cls
   tm_scope: source.java
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -157,7 +157,7 @@ ApacheConf:
 
 Apex:
   type: programming
-  color: "#EB711A"
+  color: "#91E6C3"
   extensions:
   - .cls
   tm_scope: source.java
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -157,7 +157,6 @@ ApacheConf:
 
 Apex:
   type: programming
-  color: "#91E6C3"
   extensions:
   - .cls
   tm_scope: source.java
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -164,7 +164,7 @@ Apex:
 
 AppleScript:
   type: programming
-  color: "#E0224B"
+  color: "#F71445"
   aliases:
   - osascript
   extensions:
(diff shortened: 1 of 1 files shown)


arfon:
@phase - thanks for these. Let me know when you're ready to have this merged. Also, keep an eye on the build failures please - some of these colours are clashing with existing ones.

arfon:
@phase how is this PR going? Looks like there are still some failures.

phase:
@arfon The fail system is a little weird. It doesn't tell what languages are making the original language fail, and it goes in alphabetical order, meaning adding one language could make 5 neighbors fail and you wouldn't know which one caused it if you added a lot.

phase:
Nope. Not doing this anymore.

arfon:
> Nope. Not doing this anymore.

Sorry this is so frustrating @phase. I wonder if we should consider making this a 'warn' in the test suite rather than an outright fail. Thoughts @gjtorikian?

phase:
But what if someone puts it too close and no one notices? Maybe we can make an online tester, so there aren't so many bad commits.

pchaigno:
@phase Are you able to test locally?
<br/>

> Maybe we can make an online tester

@DavidJFelix is working on something like this that could eventually become a web app. See his comment at [#2335 (Comment)](https://github.com/github/linguist/pull/2335#issuecomment-94498745).

phase:
@pchaigno I'm on my tablet most the time, and it doesn't have Ruby on it. A web app would be great.

gjtorikian:
> Thoughts @gjtorikian?

To be perfectly honest, I, too, was a bit frustrated by the test. Mostly because I would end up picking a color that would inevitably affect a different language's color than the one I was trying to fix. 

I'm not a fan of making this a warning, but I do think there could be some better reporting. Perhaps it could suggest a color that's *just far enough* to the color you like so as not to interfere with any other languages? That seems...difficult to implement.

I also wouldn't advise using a non-programming device (e.g. phone, tablet, smart fridge) to try and make quick changes. Selecting the right language [can be a time consuming process](https://github.com/github/linguist/pull/2298#issuecomment-90150424). 

I like the web app idea a lot--that way, you'd get instant feedback on your selection. I'd be perfectly happy to collaborate on that work. (Doesn't sound like @DavidJFelix, but working on this test made me discover the joy of using math to compare colors.)

phase:
Yeah, a smart fridge isn't the best for compiling code.

Could we convert it to JavaScript? We might be able to [load the Yaml file](https://raw.githubusercontent.com/github/linguist/master/lib/linguist/languages.yml) using XHR, then [parse it](https://github.com/nodeca/js-yaml).

DavidJFelix:
@gjtorikian So what I'm working on now is a bit like a traditional color picker (I was picturing something like http://www.colorpicker.com/) but rather than featuring the whole spectrum of colors, it would be like a swiss cheese of that -- holes representing areas that would fail the test.

I haven't decided the best way to represent those holes graphically yet. Right now I have a python-based tool that walks through all possible colors and tests them against the set of colors used. It would be possible to modify this function to spiral out from a hex color and search until it finds something - I wouldn't think that would be too hard, if that's the approach. I was thinking with the "swiss cheese" view, it might be easier for somebody looking to make room to do the aforementioned language color shuffle.

I'll clean up the python code a bit and get a gist or repo up for it.

phase:
@DavidJFelix You could use R, though I don't know how well it interfaces with Python.

DavidJFelix:
I haven't forgotten about this, I just hit some delays.

pchaigno:
@DavidJFelix No worries ;)
Do you have a development version hosted on GitHub?

PR #2339, opened 2015-04-16
Improve NL/NewLisp disambiguation heuristic

vitaut:
The proposed PR improves disambiguation between NL and NewLisp formats fixing cases like [this](https://github.com/rwcarlsen/pswarm/blob/c51c223e7834c7eb297d4ba96540cee50371ab29/nl/hs35mod.nl).


arfon:
:+1: looks good. Thanks.

PR #2340, opened 2015-04-17
Remove .script! hack

bkeepers:
This replaces #1786, which removes the `.script!` hack that was used for extension-less executable files. We now detect shebang before looking at the filename, so this is not necessary.

/cc @arfon @pchaigno


lib/linguist/samples.rb
```diff
@@ -50,16 +50,13 @@ def self.each(&block)
             end
           else
             path = File.join(dirname, filename)
-
-            if File.extname(filename) == ""
-              raise "#{path} is missing an extension, maybe it belongs in filenames/ subdir"
```
pchaigno: Does it mean that `filenames/` subdirectories are deprecated now?

bkeepers: Nope, it now allows files without an extension to exist _outside_ of `filenames`.

pchaigno: Okay, I see :-)

pchaigno:
:+1:

PR #2341, opened 2015-04-17
Move Linguist::Language.detect to Linguist.detect

bkeepers:
It has always bothered me that the main public method for linguist was buried in the `Language` class in the middle of a bunch of other internal methods.

This renames `Linguist::Language.detect` to `Linguist.detect` and deprecates the old method. Thoughts?

Since this is a new method, I would recommend we release this in 4.6.0, and then remove the deprecated method in 5.0.

/cc @arfon @pchaigno @larsbrinkhoff

### DIFF
--- a/lib/linguist.rb
+++ b/lib/linguist.rb
@@ -8,8 +8,85 @@ require 'linguist/shebang'
 require 'linguist/version'
 
 class << Linguist
+  # Public: Detects the Language of the blob.
+  #
+  # blob - an object that includes the Linguist `BlobHelper` interface;
+  #       see Linguist::LazyBlob and Linguist::FileBlob for examples
+  #
+  # Returns Language or nil.
+  def detect(blob)
+    # Bail early if the blob is binary or empty.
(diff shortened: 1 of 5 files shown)


pchaigno:
:+1:

larsbrinkhoff:
Great!  I was just now roaming around those areas, not finding `detect` where I expected it to be.  This is much better.

arfon:
:+1: - yeah, this has always seemed awkward to me also.

> Since this is a new method, I would recommend we release this in 4.6.0, and then remove the deprecated method in 5.0.

:+1:

arfon:
> Since this is a new method, I would recommend we release this in 4.6.0, and then remove the deprecated method in 5.0.

@bkeepers - we're currently at [`v4.7.6`](https://github.com/github/linguist/blob/master/lib/linguist/version.rb#L2) so would you suggest bumping to `v4.8.0` for this change?

bkeepers:
:+1:

PR #2342, opened 2015-04-17
Analyse an individual file using repository information.

larsbrinkhoff:
Running *Linguist* on an individual file will not take `.gitattributes` into account.  This affects the search index.

I looked into fixing this, and came up with this pull request.  It may well be wrong in many ways, so I'm mostly putting it up as a conversation item.  It turns out it will not help with the index, but perhaps it can be useful anyway.

In short, I added an optional directory argument after the file name.  If supplied, a `LazyBlob` will be created instead of a`FileBlob`.

### DIFF
--- a/bin/linguist
+++ b/bin/linguist
@@ -37,7 +37,16 @@ if File.directory?(path)
     end
   end
 elsif File.file?(path)
-  blob = Linguist::FileBlob.new(path, Dir.pwd)
+  if (ARGV[0])
+    base = ARGV[0]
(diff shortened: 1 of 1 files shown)


pchaigno:
How should we use it?
Like this? `bundle exec linguist ../path/to/file ../path/to/repository/with/gitattribute`

larsbrinkhoff:
@pchaigno Right!  And of course this only makes sense if the file is in that repository.

PR #2346, opened 2015-04-19
Add support for SSH2 and OpenSSH public keys

larsbrinkhoff:
Linguist currently recognizes `.asc` files with `-----BEGIN` as public key files (this is the "PEM" format).  [SSH2](https://tools.ietf.org/html/rfc4716#section-3.2) uses `---- BEGIN`, and [OpenSSH files](http://tools.ietf.org/html/rfc4253#section-6.6) start with `ssh-rsa` or `ssh-dss`.


lib/linguist/heuristics.rb
```diff
@@ -155,11 +155,13 @@ def call(data)
...
     disambiguate "AsciiDoc", "AGS Script", "Public Key" do |data|
-      if /^[=-]+(\s|\n)|{{[A-Za-z]/.match(data)
+      if /^---- BEGIN /.match(data)
+        Language["Public Key"]
+      elsif /^[=-]+(\s|\n)|{{[A-Za-z]/.match(data)
         Language["AsciiDoc"]
       elsif /^(\/\/.+|((import|export)\s+)?(function|int|float|char)\s+((room|repeatedly|on|game)_)?([A-Za-z]+[A-Za-z_0-9]+)\s*[;\(])/.match(data)
         Language["AGS Script"]
-      elsif /^-----BEGIN/.match(data)
+      elsif /^(-----BEGIN|ssh-(rsa|dss)) /.match(data)
```
arfon: I wonder if we should combine these into a single expression on L158?

larsbrinkhoff: Done.

arfon:
:+1: thanks @larsbrinkhoff

PR #2168, opened 2015-04-19
.run file extension

pchaigno:
There is [more than 9,900 `.run` files on GitHub](https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+NOT+djdjfkdffgjderskskvlr&type=Code&ref=searchresults). It is used by many languages. Thus, to add that extension, we need to identify all languages using it (with at least hundreds of examples).

Language | Number of files
------- | ------------
Shell | [1,900](https://github.com/search?p=1&q=extension%3Arun+bash&ref=searchresults&type=Code&utf8=%E2%9C%93)
Perl | [1,000](https://github.com/search?l=perl&q=extension%3Arun+NOT+dkfjdkfndrnjbkbh&ref=searchresults&type=Code&utf8=%E2%9C%93)
Ruby | [900](https://github.com/search?p=6&q=extension%3Arun+puts&ref=searchresults&type=Code&utf8=%E2%9C%93)
AMPL | [400](https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+solve&type=Code&ref=searchresults)

I will update this post with the list of languages as we go. I might eventually turn this issue into a pull request to add support of the extension.

This extension was requested for AMPL in #2073 (but I can't make a search query with more than 200 files without having a lot of other languages).


pchaigno:
@arfon I just turned this issue into a pull request to add support for `.run`. Could you send me sample files for the three searches linked above?

vitaut:
There are also a few hundred of AMPL .run files on GitHub. For example, https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+solve&type=Code&ref=searchresults

pchaigno:
Thanks @vitaut, I couldn't find the right keyword :/
I'm not sure 440 files will be enough to add it as a new extension in Linguist though. @arfon What do you think?

vitaut:
It could be much more than 400 as not all AMPL scripts contain the `solve` statement. I used `solve` only because it is more specific to AMPL and gives few false positives.

arfon:
> Could you send me sample files for the three searches linked above?

Done.

pchaigno:
Thanks @arfon! Also, do you think we should add the `.run` extension for AMPL?

arfon:
> Thanks @arfon! Also, do you think we should add the `.run` extension for AMPL?

Yeah let's do it.

arfon:
Just catching up on Linguist stuff - is this good to go @pchaigno?

pchaigno:
> Just catching up on Linguist stuff - is this good to go @pchaigno?

No, not yet. The results are not good enough so I'll have to use heuristic rules :/

arfon:
Just looping back on this @pchaigno - trying to clear up some old PRs. Any further thoughts on this?

pchaigno:
Current results:

|  | Shell | Perl | Ruby | AMPL
|---|---|---|---|---
| [1,000 Shell files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+then&type=Code&ref=searchresults) | 27.32% | **68.20%** | **1.50%** | **2.97%** 
| [1,000 Perl files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+use&type=Code&ref=searchresults) | **7.99%** | 88.67% | **2.48%** | **0.86%** 
| [900 Ruby files](https://github.com/search?p=6&q=extension%3Arun+puts&ref=searchresults&type=Code&utf8=%E2%9C%93) | **49.95%** | **42.27%** | 7.77% |  0.00%
| [400 AMPL files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Arun+solve&type=Code&ref=searchresults) | **13.64%** | **29.83%** | 0.00% | 56.53%

I'll see if adding new samples can improve things, but I will probably have to use heuristic rules anyway :/

pchaigno:
Going through the incorrectly identified files (see above), it appears they are essentially noise. Shell, Perl, Ruby and AMPL files are fairly well detected but there are many other types of files (I found some C files but I couldn't identify the others).
I'm not sure any of the Ruby files are actually Ruby. They seem to be related to the [Genode OS framework](http://genode.org/).
I don't think we should add the `.run` extension until we have a way to classify as `other`.

arfon:
> I don't think we should add the `.run` extension until we have a way to classify as `other`.

This sounds like a good decision. Thanks for giving this a go @pchaigno.

PR #2348, opened 2015-04-20
Extensions for the Bazel build system are written in a Python dialect

hanwen:
called Skylark, using the extension .bzl.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2579,6 +2579,7 @@ Python:
   - .tac
   - .wsgi
   - .xpy
+  - .bzl
   filenames:
   - BUILD
   - SConscript


pchaigno:
@hanwen Please see the requirements to add a new language in Linguist as per the [contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language).

Also, the first extension needs to be the primary extension. So `.bzl` should come after `.py`.

bkeepers:
It looks like there are [only 24 .bzl files on GitHub](https://github.com/search?p=1&q=extension%3Abzl+NOT+nothack&ref=searchresults&type=Code&utf8=%E2%9C%93). We normally require hundreds of samples before adding support for an extension, but http://bazel.io/ is new and seems pretty legit. 

Docs on skylark: http://bazel.io/docs/skylark/

@arfon @pchaigno `.bzl` doesn't conflict with any extensions. What do you think about adding it?

hanwen:
I didnt know about the 100s of samples policy. I'm happy to leave this as is, and add a .gitattributes to our repo for the time being.

pchaigno:
> @arfon @pchaigno .bzl doesn't conflict with any extensions. What do you think about adding it?

I don't have all the information to judge the impact it could have (on performances in particular, if we add many languages). But I sounds like a "dangerous" precedent. If we allow this language why should we refuse other legit languages?
<br/>

@hanwen Whatever the decision made, thank you for your understanding :+1:

bkeepers:
> But I sounds like a "dangerous" precedent. If we allow this language why should we refuse other legit languages?

Yeah, I think you're right.

@hanwen we're happy to add support once there's a little more adoption. Thanks for your understanding.

pspeter3:
Is it possible to add this now?

bkeepers:
> We normally require hundreds of samples before adding support for an extension…

By "samples" here I meant "repositories". Most of the samples are still only in in a handful of repositories. I think it still needs a little more adoption. 2¢

Issue #2278, opened 2015-03-26
Add list of valid (or relevant) language identifier keywords for GFM syntax highlighting


jmm:
I created this [list of language keywords](https://github.com/jmm/gfm-lang-ids/wiki/GitHub-Flavored-Markdown-%28GFM%29-language-IDs) for enabling syntax highlighting and asked GitHub support to consider linking to it from the [GitHub Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/) page. People who post issues without even marking up code are not going to be able to figure out the YAML file it currently refers them to. And for that matter, I don't know how anyone is supposed to understand it since there's no explanation of what values from it are relevant.

@gjtorikian advised me to submit a request to integrate that content into this repo.

I wrote a [node program](https://github.com/jmm/gfm-lang-ids) (it's pretty quick and dirty) to automatically generate the document from the `languages.yml` file. I made the document a wiki so that there would be a fallback in case it needed updating and I was unavailable. With the content in this repo it might make more sense to just check it in, to encourage updating it by running a script.

I looked around for a while and then gave up on figuring out where the logic is that defines the valid values. My theory is that it works like this (at least in part):

Any of the following (case-insensitive, with or without a `.` prepended) can be used as the keyword:
- top level key, with spaces replaced by hyphens
- `aliases` values that don't contain spaces
- `extensions` values

That's how I believe it works so that's how I wrote my script, but frankly I think the document it creates is massive overkill. Take JavaScript for example. My script generates the following list of keywords to enable JS highlighting:
- `_js`
- `bones`
- `es6`
- `frag`
- `gs`
- `jake`
- `javascript`
- `js`
- `jsb`
- `jsfl`
- `jsm`
- `jss`
- `jsx`
- `njs`
- `node`
- `pac`
- `sjs`
- `ssjs`
- `sublime-build`
- `sublime-commands`
- `sublime-completions`
- `sublime-keymap`
- `sublime-macro`
- `sublime-menu`
- `sublime-mousemap`
- `sublime-project`
- `sublime-settings`
- `sublime-theme`
- `sublime-workspace`
- `sublime_metrics`
- `sublime_session`
- `xsjs`
- `xsjslib`

For the purposes of enabling syntax highlighting in GFM I think there should just be `js` (_maybe_ also `javascript`). It's probably too late to not support the others, but I think only 1 or 2 are relevant to include in a general reference list. The only value I can imagine in including the others is to leave the door open to provide more specific highlighting for them in the future.

[...]

Issue #2331, opened 2015-04-13
Write 'how Linguist integrates into GitHub' docs


arfon:
> In github/markup#431 I added a section in the README on how the library is integrated into GitHub.com. I wonder if it would make sense to add something similar here.

Basically we need a much better version of this: https://github.com/github/linguist/pull/2093

arfon:
Also / cc https://github.com/github/linguist/issues/2104

larsbrinkhoff:
Plus one; thumbs up.  Suggested items to cover:

* How a repository is assigned a primary language, which is used for "Trending" lists, repository searches etc.
* How the repository language statistics are calculated and displayed in a bar.
* How individual files are indexed with respect to language.

gjtorikian:
Also /cc https://github.com/github/linguist/issues/2278. Unsure if you're thinking of a new docs folder or what but I can commit to doing that issue above.

jimmyye:
When  a new version of Linguist released, how long will it take GitHub to adapt those changes (new languages, languages renamed)?

How about a /languages or /linguist API to list all languages that GitHub is using? Something like https://github.com/github/linguist/blob/master/lib/linguist/languages.yml but in json.

gjtorikian:
> How about a /languages or /linguist API to list all languages that GitHub is using? 

That's not a bad idea. It's pretty inline with our existing [Emoji API](https://developer.github.com/v3/emojis/).

What do you think, @brandonblack? Maybe we could make a `Meta` object in GraphQL for stuff like this.

PR #2353, opened 2015-04-22
Add docs on how highlighting works

gjtorikian:
Closes https://github.com/github/linguist/issues/2278, /cc https://github.com/github/linguist/issues/2331.

This is a spike at providing a comprehensive list for which syntax highlight codes are valid on GitHub. I verified a few of the weirder ones manually. For example, something like

    ``` storyboard
    <wow>
    ```

    ``` reek
    - wow
    ```

does actually work:

``` storyboard
<wow>
```

``` reek
- wow
```

I wonder also what a test for this might look like, if we wanted one.

/cc @github/languages @jmm

### DIFF
--- a/docs/highlighting.md
+++ b/docs/highlighting.md
@@ -28,7 +28,7 @@ Whereas for AsciiDoc, this might be:
 
     [source,javascript]
 
-Below is a comprehensive list of of the language codes necessary for syntax highlighting a code block:
+Below is a comprehensive list of of the language codes necessary for syntax highlighting a code block. Generally speaking, the slugified version of a language name, as well as any of its extensions, defined the language code.
 
 | Languages | Codes
 |-----------|-------
--- a/docs/highlighting.md
+++ b/docs/highlighting.md
@@ -1,0 +1,31 @@
+# Highlighting
+
+GitHub uses Linguist to power the syntax highlighting of code blocks. You know, the difference between this
+
+```
+var util = require('util');
+var net = require('net');
+var stream = require('stream');
+var url = require('url');
+```
+
+and this
+
+``` js
+var util = require('util');
+var net = require('net');
+var stream = require('stream');
+var url = require('url');
+```
+
+In order to syntax highlight a code block, you must provide a "language name" to the code block syntax of your Markup of choice.
+
+For example, for Markdown, this would look like:
+
+    ``` javascript
+
+Whereas for AsciiDoc, this might be:
+
+    [source,javascript]
+
+Below is a comprehensive list of of the language codes necessary for syntax highlighting a code block.
--- a/docs/highlighting.md
+++ b/docs/highlighting.md
@@ -28,4 +28,363 @@ Whereas for AsciiDoc, this might be:
 
     [source,javascript]
 
-Below is a comprehensive list of of the language codes necessary for syntax highlighting a code block.
+Below is a comprehensive list of of the language codes necessary for syntax highlighting a code block:
+
+| Languages | Codes
+|-----------|-------
+| ABAP | abap
+| AGS Script | ags-script, asc, ash
+| AMPL | ampl
+| ANTLR | antlr, g4
+| APL | apl, dyalog
+| ASP | asp, asax, ascx, ashx, asmx, aspx, axd
+| ATS | ats, dats, hats, sats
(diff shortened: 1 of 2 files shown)


pchaigno:
:bow: 

This means the table will have to be updated from time to time, right? Could it be done with each new release (and add a note about it at [CONTRIBUTING.md#releasing](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#releasing))?

aroben:
> I wonder also what a test for this might look like, if we wanted one.

If the script were able to update the highlighting.md file directly, we could run the script in CI and assert that the file is not changed. If the file *did* change, it means the PR author should run the script and check in the changes.

docs/highlighting.md
```diff
@@ -0,0 +1,390 @@
...
+| desktop | desktop, desktopin
+| eC | ec, eh
+| edn | edn
+| fish | fish
+| mupad | mupad, mu
+| nesC | nesc, nc
+| ooc | ooc
+| reStructuredText | restructuredtext, rst, rest
+| wisp | wisp
+| xBase | xbase, prg
```
bkeepers: This seems prone to get out of date–even with the script to update it. I'm curious what the value is to listing all these out. Will people actually reference this?

> Generally speaking, the slugified version of a language name, as well as any of its extensions, defined the language code.

That seems like what people really want to know.

bkeepers:
> If the script were able to update the highlighting.md file directly, we could run the script in CI and assert that the file is not changed.

The annoying part of this is that it will require anyone adding a language to also update the docs. I would rather see us only update this on releases or something to take the burden off contributors.

aroben:
> it will require anyone adding a language to also update the docs

When "update the docs" is `script/update-highlighting-table && git commit docs/highlighting.md` it doesn't seem so bad.

docs/highlighting.md
jmm: I [suggested](https://github.com/github/linguist/issues/2278) linking from the GFM help page to a plain list of language codes (instead of a YAML file) that can be used for marking up fenced code blocks. For that purpose I proposed constraining the list to a handful of meaningful values, e.g. `js` for JavaScript (and *maybe* also `javascript`). For that use case I think listing all of these is overkill.

bkeepers:
Would it make sense to build out a GH pages site with this kind of docs? Right now the gh-pages branch just has some stuff I hacked together to show colors, but there's no reason we couldn't do more with it, like documenting syntax highlighting (or even integrating https://github-lightshow.herokuapp.com/).

gjtorikian:
> Would it make sense to build out a GH pages site with this kind of docs?

In correlation with @DavidJFelix's [work to implement an online color picker](https://github.com/github/linguist/pull/2337#issuecomment-95395582), I was having the same thought earlier today. Like, given the increase in responsibilities for Linguist, it might be best to move out of flat documents in the folder and towards a Pages site. As before, /cc https://github.com/github/linguist/issues/2331. 

I'm :+1: on the idea. I could make time to set the site up if @arfon / @aroben et al are into it. We'll probably need some design help but that could be addressed after the content is in.

aroben:
A Pages site sounds like a good idea if there are things we want to do that can't be done in a Markdown file.

gregorym:
@bkeepers Is the source of https://github-lightshow.herokuapp.com/ available somewhere?

aroben:
@gregorym It is not. Unfortunately the syntax highlighter itself is not public at the moment, and since Lightshow embeds it it can't be public either.

arfon:
Closing this as stale. Would still love to see this :ship: but it doesn't seem to be a focus for anyone right now.

jmm:
@arfon In light of this development perhaps you could reconsider my [original request](https://github.com/github/linguist/issues/2278) from almost a year ago to add a link to the [list of language identifier keywords](https://github.com/jmm/gfm-lang-ids/wiki/GitHub-Flavored-Markdown-(GFM)-language-IDs) I created on [Creating and highlighting code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/#syntax-highlighting) instead of just having it link to a [YAML file](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) that is very un-user-friendly :D

Glenntobby:
@ #2923 #2919

PR #2354, opened 2015-04-22
Update Objective-C check for Mathematica

### LINKED ISSUE #2197: Objective-C is being confused with Mathematica
[This file](https://github.com/d235j/360Controller/blob/master/360Daemon/360Daemon.m) is actually Objective-C, but the Linguist thinks it's Mathematica. There are no other files that I know of that are like this.

phase:
Fixes #2197 Code from @pchaigno, I just made the PR. :rabbit:

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -61,9 +61,6 @@ module Linguist
       @heuristic.call(data)
     end
 
-    # Common heuristics
-    ObjectiveCRegex = /\#(include|import|define)/
-
     disambiguate "BitBake", "BlitzBasic" do |data|
       if /^\s*; /.match(data) || data.include?("End Function")
         Language["BlitzBasic"]
@@ -81,7 +78,7 @@ module Linguist
     end
 
     disambiguate "Objective-C", "C++", "C" do |data|
-      if ObjectiveCRegex.match(data)
+      if /^[ \t]*@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/.match(data)
         Language["Objective-C"]
       elsif (/^\s*#\s*include <(cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream)>/.match(data) ||
         /^\s*template\s*</.match(data) || /^[ \t]*try/.match(data) || /^[ \t]*catch\s*\(/.match(data) || /^[ \t]*(class|(using[ \t]+)?namespace)\s+\w+/.match(data) || /^[ \t]*(private|public|protected):$/.match(data) || /std::\w+/.match(data))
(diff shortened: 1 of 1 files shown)


phase:
```objective-c
@interface StyleViewController : TTViewController {
@private
  TTStyle*  _style;
  TTStyle*  _styleHighlight;
  TTStyle*  _styleDisabled;
  TTStyle*  _styleSelected;
  NSString* _styleType;
}
@end
```
This fails. Can we separate the `.h` & `.m`? I see no reason for them to be shared in one variable.

phase:
The `#include` in `store.m` is throwing an error because it has a string of C code. I wonder if we have to have `include`, as `import|define` might be enough.

pchaigno:
> The #include in store.m is throwing an error because it has a string of C code.

We could exclude it from the test.
<br/>

> I wonder if we have to have include, as import|define might be enough.

You should check against a corpus of files. If you have a link to a GitHub search, @arfon can send you the first 1,000 file results.
<br/>

> We may also be able to check if the code is in a string, or make string-checking a standard for all languages.

There was [a GSoC idea](https://github.com/github/gsoc/blob/040f3f2c294f9e4f5bd8e1ab2269f96e23bb0fe8/README.md#use-grammar-bundles-to-tokenize-languages) related to that.

phase:
If we exclude one file, what if another file comes along somewhere with the same code? We can't just ignore every file that has a string of another language.

lib/linguist/heuristics.rb
```diff
@@ -61,9 +61,6 @@ def call(data)
       @heuristic.call(data)
     end
 
-    # Common heuristics
-    ObjectiveCRegex = /^[ \t]*@(interface|class|protocol|property|end|synchronised|selector|implementation)\b/
```
arfon: I'm not sure I understand the rationale for having a different heuristic for `Objective-C` in different scenarios. Is this necessary? There are good reasons for wanting to extract this into a variable.

phase: There are two different ways for detecting ObjC, each one for a different file type.

`.h` will be something like a class, so it has the check at line 81

`.m` will include something, so there's that.

pchaigno: It much more difficult to differentiate Objective-C from C and C++ than it is to differentiate it from Limbo, M, MUF, Mathematica, Matlab and Mercury. That's why I think we can use a simpler regex (faster?) for the second case.

arfon:
> You should check against a corpus of files. If you have a link to a GitHub search, @arfon can send you the first 1,000 file results.

Very happy to do this. If you can construct a search [here](https://github.com/search/advanced) @phase then I can go and grab a whole collection of files to test against.

arfon:

PR #2359, opened 2015-04-25
Only require heuristics to match subset of candidates

bkeepers:
Previously, a defined heuristic had to match all candidates to be used. This leads to problems when we add a language extension that conflicts with existing languages used in a heuristic. This inverts the logic so a heuristic only has to include a subset of the candidates to be considered a match.

One added benefit of this change is that we can start to add a single heuristic for languages. See  8309985 for an example, which extracts the Objective C heuristic.

/cc @arfon


lib/linguist/heuristics.rb
```diff
@@ -15,12 +15,15 @@ class Heuristics
     # Returns an Array of languages, or empty if none matched or were inconclusive.
     def self.call(blob, languages)
       data = blob.data
+      language_names = languages.map(&:name)
+      result = nil
 
       @heuristics.each do |heuristic|
-        return Array(heuristic.call(data)) if heuristic.matches?(languages)
+        result = heuristic.call(data) if heuristic.matches?(language_names)
+        break if result
```
bkeepers: These changes are subtle, but now it will keep trying heuristics if a result is not returned. Whereas, before it would just return the result from the first heuristic that matched, even if it was nothing.

bkeepers:
Sorry, one more change I forgot to note: This also removes most uses of the `Language` class. None of the heuristics were using any of the data from the language definitions, so this mostly just makes it a little cleaner.

pchaigno:
This is related to #2042. In this issue, we discussed the problem matching only a subset can bring:
Files with a `.pm` extension match the Perl6-Perl-Prolog heuristic and can potentially be classified as Prolog. This can get worst as the heuristic rules are sometimes written for a specific type of files (see for instance [XML files for translations](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L226)).

arfon:
@bkeepers - :zap: this looks good to go to me?

pchaigno:
How are we going to fix issues such as the one with `.pm` and Perl6/Perl/Prolog?

larsbrinkhoff:
Another possible solution would be to *enforce* that all languages sharing an extension also has a heuristic that matches all those languages.  Maybe as a test.

pchaigno:
> Another possible solution would be to enforce that all languages sharing an extension also has a heuristic that matches all those languages. Maybe as a test.

But that for many extension conflicts the Bayesian classifier is sufficient :/
<br>

Maybe we should associate an extension to each heuristic and then have a test to check that all language with that extension are handled by the heuristic rule. I think it would make sense because currently we have heuristic rules defined for extensions not for languages. For instance, the following heuristic rule is for `.ts`:
```ruby
disambiguate "TypeScript", "XML" do |data|
  if data.include?("<TS ")
    Language["XML"]
  else
    Language["TypeScript"]
  end
end
```
It could not be applied to other XML files...

larsbrinkhoff:
> Maybe we should associate an extension to each heuristic

[Yes](https://github.com/github/linguist/issues/2042#issuecomment-82349279).

bkeepers:
> Maybe we should associate an extension to each heuristic

I think that makes sense. I'll close this out.

Issue #2366, opened 2015-04-30
GML Geography Markup Language showing as Game Maker Language


vog:
The .gml files in my repository are classified as Game Maker Language, but these are [OGC Geography Markup Language](https://en.wikipedia.org/wiki/Geography_Markup_Language) files.

There seem to be many [users of the OGC Geography Markup Language](https://github.com/search?utf8=%E2%9C%93&q=extension%3Agml+opengis&type=Code&ref=searchresults). Please add support for them.

pchaigno:
I think we could add it as a new XML extension. Do you want to take a stab at it?

vog:
I quickly added an entry to languages.yml as well as some example files for GML 3.1 and GML 3.2, the rest should be done be somebody else. Even after skimming through CONTRIBUTING.md I have no idea what else is needed, if anything.

pchaigno:
Nothing else is needed ;)

arfon:
I re-ran the language calculation on [this repo](https://github.com/diegomigliavacca/deegree-3.3.11-gpkg) and it looks like this fix has worked.

vog:
Could you also re-run the language calculation on this repo?

https://github.com/deegree/deegree3

arfon:
> Could you also re-run the language calculation on this repo?

Done.

vog:
I'm afraid something went wrong there.  The second largest part was GML.  Now it is HTML. It seems that "Geography Markup Language" has been recognized as HTML instead of XML.

arfon:
HTML looks to be third after XML and Java - is this not correct?

https://github.com/deegree/deegree3/search?l=xml

![screen shot 2015-05-31 at 11 07 32 pm](https://cloud.githubusercontent.com/assets/4483/7903876/da76d330-07e9-11e5-93cc-27fd51f065d2.png)

vog:
That one looks correct, but the summary at the top shows a different picture that confuses me:

![screenshot](https://cloud.githubusercontent.com/assets/412749/7911185/6a65faae-085b-11e5-8dfc-5cc23c285d38.png)

arfon:
Ah OK. The language stats bar excludes languages that are defined as [`data`](https://github.com/github/linguist/blob/f380aa2e5ef1dbad4157f205e4f93a2055f1c7d7/lib/linguist/languages.yml#L3451) which means `XML` doesn't show up here (but it does show up in the search results). I'm sorry for this confusion :-\

vog:
[...]

PR #2368, opened 2015-04-30
Add languages.yml entry and sample files for OGC Geography Markup Language (#2366)

vog:


pchaigno:
Why do you want to define it as a new language if it's has the same syntax as XML?

vog:
I'm not so familar with this project, and I'm afraid I don't understand the question. What else should I have done? Simply adding the samples without changing the languages.yml?

pchaigno:
You can just add `.gml` to the list of extensions for XML and move the samples to `samples/XML/`. No need for a new GML entry here ;)

vog:
But then these files would appear in GitHub as "XML" rather than "Geography Markup Language", wouldn't they?

pchaigno:
> But then these files would appear in GitHub as "XML" rather than "Geography Markup Language", wouldn't they?

Yes they would. But, if we add `Geography Markup Language`, since it's a `data` language, it will not be displayed in the statistics of repositories (only in Explore and the search page).
<br/>

> There are so many XML formats out there that it would be confusing to treat them all the same.

It also means we can't define a language for each one of them ;) We have to find a balance.
I'm not from GitHub so I won't be the one taking that decision but I feel like it should be marked as XML because it's only a subset of XML and there aren't that many GML files on GitHub (compared to the [64 million POM files](https://github.com/search?utf8=%E2%9C%93&q=filename%3Apom.xml&type=Code&ref=searchresults) for instance).

vog:
Okay, updated branch.

pchaigno:
Could you also move the samples to `samples/XML/`? Thanks.

vog:
Sorry, forgot that one. Fixed.

arfon:
:+1: thanks for this and the discussion @vog and @pchaigno. This change will go live in the next release of Linguist.

PR #2369, opened 2015-04-30
Remove executable flag from all sample code files

vog:


pchaigno:
Why do you want to do that??

vog:
These looked like being leftovers, e.g. when the files were created/checked out on a file system that doesn't provider proper support for the executable flag.

More generally, I don't think it is wise to mark files as executable which aren't meant to be executed.

pchaigno:
> More generally, I don't think it is wise to mark files as executable which aren't meant to be executed.

Yes, but the executable flags aren't used to train the classifier and they contribute to our tests so I think it would be best to keep them.

vog:
I wasn't aware of that. In how far do these contribute to the tests? Do you check that these files can actually be run?

pchaigno:
> In how far do these contribute to the tests? Do you check that these files can actually be run?

No, we don't but Linguist uses several strategies. One of them rely on this flag (`Shebang` strategy is only run if this flag is present). Having files with this flag could eventually help us see potential bugs on this strategy.

More generally, because they play a role of samples, I prefer to keep these files close to what they were when I copied them (filename, flag and content). It doesn't hurt and gives us a good insight at what users can sometimes do :frowning:

arfon:
> More generally, because they play a role of samples, I prefer to keep these files close to what they were when I copied them (filename, flag and content). It doesn't hurt and gives us a good insight at what users can sometimes do :frowning:

Agreed. Thanks for your effort here @vog but I'd like to keep the status quo for these files for the reasons that @pchaigno outlines.

Issue #2357, opened 2015-04-23
Incorrect syntax hilighting of Ada character literals


faelys:
Hello,

there is a problem with the syntax hilighting of Ada source. I haven't really understood how linguist works, it seems language-specific hilighting is "vender provided", but for Ada it points to aroben/ada.tmbundle, which has no issue tracker, so I'm reporting it here. Please point me to the right direction if this is not the correct place to report it.

I have no idea what TextMate is or how a "bundle" works, so I can't offer a patch or a pull request, but I will do my best to describe what is wrong here.

To put it simply, it seems character literals in Ada are not recognized at all.

In Ada, much like in C, strings are arrays of characters, and string literals are written between double-quotes (`"`), while character literals are written between single-quotes (`'`). Escaping is a bit different than in C though: a double-quote character is inserted in a string literal using two consecutive double-quote character (e.g. `"This: "" is a single-double quote"`) while the character literal for the simple quote is three consecutive single-quotes (i.e. `'''`). To put it a regex way, character literals match `'.'` and the single-quote character literal is no exception.

As far as I can tell, there is currently no syntax hilighting at all for character literals. Here is an example:

```ada
declare
   Two_SL : String (1 .. 3) := "two";
   Two_CL : String (1 .. 3) := ('t', 'w', 'o');
begin
   Text_IO.Put_Line (Two_SL);
   Test_IO.Put_Line (Two_CL);
end;
```

As of this writing, `1` and `3` are colored specially, `"two"` has another color, `declare`, `begin` and `end` have keyword color, but `'t'` is in plain style like `Two_SL`.

Now in itself, it's a small problem, since character literals are small anyway. However it really hurts readability when using the double-quote character literal:

```ada
procedure Print_String_Value (Value : in String) is
begin
   Text_IO.Put_Line ("The value is: """ & Value & '"');
end Print_String_Value
```

For the reference, here `&` is the concatenation operators, and it is overloaded so that a string can be built by concatenating strings or by concatenating a string and a single character.

This causes the double-quote character literal to be considered as the beginning of a string literal, and the string coloring spans the rest of the code.

There are several real-world examples of this, [here is one of mine](https://github.com/faelys/markup-ada/blob/trunk/tools/markdown.adb#L120)

I wouldn't mind having character literals without special hilighting, if at least it wouldn't incorrectly color large sections of code.

Thanks for your time,
Natasha

arfon:
/ cc @aroben

aroben:
[...]

PR #2372, opened 2015-05-01
Use original Ada grammar

pchaigno:
We were using the branch `better-with-highlighting` on @aroben's fork of https://github.com/aroben/ada.tmbundle. textmate/ada.tmbundle#2 is now fixed and we can use the original repository again.

Discussed in [#2357 (Comment)](https://github.com/github/linguist/issues/2357#issuecomment-98129828).


arfon:
:+1:

PR #2378, opened 2015-05-03
Added support for .mdpolicy XML files

Mailaender:
Used by @MonoDevelop to define standards for source code formatting. Not very popular https://github.com/search?utf8=%E2%9C%93&q=filename%3Amdpolicy&type=Code&ref=searchresults but easy to add here as it is strict XML 1.0 and is not conflicting with anything else.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3414,6 +3414,7 @@ XML:
   - .jelly
   - .kml
   - .launch
+  - .mdpolicy
   - .mm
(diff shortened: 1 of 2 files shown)


arfon:
Thanks @Mailaender - would you mind adding an example file to go along with this change?

Mailaender:
Okay. Added one.

arfon:
Thanks @Mailaender - this will go live in the next release of Linguist.

Issue #2060, opened 2015-02-01
Common Lisp files ending in .l not being recognised


hscells:
Please see https://github.com/hscells/CSC466 for example.

Thank you.

larsbrinkhoff:
This seems tricky.  There are more than 300k `.l` files.  Traditionally, this extension is used for Lex input.  These files usually contain lots of code from the target language, so the Bayesian classifier would have a hard time doing the right thing.  I guess heuristics could latch onto the `%{` and `%}` keywords used by Lex.

I found more than 5000 `.l` files which seems to be Lisp.

pchaigno:
I agree with @larsbrinkhoff (but I'm not part of GitHub staff either). This is going to be a lot of work but it's feasible. We'll first need to identify all languages using this extension.

- [3,600 of them seem to be C#](https://github.com/search?utf8=%E2%9C%93&q=extension%3Al+namespace)
- [a good number seems to be C++/C](https://github.com/search?utf8=%E2%9C%93&q=extension%3Al+include&type=Code&ref=searchresults)

larsbrinkhoff:
@pchaigno Many of those two searches include files which are Lex files.

E.g. the first C# result: https://github.com/jk983294/Store/blob/e1e052231917815dc3f02768a69f70a2f5cfb164/CompileExp/SeuYacc/%E7%94%A8%E4%BE%8B/example1-%E7%AE%80%E5%8D%95%E8%BF%90%E7%AE%97/lex.l

Note the precense of `%{`.

larsbrinkhoff:
Which language or testing framework use `.l` and `.p`?  See for example this directory:
https://github.com/doespresso/leroy/tree/eecbeee944ebc09fde4c353a442de3a7ed892bf7/node_modules/grunt-csso/node_modules/csso/test/data/test_simpleselector

pchaigno:
> @pchaigno Those two searches include many files which are Lex files.
E.g. the first C# result. Note the precense of %{.

Arf! Good to know.

hscells:
What extension should I use if I wanted linguist to pick them up?

larsbrinkhoff:
@hscells `.lisp`is by far the most common.  `.lsp` and `.cl` are also supported by GitHub if you need something shorter.

hscells:
> larsbrinkhoff

Thank you :)

larsbrinkhoff:
May the sexp be with you!

PR #2379, opened 2015-05-03
Add EusLisp

wkentaro:
GitHub search result: https://github.com/search?utf8=%E2%9C%93&q=extension%3Al+eus+euslisp+roseus&type=Code&ref=searchresults
EusLisp GitHub Repo: https://github.com/euslisp/EusLisp

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -896,6 +896,16 @@ Erlang:
   interpreters:
   - escript
 
+EusLisp:
+  type: programming
+  color: "#08298A"
+  extensions:
+  - .l
+  ace_mode: lisp
+  tm_scope: none
+  interpreters:
+  - eus
+
 F#:
   type: programming
   color: "#b845fc"
--- a/samples/EusLisp/hello.l
+++ b/samples/EusLisp/hello.l
@@ -1,0 +1,3 @@
+#!/usr/bin/env eus
+
+(print "Hello, EusLisp!")


phase:
Your color is conflicting with Harbour, and you might want to check the [other errors](https://travis-ci.org/github/linguist/builds/61049406).

larsbrinkhoff:
Also, most uses of `.l` may be Lex.

wkentaro:
Hmm.. I'd like make `.l` files added syntax highlights as lisp filetype.
I also read #2060, and finally is that the only solution changing the extension to `.lisp` or `.lsp`?

larsbrinkhoff:
Hello @wkentaro,

First, I'm not from GitHub, so I can't make any decisions on this.  However, this would be my advice:

Your search only results 74 files (currently).  If we do this:
https://github.com/search?q=extension%3Al+defun&type=Code
we find more than 6000 `.l` files which are probably Lisp.  That number if much more compelling for GitHub to make a change.

However, there are many other `.l` files which we need to consider.  Other users might complain if they were to display as `Lisp`.

Another option beside changing the file names may be to add a `.gitattributes` to your repository with this line:

    *.l linguist-language=Lisp

wkentaro:
Thanks for your information, and these're very helpful for me.
Considering other `.l` files which is not `Lisp`, I think I should use `.gitattributes`.
I tried using it and in stats of number of codes it works:
image

but there is no syntax highlight.
Is that possible to add syntax highlights using `.gitattributes`?
image

larsbrinkhoff:
> Is that possible to add syntax highlights using `.gitattributes`?

No, it's not.  You may use a "modeline" for that: https://github.com/github/linguist#using-emacs-and-vim-modelines

wkentaro:
It works, and I think that's the best way for my request.
thanks!

larsbrinkhoff:
You're welcome!

Also, you only need one of either `vim:...` or `-*- mode:...`.

PR #2386, opened 2015-05-06
Vim and Emacs modelines are two alternatives

larsbrinkhoff:
I saw one user included both Vim and Emacs modelines in a file.  Of course, there's nothing wrong with that, but it may not have been the intention to support both editors.

This change should make the documentation more clear that there are two distinct options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,16 +53,18 @@ project-docs/* linguist-documentation
 docs/formatter.rb linguist-documentation=false
 ```
 
-### Using Emacs and Vim modelines
+### Using Emacs or Vim modelines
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,12 +57,14 @@ docs/formatter.rb linguist-documentation=false
 
 Alternatively, you can use Vim and Emacs style modelines to set the language for a single file. Modelines can be placed anywhere within a file and are respected when determining how to syntax-highlight a file on GitHub.com
```
pchaigno: Maybe changing `Vim and Emacs style modelines` to `Vim or Emacs style modelines` would make it more clear?

larsbrinkhoff: Good idea!

arfon:
:+1:

PR #2387, opened 2015-05-06
Change PHP group to HTML for TWIG

aivus:
`Twig` should be in `HTML` group like `jade`, `slim` and other template languages


larsbrinkhoff:
cc @stof

stof:
This seems sensible to me

arfon:
:+1:

PR #2391, opened 2015-05-12
Add support for SVG

stanhu:
Add a distinction between binary and text image files


arfon:
Hi @stanhu. Thanks for opening this Pull Request. Could you give a little bit more background for this change? Are you using Linguist in a project and missing this functionality?

stanhu:
Yes, we have SVG images that aren't being rendered. For example:

https://github.com/svg/svgo/blob/master/examples/test.svg

Somebody pointed out that SVG files have a potential security risks and they may need to be sanitized, so I'm not sure the best way to proceed. What do you think?

arfon:
OK thanks for the clarification. SVGs already render on GitHub - take a look at the docs [here](https://help.github.com/articles/rendering-and-diffing-images/). 

The SVG you're linking to isn't rendering as it's not a valid SVG:

```
<svg version="1.1" width="10" height="20">
    test
</svg>
```

PR #2392, opened 2015-05-12
Add Apache thrift support to generated? check

vighnesh1987:
Apache Thrift compiler generated files are currently ignored in the `generated?` check leading to massive PRs which are hard to parse. This commit fixes the same.


zfogg:
+1

pchaigno:
Could this be limited to only files with a certain extension like we did in the other methods?

vighnesh1987:
@pchaigno Good idea, added

lib/linguist/generated.rb
```diff
@@ -248,6 +249,16 @@ def generated_protocol_buffer?
...
     end
+    
+    # Internal: Is the blob generated by Apache Thrift compiler?
+    #
+    # Returns true or false
+    def generated_apache_thrift?
+      return false unless ['.rb', '.py', '.go', '.js', '.m', '.java', '.h', '.cc', '.cpp'].include?(extname)
+      return false unless lines.count > 1
+
+      return lines[0].include?("Autogenerated by Thrift Compiler") || lines[1].include?("Autogenerated by Thrift Compiler")
```
pchaigno: @arfon @bkeepers Should we change this to `lines[0..1].join("").include?("Autogenerated by Thrift Compiler")`?

arfon: Not sure which would be faster. I find the the current implementation more readable.

pchaigno: Yep, let's keep the current one (I was kind of hoping for a better way to write it in Ruby...).

arfon:
:+1: thanks for this @vighnesh1987. This should be live on GitHub with the next release of Linguist.

vighnesh1987:
Awesome!

PR #2394, opened 2015-05-12
Improving Vim modeline regex

### LINKED ISSUE #2389: vim modeline does not work?
In my repo, vim modeline does not work to highlight files.

bashrc with vim modeline:
https://github.com/wkentaro/dotfiles/blob/vim-modeline/_bashrc

bashrc with emacs modeline:
https://github.com/wkentaro/dotfiles/blob/emacs-modeline/_bashrc

arfon:
Fixes #2389

### DIFF
--- a/lib/linguist/strategy/modeline.rb
+++ b/lib/linguist/strategy/modeline.rb
@@ -2,7 +2,7 @@ module Linguist
   module Strategy
     class Modeline
       EmacsModeline = /-\*-\s*(?:(?!mode)[\w-]+\s*:\s*(?:[\w+-]+)\s*;?\s*)*(?:mode\s*:)?\s*([\w+-]+)\s*(?:;\s*(?!mode)[\w-]+\s*:\s*[\w+-]+\s*)*;?\s*-\*-/i
-      VimModeline = /\s*vim:\s*set\s*(?:ft|filetype)=(\w+):\s*/i
+      VimModeline = /vim:\s*set\s*(?:ft|filetype)=(\w+):/i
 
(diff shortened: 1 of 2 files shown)


arfon:
/ cc @pchaigno

lib/linguist/strategy/modeline.rb
```diff
@@ -2,7 +2,7 @@ module Linguist
   module Strategy
     class Modeline
       EmacsModeline = /-\*-\s*(?:(?!mode)[\w-]+\s*:\s*(?:[\w+-]+)\s*;?\s*)*(?:mode\s*:)?\s*([\w+-]+)\s*(?:;\s*(?!mode)[\w-]+\s*:\s*[\w+-]+\s*)*;?\s*-\*-/i
-      VimModeline = /\/\*\s*vim:\s*set\s*(?:ft|filetype)=(\w+):\s*\*\//i
+      VimModeline = /\s*vim:\s*set\s*(?:ft|filetype)=(\w+):\s*/i
```
pchaigno: Do we still need the two `\s*`?

arfon: Good point. No, I don't think we do.

pchaigno:
Looks good to me :+1: 

On the same subject, do we check the whole file for modelines? Could we limit it to the first 20-30 lines (in case there's a header in comments)?

arfon:
> On the same subject, do we check the whole file for modelines? Could we limit it to the first 20-30 lines (in case there's a header in comments)?

We check the whole file because there's a subset of people/editors? that add these to the bottom of files :-\

PR #2405, opened 2015-05-19
Fix Handlebars grammar

aroben:
The Handlebars repository contains the following grammar files:

* Handlebars.tmLanguage
* Handlebars.JSON-tmLanguage
* grammars/handlebars.json (symlink to Handlebars.JSON-tmLanguage)

We were processing the .tmLanguage and the .json file and complaining about a duplicate scope. This triggered a nil-handling bug (which is now fixed), but we also shouldn't complain in this case at all. Having the same grammar in multiple formats is OK.

/cc @arfon https://github.com/github/linguist/commit/6062d3b25cbe99b8660724250436eeb249483119#commitcomment-11249896

### DIFF
--- a/script/convert-grammars
+++ b/script/convert-grammars
@@ -195,7 +195,7 @@ def load_grammars(tmp_dir, source, all_scopes)
     end
     all_scopes[scope] = p.url
     grammar
-  end
+  end.compact
 end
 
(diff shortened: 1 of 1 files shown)


arfon:
:zap: thanks @aroben

PR #2419, opened 2015-05-23
Treat Carthage path as vendored

evgenyneu:
Hi, I have Carthage dependency in [my project](https://github.com/evgenyneu/moa). My project is written in Swift, but the vendored dependency is Obj-C. That results in Github treating my project as Obj-C.

Carthage is a framework manager for iOS/Swift: https://github.com/Carthage/Carthage

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -163,6 +163,9 @@
 
 ## Obj-C ##
 
+# Carthage
+- ^Carthage/
+
 # Cocoapods
 - ^Pods/
 
--- a/test/test_blob.rb
+++ b/test/test_blob.rb
@@ -422,6 +422,9 @@ class TestBlob < Minitest::Test
     # Normalize
     assert sample_blob("some/asset/path/normalize.css").vendored?
 
+    # Carthage
+    assert sample_blob('Carthage/blah').vendored?
+
     # Cocoapods
     assert sample_blob('Pods/blah').vendored?
 
@@ -473,7 +476,7 @@ class TestBlob < Minitest::Test
     # Sphinx docs
     assert sample_blob("docs/_build/asset.doc").vendored?
     assert sample_blob("docs/theme/file.css").vendored?
-    
+
     # Vagrant
     assert sample_blob("puphpet/file.pp").vendored?
   end


pchaigno:
There are only 2 repositories with a Carthage dependency on GitHub: https://github.com/search?p=1&q=Olivier+Halligon+path%3Acarthage&ref=searchresults&type=Code&utf8=%E2%9C%93

 @arfon Should there be a minimum requirement on in-the-wild usage for vendored paths/files?

evgenyneu:
@pchaigno thanks for looking into it. [Your search term](https://github.com/search?p=1&q=Olivier+Halligon+path%3Acarthage&ref=searchresults&type=Code&utf8=✓) includes 'Olivier Halligon' (@AliSoftware) which shows only Carthage frameworks from this single author. It showed the usage of the [OHHTTPStubs](https://github.com/AliSoftware/OHHTTPStubs). If we search for the `import` keyword instead we can see more hits:
https://github.com/search?utf8=✓&q=import+path%3Acarthage&type=Code&ref=searchresults

Cartage dependencies are created in the `Carthage` folder located in the project root. I think they represent 3rd-party/vendored code and should not influence language statistics of the Github repository that includes them.

pchaigno:
Oh! my bad, I though he was the author of the framework manager :/
I guess it's okay then :-)

AliSoftware:
Nope, I'm a contributor of another dependency manager (CocoaPods) that exists since longer than Carthage, but I've nothing to do with Carthage (I don't even use it myself, I just made my `OHHTTPStubs` lib compatible with it)

arfon:
[This search](https://github.com/search?p=1&q=path%3Acarthage+NOT+asdhajsabfkhvasf&ref=searchresults&type=Code&utf8=%E2%9C%93) suggests there's a bunch of these files.

@jspahrsummers - this change makes sense to me. Is there anything we're missing here?

jspahrsummers:
Seems reasonable to me. :zap:

arfon:
:+1: thanks for the review @jspahrsummers :smile:

@evgenyneu - this change will be live in the next release of Linguist (likely another week or so away)

PR #2422, opened 2015-05-27
Add Smali language

CalebFenton:
Proof that this is used frequently:
https://github.com/search?utf8=%E2%9C%93&q=extension%3Asmali+NOT+nothack&type=Code&ref=searchresults

The contribute guidelines didn't mention how I could do a full end-to-end test of this locally. But, I know the tmLanguage file works because I use it every day.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -641,3 +641,6 @@
 [submodule "vendor/grammars/Handlebars"]
 	path = vendor/grammars/Handlebars
 	url = https://github.com/daaain/Handlebars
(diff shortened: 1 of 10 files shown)


CalebFenton:
It passed tests locally, but for some reason it's including an older version of the submodule without the license. I'll fix this and do it again.

pchaigno:
Nice :+1:

arfon:
@CalebFenton thanks for this. Looks great :+1: 

This will be live with the next release of Linguist (likely in the next couple of weeks)

PR #2429, opened 2015-05-29
Make filename: Kbuild

larsbrinkhoff:
`Kbuild` is a common filename for makefiles in Linux:
https://github.com/search?q=filename%3Akbuild

The syntax is, by convention, a small subset of the *GNU Make* language.


arfon:
:+1: thanks @larsbrinkhoff

PR #2441, opened 2015-06-06
Associate heuristic rules with file extensions

### LINKED ISSUE #2042: Errors when searching for the correct heuristic rule to use

### LINKED ISSUE #2149: Some .pm categorised as Perl6, some as Perl
Hey,

We have a collection of `.pm` files, all of which are Perl 5; however about 10% are being classed as Perl 6 incorrectly:

* https://github.com/duckduckgo/zeroclickinfo-goodies/search?l=perl6
* https://github.com/duckduckgo/zeroclickinfo-goodies/search?l=perl

I've had a quick look and I can't see what the particular reason is. A specific example is:
Erroneously Perl 6:
* https://github.com/duckduckgo/zeroclickinfo-goodies/blob/f352c6c9ae215c68607c38acaa9b89402fe736a3/lib/DDG/Goodie/DateMath.pm
* https://github.com/duckduckgo/zeroclickinfo-goodies/blob/f352c6c9ae215c68607c38acaa9b89402fe736a3/lib/DDG/Goodie/PrivateNetwork.pm

Correctly Perl 5:
* https://github.com/duckduckgo/zeroclickinfo-goodies/blob/f352c6c9ae215c68607c38acaa9b89402fe736a3/lib/DDG/Goodie/BashPrimaryExpressions.pm
* https://github.com/duckduckgo/zeroclickinfo-goodies/blob/f352c6c9ae215c68607c38acaa9b89402fe736a3/lib/DDG/Goodie/TimezoneConverter.pm

cc @jagtalon @mrchrisw

pchaigno:
This pull request associates one or several file extensions to each heuristic rule instead of languages as discussed in #2359.
Fixed #2042.

be122ca also improves the heuristic rule for Perl6 and fixes #2149.


lib/linguist/heuristics.rb
```diff
@@ -13,11 +13,14 @@ class Heuristics
...
     # Returns an Array of languages, or empty if none matched or were inconclusive.
-    def self.call(blob, languages)
+    def self.call(blob, candidates)
       data = blob.data
 
       @heuristics.each do |heuristic|
-        return Array(heuristic.call(data)) if heuristic.matches?(languages)
+        if heuristic.matches?(blob.name)
+          languages = Array(heuristic.call(data))
+          return languages if languages.any? || languages.all? { |l| candidates.include?(l) }
```
pchaigno: `languages.any? || languages.all? { |l| candidates.include?(l) }` should always be true.
Maybe we should throw a warning when it's not...?

Languagelearner1111:
I probably did the pull request by accident since I don't actually know
what a pull request is. Coding noon here.

larsbrinkhoff:
I think this is an improvement.  Good job!

bkeepers:
:+1: I like this approach.

@arfon?

yyx990803:
+1

Most importantly, this allows excluding non-listed languages from a given extension.

arfon:
:+1: thanks @pchaigno

PR #2445, opened 2015-06-09
Switch to using atom-fsharp language definition

rneatherway:
The previous definition used at fsharpbinding is no longer maintained.
That repository has been split into separate parts for each editor. Atom
now has the most complete compatible grammar.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -133,9 +133,6 @@
 [submodule "vendor/grammars/fancy-tmbundle"]
 	path = vendor/grammars/fancy-tmbundle
 	url = https://github.com/fancy-lang/fancy-tmbundle
-[submodule "vendor/grammars/fsharpbinding"]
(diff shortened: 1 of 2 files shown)


pchaigno:
How come the Atom grammar is more complete than the other ones? (just curious)

rneatherway:
The Sublime text project is updating to use the new syntax, which I've been told is not textmate compatible. The other editors that were in that repo -- Vim, Emacs and Monodevelop -- all use a very different way of specifying syntax highlighting. The Atom grammar has been updated a bit recently, and will be supported going forward, so it is the best one to track now.

pchaigno:
@rneatherway Okay, thanks for the explanation!

arfon:
Thanks @rneatherway. This should be live later this week when we cut a new release of Linguist.

PR #2446, opened 2015-06-10
Added Alex Haskell (a la c2hs Haskell)

wrengr:


pchaigno:
Could you link to a GitHub search showing as much Alex Haskell files as possible? (See [the requirements to add a new language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language))

lib/linguist/languages.yml
```diff
@@ -44,6 +44,16 @@ AGS Script:
   tm_scope: source.c++
   ace_mode: c_cpp
 
+Alex Haskell:
+  type: programming
+  group: Haskell
```
pchaigno: If you make Akex Haskell part of the Haskell group it will be displayed in the statistics and in search results as being Haskell... Is that what you want?

wrengr:
[alex language:haskell extension:x](https://github.com/search?utf8=✓&q=alex+language%3Ahaskell+extension%3Ax&type=Code&ref=searchresults)

lib/linguist/languages.yml
wrengr: Yes, [Alex](https://www.haskell.org/alex/) is a Haskell preprocessor (just like c2hs). As a preprocessor language it's similar to lex/flex, except that it uses/generates Haskell rather than C/C++.

pchaigno:
What if you remove `language:Haskell`? Are [these](https://github.com/search?utf8=%E2%9C%93&q=alex+extension%3Ax&type=Code&ref=searchresults) still all Alex Haskell files?

lib/linguist/languages.yml
pchaigno: But, if Alex Haskell files are shown as being Haskell and have the same syntax highlighter, it won't change anything from the current behaviour...

wrengr:
Yes, those are all still Alex Haskell (on the first dozen pages at least; didn't look through all of them)

wrengr:
I didn't see the requirements page about adding new languages, I'll work on adding the grammar

lib/linguist/languages.yml
wrengr: Currently, Alex Haskell files are showing up as Logos in the majority of these repositories
screen shot 2015-06-10 at 3 46 55 am

pchaigno:
> Yes, those are all still Alex Haskell (on the first dozen pages at least; didn't look through all of them)

Files in subsequent result pages tend to be very similar. Try to look at the last page or the middle ones. If they only contain Alex Haskell files we can probably assume that all files in the results are Alex Haskell (it's only an approximation of the number of Alex Haskell files on GitHub after all).
<br>

> I'll work on adding the grammar

Great!

lib/linguist/languages.yml
pchaigno: Adding the `.x` file extension to the Haskell entry in `languages.yml` would be sufficient to solve that :)

wrengr: If that's all it takes, great :) I just modeled it on C2hs Haskell since I saw that preprocessor was broken out in the `languages.yml`. I'd imagine that the heuristics for identifying Haskell would identify Alex Haskell about as well as C2hs Haskell (that is, often extremely well but perhaps occasionally things go off the rails)

pchaigno: We use a Bayesian classifier (and heuristics only in some special cases). So if you open a pull request to add `.x` to Haskell with a sample files (to train the classifier) it should be okay :)

If there are too many conflicts with Logos, we'll see if we can define a heuristic.

arfon:
>  I'll work on adding the grammar

Any news on this @wrengr?

lib/linguist/languages.yml
james69miller: High

wrengr:
I haven't had a chance to work on it. After fixing my own repo via gitattributes, it hasn't been a high priority for me

arfon:
> I haven't had a chance to work on it. After fixing my own repo via gitattributes, it hasn't been a high priority for me

OK, do you think we can close this for now then?

wrengr:
Sure thing. Feel free to close

PR #2447, opened 2015-06-10
Split on comma in language name if no match is found

Ryman:
Rust allows some trailing tokens in the language name to control how rustdoc tests work, for example you may have a codeblock which need to be compiled to ensure it's valid syntax but you don't want it to run as a test as it would block indefinitely. 

Current behaviour on GitHub is that if we put ```rust,no_run then the codeblock gets no highlighting.

This PR also updates the codesample for rust, as the langauge has changed _significantly_ since 2013! I can split into a separate PR if wanted.

cc @brson - do you have any better suggestion for samples? I felt hashmap is a decent length and seems to exercise a decent amount of syntax.

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -150,7 +150,7 @@ module Linguist
     #
     # Returns the Language or nil if none was found.
     def self.find_by_name(name)
-      name && @name_index[name.downcase]
+      name && (@name_index[name.downcase] || @name_index[name.split(',').first.downcase])
     end
 
     # Public: Look up Language by one of its aliases.
@@ -164,7 +164,7 @@ module Linguist
     #
     # Returns the Language or nil if none was found.
     def self.find_by_alias(name)
-      name && @alias_index[name.downcase]
+      name && (@alias_index[name.downcase] || @alias_index[name.split(',').first.downcase])
     end
 
     # Public: Look up Languages by filename.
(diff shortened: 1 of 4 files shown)


pchaigno:
> This PR also updates the codesample for rust, as the langauge has changed significantly since 2013! I can split into a separate PR if wanted.

Are the keywords themselves very different? Should we keep the older Rust file to train the classifier?

Ryman:
> Are the keywords themselves very different? Should we keep the older Rust file to train the classifier?

Glancing at the old file:
- `pure` is no longer in use
- `uint`/`int` is now replaced with `usize`/`isize`
- `priv` is no longer in use
- `@` and `~` sigils are no longer in use
- `fail!` is now `panic!` (but that's not really a keyword)
- `do` is no longer in use

That doesn't even account for new keywords :scream_cat:. Is there a concern for highlighting pre-1.0 rust?

pchaigno:
> Is there a concern for highlighting pre-1.0 rust?

Samples are mainly used to train the classifier to then, recognize languages on GitHub. I guess many users still have pre-1.0 Rust files. So, there is a concern for those files :)
That's why I think that we should maybe keep the old Rust sample. We want the Bayesian classifier remains capable of recognizing pre-1.0 Rust files.

Ryman:
@pchaigno Fair enough, I've re-added the old sample.

brson:
@Ryman the HashMap sample looks good to me.

Ryman:
@pchaigno Any update on this? Would you like anything else adjusted?

arfon:
:+1: thanks @Ryman. This will be live in the next release of Linguist (next week sometime)

PR #2425, opened 2015-05-28
Fix Rust heuristic.

larsbrinkhoff:
The heuristic was meant to use a regexp match.

The Bayesian classifier must have picked up the slack.

Fixes #2424

Issue #2424, opened 2015-05-28
Rust misclassified as RenderScript


thelink2012:
[File in question](https://github.com/dcampbell24/iup-rust/blob/eb5724fcfd51353caec22ca4711a86257f65ea63/iup-sys/src/lib.rs).

larsbrinkhoff:
Sorry, my mistake.  Fixed in #2425.

thelink2012:
Off curiosity how much time it takes for those changes to take effect?

arfon:
@thelink2012 - we need there to be a new version of Linguist deployed first (which will probably be in the next couple of weeks). Also, reclassification of files only happens once updates have happened to the repository and so you'll need to push some changes to your repo once Linguist has been updated.

PR #2448, opened 2015-06-11
Tighten RenderScript/Rust disambiguation

chris-morgan:
RenderScript files are required to have `#pragma version(1)`, and only someone deliberately trying to mislead a filetype classifier will put it in Rust code at the start of a line, so that really is the only check needed, and defaulting to Rust is then sound.

This follows on from #2424; #2425 fixed the immediate problem, but this provides a better classification; it’s entirely possible to have a Rust file with only constants defined, for example, which the previous check would fail to catch.

These heuristics are [the same as ohcount’s][1], which were [discussed at some length][2].

[1]: https://github.com/blackducksw/ohcount/blob/d54514399013f2f9566baf40edd0e801b27fe17b/src/detector.c#L870..L897
[2]: https://github.com/blackducksw/ohcount/pull/30

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -283,10 +283,10 @@ module Linguist
     end
 
     disambiguate "Rust", "RenderScript" do |data|
-      if /^(use |fn |mod |pub |macro_rules|impl|#!?\[)/.match(data)
-        Language["Rust"]
-      elsif /#include|#pragma\s+(rs|version)|__attribute__/.match(data)
+      if /^#pragma\s+version/.match(data)
         Language["RenderScript"]
+      else
+        Language["Rust"]
       end
     end
 


larsbrinkhoff:
Thank you for this.  I wrote the current heuristic based on a set of 1000 `.rs` samples.  I don't remember the exact details anymore, but it may be that a more complex heuristic than yours is needed.

I will check your heuristic against the samples and report the results.

larsbrinkhoff:
Also, your implementation defaults to Rust and disables the bayesian classifier.  Other heuristics do this too, but I believe it's encouraged to avoid doing that.

chris-morgan:
@larsbrinkhoff As I noted (and as https://developer.android.com/guide/topics/renderscript/compute.html confirms), RenderScript requires the `#pragma version(1)`, and so I believe that to be a sufficient filter. As I understand it (though I could easily be wrong in this), most filetype detection is done purely on extension, and most extensions have an unambiguous registered filetype, thus bypassing the bayesian classifier; as noted in blackducksw/ohcount#30, RenderScript and Rust are the only notable things using the .rs extension, and RenderScript has a clear identifying signal, so anything .rs which does not contain that signal can reasonably be marked Rust. Am I correct in my understanding of things and do my conclusions seem reasonable to you?

larsbrinkhoff:
First of all, I should explain that I'm not part of GitHub staff, and have no final say in anything that goes into linguist.

I still haven't made a full run against the samples, but some quick remarks for now:

- I checked my sample files, and it seems a substantial amount of RenderScript files on GitHub do not have a `#pragma version`.

- I don't see that the link you provided documents a *requirement* for RenderScript files to include a `#pragma version`.  It says "A script *can* contain: A pragma declaration (#pragma version(1))" (emphasis mine).  Is that what you are referring to?

- Even if it were a requirement, it could be imported into a file through `#include` and thus not appear in the file itself.

- I agree it may be reasonable to default to Rust as you did.

larsbrinkhoff:
By the way, do you know if Rust files can contain `#include`, or do people use a C preprocessor with Rust?

retep998:
Rust has attributes in the form of `#[attribute]` or `#![attribute], and it also uses # for `r#"raw strings"#`, but Rust definitely does not use the C preprocessor with things like `#include` or `#pragma`.

chris-morgan:
@larsbrinkhoff Oops, I didn’t reread it carefully enough—I *think* the wording indicated that `#pragma version(1)` was required in the past, but it definitely doesn’t now. I guess I need to add back more checks than just the `#pragma version`, then. You know, that original regex doesn’t look so bad now…

@retep998 Macro bodies could also have `#` followed by an identifier, and the form could also be used in the future (and it wouldn’t surprise me in the least if it is used in some form in the future). Neither of these should particularly concern us *now*, though.

larsbrinkhoff:
What language is this?

https://github.com/jmaih/RISE_toolbox/tree/master/examples/MarkovSwitching/LiuWaggonerZha2009/Tutorial2/Models

larsbrinkhoff:
I guess this is moot now? but for the record:

| Language | Before | After | Actual
| --- | --- | --- | ---
| RenderScript | 741 | 464 | 707
| Rust | 254 | 532 | 225
| javac output? | | | 45
| unknown | 1 | 1 | 12
| RedScript | | | 7
| ResPack | 4 | 4 | 4

larsbrinkhoff:
If anyone knows what these are, let me know!

https://github.com/search?q=extension%3Ars+NOT+slaribartfaxt+path%3Abuild

larsbrinkhoff:
@chris-morgan If you want to examine the samples yourself, I can send you a zip file.

nightpool:
@larsbrinkhoff those seem to be some netbeans specific build file, although I may be wrong.

larsbrinkhoff:
@nightpool Thanks, that's my conclusion too.

arfon:
Just looping back on this PR - what's the current status?

larsbrinkhoff:
I believe the latest word from @chris-morgan is:
> You know, that original regex doesn’t look so bad now…

And my test run indicates that **Before** is a more accurate classification than **After**.

arfon:
Ha! Ok. So are we good to close this then?

larsbrinkhoff:
Let's hope Chris can check in and confirm that.

chris-morgan:
Yeah, let’s just close this.

arfon:
:+1: thanks for confirming @chris-morgan

PR #2457, opened 2015-06-18
Adding some more vendered expressions

drewcrawford:
* `Vendor/`, the capitalized path
* Crashlytics, a common iOS framework
* Fabric, a common iOS framework


arfon:
:+1: thanks @drewcrawford. This will be live in the next release of Linguist (likely later this week).

PR #2463, opened 2015-06-21
Instructions to add a new extension to a language

pchaigno:
This pull request adds a new entry in `CONTRIBUTING.md` on how to add a new extension to a language.

It's mostly the same as `Adding a language` but, in many cases, new contributors don't seem to feel concerned by the existing instructions/requirements when they just add an extension. It will also give us something we can directly point to when necessary.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ In addition, if this extension is already listed in [`languages.yml`][languages]
 
 ## Adding a language
 
-We try only to add languages once they have some usage on GitHub. In most cases we prefer that languages be in use in hundreds of repositories before supporting them in Linguist.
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
> We try only to add new extensions once they have some usage on GitHub. In most cases we prefer that extensions be in use in hundreds of repositories before supporting them in Linguist.

There seem to be ample precedent for adding a new extension to popular languages, even though the new extension aren't used in hundreds of repositories.

pchaigno:
I thought that *In most cases we prefer* was sufficient to describe this. Do you think we should change this is both *Adding an extension to a language* and *Adding a language*?

larsbrinkhoff:
I think a case could be made for having less strict requirements for individual extensions, than for new languages.

I suggest: "We prefer that extensions are used in hundreds of *files* before supporting them in Linguist."  This also has the benefit of being easy to verify with a code search.

arfon:
:+1: thanks for this @pchaigno.

PR #2466, opened 2015-06-22
Added .crunit suffix to Python language

NiklasRosenstein:
see https://github.com/NiklasRosenstein/creator

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2640,6 +2640,7 @@ Python:
   extensions:
   - .py
   - .cgi
+  - .crunit
   - .fcgi
   - .gyp
   - .lmi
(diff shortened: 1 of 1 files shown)


pchaigno:
It looks like you're the only user of this extension on GitHub: https://github.com/search?utf8=%E2%9C%93&q=extension%3Acrunit+NOT+sjfhdkfdgfndjg&type=Code&ref=searchresults :/
We try to only add new extensions/languages once they have some usage on GitHub. You can probably fix this using [Linguist Overrides](https://github.com/github/linguist#overrides).

arfon:
> We try to only add new extensions/languages once they have some usage on GitHub.

:+1: - @NiklasRosenstein please take a look at the Linguist Overrides as @pchaigno suggests.

PR #2471, opened 2015-06-24
Update generate_apache_thrift? check to include html/php files

vighnesh1987:


arfon:
Thanks @vighnesh1987. Could you link to some example files that this addition would be a fix for?

arfon:
Closing this as stale.

PR #2474, opened 2015-06-26
Semantic language snippets

rafaelstz:


pchaigno:
Unfortunately, this will only highlight Bash keywords, which doesn't make sense here:

```sh
$ bundle exec linguist --breakdown

100.00% Ruby

Ruby:
Gemfile
Rakefile
bin/linguist
github-linguist.gemspec
lib/linguist.rb
…
```

rafaelstz:
Rigth! :disappointed:

PR #2476, opened 2015-06-30
Add .tsx as a TypeScript extension

RyanCavanaugh:
TypeScript now supports JSX in .tsx files.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3335,6 +3335,7 @@ TypeScript:
   - ts
   extensions:
   - .ts
+  - .tsx
   tm_scope: source.ts
   ace_mode: typescript
 
--- a/samples/TypeScript/world.tsx
+++ b/samples/TypeScript/world.tsx
@@ -1,0 +1,2 @@
+var x = <div></div>;
+console.log("Hello, World!");


arfon:
@RyanCavanaugh - thanks for this. It looks like there's a fair amount of usage of this extension for [XML files](https://github.com/search?utf8=%E2%9C%93&q=NOT+aksjdbsdkfbej+extension%3Atsx&type=Code&ref=searchresults) - do you know of other uses for this extension?

RyanCavanaugh:
Wow, that's news to me. I'm not aware of any other uses myself, but I wasn't aware of .tsx for a 'tileset' XML file either, so I'm not a good person to ask.

It's worth clarifying that this is a totally new file extension for TypeScript so we don't expect to see any uses of it in the wild yet. Would it be appropriate to add a disambiguation rule now, or wait for further adoption?

arfon:
> or wait for further adoption?

If it's OK with you then I'd rather wait until there's been some additional usage of this extension on GitHub. In the interim, if you'd like to enable syntax highlighting then the [Emacs/Vim overrides](https://github.com/github/linguist#using-emacs-or-vim-modelines) should accomplish this.

RyanCavanaugh:
Sounds good, thanks for taking a look.

markogresak:
What is the status on this? I'm using this ext with my new TypeScript + React project and it's not detecting the language, nor is it enabling syntax highlighting of any kind.

PR #2478, opened 2015-07-01
Append Test::More into strict.t

mattn:
Because, most of tap scripts on Perl5 are detected as Perl6.

### DIFF
--- a/samples/Perl/strict.t
+++ b/samples/Perl/strict.t
@@ -1,4 +1,5 @@
 use Test::Base;
+use Test::More;
 
 __DATA__
 === Strict Test
--- a/samples/Perl6/test.p6
+++ b/samples/Perl6/test.p6
@@ -73,26 +73,6 @@ Here's some POD!  Wooo
 
 say('this is not!');
 
-=table
-    Of role things
-
-say('not in your table');
-#= A single line declarator "block" (with a keyword like role)
-#| Another single line declarator "block" (with a keyword like role)
-#={
-    A declarator block (with a keyword like role)
-  }
-#|{
-    Another declarator block (with a keyword like role)
-  }
-#= { A single line declarator "block" with a brace (with a keyword like role)
-#=«
-    More declarator blocks! (with a keyword like role)
-  »
-#|«
-    More declarator blocks! (with a keyword like role)
-  »
-
 say 'Moar code!';
 
 my $don't = 16;


pchaigno:
Thanks for the pull request!

Have you tested the impact of your change on other Perl files?

mattn:
Yes, code in above detected as Perl5 with this change. BTW, the word `More` is counted as heuristic level 2 for Perl6. But it's not code. it's in comments. So it's not fair.

https://github.com/github/linguist/blob/master/samples/Perl6/test.p6#L90-L93

pchaigno:
> Yes, code in above detected as Perl5 with this change.

In that case, could you explain a little more whats happening here? I don't know what is a tap script in Perl. Also, is `Test::More` specific to one of the versions or is it common to both?

> BTW, the word More is counted as heuristic level 2 for Perl6. But it's not code. it's in comments. So it's not fair.

So the comment starts [here](https://github.com/github/linguist/blob/master/samples/Perl6/test.p6#L76) but I can't see where it's supposed to end...?

mattn:
> In that case, could you explain a little more whats happening here? I don't know what is a tap script in Perl. Also, is Test::More specific to one of the versions or is it common to both?

Perl6 doesn't have `Test::More`. But the script using `Test::More` is handled as Perl6. Because the word `More` is counted as 2 in Perl6 in the code as I said in above. My change add `More` word into Per5 script is fixing this problem.

> So the comment starts here but I can't see where it's supposed to end...?

next /^=/. but not. so it's EOF.

pchaigno:
Thanks for the explanation :+1: 

> next /^=/. but not. so it's EOF.

Okay, then, could you also remove [those lines](https://github.com/github/linguist/blob/master/samples/Perl6/test.p6#L76-L94) in your pull request? I'm not a big fan of changing samples but it looks like we don't have a choice here since the tokenizer doesn't handle Pod comments :/

mattn:
Best way to fix this issue is handling string literals well, parse pod correctly, and ignoring all of non-syntax literals. If you think it's harder to do, I'll remove this as your said.

pchaigno:
> Best way to fix this issue is handling string literals well, parse pod correctly, and ignoring all of non-syntax literals. If you think it's harder to do, I'll remove this as your said.

Take a look at [`tokenizer.rb`](https://github.com/github/linguist/blob/master/lib/linguist/tokenizer.rb), you'll see that it's not that simple. We would have to parse all languages and not only Perl ;)
There was a project for Google Summer of Code about this (using the Sublime Text/TextMate grammars we know to parse files) but no student was selected to work on it.

For the time being, I think you should just remove this section ;)

mattn:
okay, will do. then, I should revert adding `Test::More`?

pchaigno:
> then, I should revert adding Test::More?

As you wish.

mattn:
I will remains `Test::More` because `Test::More` is used in Perl5 but not in Perl6. The word `More` should be prevailed to Perl5.

pchaigno:
@arfon This looks ready to be merged to me :-)

arfon:
:+1: thanks @mattn & @pchaigno. This should be live in the next release of Linguist (next week).

PR #2481, opened 2015-07-04
Isabelle ROOT support (updated PR)

arfon:
This replaces https://github.com/github/linguist/pull/1823. As Linguist is case-sensitive I think we should be OK here for the potential concerns raised in https://github.com/github/linguist/pull/1823#issuecomment-91523375. That said, I've added a sample `root.php` file for completeness.

/ cc @lsf37


arfon:
@lsf37 - this should be live in the next release of Linguist (this week sometime)

PR #2180, opened 2015-07-05
.inc file extension

pchaigno:
There is [more than 10 millions `.inc` files on GitHub](https://github.com/search?q=extension%3Ainc+NOT+sfhdkfhdkhfdjfudhfkd&ref=reposearch&utf8=%E2%9C%93) and it's used by many languages. Thus, to add that extension, we need to identify all languages using it (with at least hundreds of examples).

I will update this post with the list of languages as we go. I might eventually turn this issue into a pull request to add support of the extension.

Language | Number of files
------ | ------ | -----
PHP | [9,241,000](https://github.com/search?q=extension%3Ainc+php&ref=reposearch&utf8=%E2)
Pascal | [206,000](https://github.com/search?p=96&q=extension%3Ainc+end+begin&ref=searchresults&type=Code&utf8=%E2%9C%93)
Assembly | [at least 69,000](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ainc+NOT+php+mov+OR+jmp+OR+lda+OR+globl)
MySQL | [49,000](https://github.com/search?p=1&q=extension%3Ainc+select+where+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)
C++ | [8,000](https://github.com/search?p=1&q=extension%3Ainc+class+void+include+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)
HTML | [5,000](https://github.com/search?p=100&q=extension%3Ainc+body+html+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)*
SourcePawn | [2,300](https://github.com/search?p=1&q=extension%3Ainc+native+forward+stock&ref=searchresults&type=Code&utf8=%E2%9C%93)*
Clarion | [at least 165](https://github.com/search?p=1&q=extension%3Ainc+destruct+OR+construct+procedure+NOT+begin+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)*

`*` I checked these by downloading 1,000 samples from the search results and using [simple heuristics](https://gist.github.com/pchaigno/b50ec322afd3cdcdafdf) to triage them.


pchaigno:
So, I finally turned this into a pull request and here are the results:

                        | PHP        | SourcePawn | Assembly   | C++       | Pascal     | HTML       | SQL        |
------------------------|------------|------------|------------|-----------|------------|------------|------------|
[862 PHP samples](https://github.com/search?q=extension%3Ainc+php&ref=reposearch&utf8=%E2)        | **98.79%** | 0.89%      |            |           | 0.32%      |            |            |
[952 SourcePawn samples](https://github.com/search?p=1&q=extension%3Ainc+native+forward+stock&ref=searchresults&type=Code&utf8=%E2%9C%93) | 0.05%      | **98.89%** |            | 1.06%     |            |            |            |
[898 Assembly samples](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ainc+NOT+php+mov+OR+jmp+OR+lda+OR+globl)   | 0.29%      |            | **99.24%** | 0.27%     | 0.20%      |            |            |
[787 C++ samples](https://github.com/search?p=1&q=extension%3Ainc+class+void+include+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)        |            | 0.26%      |            | **9.74%** |            |            |            |
[808 Pascal samples](https://github.com/search?p=96&q=extension%3Ainc+end+begin&ref=searchresults&type=Code&utf8=%E2%9C%93)     | 0.19%      |            | 0.16%      | 1.20%     | **98.46%** |            |            |
[897 HTML samples](https://github.com/search?p=100&q=extension%3Ainc+body+html+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)       | 0.56%      | 1.30%      | 0.64%      | 1.81%     | 0.42%      | **95.27%** |            |
[938 SQL samples](https://github.com/search?p=1&q=extension%3Ainc+select+where+NOT+php&ref=searchresults&type=Code&utf8=%E2%9C%93)        | 0.26%      |            |            |           |            |            | **99.74%** |

It looks to me like heuristic rules won't be needed for `.inc` :-)

@arfon This is ready for review ;)

arfon:
Epic effort @pchaigno - this is awesome :zap:

Given the coverage here of languages (and the classifier accuracy) I think this should be good to go. Would love a :+1: from @bkeepers here too for completeness.

bkeepers:
Wow, I'm impressed the results are that accurate.  I'm :+1:.

This actually got me thinking that it would be interesting to add an extended test suite that runs against these other corpuses and ensures they match a certain threshold.

pchaigno:
> Wow, I'm impressed the results are that accurate.

Yep, I am too. I added more samples than usually so that's one way to explain it but I'm afraid we could be overfitting. Although, only C++/SourcePawn/Pascal really needed more samples. C++ vs. SourcePawn needed more samples because only a handful of keywords are different. If there's overfitting somewhere I'd expect it to be for Pascal...
<br>

> This actually got me thinking that it would be interesting to add an extended test suite that runs against these other corpuses and ensures they match a certain threshold.

Sounds like a good idea to me. That would be in a separate PR right? Should we open an issue to track progress on that idea?

bkeepers:
> Sounds like a good idea to me. That would be in a separate PR right? Should we open an issue to track progress on that idea?

:+1:

arfon:
:+1: thanks @pchaigno.

cperciva:
Please add Makefile to the set of languages which can have a .inc suffix.  For that matter, a file named "Makefile.inc" should probably be automatically classified as "Makefile" without even considering other options.

Right now github is telling me that Tarsnap/spiped's Makefile.inc is a sourcepawn file! https://github.com/Tarsnap/spiped/search?l=sourcepawn

PR #2493, opened 2015-07-11
Add .kicad_pcb extension for KiCad PCB design files

miek:
Mentioned in http://www.kicad-pcb.org/display/KICAD/File+Formats. Along with .sch (schematic) files, .kicad_pcb files make up the bulk of a KiCad project.

[~4400 search results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Akicad_pcb+NOT+nothack&type=Code&ref=searchresults)


lib/linguist/languages.yml
```diff
@@ -1637,6 +1637,7 @@ KRL:
 KiCad:
   type: programming
   extensions:
+  - .kicad_pcb
   - .sch
```
pchaigno: The first extension should be the primary extension. I think `.sch` should stay the primary one ;)
(It's not very well documented, sorry about that.)

miek: Fixed :)

arfon:
:+1: thanks @miek - this will be live in the next release of Linguist (likely the next week or so)

PR #2497, opened 2015-07-12
Detect the language of a code snippet in memory

pchaigno:
This pull request adds a new class to Linguist to allow for the analysis of code snippets in memory (i.e. without a corresponding file on disk). It expands the API so that it is possible to do this:
```ruby
require 'linguist'

blob = Linguist::MemoryBlob.new('README.asc', '==== Test\n\nTest app')
Linguist::Language.detect(blob)
```

It will make it possible to use Linguist in [github/markup](https://github.com/github/markup) (as discussed in github/markup#468) without restricting the API (See [usage of markup](https://github.com/github/markup#usage)).


pchaigno:
@bkeepers @arfon What do you think of this?

bkeepers:
I'm at OSCON this week and slammed with some other stuff, so I won't get a chance to look at this until at least next week. Sorry!

pchaigno:
> I'm at OSCON this week and slammed with some other stuff, so I won't get a chance to look at this until at least next week. Sorry!

No problem! It can wait.

lib/linguist/memory_blob.rb
```diff
@@ -0,0 +1,73 @@
+require 'linguist/blob_helper'
+
+module Linguist
+  # A MemoryBlob is a wrapper around the content of a file to make it quack
+  # like a Grit::Blob. It provides the basic interface: `name`,
+  # `data`, `path` and `size`.
+  class MemoryBlob
```
bkeepers: What about just calling this `Linguist::Blob`?

bkeepers:
This looks pretty good.

I would love to see an implementation of https://github.com/github/markup using this API before this gets merged, just to make sure this makes sense.

lib/linguist/memory_blob.rb
pchaigno: Yep, that could make sense since `Linguist::FileBlob` inherits it.
I'll change it.

pchaigno:
> I would love to see an implementation of https://github.com/github/markup using this API before this gets merged, just to make sure this makes sense.

I just opened a pull request on [github/markup](https://github.com/github/markup). It's not complete but it'll give you an idea of how `Linguist::Blob` can be used in github-markup.

pchaigno:
@bkeepers Did you get a chance to look at the implementation?

bkeepers:
@pchaigno I just checked it out and left some comments. Thanks for taking a crack at that.

arfon:
@bkeepers @pchaigno - I found myself wanting exactly this functionality in Linguist this morning. What's the status of this as far as you are both concerned?

pchaigno:
This branch is ready on my side. The [PoC in github/markup](https://github.com/github/markup/pull/537) works for me locally (I have to generate the `samples.json` file manually) but [I wasn't able to make it work with Travis CI](https://github.com/github/markup/pull/537#discussion_r38753566).

arfon:
Closing in favour of #2694

PR #2498, opened 2015-07-13
xBase: add .ch extension + sample

vszakats:
Usage in GitHub:
https://github.com/search?utf8=✓&q=extension%3Ach+define&type=Code&ref=searchresults

Referring to:
https://github.com/github/linguist/commit/6fb4e6836cd891d5b95192dbbd277e97c503cfe1#commitcomment-12127776

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3783,5 +3783,6 @@ xBase:
   color: "#403a40"
   extensions:
   - .prg
+  - .ch
   tm_scope: none
   ace_mode: text
--- a/samples/xBase/sample.ch
+++ b/samples/xBase/sample.ch
@@ -1,0 +1,30 @@
+#ifndef __HARBOUR__
+   #ifndef __XPP__
+      #ifndef __CLIP__
+         #ifndef FlagShip
+            #define __CLIPPER__
+         #endif
+      #endif
+   #endif
+#endif
+
+/* File create flags */
+#define FC_NORMAL          0  /* No file attributes are set */
+#define FC_READONLY        1
+#define FC_HIDDEN          2
+#define FC_SYSTEM          4
+
+// New-style comment
+#command SET DELETED <x:ON,OFF,&>      => Set( _SET_DELETED, <(x)> )
+#command SET DELETED (<x>)             => Set( _SET_DELETED, <x> )
+#command @ <row>, <col> SAY <exp> [PICTURE <pic>] [COLOR <clr>] => ;
+         DevPos( <row>, <col> ) ; DevOutPict( <exp>, <pic> [, <clr>] )
+         
+#command ENDIF <*x*> => endif
+
+#ifdef __CLIPPER__   
+   #xtranslate hb_MemoWrit( [<x,...>] )  => MemoWrit( <x> )
+   #xtranslate hb_dbExists( <t> )        => File( <t> )
+   #xtranslate hb_dbPack()               => __dbPack()
+   #xtranslate hb_default( @<v>, <x> )   => iif( StrTran( ValType( <v> ), "M", "C" ) == StrTran( ValType( <x> ), "M", "C" ),, <v> := <x>, )
+#endif


pchaigno:
Are [these](https://github.com/search?p=1&q=extension%3Ach+NOT+define+NOT+command&ref=searchresults&type=Code&utf8=%E2%9C%93) xBase files too?

vszakats:
No, those aren't.

Here is a list of the originally (since Clipper 5 in 1990) supported directives in `.ch` files (there are some extensions offered by certain implementations since):
http://x-hacker.org/ng/53guide/ng11fd89.html

Here's a description of the comment styles supported by the original language (`/* */` and `//` being the most popular ones):
https://sourceforge.net/p/cloc/feature-requests/67/#ad95

pchaigno:
> No, those aren't.

Okay, so we should identify those files. With the pull request as-is, they will all be detected as xBase :/
<br>

By the way, would you know a grammar for xBase? (Sublime Text package or Atom/TextMate bundle)

vszakats:
Having a closer look, I can see two files in an Android project, some files in the __attic__ in OpenBSD (which means deleted/inactive files), and most of the hits coming from the single project LZO, where it's used for non directly compilable `.c` or `.h` files. And there is one IMO true candidate for collision, project: https://github.com/mietek/charity-language, which uses `.ch` as the main extension for its language files. I'd argue this language ever had the impact of `xBase`, but anyhow, what is the procedure if there are some misuses and random files using a canonical extension name?

BTW, this extension is also used in private repositories which this public search won't display.

vszakats:
As for the xBase grammar: I haven't seen a perfect one yet, but here's working one for Atom:
https://github.com/hernad/atom-language-harbour/blob/master/grammars/harbour.cson

vszakats:
There are these two as well for Clipper and FoxPro flavors:
https://sourceforge.net/p/colorer/svn/1956/tree/branches/lazyeugene/hrc/hrc/db/clipper.hrc
https://sourceforge.net/p/colorer/svn/1956/tree/branches/lazyeugene/hrc/hrc/db/foxpro.hrc

vszakats:
In the above referenced grammar, it is applied to both `.prg` and `.ch` files inside GitHub's Atom editor:
https://github.com/hernad/atom-language-harbour/blob/master/grammars/harbour.cson
(even though this is a "Community" package, not an official one)

vszakats:
Comparing 'Charity Language' with xBase flavors:

Charity Language (since 2000): https://github.com/mietek/charity-language
Original Clipper (since 1990): https://en.wikipedia.org/wiki/Clipper_(programming_language)

xBase incarnations in active use/development as of now:
- Harbour: https://github.com/harbour/core https://sourceforge.net/projects/harbour-project/ https://en.wikipedia.org/wiki/Harbour_(software)
- Harbour fork: https://github.com/vszakats/harbour-core
- xHarbour: https://sourceforge.net/projects/xharbour/ http://www.xharbour.com
- CLIP: https://sourceforge.net/projects/x-clip/
- Flagship: https://en.wikipedia.org/wiki/Flagship_compiler http://fship.com
- Xbase++: https://en.wikipedia.org/wiki/XBase%2B%2B https://www.alaska-software.com
Clipper: 

[ Disclaimer: I'm major contributor to Harbour since 1999 ]

pchaigno:
> what is the procedure if there are some misuses and random files using a canonical extension?

We'll only add languages that use this extension and are currently present in at least a hundred repositories on GitHub (usually 600-900 files).
<br>

> As for the xBase grammar: I haven't seen a perfect one yet, but here's working one for Atom:
https://github.com/hernad/atom-language-harbour/blob/master/grammars/harbour.cson

[It's not great](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fhernad%2Fatom-language-harbour%2Fmaster%2Fgrammars%2Fharbour.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fvszakats%2Flinguist%2Fextch%2Fsamples%2FxBase%2Fsample.ch&code=) :/
Is Harbour's syntax close to xBase's?
<br>

> There are these two as well for Clipper and FoxPro flavors:
https://sourceforge.net/p/colorer/svn/1956/tree/branches/lazyeugene/hrc/hrc/db/clipper.hrc
https://sourceforge.net/p/colorer/svn/1956/tree/branches/lazyeugene/hrc/hrc/db/foxpro.hrc

I'm afraid GitHub doesn't support this format.

vszakats:
I know it's not great, that's just the best I could cite, it's certainly better than nothing though. And a good base for improvement. (till it's not used, it surely won't be improved)

Harbour project itself has 147 files. This project has 90 forks (with at least two major ones). In private projects I host a multiple number of those files (won't disclose it publicly, but GitHub Staff members can verify it).

vszakats:
xBase is an umbrella term for all languages in this family. Harbour aims to be 100% Clipper compatible, and Clipper was the most compatible with the original `dBase` language, so the answer is yes, the Harbour grammar covers the full base "xBase" syntax plus some extensions.

pchaigno:
> xBase is an umbrella term for all languages in this family. Harbour aims to be 100% Clipper compatible, and Clipper was the most compatible with the original dBase language, so the answer is yes, the Harbour grammar covers the full base "xBase" syntax plus some extensions.

Okay, thanks for the explanation!
<br>

> I know it's not great, that's just the best I could cite, it's certainly better than nothing though. And a good base for improvement. (till it's not used, it surely won't be improved)

I agree it's better than nothing.
Could you add this grammar to your pull request for both [Harbour](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1352) and xBase?
<br>

> And a good base for improvement. (till it's not used, it surely won't be improved)

You should report the issues too ;)

vszakats:
I have an open PR for this grammar, so far to cover a specific Harbour extension: https://github.com/hernad/atom-language-harbour/pull/2

Once I shift my editing to Atom, and mis-parsing starts to itch, I'll certainly make PRs. I already spotted a parsing error in misinterpreting a backslash in a string. Notice that xBase languages are rather complicated to parse due to the many layers of legacy baggage and the all-too-powerful preprocessor it features, so I'm generally satisfied if it does some basic coloring and it doesn't fail badly on "common" syntax.

I'm looking into adding the grammar as a PR. Never done that, so I need to investigate.

pchaigno:
> I'm looking into adding the grammar as a PR. Never done that, so I need to investigate.

You can find some help in [the guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language). If you get an error, don't hesitate to ask, I'll try to help :)

vszakats:
Thank you Paul for your help.
I've PR-d various extensions and fixes to the grammar cited above: https://github.com/hernad/atom-language-harbour/pull/3
Also issued a PR for including this grammar in Linguist: https://github.com/github/linguist/pull/2499

vszakats:
From Issue https://github.com/github/linguist/issues/2504#issuecomment-122485051:

> Okay, so we should identify those files. With the pull request as-is, they will all be detected as xBase :/
> Before #2498 can be merged we need to be confident we're not going to mis-classify these ~7000 files > .ch files that aren't XBase

Of these ~7000, it will be fairly complicated to do that for the thousands of duplicates of the `lzo1*.ch` files, which use similar format as `.ch` with some extra C code in them. `.ch` files (which stand for *C*lipper *H*eaders) use similar PP directive syntax as `.h` headers for C do. I think LZO misuses `.ch` extension for their `.c` files and I'm not sure it's the best choice to design around the exception.

Apart from this peculiar case, it's quite easy to identify `.ch` files using [this](https://www.debuggex.com/r/dMIeOfYAk1y0Nbd-) regexp: 
```
^\s*#\s*(if|ifdef|ifndef|endif|define|command|xcommand|translate|xtranslate|include|elif|else|pragma|error|warning|stdout|line|undef)\b
```

UPDATE: or a shortened version which should cover most cases:
```
^\s*#\s*(if|ifdef|ifndef|define|command|xcommand|translate|xtranslate|include|pragma|undef)\b
```

vszakats:
@pchaigno Can you help in adding disambiguation heuristics using this regexp to help identify xBase `.ch` files:
`^\s*#\s*(if|ifdef|ifndef|define|command|xcommand|translate|xtranslate|include|pragma|undef)\b`

[ This will return positive for those few LZO source files, but I'd argue those are wrongly named and the correct fix (if LZO authors are interested) would be to rename them because `.ch` is not an expected extension for C source code. It's also possible to extend the rules to filter out C sources using some patterns, but I'm at loss what these patterns would be. ]

vszakats:
Just noticed that `.ch` was unintentionally added back as an extension as part of my xBase grammar patch: https://github.com/github/linguist/pull/2499/files

It was nice to see `.ch` files colored, but I by no means intended to add this back as part of above patch, while having this under discussion and pending in this patch. My apologies for it :( Could you please remove this or should I PR?

arfon:
> Just noticed that .ch was unintentionally added back as an extension as part of my xBase grammar patch: https://github.com/github/linguist/pull/2499/files

Yikes, good spot. Thanks @vszakats. Rather than revert this I think we should push ahead and add the heuristic you suggest: 

> ^\s*#\s*(if|ifdef|ifndef|define|command|xcommand|translate|xtranslate|include|pragma|undef)\b

Would you mind taking a look at [`heuristics.rb`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L287-L291) and seeing if you can write one for `xBase`? 

I'm also happy to do this if you're not comfortable working with the heuristics code.

vszakats:
@arfon Thanks for jumping in and for the pushing ahead.

I've taken a quick look to the rules; Being a total newbie in this area, I'd rather trust your patch than start with it from zero.

[ I'd guess first all the other language would need to be added to `languages.yml` so that we can add a disambiguate rule between them. And next these two lines, will probably be needed at some point (not much, I know):
```ruby
      elsif /^\s#\s(if|ifdef|ifndef|define|command|xcommand|translate|xtranslate|include|pragma|undef)\b/i.match(data)
        Language["xBase"]
```
]

arfon:
Closing in favour of #2515

PR #2512, opened 2015-07-20
Pawn TextMate-compatible grammar

ziggi:
Just good TextMate-compatible grammar for Pawn.

Some changes:
- ~~Renamed to Pawn, because it is right name.~~
- Added .inc extension, because [Pawn have more than 40K result in search](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ainc+samp+OR+sa-mp+OR+pawn&type=Code&ref=searchresults) (but if you want I can remove it).


pchaigno:
> Renamed to Pawn, because it is right name.

Unfortunately, we can't change the names of languages yet :/ (see #2034 and #995).
<br>

> Added .inc extension, because Pawn have more than 40K result in search (but if you want I can remove it).

I added the `.inc` extension to SourcePawn in #2180. Was that incorrect or is it also true?
Is SourcePawn a Pawn dialect?

ziggi:
> Unfortunately, we can't change the names of languages yet :/ (see #2034 and #995).

Ok. Returned to the old name.

> I added the .inc extension to SourcePawn in #2180. Was that incorrect or is it also true?
Is SourcePawn a Pawn dialect?

Your changes in #2180 correct, but yes, SourcePawn is a Pawn dialect ([proof in README.md](https://github.com/alliedmodders/sourcepawn)), and these languages are very similar.

samples/PAWN/grandlarc.pwn
```diff
@@ -1,520 +0,0 @@
-//----------------------------------------------------------
```
pchaigno: Why did you remove this file? Is it not a Pawn file?

ziggi: > Why did you remove this file? Is it not a Pawn file?

It's Pawn file, but this code very similar to C++ and `.inc` test do not pass.

pchaigno:
> SourcePawn is a Pawn dialect (proof in README.md), and these languages are very difficult to discern.

@ziggi Does that mean that your search page (in the original post) contains both Pawn and SourcePawn files?
Should we group these two languages under Pawn? How different is the syntax?
<br>

@arfon We can't rename a language, so does that mean that we can't remove one either?

ziggi:
> @ziggi Does that mean that your search page (in the original post) contains both Pawn and SourcePawn files?

I think no, because i use "sa-mp", "samp" and "pawn" keywords.

> Should we group these two languages under Pawn? How different is the syntax?

You can read [here](https://github.com/alliedmodders/sourcepawn) about changes, but i think the syntax is almost identical.

Maybe Linguist may set language for `.inc` files by existing `.sp` or `.pwn` file in the repository?

arfon:
> @arfon We can't rename a language, so does that mean that we can't remove one either?

I think we probably could remove one in theory but in practice though I think this would leave us with a whole bunch of stale indexed files for which the GitHub UI no-longer recognizes.

@twp - would this be a reasonable assessment?

50Wliu:
Just saw this PR and I'd like to give a bit of clarification.

SourcePawn is indeed based off of Pawn, but it's started to diverge significantly from it (if I recall they don't sync with upstream Pawn changes - I can only think of one exception).  Most noticeable is the addition of "newdecl" syntax which is completely different from Pawn and is designed to move Sourcepawn into a C#/Java-like direction (complete with some object-oriented sugar).

Old:
```sp
new Float:hello;  //Identical to Pawn
```

New:
```sp
float hello;  //Identical to Java, and isn't even tokenized correctly by the existing SP grammar
```

Now the largest problem is that *both syntaxes are supported*, and compile to identical bytecode.  It's nearly impossible to distinguish Pawn from old-style Sourcepawn (since most of the differences are in advanced features that barely anyone uses), but it should be easy to distinguish it from the new syntax.  Not sure if that helps.

I'll probably create a new issue in the coming days with a more detailed breakdown of Sourcepawn and recommendations for a new grammar (hint: it's mine, though I haven't spent much time updating it).

arfon:

arfon:
Replaced by #2882

PR #2527, opened 2015-07-30
Parse HCL as Ruby, not JavaScript

radeksimko:
cc @sethvargo 

https://github.com/github/linguist/pull/1899

Before we get a proper HCL grammar, I believe it's much better to parse HCL as Ruby, so at least people don't see the annoying red background everywhere:

screen shot 2015-07-30 at 11 32 25

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1273,8 +1273,8 @@ HCL:
   extensions:
     - .hcl
     - .tf
-  ace_mode: javascript
-  tm_scope: source.json
+  ace_mode: ruby
+  tm_scope: source.ruby
 
 HTML:
   type: markup


pchaigno:
@radeksimko Thanks for the report and pull request :+1: 

Could you show an example of HCL file highlighted as Ruby with [Lightshow](https://github-lightshow.herokuapp.com/) for reviewers?

radeksimko:
@pchaigno Sure, here you go:

[screen shot 2015-07-30 at 12 06 42](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=source.ruby&grammar_url=&grammar_text=&code_source=from-text&code_url=&code=provider+%22aws%22+%7B%0D%0A++++region+%3D+%22us-east-1%22%0D%0A%7D%0D%0A%0D%0Aresource+%22terraform_remote_state%22+%22shared%22+%7B%0D%0A++backend+%3D+%22s3%22%0D%0A%0D%0A++%2F%2F+Here+is+a+comment+that+is+improperly+highlighted+%3A%28%0D%0A++%23+Here+is+a+comment%0D%0A++config+%7B%0D%0A++++bucket+%3D+%22terraform-test-bucket%22%0D%0A++++key+%3D+%22terraform.tfstate%22%0D%0A++++region+%3D+%22us-east-1%22%0D%0A++%7D%0D%0A%7D%0D%0A%0D%0Amodule+%22example%22+%7B%0D%0A++source+%3D+%22.%2Fmy_module%2F%22%0D%0A++environment+%3D+%22%24%7Bterraform_remote_state.shared.output.env_name%7D%22%0D%0A%7D)

It's not perfect (proper HCL grammar would be), but still a lot less obtrusive compared to Javascript.

pchaigno:
:+1:

sethvargo:
Hi all, 

We noticed this yesterday. For some reason (I don't know if maybe linguist changed between when I originally opened the PR and when it was merged), but the syntax highlighting is a bit off. I was going to play around with some other dialects, but it looks like Ruby definitely works the best.

Thanks @radeksimko and :+1:

pchaigno:
> I don't know if maybe linguist changed between when I originally opened the PR and when it was merged

Nop, [nothing changed on our side](https://github.com/github/linguist/blame/16a4b4947f4ba59800ee1b93aa9b5f943727b7c3/lib/linguist/languages.yml#L1271-L1278). Maybe the grammar changed though ;)

sethvargo:
Okay @pchaigno :smile: 

I'm in favor of merging this as-is until we have a real grammar for Terraform and sorry about the mishap!

arfon:
HCL support was shipped yesterday in [`v4.5.10`](https://github.com/github/linguist/releases/tag/v4.5.10). Happy to merge this now but I probably won't cut a new gem until next week sometime.

phinze:
Hey folks - quick question:

Over in Terraform land we've been regularly using `js` for [code blocks in comments](https://github.com/hashicorp/terraform/issues/2677#issuecomment-120498012) and [gists](https://gist.github.com/phinze/f7bf1c0753fd5e85bbca) and things have looked decent - that's why we thought the `js` grammar would be a good start.

What's different in those places that they don't get all red looking?

radeksimko:
@phinze Good question...

I can see you're setting the language explicitly to `js`, so the conclusion is that the original PR may not be setting the syntax to Javascript correctly.

https://github.com/github/linguist/blob/1c73db499fc11a20d780770524be7d2d383e722a/lib/linguist/languages.yml#L1592-L1635

the only difference I see is `tm_scope: source.json` VS `tm_scope: source.js`. Does that mean the original PR was trying to parse it as JSON?

`tm_scope: source.js` might be the solution, but I'd have to test it... It does look better for comments.

phinze:
> Does that mean the original PR was trying to parse it as JSON?

That's a good theory, the red highlighting does seem to correlate with JSON's stricter notion of quoted object attribute keys.

sethvargo:
screen shot 2015-07-30 at 11 47 55 am
screen shot 2015-07-30 at 11 48 11 am

@radeksimko's theory sounds correct, but I think Ruby is handling the comments a bit better than javascript

radeksimko:
@sethvargo I think it depends which comment syntax allowed in HCL you use. Javascript does not know `#` and Ruby doesn't know `//` nor `/* */`...

PR #2535, opened 2015-08-04
Case-insensitive extension match for heuristic rules

pchaigno:
This pull request makes the extension match for heuristic rules case-insensitive.
I didn't seem coherent since the detection by extension is case-insensitive (see #2533).

I noticed it wasn't with the [new heuristic](https://github.com/github/linguist/pull/2533/files#diff-e56a8d50b1a75fc4212ffb3504faa007R354) from @arfon.

[`samples/R/hello-r.R`](https://github.com/github/linguist/blob/master/samples/R/hello-r.R) is, indirectly, a test for this.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -56,7 +56,7 @@ module Linguist
 
     # Internal: Check if this heuristic matches the candidate languages.
     def matches?(filename)
-      @extensions.any? { |ext| filename.end_with?(ext) }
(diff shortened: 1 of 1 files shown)


arfon:
:+1: thanks @pchaigno - I meant to dig into this a little today then forgot. Then you fixed it :smile:

pchaigno:
:stuck_out_tongue_closed_eyes:

PR #2539, opened 2015-08-05
Updates for the Puppet language

rji:
This PR resolves two issues with the Puppet language in Linguist:
  - Updates the language color based on the [Puppet Labs style guide](https://puppetlabs.com/styleguide/brand#color)
  - Adds an additional sample to prevent files containing only the line `hiera_include('classes')` from being categorized as Pascal

For an example of the Pascal mis-classification, please see https://github.com/garethr/puppet-mesos-example/search?l=pascal

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2653,7 +2653,7 @@ Public Key:
 
 Puppet:
   type: programming
-  color: "#332A77"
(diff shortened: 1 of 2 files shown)


arfon:
:+1: thanks @rji - this will be live in the next release of Linguist (next week some time)

PR #2541, opened 2015-08-07
Reworking documentation paths

arfon:
Replaces #2164

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -179,6 +179,10 @@
 # Fabric
 - Fabric.framework/
 
+# git config files
+- gitattributes$
+- gitignore$
+- gitmodules$
 
 ## Groovy ##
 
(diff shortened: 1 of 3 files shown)


arfon:
/ cc @aroben for a quick review based on his comment [here](https://github.com/github/linguist/pull/2164#issuecomment-75966654)

lib/linguist/vendor.yml
```diff
@@ -224,19 +228,6 @@
 # Html5shiv
 - (^|/)html5shiv\.js$
 
-# Samples folders
-- ^[Ss]amples/
-
-# LICENSE, README, git config files
-- ^COPYING$
-- LICENSE$
-- License$
```
aroben: I think "License" won't be matched by `documentation.yml`. Can you add a test for it, and fix it if the test fails?

lib/linguist/vendor.yml
```diff
@@ -224,19 +228,6 @@
...
-- gitattributes$
-- gitignore$
-- gitmodules$
-- ^README$
-- ^readme$
```
aroben: Ditto for "readme"

aroben:
Looks pretty good! Just two small comments above.

arfon:
Thanks for the review @aroben :zap:

PR #2542, opened 2015-08-07
Highlight Brewfile as Ruby

josh:
Hey @arfon!

First time linguist contributor here! Does this look okay?

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2934,6 +2934,7 @@ Ruby:
   - .pryrc
   - Appraisals
   - Berksfile
+  - Brewfile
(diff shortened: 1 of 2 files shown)


pchaigno:
With a sample `Brewfile` in `samples/Ruby/filenames/` it will be perfect!
Sample files are used for testing (and training) so we consider it good practice to add a new sample for each new extension/filename ;)

josh:
Added sample Brewfile.

arfon:
> First time linguist contributor here!

:-) thanks @josh - looks great.

arfon:
:+1: thanks. I'll probably cut a new gem early next week.

PR #2546, opened 2015-08-10
Make regex for vim modeline more lenient

ammaraskar:
The current regex looks strictly for set lines containing only a filetype parameter. Unfortunately this doesn't play too well with files which also set additional vim settings. An example of such a file can be seen here: https://github.com/alliedmodders/sourcemod/blob/master/AMBuildScript

It'd be cool to have syntax highlighting on that file but currently that would involve making an additional line containing either the emacs style modeline or a vim modeline containing the ``ft`` exclusively. Both of these seem hacky.

I've also added two tests with more complex modelines and the filetype in different positions.


pchaigno:
Nice PR :+1:

arfon:
Thanks @ammaraskar. This will be live later this week when we cut a new gem.

PR #2555, opened 2015-08-13
Add .arpa to DNS Zone.

larsbrinkhoff:
Addition to #2554.  There are [753](https://github.com/search?q=extension%3Aarpa+soa+OR+root+OR+CNAME) `.arpa` files which are zone files.

(There are also [1163](https://github.com/search?q=extension%3Aarpa+NOT+soa+NOT+root+NOT+CNAME) `.arpa` files with some other content.  This in on par with the `.zone` file extension: [2698](https://github.com/search?q=extension%3Azone+soa+OR+root+OR+CNAME) vs [4084](https://github.com/search?q=extension%3Azone+NOT+soa+NOT+root+NOT+CNAME).)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -754,6 +754,7 @@ DNS Zone:
   type: data
   extensions:
   - .zone
+  - .arpa
   tm_scope: text.zone_file
   ace_mode: text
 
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
Note: there is also `.db`, but the numbers are not as favourable: 15k vs 833k.

arfon:
:+1: thanks @larsbrinkhoff. 

> `.com` and the other top-level domain names are also used for zone files, but that's a huge can of worms.

Agreed.

PR #2560, opened 2015-08-14
Add .sexp to Common Lisp.

larsbrinkhoff:
There are [21000](https://github.com/search?q=extension%3Asexp+NOT+slartibartfast) `.sexp` files.  They contain "[Symbolic Expressions](https://en.wikipedia.org/wiki/S-expression)", which is the syntax of Lisp.

Arguably, these files are not in any particular dialect of Lisp.  I had to choose one, so I picked Common Lisp.

This might be an opportunity to introduce a generic "Lisp" language, which was discussed in #1827 and #1837.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -616,6 +616,7 @@ Common Lisp:
   - .lsp
   - .ny
   - .podsl
+  - .sexp
   interpreters:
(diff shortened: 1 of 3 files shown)


arfon:
:+1: looks good. Thanks @larsbrinkhoff

PR #2561, opened 2015-08-14
Highlight LabVIEW files as XML

pchaigno:
LabVIEW files are [XML documents](https://github.com/search?utf8=%E2%9C%93&q=extension%3Alvproj+xml&type=Code&ref=searchresults).


arfon:
:+1:

PR #2565, opened 2015-08-18
Grammar for AspectJ

pchaigno:
This pull request adds a grammar for AspectJ.


arfon:
:+1: thanks @pchaigno

PR #2567, opened 2015-08-19
#import "*.h" detection for Objective-C

pcantrell:
Apple’s standard generated .h file for frameworks is incorrectly being identified as C++. This fix correctly identifies it as Objective-C.

The `#import` (as opposed to `#include`) directive is used by both Obj-C and Microsoft C++. However, the Microsoft version is [only for importing libraries](https://msdn.microsoft.com/en-us/library/8etzzkb6(VS.71).aspx) and would never refer to a file with an `.h` extension. If we see `#import "foo.h"` or `#import <foo.h>`, we can therefore conclude the file is Obj-C.

Helps with #1626.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -65,7 +65,7 @@ module Linguist
     end
 
     # Common heuristics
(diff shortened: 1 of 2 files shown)


pcantrell:
Rebased & ready to merge….

arfon:
:+1: thanks @pcantrell. This will be live in the next release of Linguist.

PR #2568, opened 2015-08-19
Use Licensee to classify submodule licenses

benbalter:
This pull requests updates the grammars test to lean on Licensee to do much of the license-detection heavy lifting. Specifically:

1. Look to Licensee first to see if it can detect a license
2. If licensee couldn't detect the license, but found a license file, use [Licensee's more nuanced license file detection](https://github.com/benbalter/licensee/blob/master/lib/licensee/project.rb#L37-L50), passing the license file to the existing license detection
3. If all else fails, look to the readme

Beyond the updated logic, rather than break things out into `:unlicensed` and `:unrecognized`, the test now contains an explicit whitelist of approved licenses which it compare the detected license, if any against. This allows for more descriptive failure messages (e.g., `the submodule FOO is licensed under the XYZ license, but the XYZ license isn't an approved license`). 

To implement this, I added a private `submodule_license` method to separate the license detection logic, and a private `submodule_licenses` which returns a hash containing each submodule and its license.


arfon:
Awesome @benbalter. Thanks so much for working on this. Looks pretty sane to me but given @aroben worked on this I'd love to :+1: from him too.

aroben:
Is this fixing a bug in the current test? If we have to look at README files for some repos anyway, what's the benefit to using Licensee?

arfon:
IMO the value here @aroben is that this in standardizing the way that we detect licenses in GitHub generally. Licensee is used for the [license API](https://developer.github.com/v3/licenses/) so it makes sense (to me at least :smile) to use the same approach where possible.

pchaigno:
Licensee also recognizes more licenses than the current test does.
And the problem of recognizing licenses is not as straightforward as it might appear (see #2575 for instance).

aroben:
Standardizing on Licensee would be great. My concern is that this PR adds Licensee but doesn't remove the existing

aroben:
@pchaigno Thanks for pointing out that bug.

I wish that using Licensee would let us get rid of the existing buggy license detection logic in here. @benbalter Do you think Licensee could adopt some of the README logic we have here?

test/test_grammars.rb
```diff
@@ -63,47 +80,35 @@ def test_local_scopes_are_in_sync
...
-      elsif classify_license(license)
-        :licensed
+    submodule_licenses.each do |submodule, license|
+      next if PROJECT_WHITELIST.include?(submodule)
+
+      license_file = Licensee::Project.new(submodule).license_file
+      submodule_name = submodule.split("/").last
+
+      if license_file
+        assert license, "Submodule #{submodule_name} contains an unrecognized license. Please update #{__FILE__} to recognize the license"
```
aroben: The existing test would print out all errors at once. The way the test is currently written it will print out the very first error and stop. It'd be nice to keep showing all errors like we did before.

test/test_grammars.rb
```diff
@@ -63,47 +80,35 @@ def test_local_scopes_are_in_sync
...
-    unrecognized = categories[:unrecognized] || []
-    disallowed_unlicensed = unlicensed - LICENSE_WHITELIST
-    disallowed_unrecognized = unrecognized - LICENSE_WHITELIST
-    extra_whitelist_entries = LICENSE_WHITELIST - (unlicensed | unrecognized)
+    unlicensed   = submodule_licenses.select { |k,v| v.nil? }
+    unrecognized = submodule_licenses.select { |k,v| v.nil? && Licensee::Project.new(k).license_file}
+    licensed     = submodule_licenses.reject { |k,v| v.nil? }
+    unapproved   = licensed.reject { |k,v| LICENSE_WHITELIST.include?(v) }
+    extra_whitelist_entries = PROJECT_WHITELIST - unlicensed.keys - unrecognized.keys
+    assert unapproved.empty?
```
aroben: This assertion could use an error message.

pchaigno:
> Thanks for pointing out that bug.

@aroben I introduced it, so the least I can do is report it :stuck_out_tongue_closed_eyes:

benbalter:
@aroben very valid and to be sure, the idea's far from perfect.

* Accuracy - As @pchaigno there's a handful of mis-identified licenses, MPL as GPL being the biggest I noticed. It doesn't matter when we accept both MPL and GPL, but in a world where we accept the misidentified license, but not the real one, Licensee could prevent the mistake.

* Consistency - As @arfon pointed out, there's something to be said for consistency. Licensee has a hierarchy of license file preferences, where multiple potential license files exist, and that preference will match .com (as well as the license categorization where Licensee detects the license).

* Dogfooding - I already found [not one](https://github.com/benbalter/licensee/releases/tag/v4.7.3), but [two](https://github.com/benbalter/licensee/releases/tag/v4.7.4) bugs while working on this PR.

* Future proof - As new licenses are added, we're less likely to update the heuristics that detect them, but rather just white list the license id.

* Potentially really future proof - If we wanted, rather than white listing licensing, regardless of how it's detected, since Licensee knows information about each license, we could say all licenses must allow modification but not require share alike. So long as the license meets those two requirements, the test could pass.

* Standards - Internally, we move to SPDX standards, rather than a roll-your-own license description

All that said, from a DRY standpoint, you're 100% right, we make the code less likely to be used, and rely on it less, but don't remove it entirely. 

> Do you think Licensee could adopt some of the README logic we have here?

It's [address in the Licensee readme a bit](https://github.com/benbalter/licensee#what-about-looking-to-see-if-the-author-said-something-in-the-readme), but in short, there's three issues with the regex approach:

1. Computers are terrible at reading english. So phrases like "This one file is MIT, but the entire project is otherwise GPL" is hard to reconcile. Perhaps it could look for one license name, but not another, but even then, you could say "the MIT license is a joke, do whatever you want with my code" in the readme.

2. If I say "This project is licensed under the GPL license" in my readme, is it GPLv2? v3? v2 or later? It's nearly impossible to know what rights you actually get.

3. Most licenses are actually license templates. They have blank spaces for things like the project name and copyright date. Linking to a template can go a long way in terms of manifesting author intent, but in terms of building tools to encourage norms, we'd want the tooling to optimize for the ideal use case, not the lowest common denominator.

All that said, it's going to continue to get better. Whether it's readme detection, or submitting a PR upstream to properly license the project, I'd love for the bespoke logic to eventually go away, no doubt.

(will work on fixing the tests per your feedback either way).

benbalter:
Went through and did a quick breakdown of how the existing licenses are being classified:

* Licensee detect 225 of 329 (~70%)

* The fallback license classifier identifies 17 of 329 (~5%). For all of these, licensee is properly refusing to call the license its closest match because each have added text (in the case of Mako, an explicit trademark reservation).

* The readme classifier identified 87 of 329 (~25%). In fbb3ab22920e2ffb305aa3e1806e8348b950e76b I was able to slim it down by removing unlicensed and bsd as they weren't being detected by the native detector anymore.

aroben:
> The readme classifier identified 87 of 329 (~25%).

Huh? The test was previously asserting that all submodules had a detected license (except for the whitelisted ones), but it sounds like you're saying that it only detected 25% of submodules had a license.

benbalter:
>  it sounds like you're saying that it only detected 25% of submodules had a license.

100% of submodules have detected licenses. Of those 100%, 75% were detected by licensee, failing that, 5% were detected by looking at the license file, and still failing that, the remaining 25% were detected by looking at the readme.

test/test_grammars.rb
```diff
@@ -112,30 +113,57 @@ def submodule_paths
...
+
+    # We know a license file exists, but Licensee wasn't able to detect the license,
+    # Let's try our own more permissive regex method
+    if project.license_file
+      path = File.expand_path project.license_file.path, submodule
+      license = classify_license(path)
+      return license if license
+    end
+
+    # Neither Licensee nor our own regex was able to detect the license, lets check the readme
```
aroben: Typo: lets -> let's

aroben:
Thanks for all the explanation @benbalter. This looks good to me.

benbalter:
@aroben just to make it explicit, batched the assertions (so they all fail at once with a descriptive error message, not just on the first submodule), per your suggestion, via https://github.com/github/linguist/commit/fbb3ab22920e2ffb305aa3e1806e8348b950e76b.

aroben:
Yup, thanks so much for doing that!

arfon:
:zap:

benbalter:
"Long-time user, first time contributor."

PR #2572, opened 2015-08-19
TextMate scope for RMarkdown

pchaigno:
Probably an edge case, but it can be useful for RMarkdown code snippets in Markdown documents.


arfon:
Oh nice! / cc @karthik

PR #2578, opened 2015-08-20
Make RAML a markup language

blakeembrey:
RAML was originally merged as a data language, but this seems like an incorrect definition. I changed it to be markup instead, which will also result in RAML appearing in repo statistics.

For example, API Blueprint is another similar API definition language that was recently merged as `markup`.


arfon:
:+1: thanks @blakeembrey. This will be live in the next release of Linguist (next week sometime)

PR #2585, opened 2015-08-23
Sass is an actual language

hcatlin:
I'd like to request that Sass/SCSS both be removed from the CSS group and have it's own group created. Sass has a ton of usage these days, outranking a lot of the other 'top level' languages, and while it's a server-side preprocessor, both of it's syntaxes are full blown languages at this point and think they deserve differentiation from their W3C brethren.

Thanks!

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2985,7 +2985,7 @@ SAS:
 SCSS:
   type: markup
   tm_scope: source.scss
-  group: CSS
+  group: Sass
   ace_mode: scss
   extensions:
   - .scss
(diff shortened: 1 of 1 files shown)


arfon:
Hi @hcatlin. We've been going back and forth a little on what to do with the grouping functionality in Linguist as I think there are some UI concerns when looking at a repository that contains languages defined as part of a larger group (see similar discussion here for `Jade` https://github.com/github/linguist/pull/2978#issuecomment-217525763)

I'd like us to fix the UI issue on GitHub rather than removing the grouping functionality so I'm going to close this.

PR #2586, opened 2015-08-23
Haml is independent of HTML

hcatlin:
I'm just on a tear today separating out my languages... ;)

My argument is that Haml should be independent of HTML, because it doesn't look at all like HTML, has a ton of other features, helpers, etc that make it different than HTML.

Also, Handlebars is considered distinct from HTML, and then so should Haml.

.haml: 1.1 million files https://github.com/search?utf8=%E2%9C%93&q=extension%3Ahaml+NOT+nothack&type=Code&ref=searchresults
.hbs: 300k https://github.com/search?utf8=%E2%9C%93&q=extension%3Ahbs+NOT+nothack&type=Code&ref=searchresults
.handlebars: 98k https://github.com/search?utf8=%E2%9C%93&q=extension%3Ahandlebars+NOT+nothack&type=Code&ref=searchresults

We'd really like to be able to track usage of Haml around the internet better than is indicated by keeping it as just a flavour of HTML.

Thanks!

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1357,7 +1357,6 @@ Hack:
   tm_scope: text.html.php
 
 Haml:
-  group: HTML
   type: markup
   extensions:
   - .haml


arfon:
> Also, Handlebars is considered distinct from HTML, and then so should Haml.

I'd actually argue this is an error in Linguist (that we should fix). Currently we have lots of the HTML templating languages under `group: HTML`:

- [HTML+Django](https://github.com/github/linguist/blob/e128c3fa82a2ea2bca8f529a970ec8003a6f5da3/lib/linguist/languages.yml#L1312)
- [erb](https://github.com/github/linguist/blob/e128c3fa82a2ea2bca8f529a970ec8003a6f5da3/lib/linguist/languages.yml#L1325)
- [Jade](https://github.com/github/linguist/blob/e128c3fa82a2ea2bca8f529a970ec8003a6f5da3/lib/linguist/languages.yml#L1586)
- [Latte](https://github.com/github/linguist/blob/e128c3fa82a2ea2bca8f529a970ec8003a6f5da3/lib/linguist/languages.yml#L1757)
- [Scaml](https://github.com/github/linguist/blob/e128c3fa82a2ea2bca8f529a970ec8003a6f5da3/lib/linguist/languages.yml#L3117)

> We'd really like to be able to track usage of Haml around the internet better than is indicated by keeping it as just a flavour of HTML.

I'm definitely sympathetic to this but I'm curious what 'tracking' you're talking about here. Being in the `HTML` group doesn't stop `HAML` from being segmented in our search:

screen_shot_2015-09-08_at_12_20_50_pm

cc @aroben here as he might have some thoughts on the segmentation of markup languages.

aroben:
> I'm curious what 'tracking' you're talking about here. Being in the HTML group doesn't stop HAML from being segmented in our search:

I think this only affects the language stats bar on the repo page.

hcatlin:
Referring to this: https://github.com/blog/2047-language-trends-on-github

hcatlin:
Since you've moved Handlebars into the same type of group, we're all good. ;)

arfon:
> Since you've moved Handlebars into the same type of group, we're all good. ;)

:+1:

PR #2590, opened 2015-08-25
Adding support for recognizing F* files  ... no syntax highlighting yet

catalin-hritcu:
F* is an ML-like functional programming language aimed at program verification:
https://en.wikipedia.org/wiki/F*_(programming_language)
https://www.fstar-lang.org/

Here is a PR adding support for recognizing F* files based on their extension (`.fst` and `.fsti`). These extensions are not currently in `languages.yml`.

Here is a GitHub search result showing in-the-wild usage:
https://github.com/search?l=&q=module+extension%3Afst&ref=advsearch&type=Code&utf8=%E2%9C%93

This PR only provides support for recognizing F* files based on extension. We're also working on contributing F* syntax highlighting back to ace editor:
https://github.com/FStarLang/FStar/issues/333
https://www.fstar-lang.org/ace/mode-fstar.js
and we also have an Atom grammar we could contribute:
https://github.com/FStarLang/atom-fstar/blob/master/grammars/fstar.cson

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -947,6 +947,18 @@ F#:
   tm_scope: source.fsharp
   ace_mode: text
 
+F*:
+  type: programming
+  color: "#b845fc"
+  search_term: fstar
+  aliases:
+  - fstar
+  extensions:
+  - .fst
+  - .fsti
+  tm_scope: none
+  ace_mode: text
+
 FLUX:
   type: programming
   color: "#88ccff"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -955,7 +955,6 @@ F*:
   - fstar
   extensions:
   - .fst
-  - .fsti
   tm_scope: none
   ace_mode: text
 
--- a/samples/F*/qsc.fst
+++ b/samples/F*/qsc.fst
@@ -1,0 +1,93 @@
+(*--build-config
+    options:--admit_fsi Set --z3timeout 20;
+    other-files:set.fsi heap.fst st.fst all.fst list.fst
+  --*)
+module QuickSort
+open List
+#set-options "--initial_ifuel 2 --initial_fuel 1 --max_ifuel 2 --max_fuel 1"
+
+(* Specification of sortedness according to some comparison function f *)
+val sorted: ('a -> 'a -> Tot bool) -> list 'a -> Tot bool
+let rec sorted f l = match l with
+    | [] -> true
+    | [x] -> true
+    | x::y::xs -> f x y && sorted f (y::xs)
+
+
+type total_order (a:Type) (f: (a -> a -> Tot bool)) =
+    (forall a. f a a)                                           (* reflexivity   *)
+    /\ (forall a1 a2. (f a1 a2 /\ f a2 a1)  ==> a1 == a2)       (* anti-symmetry *)
+    /\ (forall a1 a2 a3. f a1 a2 /\ f a2 a3 ==> f a1 a3)        (* transitivity  *)
+    /\ (forall a1 a2. f a1 a2 \/ f a2 a1)                       (* totality      *)
+
+
+val count : 'a -> list 'a -> Tot nat
+let rec count (x:'a) (l:list 'a) = match l with
+  | hd::tl -> (if hd=x then 1 else 0) + count x tl
+  | [] -> 0
+
+
+val mem_count : x:'a -> l:list 'a ->
+      Lemma (requires True) (ensures (mem x l == (count x l > 0)))
+      [SMTPat (mem x l)] (decreases l)
+let rec mem_count x l = match l with
+  | [] -> ()
+  | _::tl -> mem_count x tl
+
+
+val append_count: l:list 'a -> m:list 'a -> x:'a
+               -> Lemma (requires True)
+                        (ensures (count x (append l m) = (count x l + count x m)))
+                        [SMTPat (count x (append l m))]
+let rec append_count l m x = match l with
+  | [] -> ()
+  | hd::tl -> append_count tl m x
+
+
+val partition: ('a -> Tot bool) -> list 'a -> Tot (list 'a * list 'a)
+let rec partition p = function
+  | [] -> [], []
+  | hd::tl ->
+     let l1, l2 = partition p tl in
+     if p hd
+     then hd::l1, l2
+     else l1, hd::l2
+
+opaque logic type trigger (#a:Type) (x:a) = True
+val partition_lemma: f:('a -> Tot bool) -> l:list 'a -> Lemma (requires True)
+      (ensures (forall hi lo. (hi, lo) = partition f l
+                ==>  (length l = length hi + length lo
+                 /\ (forall x.{:pattern (trigger x)}
+                              trigger x ==>
+                               (mem x hi ==> f x)
+                           /\  (mem x lo ==> not (f x))
+                           /\  (count x l = count x hi + count x lo)))))
+      [SMTPat (partition f l)]
+let rec partition_lemma f l = match l with
+    | [] -> ()
+    | hd::tl -> cut (trigger hd); partition_lemma f tl
+
+
+val sorted_app_lemma: #a:Type
+                      -> f:(a -> a -> Tot bool){total_order a f}
+                      -> l1:list a{sorted f l1}
+                      -> l2:list a{sorted f l2}
+                      -> pivot:a
+                      -> Lemma (requires ((forall y.{:pattern (trigger y)} trigger y ==>
+                                                        ((mem y l1 ==> not (f pivot y))
+                                                      /\ (mem y l2 ==> f pivot y)))))
+                               (ensures (sorted f (append l1 (pivot::l2))))
+                               [SMTPat (sorted f (append l1 (pivot::l2)))]
+let rec sorted_app_lemma f l1 l2 pivot = match l1 with
+    | [] -> if is_Cons l2 then cut (trigger (Cons.hd l2)) else ()
+    | hd::tl -> cut (trigger hd); sorted_app_lemma f tl l2 pivot
+
+val sort: f:('a -> 'a -> Tot bool){total_order 'a f}
+       -> l:list 'a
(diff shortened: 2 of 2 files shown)


larsbrinkhoff:
Thank you for your contribution.

Did you check the [contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md)?  Supposedly, a new language should appear in "hundreds of repositories".

Also, 99% of all `.fst` files appear to be in [FASTA](https://en.wikipedia.org/wiki/FASTA_format).  That would make it harder to do a search and find actual F* results.

catalin-hritcu:
@larsbrinkhoff  I guess the only thing we could do about FASTA is change our extension, right? It doesn't matter that FASTA is not a programming language and can be easily distinguished syntactically from F\*, does it?

We could consider changing our extension, but it would still not solve the "hundreds of repositories" problem. It's hard for a new research language with a focus on verification to achieve this goal in a reasonable amount of time, still if I look at your list I see a couple such languages there, like Alloy, Idris and Lean.

Anyway, if getting on this list is not an option for F\*, is there anything we can do to get language recognition and syntax highlighting to work well for the F\* github repository (https://github.com/FStarLang/FStar)? We tried setting `linguist-language=FStar` in `.gitattributes` for `fst` and `fsti` files, but that only works if we're in FStar was on this global list, right?

larsbrinkhoff:
I should explain that I'm just a contributor, not part of GitHub staff, so I don't decide what goes into Linguist or not.

Many languages do share file extensions, so that problem is solvable.  However, you're right that research languages usually are not accepted, until they see more widespread use.

larsbrinkhoff:
I downloaded 962 `.fst` sample file somewhat randomly, and this is the breakdown into various languages and formats:

Language | Files | Comment
--- | --- | ---
F* | 564 | Verification
FASTA | 230 | Bioinformatics
FAST | 66 | Verification
Binary | 43
Unknown | 43
OpenFST | 16 | Speech recognision

lib/linguist/languages.yml
```diff
@@ -947,6 +947,18 @@ F#:
...
 
+F*:
+  type: programming
+  color: "#b845fc"
+  search_term: fstar
+  aliases:
+  - fstar
+  extensions:
+  - .fst
+  - .fsti
```
pchaigno: I think you should remove that extension. There are [only 6 `.fsti` files](https://github.com/search?utf8=%E2%9C%93&q=NOT+djchfdjbhdjfbfnv+extension%3Afsti&type=Code&ref=searchresults) ;)

catalin-hritcu:
@pchaigno  I've removed .fsti from my PR until it gets more popular. Interface files are a relatively new feature of F* and adoption started only recently ... there are 18 files now :)

catalin-hritcu:
Any news about this PR?

catalin-hritcu:
This seems to be taking really long. Would it be a reasonable compromise to remove even the `fst` extension from my patch, so that we can at least get language recognition and syntax highlighting to work well for the F* repo (using `.gitattributes`)? Could this work technically?

larsbrinkhoff:
I tried your search `module+extension:fst` and found 12 repositories.  The guideline says "hundreds of repositories".  So my guess would be that F* is not yet popular enough to add to Linguist.

catalin-hritcu:
Requiring a language to be used in hundreds of repos just to be able to allow a single project to get syntax highlighting working for its own files seems completely broken. How about freeing up syntax highlighting on github from this tyranny?

larsbrinkhoff:
Repository-local language highlighting has been suggested: #2627.

arfon:
> So my guess would be that F* is not yet popular enough to add to Linguist.

Yeah I'm afraid this isn't enough to justify an addition to Linguist yet. 

@catalin-hritcu - I'm sorry that this is a frustrating situation for you. As @larsbrinkhoff pointed out, custom repository-level highlighting is something that's been requested already (and would be a very nice feature to have). 

For now though, `F*` does not meet our contributor guidelines for GitHub usage. I'd suggest we revisit this in a few months when we can evaluate the adoption levels then.

beurdouche:
@arfon @pchaigno @larsbrinkhoff Annual revisit ? ;) 
F* is much more used in the industry now and probably more on Github too (for a verification language). Doubled the number of followers (more than Coq now...). Doubled the number of commits. More publications... WDYT ? Thanks for your time ;)

catalin-hritcu:
@beurdouche That's not what they count officially though.
> we prefer that extensions be in use in hundreds of repositories before supporting them in Linguist.
So how many repositories do we have now? :)

@larsbrinkhoff  How did you get the 12 number last time? I'm quite sure it's a lot more than that now.

larsbrinkhoff:
I don't recall exactly, but I probably used a script which downloaded all files found by searching for `modules extension:fst`.  Unfortunately, that script doesn't work any more due to changes at GitHub.  @Alhadis, did you fix your search script yet?

Alhadis:
If by "fix" you mean bring it to a point where it can run semi-automatically, yeah.

**Usage:**
- [Copy+paste this into developer console](https://github.com/Alhadis/Harvester/blob/js/harvester.js)
- Run `harvest("extension:fst", "modules").then(_=> new Notification("Done"))`
- When done, run `copy(silo.reap("extension:fst"));` to copy to the system clipboard

That'll give you a sorted list of unique URLs to run through `wget` or whatever.

larsbrinkhoff:
I'm not sure if I used it right (how do I know it's done?), but I think I found five repositories:

https://github.com/FStarLang/FStar/
https://github.com/YaccConstructor/YC_FStar/
https://github.com/fstar-spbu-edu/fstar-edu/
https://github.com/mitls/hacl-star/
https://github.com/mitls/mitls-fstar/

Alhadis:
It'll display a desktop notification when it's finished. If you leave the dev console open, you can sorta monitor the harvest's progress. Assuming you have `Log XMLHttpRequests` enabled (and that you're running this from a GitHub-hosted URL).

Alhadis:
I should also point out that it's slow on purpose, because GitHub will block access if it detects a perceived abuse attempt.

Which is ironic, because it's stopping us from doing the exact opposite.

catalin-hritcu:
@Alhadis  Finally got to trying this out and it works quite well. Thanks a lot!

One change I needed to do is replace a `modules` to `module` in the harvest command that should look like this:
```
harvest("extension:fst", "module").then(_=> new Notification("Done"))
```

Finally, I pasted the results in a file
https://gist.github.com/catalin-hritcu/aae41bb3e34acb7b34c388c635a8c5d5
and run this command:
```
cut -f1-5 -d / repos.txt | sed -e 's/raw.githubusercontent.com/github.com/g' | sort | uniq
```
to obtain the following list of 53 repos:
https://gist.github.com/catalin-hritcu/52c8aac5c8c946fec76f19760ed59aa0

That's a 4x improvement to the 12 repos @larsbrinkhoff reported last year, but it's still not yet "hundreds of repositories".

We have 45 F* contributors though (https://github.com/FStarLang/FStar/graphs/contributors) so we could also get there by asking everyone to fork our repo part of their development process. Would that count? :)

Alhadis:
> We have 45 F* contributors though (https://github.com/FStarLang/FStar/graphs/contributors) so we could also get there by asking everyone to fork our repo part of their development process. Would that count? :)

I don't think forks are considered, but I could be wrong. @pchaigno, would you know?

> Finally, I pasted the results in a file
https://gist.github.com/catalin-hritcu/aae41bb3e34acb7b34c388c635a8c5d5
and run this command:

You'll get sick of doing that, trust me. =) [Keep this handy](https://github.com/Alhadis/Harvester/blob/js/summary.pl) if you find yourself running the harvester often.

(I regret not streamlining the process further, BTW... ideally, the script should just dump the URLs to a text file somewhere on the user's system, but that's not so easy when running in a browser context...)

catalin-hritcu:
@Alhadis  We might be nearing 100s of repos using F*, just that your scripts don't seem ready do acknowledge it :)

The outputs I get this time are these:
https://gist.github.com/catalin-hritcu/67003b4d7d17e1fbc557b7d20a1e0572
https://gist.github.com/catalin-hritcu/7d393ec9dab2b3da71d7f193a4c772fb
Implying that we now have only 39 unique repos, which is **significantly lower** than last time I tried (53). This is simply not right. Looked at the diff between the two, and the repos that are now "gone" from the output are definitely still there.

screenshot from 2017-07-21 10-06-19

What's happening here? Could it be because of the capping of the number of search results or is it something else?

catalin-hritcu:
Tried to run this once more and I see differences from the previous run
screenshot from 2017-07-22 09-23-02
This seems to imply that the script is a bit flaky, but still doesn't explain the much bigger difference I'm seeing compared to November.

Alhadis:
@catalin-hritcu The script only collates **public** search results. So if people have private F* repositories, it won't be able to accommodate them in the usage statistics.

Granted, the script isn't perfect, but it should by no means be an accurate, black-and-white gauge of a language's popularity in-the-wild. It was authored with intent of helping Linguist contributors add new languages and determine whether there's a risk of misclassifications/conflict.

catalin-hritcu:
@Alhadis  After a bit more thinking, I'm starting to see that your script does the best job that can be done given the data that GitHub exposes in its search results. The problem with getting a complete list of GitHub projects using our language this way seems to be that GitHub caps the number of search results it exposes to 1000. In the case of F\* there are however 3497 hits currently, and even if your script tries the 3 different orders "Best match", "Recently indexed", and "Least recently indexed", it still only manages to gather ~2380 hits, i.e. 68%. I'm not sure what kind of trickery one can play to get a more complete list.

> The script only collates public search results. So if people have private F* repositories, it won't be able to accommodate them in the usage statistics.

Don't think this has to do with private repos, and I have already checked that the now missing sites are indeed public and do contain `fst` files.

Alhadis:
Accidentally copy+pasting uncompressed JavaScript code into your terminal:

yeah.nah();

Not a fan.

@pchaigno You think that's how Stephane Chazelas managed to write [this beauty](https://github.com/stephane-chazelas/misc-scripts/blob/master/which_interpreter)? =) Also, the last time I said I was gonna redevelop this as a command-line solution, this was actually the real impetus for doing so - I kept forgetting the damn code was on my clipboard and found myself swimming up a waterfall of shell errors... for the the third time.

PR #2591, opened 2015-08-26
Vendored definitions for Xcode-related files

jtbandes:
A couple Xcode project–related things that shouldn't count towards language stats.


arfon:
Thanks @jtbandes. This will be live later this week when we cut a new Linguist gem.

PR #2592, opened 2015-08-26
Fix typo in Obj-C heuristic keyword

jtbandes:
`synchronised` → `synchronized`


arfon:
:+1: thanks @jtbandes

PR #2589, opened 2015-08-24
Add the X10 language.

mandel:
Add support for the [X10 language](https://en.wikipedia.org/wiki/X10_%28programming_language%29).

According to [this search](https://github.com/search?o=asc&p=1&q=class+extension%3Ax10&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93) there are a little bit less than 12,000 code results.

PR #2594, opened 2015-08-26
Add some additional help text to license test

benbalter:
This pull request adds some additional help text to the license list, to make scenarios like the one in https://github.com/github/linguist/pull/2589 a bit friendlier for contributors.

It instructs users to add a LICENSE file, if none exists, and to add *the text* of the license to the license file, if we can't detect the license from the license file.


arfon:
:+1: looks good to me @benbalter

PR #2601, opened 2015-08-29
An example comparison for docs

james69miller:


james69miller:
Activate my account at less for 1 mouth

PR #2603, opened 2015-08-29
Add support for the venerable M4 language

matejak:
m4 is a general-purpose macro processor designed by Brian Kernighan and Dennis Ritchie. GNU M4 (the interpreter) also has built-in functions for including files, running shell commands, doing arithmetic, etc. 

Since the `autoconf` project uses `m4`, the language is very often used to fine-tune the build process of some `C/C++` projects. However, `m4` can be used on its own.

On the technical side, I think that it is a big luck that the `.m4` extension is unoccupied. The language has a very free-form syntax and it is often used as code generator, so identifying `m4` code using statistical methods seems impossible to me.

A huge number of projects make use of `m4`: 
* https://github.com/search?p=1&q=extension:m4+NOT+nothack&ref=searchresults
* https://github.com/search?utf8=✓&q=filename:configure.ac+NOT+nothack&ref=searchresults


pchaigno:
Are [these](https://github.com/search?o=desc&p=1&q=extension%3Am4+dnl&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93) m4 files too?

Do you know a grammar (Sublime Text package or Atom/TextMate bundle) we could use to highlight this language?

samples/M4/configure.ac
```diff
@@ -0,0 +1,143 @@
+#                                               -*- Autoconf -*-
```
pchaigno: This file should be in `samples/M4/filenames/` ;)

larsbrinkhoff: Also, because of the Emacs modeline, maybe one of:

- Add an `autoconf` alias to M4.
- Add a separate Autoconf language with `group: M4`.

matejak:
@pchaigno : Yes, those are m4 files. More precisely, they are m4sugar files.

I have introduced a new language - m4sugar. m4sugar is used in autoconf (such as configure.ac files and support m4 files such as those pointed out by pchaigno). However, m4sugar syntaxi is valid m4 syntax, although it looks quite different.
I am not knowledgeable of grammar files, so I propose to leave them for later if it is possible. I think that the practical approach would be to have separate grammar files for m4 and m4sugar (autoconf).

larsbrinkhoff:
It seems `configure.ac` is ignored by Linguist.  So while useful as a sample file, it will have no effect in `filenames:`.

Or so I believe.  It would be good to check this.

matejak:
Wow, good point. `configure.ac` is, however, hand-creafted `m4sugar` and should be seen as part of the project codebase. So I have removed it from the list, however some .m4 files ended up there.

arfon:
Closing in favour of https://github.com/github/linguist/pull/2868

PR #2610, opened 2015-09-01
Add color for the PL/SQL language.

ismail-arilik:
I have added color to the PL/SQL language: #dad8d8. I take this color from a window background which evokes the interface of the language. Also I searched for this color code in this file but I couldn't find any match.


arfon:
:+1: thanks @ismail-arilik. This will be live later this week when we cut a new version of Linguist.

PR #2616, opened 2015-09-02
Ignore spec fixtures

scttnlsn:


scttnlsn:
Example of a repository that has the language mislabelled because of this: https://github.com/scttnlsn/dandelion

arfon:
:+1: makes sense. This will be live with the next release of Linguist (likely another week or so away)

scttnlsn:
Great.  Thanks!

PR #2620, opened 2015-09-03
Treat Go dependency tree as generated and vendored code

kr:
This PR is like #1549, but as of Go 1.5, the go tool has a vendoring mechanism, and it uses `vendor` as the directory name.

For more details on the Go 1.5 vendor functionality, see
https://groups.google.com/d/msg/golang-dev/74zjMON9glU/4lWCRDCRZg0J
and the description of the change introducing the feature,
https://go.googlesource.com/go/+/183cc0cd41

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -56,6 +56,7 @@ module Linguist
       generated_net_specflow_feature_file? ||
       composer_lock? ||
       node_modules? ||
+      go_vendor? ||
       godeps? ||
       generated_by_zephir? ||
       minified_files? ||
@@ -278,6 +279,14 @@ module Linguist
     # Returns true or false.
     def node_modules?
       !!name.match(/node_modules\//)
+    end
+
+    # Internal: Is the blob part of the Go vendor/ tree,
+    # not meant for humans in pull requests.
+    #
+    # Returns true or false.
+    def go_vendor?
+      !!name.match(/vendor\//)
     end
 
     # Internal: Is the blob part of Godeps/,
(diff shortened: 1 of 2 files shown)


pchaigno:
The `vendor` directory is [already marked as vendored](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml#L71). Why do you want to mark it as generated too?

kr:
> The vendor directory is already marked as vendored.

Ah, excellent. I'll revise this PR.

> Why do you want to mark it as generated too?

For the same reason that `node_modules` and `Godeps` directories are marked as generated.

arfon:
Hi @kr - thanks for submitting this. Unfortunately I don't think we can merge this as it's going to mark all `vendor` paths as generated for all users of GitHub (not just people vendorizing Go dependencies).

In the future we plan to support [overrides](https://github.com/github/linguist#using-gitattributes) for custom behaviour like this (i.e. doing something like `vendor/* linguist-generated=true`) but for now I think we should keep things as they are.

parkr:
> In the future we plan to support overrides for custom behaviour like this (i.e. doing something like vendor/* linguist-generated=true) but for now I think we should keep things as they are.

Hey @arfon, with the [go 1.6 release making vendoring in `vendor/` the default](https://golang.org/doc/go1.6#go_command), I believe we should rethink this. Any `.go` files in `vendor/` in 1.6 and later will be treated as vendored dependencies.

bkeepers:
:+1: I agree.

Could it be as easy as changing the regex to `/\Avendor\/.*\.go\z/`?

PR #2622, opened 2015-09-04
Language bar clarification in README.md

miksen:
Clarified what the percentages in the Language bar are based on.


arfon:
:+1: thanks @miksen

PR #2624, opened 2015-09-05
New JSX language under JavaScript group

pchaigno:
A specific grammar is needed to highlight `.jsx` files.
Thus, these files are now grouped under a distinct language but still in the JavaScript group.
See discussion at atom/language-javascript#220 for background.

Current highlighting of JSX code:
```js
'use strict';

const React = require('react')

module.exports = React.createClass({
  render: function() {
    let {feeds, log} = this.props;

    log.info(feeds);
    return <div className="feed-list">
      <h3>News Feed's</h3>
      <ul>
        {feeds.map(function(feed) {
          return <li key={feed.name} className={feed.fetched ? 'loaded' : 'loading'}>
            {feed.data && feed.data.length > 0 ?
              <span>{feed.name} <span className='light'>({feed.data.length})</span></span>
              : 'feed.name' }
          </li>
        })}
      </ul>
    </div>;
  }
});
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -677,3 +677,6 @@
 [submodule "vendor/grammars/X10"]
 	path = vendor/grammars/X10
 	url = git@github.com:x10-lang/x10-highlighting.git
+[submodule "vendor/grammars/language-babel"]
+	path = vendor/grammars/language-babel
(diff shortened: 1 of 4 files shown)


arfon:
:+1: thanks @pchaigno

PR #2630, opened 2015-09-14
4.6.0: Reduce memory pressure

vmg:
Here's a small release that reduces memory pressure when calculating languages (I ran memory benchmarks and fixed a couple trivial allocation spots that were *really* showing up on the graphs). I also added a `git-linguist` helper that lets you run language statistics as a separate process spawned by git.

cc @github/linguist


arfon:
:+1: looks great.

lib/linguist/generated.rb
```diff
@@ -241,22 +241,26 @@ def generated_go?
       return lines[0].include?("Code generated by")
     end
 
+    PROTOBUF_EXTENSIONS = ['.py', '.java', '.h', '.cc', '.cpp']
```
aroben: Is there any benefit to `.freeze`ing these arrays?

vmg: Just logic safety. It shouldn't affect performance.

PR #2631, opened 2015-09-14
Add Biml scripts to languages.yml

matt40k:
Add Biml scripts - http://bimlscript.com/GetStarted/AboutBimlScript
Schema - http://bimlscript.com/Content/Docs/biml.xsd

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3608,6 +3608,7 @@ XML:
   - .xml
   - .ant
   - .axml
+  - .biml
   - .ccxml
   - .clixml
   - .cproject


arfon:
Hi @matt40k. I'm afraid I can only find [30 results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Abiml+NOT+nothack&type=Code&ref=searchresults) for this extension which isn't really sufficient (see our contributing guidelines [here](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language)

If you're looking to override the language statistics on your repository to include `.biml` files then you can do that [like this](https://github.com/github/linguist#overrides).

Feel free to re-open this if the number of `.biml` files increases significantly.

PR #2634, opened 2015-09-15
Add the Solidity Language

whitj00:
Examples in wild: https://github.com/search?q=extension%3Asol+contract&ref=searchresults&type=Code&utf8=✓

Documentation:  https://github.com/ethereum/wiki/wiki/Solidity-Tutorial https://github.com/ethereum/wiki/wiki/Solidity-Features

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3239,6 +3239,13 @@ Smarty:
   ace_mode: smarty
   tm_scope: text.html.smarty
 
+Solidity:
+  type: programming
+  extensions:
+  - .sol
+  ace_mode: text
+  tm_scope: source.sol
+
 SourcePawn:
   type: programming
   color: "#5c7611"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3244,7 +3244,7 @@ Solidity:
   extensions:
   - .sol
   ace_mode: text
-  tm_scope: source.sol
+  tm_scope: source.solidity
 
 SourcePawn:
   type: programming
--- a/.gitmodules
+++ b/.gitmodules
@@ -680,3 +680,6 @@
 [submodule "vendor/grammars/language-babel"]
 	path = vendor/grammars/language-babel
 	url = https://github.com/gandm/language-babel
+[submodule "vendor/grammars/SublimeEthereum"]
+	path = vendor/grammars/SublimeEthereum
+	url = https://github.com/davidhq/SublimeEthereum
--- a/grammars.yml
+++ b/grammars.yml
@@ -130,6 +130,8 @@ vendor/grammars/SublimeBrainfuck:
 - source.bf
 vendor/grammars/SublimeClarion/:
 - source.clarion
+vendor/grammars/SublimeEthereum:
+- source.solidity
 vendor/grammars/SublimePapyrus/:
 - source.compiled-papyrus
 - source.papyrus
--- a/samples/Solidity/crowdsale.sol
+++ b/samples/Solidity/crowdsale.sol
@@ -1,0 +1,73 @@
+contract token { 
+    mapping (address => uint) public balance; 
+    function token() {}  
+    function sendToken(address receiver, uint amount) returns(bool sufficient) {  } 
+}
+
+contract CrowdSale {
+    
+    address public admin;
+    address public beneficiary;
+    uint public fundingGoal;
+    uint public amountRaised;
+    uint public deadline;
+    uint public price;
+    token public tokenReward;   
+    Funder[] public funders;
+    
+    /* data structure to hold information about campaign contributors */
+    struct Funder {
+        address addr;
+        uint amount;
+    }
+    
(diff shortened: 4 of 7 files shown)


pchaigno:
There are [many data files under the `.sol` file extension](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+NOT+contract&type=Code&ref=searchresults). We should probably recognize them if we don't want to end up with many miss-classifications...

whitj00:
How would you recommend excluding those?

pchaigno:
Sorry for the delay. Busy weeks.

We can't only exclude them, we have to recognize their language.

I downloaded 913 of [these files (`.sol NOT contract`)]. 660 of them match the same pattern: a file with only numbers. So this looks like some kind of data file but we still need to identify it. Once that's done, we can add it to Linguist with eventually a heuristic rule to distinguish the two languages.

Is the `contract` keyword mandatory in Solidity files?

whitj00:
yes

arfon:
@pchaigno - are you still planning on looking into this one?

u2:
We are keen for it. :+1:

arfon:

pchaigno:
It looks like I was wrong, few of [these files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+NOT+contract&type=Code&ref=searchresults) match the language identified by @arfon. If we exclude Solidity files and files related to the circuit board soldering diagrams, we still have [many unidentified files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+NOT+contract+NOT+M02&type=Code&ref=searchresults).
I don't think we can add the Solidity language without a `not-a-language` option. Otherwise, we will have many false positives :/

arfon:
> I don't think we can add the Solidity language without a `not-a-language` option. Otherwise, we will have many false positives :/

:disappointed: I think you're right @pchaigno. Thanks for the followup. @whitj00 - I'm afraid this is a limitation with Linguist right now - if one (and only one) language defines a file extension (`.sol` in this case) then any/all files with that extension will be automatically classified as the language that is listing the `.sol` extension.

As @pchaigno mentioned, the long-term solution is for us to have an option to run our heuristics and return `no language` when it doesn't match. Right now though we don't have support for such a behaviour I'm afraid.

The only solution that we can consider in the short term is to add `Solidity` but without defining the `.sol` extension. This will allow you to use the [Linguist overrides](https://github.com/github/linguist#overrides) to manually set the language of _your_ repositories in the `.gitattributes` data. Would this be of any value to you?

arfon:
> The only solution that we can consider in the short term is to add Solidity but without defining the `.sol` extension. This will allow you to use the Linguist overrides to manually set the language of your repositories in the `.gitattributes` data. Would this be of any value to you?

Ping @whitj00. Any thoughts on my last comment :point_up: ?

graup:
@arfon, I think the last thing you proposed would be quite helpful, and it would also prepare for the day when you do add support for `no language`.

LogvinovLeon:
As a guy comming from a competitive programming background I can say, that *.sol files are often used to store **sol**utions to tasks. Solutions in general can have any format, so it's not possible to identify all of them.

Also from the official Sodility grammar I can claim, that any Solidity file would have `contract` or `library` keyword, so implementing a heuristicks would be easy, after we would have `not-a-language`

lib/linguist/languages.yml
```diff
@@ -3239,6 +3239,13 @@ Smarty:
   ace_mode: smarty
   tm_scope: text.html.smarty
 
+Solidity:
+  type: programming
+  extensions:
+  - .sol
```
frangio: According to the discussion in #2634 the extension would have to go until the `not-a-language` feature is implemented.

frangio:
@pchaigno @arfon Can we get this PR merged without the `.sol` extension? This would allow us to use the override as @arfon suggested, which is still not currently possible.

arfon:
@frangio - I no longer maintain this project but you can now use [`.gitattribute` overrides](https://github.com/github/linguist#using-gitattributes) to change pretty much anything about GitHub's default behaviour.

frangio:
@arfon, as I understand it, forcing language X in an override doesn't work if language X isn't declared as such in the linguist library (i.e. in the `languages.yml` file).

I created [a repo](https://github.com/frangio/test-linguist-language) to test the override with Solidity, but it doesn't seem to work.

arfon:
Ah yes. My mistake. I didn't realise that Solidity wasn't defined.

maraoz:
+1 to @frangio's request

Vishesh-Gupta:
I recently added a solidity file, it was not recognized with the languages and got to know about linguist. I think even though there are a lot of files with the same extension now but then most of them are solidity files so I think we can add it as a language to [languages.yml](https://github.com/github/linguist/pull/2634/files/10c1e315b36312805856e43c23eababcc3fad2ba#diff-3552b1a64ad2071983c4d91349075c75) and for other files they can use a .gitignore? as classifying a language helps out a lot.

Issue #2645, opened 2015-09-23
Add Support for PL/SQL


jaredpetersen:
It would be nice if PL/SQL was supported, though I'm not sure if it is used enough to warrant it. Thoughts?

arfon:
Pretty sure we already do: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L2452-L2463

jaredpetersen:
Huh, interesting. I've always save my PL/SQL code as a .pck, which is a valid file extension, but it's not listed there.

arfon:
Looks like there are a [fair few of these](https://github.com/search?l=&q=create+extension%3Apck&ref=advsearch&type=Code&utf8=%E2%9C%93) files on GitHub. Feel free to open a new Pull Request adding this extension (take a look at our [contributing guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language) together with a sample or I'll likely get to this in the coming weeks.

jaredpetersen:
Okay, that sounds good! I'll try to look into that further tonight. Thanks for your help!

jaredpetersen:
@arfon Would you mind taking a look at my pull request? :grinning:

arfon:
> @arfon Would you mind taking a look at my pull request?

:+1:  Done

PR #2648, opened 2015-09-24
Added .pck PLSQL file extension

jaredpetersen:
Added the .pck file extension to PLSQL. Fixes issue #2645.

In-the-wild usage of the file extension:
https://github.com/search?l=&q=create+extension%3Apck&ref=advsearch&type=Code&utf8=%E2%9C%93

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2456,6 +2456,7 @@ PLSQL:
   color: "#dad8d8"
   extensions:
   - .pls
+  - .pck
   - .pkb
   - .pks
   - .plb
--- a/samples/PLSQL/plsqlguide.pck
+++ b/samples/PLSQL/plsqlguide.pck
@@ -1,0 +1,90 @@
+create or replace package plsqlguide is
+
+-- Author  : Jared Petersen
+-- Created : 9/22/2015 12:26:22 AM
+-- Purpose : Basic PLSQL template/guide
+
+/* Procedures */
+procedure p_main;
+
+end plsqlguide;
+/
+create or replace package body plsqlguide is
+
+/* Main entry point (homepage) */
+procedure p_main
+	is
+begin
+
+htp.prn('
+	<!DOCTYPE html>
+		<html lang="en">
+		<head>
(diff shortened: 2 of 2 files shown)


jaredpetersen:
The existing file extensions were not completely sorted alphabetically, which wasn't mentioned as a requirement in the contributing guidelines. I fixed the issue in my second commit. Will it be necessary for me to submit another pull request with all of the changes in a single commit?

larsbrinkhoff:
You can rebase your pull request against Linguist's master branch.  This way, you can correct errors, merge or split commits, and add more changes as you see fit.

jaredpetersen:
@larsbrinkhoff Thanks for your help!

larsbrinkhoff:
You're welcome!

One thing though: The first position in `extensions` is special.  It's considered the "primary file extension".  (I'm not sure exactly what that means.)  So `.pls` should probably stay where it is.

jaredpetersen:
Okay, made the change and rebased again! That information really should be publicly documented.

jaredpetersen:
@larsbrinkhoff When do you think this can be reviewed and possibly merged into the repository?

larsbrinkhoff:
@jaredpetersen Sorry, I'm not a GitHub staff member, so I don't have any say in the matter.  It looks good to me, though.

arfon:
:+1: thanks folks. @jaredpetersen this will be live later this week when we cut a new Linguist gem.

jaredpetersen:
Is this live yet? Linguist still isn't picking up on the language of my [PL/SQL repository](https://github.com/jaredpetersen/plsql-sample-webapp). The code is the same as the one used for the sample.

arfon:
> Linguist still isn't picking up on the language of my PL/SQL repository. 

Linguist only updates languages when there are changes to the repo. I've just manually re-calculated the stats and they're showing up properly now.

jaredpetersen:
> Linguist only updates languages when there are changes to the repo. I've just manually re-calculated the stats and they're showing up properly now.

Huh, interesting. That makes sense. Thanks for the help!

PR #2655, opened 2015-09-25
Detect Pickle data dumps (.p, .pkl, .pickle)

iblech:
Some repositories, especially in the numerical/scientific Python community, include data dumps serialized with Pickle. Currently those are misclassified as OpenEdge APL. This pull request adds Pickle to the language list (with `type: data`), along with several samples. I hope, but wasn't able to test, that the Bayesian classifier will learn the difference to OpenEdge APL.

Examples (includes only Pickle files containing numpy arrays):
https://github.com/search?utf8=%E2%9C%93&q=numpy+extension%3Apkl&type=Code&ref=searchresults
https://github.com/search?utf8=%E2%9C%93&q=numpy+extension%3Ap&type=Code&ref=searchresults
https://github.com/search?utf8=%E2%9C%93&q=numpy+extension%3Apickle&type=Code&ref=searchresults


larsbrinkhoff:
You may ask @arfon for a large sample of `.p` files to check that the Bayesian classifier does its job.

pchaigno:
The contribution guidelines state:
> In most cases we prefer that extensions be in use in hundreds of repositories before supporting them in Linguist.

According to your searches there are 16 `.p` and 10 `.pickle` files. I don't think that will be enough ;)

Could you also add a sample for `.pkl`?

Do you know a grammar (Sublime Text package or TextMate/Atom bundle) that could be used to highlight those files?

iblech:
Sorry, didn't notice your reply sooner.

There are 25,740 code results for `.pkl`. :-) The extensions `.p` and `.pickle` are just alternative ones, the data format is exactly the same.

Sure, I could rename some of the samles to `.pkl`.

I don't know a grammar for highlighting. Highlighting Pickle data dumps doesn't make much sense, since they are mostly semi-binary blobs. You don't read those manually, but instead load and inspect them from Python.

arfon:
> According to your searches there are 16 `.p` and 10 `.pickle` files. I don't think that will be enough

@iblech - would you mind removing these two extensions from this pull request? I think we should just stick to `.pkl` for now as it's clearly the dominant extension. Also, if you could rename the `.p` samples to `.pkl` that would be excellent too.

iblech:
> @iblech - would you mind removing these two extensions from this pull request? I think we should just stick to `.pkl` for now as it's clearly the dominant extension. Also, if you could rename the `.p` samples to `.pkl` that would be excellent too.

Sure, did so. (It seems that the Python documentation is not entirely consistent on what the recommended extension is. See https://docs.python.org/2/library/pickle.html, https://docs.python.org/3/library/pickle.html, and https://wiki.python.org/moin/UsingPickle.)

arfon:
:+1: thanks @iblech. This will be live in a couple of weeks when we cut a new Linguist gem.

iblech:
Great! Thanks for your guidance.

arfon:
> Great! Thanks for your guidance.

My pleasure. Thanks for sticking with us :smile:

PR #2656, opened 2015-09-25
Clarified that only nonprimary extensions should be sorted

iblech:
While preparing a different, unrelated pull request I was confused that the test `extensions_are_sorted` failed, since the first extension is interpreted as the primary one (and therefore should not be sorted among the other, nonprimary extensions).

Turns out that the test (correctly) only checks that the nonprimary extensions are sorted. This pull request renames the test to `nonprimary_extensions_are_sorted`, thus better documenting its purpose, and adds a short comment to `languages.yml`.


arfon:
:+1: thanks @iblech

PR #2666, opened 2015-10-03
Use all extension segments

gabrielatoney:
A filename like `index.html.fr` would previously return `.html.fr` and
`.fr` as possible extensions. This makes it so it will also return
`.html`.


gabrielatoney:
10584736_866627353376084_146812061_n
10684204_902720219784680_1317286963_n
10919643_1583814761893857_1482791735_n
11078428_813391088754632_1674118377_n
11098318_1705937096300209_1819190563_n
11098661_1431503323827943_1077605040_n

PR #2667, opened 2015-10-03
Add Ur/Web support

tokenrove:
This adds support for Ur/Web, using Gary Walborn's MIT-licensed Sublime mode.

Here's [a search that shows in-the-wild usage](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aur+fun&type=Code&ref=searchresults).

Let me know if there's anything else I should do.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3475,6 +3475,17 @@ UnrealScript:
   tm_scope: source.java
   ace_mode: java
 
+UrWeb:
+  type: programming
+  aliases:
+  - Ur/Web
+  - Ur
+  extensions:
+  - .ur
+  - .urs
+  tm_scope: source.ur
+  ace_mode: text
+
 VCL:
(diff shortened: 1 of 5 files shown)


lib/linguist/languages.yml
```diff
@@ -3475,6 +3475,17 @@ UnrealScript:
   tm_scope: source.java
   ace_mode: java
 
+Ur/Web:
```
pchaigno: This needs to be the same as the directory for the samples. So we can't use `/`.
Could it be UrWeb or maybe just Ur?

tokenrove: I was going to modify things to accommodate this but I see it would be better to go with the flow.  I'll swap the primary name and the main alias.  Sorry for the noise.

arfon:
Thanks for this @tokenrove. I've got #2758 going to replace this.

PR #2668, opened 2015-10-03
Accommodate samples whose primary name contains / by checking aliases

tokenrove:
This came up when creating #2667, but I didn't want it to be an impediment to merging that PR.  However, I decided to open this in case there was interest in supporting slashes in the primary names of languages.  It would be nice, at least presentation-wise.


tokenrove:
I think it's too early in the morning for me to be submitting pull requests.

PR #2671, opened 2015-10-04
Change Stata grammar repo

pchaigno:
Change the repository for the Stata grammar from [statatmbundle/Stata.tmbundle](https://github.com/statatmbundle/Stata.tmbundle) (fork) to [pschumm/Stata.tmbundle](https://github.com/pschumm/Stata.tmbundle) (original repo).
The two repository are identical so it probably make more sense to use the original one :-)


arfon:
:+1:

PR #2675, opened 2015-10-09
Add .phps extension for PHP

fbonzon:
**.phps** stands for "PHP Source"

Useful to show examples of PHP code. A properly-configured server will output a .phps file as is, with color-formated source code instead of the HTML that would normally be generated.

[Usage in GitHub](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aphps+NOT+nothack&type=Code)

[Source of example file](https://github.com/PHPMailer/PHPMailer/blob/master/examples/mail.phps)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2441,6 +2441,7 @@ PHP:
   - .php3
   - .php4
   - .php5
+  - .phps
   - .phpt
   filenames:
   - Phakefile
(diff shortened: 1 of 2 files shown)


pchaigno:
Nice pull request! :+1:

arfon:
:+1: thanks @fbonzon. This will be live in the next week or so when we cut a new release of Linguist.

PR #2676, opened 2015-10-09
Support for files generated by Racc.

lucky:
This changeset includes a sample racc file from [this auto-generated
file](https://github.com/tenderlove/rjson/blob/44e9bf0440508b78b2bc3376da22df11abbb5483/lib/rjson/parser.rb)
(MIT-licensed).

[Racc](https://github.com/tenderlove/racc) Racc is an LALR(1) parser
generator. It is written in Ruby itself, and generates ruby programs.


arfon:
:+1: thanks @lucky. This will be live in the next release of Linguist (later this week)

PR #2681, opened 2015-10-11
Add .brd extension for KiCad

miek:
This adds the .brd extension used by older versions of KiCad's PCB layout application PcbNew.

The extension is also used by Eagle (which has a sample) and there are currently [~4300 misclassified KiCad files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Abrd+pcbnew&type=Code&ref=searchresults).


arfon:
:+1: thanks @miek

PR #2689, opened 2015-10-16
Add FreeMarker syntax highlighting

edendramis:
FreeMarker is a very popular Java templating language. Here are some "in the wild" usages: https://github.com/search?q=%23ftl&ref=searchresults&type=Code&utf8=%E2%9C%93

FreeMarker docs: http://freemarker.org/docs/index.html

I'll be updating the TextMate bundle with the newest specs, but this should suffice for now.


lib/linguist/languages.yml
```diff
@@ -1036,6 +1036,18 @@ Frege:
   - .fr
   tm_scope: source.haskell
   ace_mode: haskell
+  
+FreeMarker:
+  type: programming
+  color: "#0050b2"
+  group: Java
+  search_term: ftl
```
pchaigno: `search_term` is now deprecated and it should be necessary here since `ftl` is already an alias.

lib/linguist/languages.yml
```diff
@@ -1036,6 +1036,18 @@ Frege:
   - .fr
   tm_scope: source.haskell
   ace_mode: haskell
+  
+FreeMarker:
+  type: programming
+  color: "#0050b2"
+  group: Java
```
pchaigno: This means that FreeMarker files will be counted as Java files.
Is that what you want?

pchaigno:
This search might be a better indication for reviewers: https://github.com/search?utf8=%E2%9C%93&q=macro+extension%3Aftl&type=Code&ref=searchresults (limited to `.ftl` files)

lib/linguist/languages.yml
edendramis: @pchaigno Oh, no I don't. It's not really HTML either since FreeMarker can be used to output anything. I'll remove the group unless you think it should e HTML like some of the other templating languages.

lib/linguist/languages.yml
edendramis: removed, thanks

edendramis:
@pchaigno I updated the link, thanks. I didn't think to use `extension:` for the search options.

arfon:
:+1: thanks @edendramis. This will be live in the next release of Linguist (hopefully later this week)

edendramis:
@arfon :grin: Thanks!

PR #2690, opened 2015-10-16
Upgrade to licensee 6

bkeepers:
This upgrades to [licensee 6.0.0b1](https://github.com/benbalter/licensee/releases/tag/v6.0.0b1). The only change appears to be that the license of the factor grammar is now properly detected.

/cc @benbalter @vmg


bkeepers:
Looks like licensee 6 is using some Ruby 2 syntax, causing the 1.9.3 build to fail.

Anyone opposed to dropping 1.9.3 support?

/cc @arfon @pchaigno

pchaigno:
> Anyone opposed to dropping 1.9.3 support?

:+1:

(This will mean dropping support in [github/markup](https://github.com/github/markup) too if we continue with github/markup#537 approach.)

arfon:
:+1: fine by me.

arfon:
@bkeepers - any idea why these commits (https://github.com/github/linguist/commit/317219e479fbf23cca232dca4d9651984aaa46aa & https://github.com/github/linguist/commit/932726863fe553ab38a40a91fdf2078a4e5c0a5a) look like they're in this branch but the 'Files Changed' tab doesn't include them?

PR #2694, opened 2015-10-20
Memory blob

arfon:
Supersedes https://github.com/github/linguist/pull/2497


arfon:
@bkeepers - I'd like to get this merged. Whaddya think?

test/helper.rb
```diff
@@ -19,6 +26,12 @@ def samples_path
 end
 
 def sample_blob(name)
-  name = File.join(samples_path, name) unless name =~ /^\//
-  Linguist::FileBlob.new(name, samples_path)
+  filepath = (name =~ /^\//)? name : File.join(samples_path, name)
+  Linguist::FileBlob.new(filepath, samples_path)
+end
```
bkeepers: Is this still used anywhere? It took me a while to parse the difference between `fixture_blob`, `fixture_blob_memory`, `sample_blob`, and `sample_blob_memory`. I get it now, but I'm not sure if there's a way to simplify that.

bkeepers:
:+1:

test/helper.rb
arfon: > Is this still used anywhere?  

Yep, a bunch. For example: https://github.com/github/linguist/blob/b3ff84872be787551556d144ee4ef133cf8b1a33/test/test_file_blob.rb#L34-L37

> It took me a while to parse the difference between

Yeah, it might be worth trying to rename these at some point.

arfon:
:+1: thanks for this @pchaigno

PR #2698, opened 2015-10-21
Change D color to match the website

phase:
The original yellow color doesn't fit D very well, and it looks an awful lot like the color for JavaScript. This changes it to the color used on the website's sidebar, a nice dark maroon: http://www.colorpicker.com/2e2324

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -720,7 +720,7 @@ Cython:
 
 D:
   type: programming
-  color: "#fcd46d"
+  color: "#2e2424"
   extensions:
   - .d
   - .di
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -720,7 +720,7 @@ Cython:
 
 D:
   type: programming
-  color: "#2e2424"
+  color: "#ba595e"
   extensions:
   - .d
   - .di


davidzchen:
That color seems very close to black. Would it be better to use the red used for the D logo?

phase:
@davidzchen Maybe something like http://www.colorpicker.com/ba595e ?

davidzchen:
Looks good to me.

arfon:
Thanks! This change will be live when we next cut a Linguist gem (another couple of weeks)

phase:
@arfon One thing I noticed in the releases is that you don't @tag the PR's creators, only link to their accounts. I think tagging them would be better, so they get notified that their contributions have been released.

arfon:
@phase - I keep meaning to modify the script I'm using to generate this summary. Thanks for the reminder!

pchaigno:
>  I keep meaning to modify the script I'm using to generate this summary.

@arfon Is it open source? I'd be interested to use it in some of my projects :-)

arfon:
@pchaigno - yeah, I use this one: https://github.com/skywinder/github-changelog-generator

phase:
https://github.com/skywinder/github-changelog-generator/blob/master/lib/github_changelog_generator/generator/generator_generation.rb#L175 seems to be the culprit.

pchaigno:
> @pchaigno - yeah, I use this one: https://github.com/skywinder/github-changelog-generator

Nice, thanks!

skywinder:
> @arfon One thing I noticed in the releases is that you don't @tag the PR's creators, only link to their accounts. I think tagging them would be better, so they get notified that their contributions have been released.

Seems, there is no notifications comes, when some username mentioned in project files.

PR #2699, opened 2015-10-22
Fix typo

IonicaBizau:
"Information" instead of "informations" :dizzy:


arfon:
Thanks for this pull request - as Linguist incorporates sample code from other repositories this is actually a typo somewhere upstream (although I'm not sure exactly where sorry). 

Regardless I'm not sure it makes sense to be fixing typos in these samples.

IonicaBizau:
Totally agree! I just wasn't sure if this is (not) written by you. :+1:

Issue #2695, opened 2015-10-20
Django Template highlighting sees " in comment as error.


morganwahl:
I have a file that seems to confuse the highlighting for Django Templates. It's highlighting everything after a double-quote with a red background, as if it's an error. See:

https://gist.github.com/morganwahl/c93fe2e29f9b806a45e5

morganwahl:
Oddly, the gist site seems to just highlight the files as if they were HTML, regardless of modeline. However, on the main github site I see the same every-red behavior even when the template code is being highlighted due to the modelines.

pchaigno:
> Oddly, the gist site seems to just highlight the files as if they were HTML, regardless of modeline.

I don't think Gists support modeline overrides (/cc @arfon).
The language is determined based on your file extension. You could use [`.mustache`](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1323-L1334) to ensure that your snippet is highlighted as a Django Template file.

> However, on the main github site I see the same every-red behavior even when the template code is being highlighted due to the modelines.

This is probably an issue with the grammar used for highlighting (Linguist only selects the appropriate grammar, [textmate/python-django.tmbundle](https://github.com/textmate/python-django.tmbundle) for Django Templates). Just to be sure, do you have an example of incorrectly highlighted file?

arfon:
> I don't think Gists support modeline overrides

It should do. Take a look at this for example: https://gist.github.com/arfon/1e34d84ae570efafd225

pchaigno:
Hum. The modeline is not recognized for two reasons.
- Linguist currently [only looks at the beginning of the file](https://github.com/github/linguist/blob/master/lib/linguist/strategy/modeline.rb#L26) for a modeline.
- `django` is [not defined as an alias for HTML+Django](https://github.com/github/linguist/blob/6e934067a87103b122dff03d9517940da99d9aa3/lib/linguist/languages.yml#L1331-L1333). See [these two files with different modelines](https://gist.github.com/pchaigno/97a261d8265c83295206). #2700 should fix this.

arfon:
> - Linguist currently only looks at the beginning of the file for a modeline.

I'm not sure that's the case. See [this file](https://gist.github.com/arfon/dfa843b00dac21f793f7#file-test3-html) for example.

> #2700 should fix this.

:+1: Looks good to me. Let's get it merged in.

pchaigno:
> I'm not sure that's the case. See this file for example.

Oh, right :/
[...]

PR #2700, opened 2015-10-22
Django as an alias for HTML+Django

pchaigno:
This pull request adds `django` as an alias for `HTML+Django`.
Discussed in #2695.


arfon:
Thanks @pchaigno. We just cut a new Linguist gem yesterday and so this probably won't be live for another week or two.

PR #2707, opened 2015-10-30
Add .cake extension for C#

Redth:
This adds the .cake file extension to the C# language.

Here is a search in the wild: https://github.com/search?q=extension%3Acake+NOT+coffee&type=Code

Cake (C# Make) is a cross platform build automation system with a C# DSL to do things like compiling code, copy files/folders, running unit tests, compress files and build NuGet packages.

You can find out out more about cake here: http://cakebuild.net/

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -392,6 +392,7 @@ C#:
   aliases:
   - csharp
   extensions:
+  - .cake
   - .cs
   - .cshtml
   - .csx
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -392,8 +392,8 @@ C#:
   aliases:
   - csharp
   extensions:
-  - .cake
   - .cs
+  - .cake
   - .cshtml
   - .csx
 
--- a/samples/C#/build.cake
+++ b/samples/C#/build.cake
@@ -1,0 +1,86 @@
+///////////////////////////////////////////////////////////////////////////////
+// ARGUMENTS
+///////////////////////////////////////////////////////////////////////////////
+
+var target = Argument<string>("target", "Default");
+var configuration = Argument<string>("configuration", "Release");
+
+///////////////////////////////////////////////////////////////////////////////
+// GLOBAL VARIABLES
+///////////////////////////////////////////////////////////////////////////////
+
+var solutions = GetFiles("./**/*.sln");
+var solutionPaths = solutions.Select(solution => solution.GetDirectory());
+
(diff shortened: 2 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -392,6 +392,7 @@ C#:
   aliases:
   - csharp
   extensions:
+  - .cake
```
pchaigno: The first extension should be the primary one, `.cs` ;)

pchaigno:
Using your search query, I found that there are 116 projects that use Cake files dispatched among 61 users. Not sure that it will be enough to make it a language in Linguist.

lib/linguist/languages.yml
Redth: Thanks, I had assumed entirely alphabetical, fixed now :)

Redth:
@pchaigno if it's any help, support for cake was merged into Atom already https://github.com/atom/language-csharp/pull/30

pchaigno:
> support for cake was merged into Atom already

Does that mean that Cake doesn't have the same syntax as C#?

Redth:
It was added with the same syntax as C#. Cake is in fact C# (cake scripts are run under the roslyn script host engine - mono script host on mac/unix).  Cake does contain a few more possible preprocessor directives but the syntax highlighting is the same as normal C#.

pchaigno:
I see, it's a second grammar file referring the C# one. We'll have to declare Cake as a second file with C# as the parent language (meaning Cake will be displayed as C# but we'll be able to use a different grammar file for it).

First we should make sure that 116 projects with Cake files is enough... @arfon?

Redth:
Hmm, I didn't realize that this already exists: `/vendor/grammars/language-csharp/source.cake`

Does that make things easier?

pchaigno:
> Does that make things easier?

Yes, we won't have to add the grammar but we'll still have to declare a separate language.

> EDIT: Looks like @arfon already added this? 0a6165c

That's just because he updated the grammar submodule ;)

gep13:
@pchaigno @arfon  I was just wondering what the status of this is?  After yesterday's announcements, Cake is now a first class citizen in both the VS Code and Visual Studio Team Services (formerly VSO) extension galleries:

https://marketplace.visualstudio.com/items/cake-build.cake-vscode
https://marketplace.visualstudio.com/items/cake-build.cake

Just wondering if this helps the case for getting this support included.

arfon:
Thanks @Redth - I've got an updated version of this Pull Request going in #2733

PR #2709, opened 2015-10-31
Mainly fixing problems with Perl heuristics

JJ:
And also adding a little bit of text to the README file to help with local use and test.

It was returning incorrectly some files belonging to Perl6. It's fixed now.


lib/linguist/heuristics.rb
```diff
@@ -274,7 +274,7 @@ def call(data)
     end
 
     disambiguate ".pl" do |data|
-      if /^(use v6|(my )?class|module)/.match(data)
+      if /^(use v6|(my)? class|module)/.match(data) #class might be a variable name in Perl 5
```
pchaigno: Could you add a test for this? I don't really understand the reasoning behind this :/

pchaigno:
`CONTRIBUTING.md` might be a better place for the new guidelines. What do you think?

lib/linguist/heuristics.rb
JJ: 2015-11-01 11:25 GMT+01:00 Paul Chaignon <notifications@github.com>:

> In lib/linguist/heuristics.rb
> <https://github.com/github/linguist/pull/2709#discussion_r43581107>:
>
> > @@ -274,7 +274,7 @@ def call(data)
> >      end
> >
> >      disambiguate ".pl" do |data|
> > -      if /^(use v6|(my )?class|module)/.match(data)
> > +      if /^(use v6|(my)? class|module)/.match(data) #class might be a variable name in Perl 5
>
> Could you add a test for this? I don't really understand the reasoning
> behind this :/
>

When you have a variable called $class it gives a false positive. Yep, I
guess I can.

JJ:
2015-11-01 11:27 GMT+01:00 Paul Chaignon <notifications@github.com>:

> CONTRIBUTING.md might be a better place for the new guidelines. What do
> you think?
>

Whatever. Should I move it there?

pchaigno:
Not sure, let's ask @arfon :-)

JJ:
This new commit adds `Any.pm` which was incorrectly classified here https://github.com/JJ/Algorithm-Evolutionary/blob/434ae5dd785da05c38b135ff7edebd7a3f6e7d8e/lib/Algorithm/Evolutionary/Individual/Any.pm as perl6 because it included a `$class` variable (same as the other files with the same problem). Tested now with this new version:

```
$ bundle exec linguist ../../CPAN/Algorithm-Evolutionary
99.33%  Perl
0.57%   R
0.10%   Shell
```
It is correctly classified.

arfon:
> Whatever. Should I move it there?

Yes, thanks for adding these steps. I'd rather this was in contributing.md - the README is mainly aimed at people troubleshooting issues.

lib/linguist/heuristics.rb
pchaigno: Wouldn't the variable start with `$` in that case? This regular expression checks that `class` it at the very beginning of the line.

JJ: Well, it issues a false positive indeed, that's what started it all. The problem is I guess in the optional my, which maybe drops the ^.

pchaigno: > The problem is I guess in the optional my, which maybe drops the ^.

I'm not sure I follow you... What do you mean by that?

> Please check the file I've added to the test suite.

Which one is generating a false positive?

JJ: > I'm not sure I follow you... What do you mean by that?

First, Any.pm which I have added in the test suite indeed is recognized as Perl6, which is it not. Since it does not have module or v6, it must be because line 49 fires that regular expression. Please note that ^ might not be working as intended, since *anyway* `class` or `module` or `use v6` might not be at the very beginning of a line, but indented by whitespace. So my hunch is that the beginning of a line ^ is anchored to the optional `my`. If that is not pressent, it fires whenever it finds `class` anywhere in a line (which indeed it does). 
Anyway, experience is the mother of science. Please check the file I provide (`Any.pm`) with the former and current regex. If it checks out as perl6, then it's wrong. If it's perl5, then it's correct! :-)

pchaigno: > First, Any.pm which I have added in the test suite indeed is recognized as Perl6, which is it not. Since it does not have module or v6, it must be because line 49 fires that regular expression.

I ran the last released version of Linguist (`v4.7.3`) against `Any.pm` and it detected it as Perl5:
```
$ linguist Any.pm 
Any.pm: 101 lines (66 sloc)
  type:      Text
  mime type: application/x-pagemaker
  language:  Perl
```
Maybe the repository where that file is hasn't been updated in a long time...
<br>

> Please note that ^ might not be working as intended, since anyway class or module or use v6 might not be at the very beginning of a line, but indented by whitespace.

Good point. We could probably add something like `\s*` before `(my )?`.
<br>

> If that is not pressent, it fires whenever it finds class anywhere in a line (which indeed it does).

The four possible matching lines with `^(use v6|(my )?class|module)` are `^use v6`, `^my class`, `^class` and `^module`. If `class` is not at the beginning of the line it won't ever match.
<br>

> Anyway, experience is the mother of science. Please check the file I provide (Any.pm) with the former and current regex. If it checks out as perl6, then it's wrong. If it's perl5, then it's correct! :-)

I tried with https://regex101.com/ and it's not matching the Perl6 regular expression. It is, however, matching the Perl5 one.

JJ:
I have updated from upstream and added new tests. It works on my perl and perl6 samples, and I have added a new sample to the perl6 set.

JJ:
I have updated a sample file position and also the regex for detecting class.

lib/linguist/heuristics.rb
JJ: I can retire this small modification if you want. Still, there are other files, samples and tests modified in this pull request; or I can close it and create a new one if needed.

pchaigno: What do you think of changing the regex to `^\s*(use v6|(my )?class|module)`?

JJ: 2015-12-13 11:57 GMT+01:00 Paul Chaignon <notifications@github.com>:

> In lib/linguist/heuristics.rb
> <https://github.com/github/linguist/pull/2709#discussion_r47443025>:
>
> > @@ -274,7 +274,7 @@ def call(data)
> >      end
> >
> >      disambiguate ".pl" do |data|
> > -      if /^(use v6|(my )?class|module)/.match(data)
> > +      if /^(use v6|(my)? class|module)/.match(data) #class might be a variable name in Perl 5
>
> What do you think of changing the regex to ^\s*(use v6|(my )?class|module)
> ?
>

I can check it, if you want. There are also some other changes
(Makefile.PL) and stuff that are being blocked because of this single
thing. I think that since Perl6 is about to be released any time now, it's
important to get it right.

JJ:
I have tested this on a bunch of CPAN modules I maintain, and it still issues some false perl6 positives (mainly on .t files, I can't fathom why) but it does not fail in any of the modules it was failing so far. Thanks for the feedback.

pchaigno:
> I have tested this on a bunch of CPAN modules I maintain, and it still issues some false perl6 positives (mainly on .t files, I can't fathom why)

I'm afraid that we won't ever reach a 100% accuracy but if you want to open an issue with a few false positive examples, I could take a look into it, see if there's something else we can improve. 
Otherwise, there is always the [overrides](https://github.com/github/linguist#overrides).

> Thanks for the feedback.

Thank you for your patience :+1:

JJ:
Hi

2015-12-13 12:37 GMT+01:00 Paul Chaignon <notifications@github.com>:

> I have tested this on a bunch of CPAN modules I maintain, and it still
> issues some false perl6 positives (mainly on .t files, I can't fathom why)
>
> I'm afraid that we won't ever reach a 100% accuracy but if you want to
> open an issue with a few false positive examples, I could take a look into
> it, see if there's something else we can improve.
> Otherwise, there is always the overrides
> <https://github.com/github/linguist#overrides>.
>

Yep, I know, but the thing is that I can't figure out the heuristics it's
applying in that case. A --lingsplainer option would help, but that, I
guess, is another completely different issue.

JJ:
So, not accepted? Did I miss something?

arfon:
> So, not accepted? Did I miss something?

Sorry, wrong button. This is stale because I need to do some work on this :-)

JJ:
OK, if you need some help, please let me know...

CONTRIBUTING.md
```diff
@@ -4,7 +4,42 @@
...
+The majority of contributions won't need to touch any Ruby code at
+all.
+
+## Installation how to
+
+You might want to get Linguist running in order to test locally whatever
+changes you might be doing to it. You will need to install some
+prerrequisite libraries first. Run:
+
+	sudo apt-get install cmake
```
aaronpuchert: Not everybody is running a Debian-based Linux. Others will have `yum`, `pacman`, `zypper`, `brew`, or no package manager at all.

JJ: Good. I'll try to find out how others work.

aaronpuchert: I think it's Ok if you just say "install CMake", people will hopefully know how that works in their OS.

JJ: OK.

brandonblack:
There are some good changes in here and the contribution is very much appreciated. However, this unfortunately sat for so long that some of the information is out-dated (eg. `script/bootstrap` is now a good solution for local dev setup) and I'd prefer to to deal with the perl heuristics changes in a separate PR.

I'm going to close this, but if you feel like tackling the perl changes separately in a follow-up PR I'd be in favor of that.

PR #2714, opened 2015-11-04
Remove mentions of treatment of vendored files in diffs

aroben:
Vendored files are treated just like any other files when diffing on github.com.

/cc @github/linguist


arfon:
:+1: thanks.

PR #2715, opened 2015-11-05
demo of comment symbol in APL

alexcweiner:


bkeepers:
:wave: @alexcweiner 

From the [test failure](https://travis-ci.org/github/linguist/builds/89366565):

```
APL has a sample with an interpreter (apl) that isn't explicitly defined in languages.yml
```

This is failing because the shebang in the sample uses `apl` as an interpreter, but that isn't [explicitly defined](https://github.com/alexcweiner/linguist/blob/8fc5d02cb2182d676d66095dd57b8daf0fffafab/lib/linguist/languages.yml#L72-L79) as an interpreter for APL.

Also, can you explain what changes you're hoping to see in how Linguist classifies APL by adding this sample?

arfon:

alexcweiner:
I'm hoping to see the lamp symbol parsed and counted as a comment line,
like the pound symbol

arfon:
@alexcweiner - the comment lines appear to be correctly recognized [here](https://github.com/alexcweiner/linguist/blob/patch-1/samples/APL/Comments.apl)

screen shot 2015-12-15 at 7 27 58 am

If your comment is about the reported `sloc` stats, Linguist only ignores blank lines when calculating this:
https://github.com/github/linguist/blob/master/lib/linguist/blob_helper.rb#L306-L308

arfon:

PR #2716, opened 2015-11-05
Use negative lookbehind when tokenizing string literals

aroben:
This can double the speed of tokenizing large RTF files that use `\'hh` escape sequences.

/cc @josh @vmg @github/linguist

### DIFF
--- a/lib/linguist/tokenizer.rb
+++ b/lib/linguist/tokenizer.rb
@@ -86,13 +86,13 @@ module Linguist
           if s.peek(1) == "\""
             s.getch
           else
-            s.skip_until(/[^\\]"/)
+            s.skip_until(/(?<!\\)"/)
           end
         elsif s.scan(/'/)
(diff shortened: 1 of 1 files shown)


vmg:
Nice profiling @aroben. :+1:

aroben:
Travis seems unable to install `byebug`. Not sure what that's about.

arfon:
> Travis seems unable to install `byebug`. Not sure what that's about.

It's only on Ruby 2.0.0 that we're getting this error. I've just merged a PR with the same issue, happy to do the same here.

aroben:
@arfon If you're comfortable, feel free to merge this whenever. Thanks!

arfon:
@aroben - I'll cut a new Linguist gem early next week (unless this is more urgent?)

aroben:
Nope, there's no particular rush. Thanks!

PR #2718, opened 2015-11-08
Fetch libicu via Travis, switch back to container-based

meatballhat:
This is a test.  I don't understand the details of how the `bundle config
build.charlock_holmes` bit works/fails, so the intent here is to get a PR posted
for purposes of getting feedback via the full Travis machinery.


meatballhat:
Appears to work (???) :smiley_cat:

pchaigno:
Nice :+1:

arfon:
Awesome. Thanks @meatballhat!

meatballhat:
@arfon :heart_eyes_cat: :metal: thanks for the merge!

PR #2720, opened 2015-11-09
add .jsonldf extension for JSON-LD (frames)

neothemachine:
see http://json-ld.org/spec/latest/json-ld-framing/#iana-considerations

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1588,6 +1588,7 @@ JSONLD:
   ace_mode: javascript
   extensions:
   - .jsonld
+  - .jsonldf
   tm_scope: source.js
 
 JSONiq:


pchaigno:
You seem to be the [only user of `.jsonld` files on GitHub](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ajsonldf+NOT+djfjdkjfhjd); it will probably not be enough since the requirement is usually ["hundreds of repositories"](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language).

neothemachine:
Haha, nice query! Ok, didn't know about the requirement. Let's close this and I'll come back to it once the draft standard I linked to becomes a real W3C rec.

PR #2728, opened 2015-11-18
Add new .eam.fs sample for Formatted.

larsbrinkhoff:
Apparently, there are many `.fs` files mostly filled with numbers.  Currently, they are detected as Forth.

https://github.com/lammps/lammps/search?q=extension%3Afs (plenty of [copies](https://github.com/search?q=%22LAMMPS+software%22+in%3Areadme))
https://github.com/lmhale99/iprPy/search?q=extension%3Afs

This adds a sample to the Formatted language, which corrects the language detection.


pchaigno:
Most of these files use `.eam.fs`. Could this segmented extension be used instead of just `.fs`? It would ensure that we don't get false positives...

larsbrinkhoff:
Good observation!  Updated accordingly.

arfon:
:+1: thanks @larsbrinkhoff

PR #2729, opened 2015-11-19
Recognise END statement in FORTRAN files.

larsbrinkhoff:
This adds the `END` statement to the Fortran/Forth heuristic.

Also allows Fortran comments `CCC` as seen here: https://github.com/search?q=language%3Afortran+ccc&type=Code

Fixes bad detection of these files:
https://github.com/jeiros/FORTRAN_course/search?l=forth


arfon:
:+1:

Issue #2724, opened 2015-11-13
Detect generated files of any language


cespare:
Hi,

I propose to add a new generated file check by extending the [generated go check](https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L238-L243) to apply to all languages. (This check looks for the text `"Code generated by"` in the first line of the file.)

This is useful for us because we're generating protobuf code for non-standard languages and we'd like to have diff suppression. Having the `generated_go?` check has been handy for our generated Go code and it'd be great to have this option in other languages as well.

If this sounds okay, I'll send along a PR. Thanks for your consideration!

cespare:
Friendly ping :)

deathbeam:
+1

arfon:
@cespare - this sounds like a useful addition. Please feel free to open a Pull Request implementing this support and continue the discussion there :+1:

cespare:
Thanks. Let's continue discussion over on #2734.

cespare:
Reopening as my PR was closed, yet the problem remains.

arfon:
Copying in my comment from #2734:

> Ultimately I believe that a solution using .gitattributes data would be a good solution here but as outlined in #1792 and #2206 we don't get have support for reading in .gitattributes data on our application servers.

lildude:
I think we can close this now as you can now mark files as generated in your `.gitattributes` file. Whilst a generic method would be nice, it's too error prone whilst we don't have an industry-wide, cross-language format or method of declaring a file is generated. When one appears, we can always look into it then.

Until then, `*.whatever linguist-generated` in your `.gitattributes` file is your friend.

arfon:
🎉👏🏻

mimoo:
is this possible to have this in any languages? I have OCaml files that start with a similar pattern (except that comments start with `(*` instead of `//`)

lildude:
@mimoo It's certainly possible, but very likely to lead to incorrect classification if someone happens to have added a comment that matches the regex we use. As such, we implement per-language checks to reduce the chances of this. See https://github.com/github/linguist/blob/master/lib/linguist/generated.rb for all those implemented so far.

Issue #2206, opened 2015-03-06
.gitattributes linguist-vendored overrides not suppressing vendored files in diff


dougnukem:
I'm trying to ensure that certain paths are suppressed in pull-request diffs and are treated as vendored files.

I setup a `.gitattributes` file specifying those paths as `linguist-vendored` as indicated here https://github.com/github/linguist#overrides

"Vendored files are also hidden by default in diffs on github.com."

```
node_modules/* linguist-vendored
cordova/* linguist-vendored
Godeps/_workspace/* linguist-vendored

```

Then I created a pull-request that adds a file in `node_modules/` but I'm still seeing that file unsuppressed in the pull-request diff.
- https://github.com/dougnukem/linguist-test/pull/2/files

Am I doing something wrong or misunderstanding the purpose of the `.gitattributes` `linguist-vendored`?

seanhealy:
I'm also seeing this and wondering the exact same thing.

dougnukem:
What's odd is that it seems like `node_modules/*` is an explicit ignored file already so I shouldn't need to add it to the `linguist-vendored` but I'd like to be able to use the `.gitattributes` to manage per-project diff/pull-request ignored files.

arfon:
Hi @dougnukem thanks for the question. Unfortunately the diff suppression isn't customizable, it's handled by the `generated` class here: https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L53-L69

`vendored` is only really used for language classifications/statistics and not for behaviour on GitHub.com. We'd eventually like to add support for overriding behaviour like this using the `.gitattributes` data but it's more complicated than just adding a new attribute similar to [this one](https://github.com/github/linguist/blob/master/lib/linguist/lazy_blob.rb#L32-L38) as we don't actually have git data available on the front-end machines when generating these diffs.

Also, incase it isn't obvious from this comment, this statement is unfortunately incorrect and we need to update the docs:

> Vendored files are also hidden by default in diffs on github.com.

dougnukem:
Thanks for the clarification @afron.

So even without the `.gitattributes` `linguist-vendored` suppressing the diffs it looks like specifically `node_modules` is listed in `generated.rb`. So shouldn't anything in the `node_modules` be suppressed in the pull-request diff (https://github.com/dougnukem/linguist-test/pull/2/files)?

https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L57

[...]

PR #2734, opened 2015-11-21
Detect a special header that marks any file type as generated

cespare:
This is a first attempt at a solution for #2724.

This change has the downside of reading + line-splitting all data in all files, instead of just some files as before.

A better solution would just peek the first, say, 200 bytes from the file and look for the header there. But the FileBlob interface is extremely limited. Should I extend it?

Thanks.


cespare:
I accepted the CLA. How do I get the check to rerun?

pchaigno:
> I accepted the CLA. How do I get the check to rerun?

Try to close and reopen ;)

We should also check the performance impact of this change...

logiclrd:
It would be awesome if there were a way to configure this in `.gitattributes`. We have an in-house code generator that puts text at the top of every file, but that text does not read "Code generated by", and we don't really want to change it to say that because we'd have to regenerate every single project, which, across all branches of our entire codebase, amounts to tens of thousands of files.

cespare:
Friendly ping.

arfon:
Thanks for the ping @cespare. 

> This change has the downside of reading + line-splitting all data in all files, instead of just some files as before.

On reflection I'm not sure that we should proceed with this PR for this reason. I realise it's not idea but I think I'd rather us continue with the current pattern, that is, if there's a new file type we should be marking as `generated?` then let us know and we'll add support for it.

cespare:
@arfon this is not a good resolution. Like @logiclrd, we have various internal generated files that are not in widespread use. They would not justify changes to the `generated?` check. Instead, there needs to be some generic mechanism.

As I mentioned above, this solution could be fixed if the FileBlob interface allowed for peeking files. Can we discuss that? I understand that this PR is too expensive as-is -- I had hoped that opening it would be the start of some fruitful discussion.

Or, let's consider some alternative mechanism such as gitattributes proposed by @logiclrd.

arfon:
> As I mentioned above, this solution could be fixed if the FileBlob interface allowed for peeking files. Can we discuss that? I understand that this PR is too expensive as-is -- I had hoped that opening it would be the start of some fruitful discussion.

arfon:
> As I mentioned above, this solution could be fixed if the FileBlob interface allowed for peeking files. Can we discuss that? I understand that this PR is too expensive as-is -- I had hoped that opening it would be the start of some fruitful discussion.

Sorry, my bad @cespare. I'm trying to clean up a bunch of issues and PRs on Linguist right now and I realise I didn't address this point.

> Or, let's consider some alternative mechanism such as gitattributes proposed by @logiclrd.

Ultimately I believe that a solution using `.gitattributes` data would be a good solution here but as outlined in https://github.com/github/linguist/issues/1792 and https://github.com/github/linguist/issues/2206 we don't get have support for reading in `.gitattributes` data on our application servers.

PR #2735, opened 2015-11-21
New extension support for PL/SQL language

tschf:
I recently converted the Oracle text mate bundle [[1](https://github.com/mulander/oracle.tmbundle)] to an Atom grammar [[2](https://atom.io/packages/language-oracle)]. As such, I thought it would be good to also bring all the related extensions that have been introduced, into linguist, for the Oracle PL/SQL language.

Below is the list of extensions identified along with the number of results when searching that extension in GitHub with the string `create or replace` which is common in any database object.
- bdy - 732 entries [[3](https://github.com/search?p=1&q=%22create+or+replace%22+extension%3Abdy&ref=searchresults&type=Code&utf8=%E2%9C%93)]
- spc - 748 entries [[4](https://github.com/search?utf8=%E2%9C%93&q=%22create+or+replace%22+extension%3Aspc&type=Code&ref=searchresults)]
- trg - 1489 entries [[5](https://github.com/search?utf8=%E2%9C%93&q=%22create+or+replace%22+extension%3Atrg&type=Code&ref=searchresults)]
- tps - 193 entries [[6](https://github.com/search?p=19&q=%22create+or+replace%22+extension%3Atps&ref=searchresults&type=Code&utf8=%E2%9C%93)]
- vw - 1317 entries [[7](https://github.com/search?utf8=%E2%9C%93&q=%22create+or+replace%22+extension%3Avw&type=Code&ref=searchresults)]
- fnc - 654 entries [[8](https://github.com/search?utf8=%E2%9C%93&q=%22create+or+replace%22+extension%3Afnc&type=Code&ref=searchresults)]

The extensions `ddl` and `prc` both with an existing mapping to `SQL`, but I feel should also have cases where they are mapped to PL/SQL - it seems `create`, `replace` for example aren't syntax highlighted in the phrase `create or replace`. Example case - https://github.com/gordynor/Demo/blob/0ddb72599c26b935d5731cf2d5e9680f6ac73c5a/sql/test.prc
- ddl - 417 entries [[9](https://github.com/search?p=41&q=%22create+or+replace%22+extension%3Addl&ref=searchresults&type=Code&utf8=%E2%9C%93)].
- prc - 21777 entries [[10](https://github.com/search?utf8=%E2%9C%93&q=%22create+or+replace%22+extension%3Aprc&type=Code&ref=searchresults)]

I added opposing samples in the samples directory for `prc` and `ddl` which contain a `create or replace` statement. https://github.com/tschf/linguist/tree/master/samples/PLSQL. Is this sufficient?

The other extension added to the grammar was `dml`. Upon searching with phrases like `create or replace` (only 5 code files - expected to be low, since it's DML) or `insert into` (still only 99 code entries). As such left this one out.

The contributing guide states that extensions should be in use in hundreds of repositories, so I'm not sure if the above analysis means the majority of these extensions fail that requirement or not.

Also, the contributing guidelines mention about testing the performance of the Bayesian classifier. I wasn't really sure where to begin with this?


pchaigno:
> The contributing guide states that extensions should be in use in hundreds of repositories, so I'm not sure if the above analysis means the majority of these extensions fail that requirement or not.

I downloaded the first 1,000 files of each search query above and counted the number of users and repositories. `+` signs mean that there was more than 1,000 files so it's just an estimation a minima.

Extension | Number of repositories | Number of users
---|---|---
`.bdy` | 34 | 27
`.spc` | 43 | 36
`.trg` | 83+ | 73+
`.tps` | 35 | 29
`.vw` | 69+ | 61+
`.fnc` | 65 | 59
`.ddl` | 136 | 124
`.prc` | 206+ | 194+

I'm not from GitHub so I can't say which extensions will be accepted and which won't :/
The requirement in the guidelines is more of a rule of thumb though.
<br>

> Also, the contributing guidelines mention about testing the performance of the Bayesian classifier. I wasn't really sure where to begin with this?

Here are the results of the detection for the first 1,000 `.ddl` and `.prc` PL/SQL files:

Extension | SQL | PL/SQL
---|---|---
`.ddl` | **61.36%** | 38.64%
`.prc` | **3.37%** | 96.63%

(I'd like to make the same tests for SQL files but I don't really know what keyword use. Is there an SQL keyword that doesn't exist in PL/SQL?)

I don't know why the results are so different between the two extensions :/
It looks like we'll have to use a heuristic rule. I'll see if we can use the [regular expressions that already exist for `.sql`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L337-L342).

tschf:
Sorry I've been slow to respond. I suppose since there is currently not other mapping for `.ddl` it would make most sense just to combine all the regular expressions defined for `.sql`, with an `elsif` for plsql. 

With `.prc`, it should really only be containing definitions for procedures.

```plsql
create or replace procedure foo--`or replace` is optional, but more common
authid definer--optional
as
  l_bar NUMBER;
begin
  --todo
  null;
end foo;--or `end;`
```

Just looking at the documentation for other rdbms's

* plpgsql

Doesn't have the `procedure` keyword, but rather `function` (PL/SQL has function also, but a procedure and function are different). And always ends with the line: `language 'plpgsql';` Doing a search returns [13 results](https://github.com/search?utf8=%E2%9C%93&q=%22language+%27plpgsql%22+extension%3Aprc&type=Code&ref=searchresults)

It supports the `or replace` clause.

Can have the return/returns statement - typically `void` in this case

http://www.postgresql.org/docs/9.1/static/sql-createfunction.html

* SQL Server

Parameters prefixed with `@`

No `or replace` clause

`create proc` or `create procedure`

Seems quite comment to prefix with `dbo`. e.g. `CREATE PROC dbo.ClientSel` or `CREATE PROCEDURE dbo.SECONDPRC` or (wrap names with square brackets) `CREATE PROCEDURE [dbo].[AuthorsGetAll] `.

It also doesn't seem uncommon for the files to end with `go` (but not always).

Sometimes it starts with a check to see if the file exists:

```
IF EXISTS (SELECT * FROM DBO.SYSOBJECTS WHERE ID = OBJECT_ID(N'dbo.AvailableInSearchSel') AND OBJECTPROPERTY(id, N'IsProcedure') = 1)
    DROP PROCEDURE dbo.AvailableInSearchSel
GO
CREATE Procedure AvailableInSearchSel
AS
```

I also noticed there are files with this extension using the function keyword for SQL server:

"create procedure" @ extension:prc - [1500 results](https://github.com/search?p=3&q=%22create+procedure%22+%40+extension%3Aprc&ref=searchresults&type=Code&utf8=%E2%9C%93)
"create proc" @ extension:prc - [521 results](https://github.com/search?utf8=%E2%9C%93&q=%22create+proc%22+%40+extension%3Aprc&type=Code&ref=searchresults)
"create function" @ extension:prc - [188 results](https://github.com/search?utf8=%E2%9C%93&q=%22create+function%22+%40+extension%3Aprc&type=Code&ref=searchresults)

https://msdn.microsoft.com/en-AU/library/ms187926.aspx
https://msdn.microsoft.com/en-au/library/ms186755.aspx

* MySQL

Parameters optionally start with whether they are `in`, `out`, or `inout`. 
e.g. `proc_name(in paramName paramType)`. The differs from Oracle where the in/out comes after the parameter name and before the parameter type. 

Also no `or replace` argument

Usually sets some other delimiter than `;`.

Can start with the `drop if exists` clause: `drop procedure if exists ui_expire_invalid_sessions;`; This syntax is not valid for PL/SQL.

Found [162 results](https://github.com/search?p=16&q=%22create+procedure%22+delimiter+%40+extension%3Aprc&ref=searchresults&type=Code&utf8=%E2%9C%93) with the delimter keyword - the majority belonging to 1 user/repo

http://dev.mysql.com/doc/refman/5.7/en/create-procedure.html
..

So I suppose with all that, the disambiguation would be something like:

```
contains(delimiter and/or drop procedure if exists and create procedure) or --MySQL
contains(language 'plpgsql') and contains(create[ or replace] procedure)--PostgreSQL
contains - create proc[edure] or create function and/or go)--SQL Server 
```

Just some thoughts.

Is it the common for people unfamiliar with getting the tests set up in Ruby to push commits to see how the tests run with Travis (as is mentioned on docs)? I was able to run the test as mentioned on the contributing doc, but wouldn't know where to begin to test the heurisitics.

brandonblack:
Flagging this as stale. 🏴

tschf:
I was just attempting to merge from master (to resolve the "Stale" flag), but think I did something wrong as it's removed the "vendor/grammars/assembly" folder when looking at the files changed.

lildude:
Giving this a nudge. @tschf it's been a whilst since this was last touched. Do you want to try merging master again and see how things go?

tschf:
@lildude Thanks for the nudge. Merged from master, and restored the `assembly` module (by removing the folder `vendor/grammars/assembly` and re-adding with `git submodule add https://github.com/nanoant/assembly.tmbundle vendor/grammars/assembly`). It looks like the checks are passing now.

lib/linguist/languages.yml
```diff
@@ -3047,13 +3047,21 @@ PLSQL:
   tm_scope: none
   color: "#dad8d8"
   extensions:
-  - ".pls"
```
lildude: Is `.pls` really no longer considered the primary extension?  

The primary extension should be the first in the list and the only one that doesn't need to be in alphabetical order.

tschf: > The primary extension should be the first in the list..

Oh, I didn't realise this was the way it worked - thought they all needed to be sorted alphabetically.

> Is `.pls` really no longer considered the primary extension?

I also didn't realise `pls` was considered the primary. I know personally, I never use this extension. Looking up, I see its for `PL/SQL Source`, even so, I didnt think it was that common. If I do a search for `create or replace` alongside that extension, I get 1260 results. Compare for instance to pks (package spec), there are 7 089. Not saying the number of results is a clear indication of what is "primary" and what isn't - just thought it's worth pointing out there are far more instances of an extension that is not `pls`.

In fact, if I open up SQL Developer and go to save a new source file, it defaults onto SQL

image

This could just be the fact the SQL Worksheet is typically to run SQL*Plus scripts, that support sql plus directives, etc, Still I would consider these PL/SQL source files as well.

The other options do indeed include `pls`: 

image

One final comment I would make is that a series of [example scripts](http://stevenfeuersteinonplsql.blogspot.com.au/2014/06/new-link-to-my-demozip-plsql-script.html) provided by Steven Feurstein, author of the popular Oracle PL/SQL Programming book, mostly use the extension `sql` - one file has the extension `pls`.

Reviewing all the other SQL languages defined in `languages.yml` seem to also put sql at the top of the extension list.

With all that said, [orafaq](http://www.orafaq.com/wiki/File_extensions) does list `pls` under the heading PL/SQL source files, so happy to go with what you had flagged as the primary source file extension.

tschf:
Just reviewing, and I had added tps (Type spec) but seemingly left off tpb - since they are corresponding (Spec and body), I've just added that in too.

There have been a couple of contributions to the Atom Oracle grammar to add more extensions since I logged this, with a total of 26 different extensions that folks use (11 not added into this fork yet) - but may be better to leave as is, to get it in as is.

What's not added in, that is in the grammar:

- "dml"
- "idx"
- "mv"
- "pkh"
- "prg"
- "sch"
- "seq"
- "syn"
- "tab"
- "tbl"
- "typ"

lildude:
Thanks. I think it's best to leave the `.pls` as the default for the moment as there's no real reason to change it at this time.

We can probably hold off on the other extensions too, unless anyone else thinks they should be added now.

tschf:
> We can probably hold off on the other extensions too, unless anyone else thinks they should be added now.

Yeah, that sounds good.

lildude:
@pchaigno & @Alhadis & @larsbrinkhoff (sorry I've missed you on earlier pings), any thoughts or objections?

Alhadis:
No objections, LGTM. 👍

PR #2736, opened 2015-11-22
add jquery.fancybox.css to vendor list

ppwwyyxx:
fancybox has both js and css


arfon:
:+1: thanks @ppwwyyxx. This will be live later in the week when we cut a new Linguist gem.

PR #2746, opened 2015-11-26
Adding 'Jupyter Notebook' language

marceloramires:
Here's a search result on github for .ipynb, which is an extension for Jupyter Notebooks, a very popular way to code in python more visually.

Here's a github pull request showing its popularity, as requested in que contributions.md file:
https://github.com/search?utf8=%E2%9C%93&q=extension%3Aipynb+NOT+nothack&type=Code&ref=searchresults


pchaigno:
The syntax is very different from Python's. I think you should add it as a separate language (otherwise it will be highlighted incorrectly). Is there a language name or should we use `Jupyter Notebook`?

marceloramires:
The file itself looks really different because of all the metadata, but the way it's developed is by running Jupyter Notebook and coding in pure python, not by editing the file.

Look at this example: https://github.com/marceloramires/ImageProcessing/blob/master/MorphologicalOperations.ipynb

Should we add a separate language in this case also? If so, would this case also need a language grammar?

pchaigno:
> The file itself looks really different because of all the metadata, but the way it's developed is by running Jupyter Notebook and coding in pure python, not by editing the file.

In that case, we could probably recognize Jupyter Notebooks as JSON files (since they seem to be plain JSON) and mark them as generated.

> Should we add a separate language in this case also? If so, would this case also need a language grammar?

Adding a grammar is not mandatory. We could probably use the JSON grammar here. Since [Jupyter Notebooks are rendered](https://github.com/blog/1995-github-jupyter-notebooks-3), it won't really matter anyway (only for their edition on github.com).

Is your goal to take into account Jupyter Notebooks in language statistics as Python code? If so, I'm afraid it won't be possible (keep in mind that I'm not part of GitHub staff though). You could always change that default behaviour for your own repositories using [Linguist overrides](https://github.com/github/linguist#overrides) ;)

marceloramires:
The initial idea was, in fact, to recognize it as Python (as that wat you code in when you work with Jupyter Notebooks. Although that syntax argument really seems strong.

Jupyter notebooks are not recognized as anything (in statistics), and that's what I'd like to improve.

At first you suggested creating a language "Jupyter Notebook", which seemed wrong, but thinking about it now, seems to be a good option. The syntax would be, in fact, plain JSON, but that isn't either what people code in.

My suggestion: Create the "Jupyter Notebook" language in languages.yml and add JSON's grammar as you suggested (or leaving it with no grammar).

If you agree, I'll be happy to do it :)

arfon:
> My suggestion: Create the "Jupyter Notebook" language in languages.yml and add JSON's grammar as you suggested (or leaving it with no grammar).

This sounds like a good way to proceed.

marceloramires:
Thanks! Done.

For the color, I chose the orange from [Jupyter Notebook](http://jupyter.org/)'s logo.

Jupyter Notebook Logo

samples/Jupyter-Notebook/JupyterNotebook.ipynb
```diff
@@ -0,0 +1,57 @@
+{
```
pchaigno: The directory in `samples/` needs to match exactly the name for the language in `languages.yml` ;)

pchaigno:
The type `programming` seems incorrect to me. From what I've read, `.ipynb` files can contain Python code but there are not programs in themselves. Maybe the type should be `markup` or `prose`...?

Also, for the sample file, it's better if it is a "real" file from a real application. It's unlikely that there will ever be a conflict on the file extension but I think it's still good practice since Linguist uses it for tests and the classifier may use it in its learning phase.

arfon:
I think these are either `data` or `markup`.

marceloramires:
Comments addressed:

* Renamed the sample folder to "Jupyter Notebook" (no hyphen)
* Added a real working file with python code, instead of a dummy one, as a sample
* Made the language type 'markup'

lib/linguist/languages.yml
```diff
@@ -1694,6 +1694,17 @@ Julia:
   color: "#a270ba"
   ace_mode: julia
 
+ Jupyter Notebook:
+  type: markup
+  ace_mode: json
```
pchaigno: You will also need to set the `tm_scope` to `source.json` ;)

marceloramires:
Done!

lib/linguist/languages.yml
```diff
@@ -1694,6 +1694,18 @@ Julia:
   color: "#a270ba"
   ace_mode: julia
 
+ Jupyter Notebook:
```
pchaigno: There is an extra space at the beginning of this line. It might be what's causing the Travis build to fail.

marceloramires:
The test 'test_color_proximity' was failing, and pointed out that the color I chose for Jupyter Notebook was too close to an existing one (Mask language). I've darkened it a bit, and bumped into another language (Alloy).

I haven't been able to run the tests locally (I'll look better into that, and maybe improve the documentation, but maybe it's something sketchy in my environment), but after reading the test code, and seeing the needed threshold, I've rewritten the test in python and compared some colors with all the used ones until I found a clean green one.

Does not resemble the logo anymore, but that ship has sailed. :(

All tests are now passing. :)

arfon:
> After reading the test code, and seeing the needed threshold, I've rewritten the test in python and compared some colors with all the used ones until I found a green one I liked.

I'd be interested in seeing that test (even in a Gist) if you have some time. This `test_color_proximity` is a real pain for contributors like yourself - thanks for sticking with us.

arfon:
Also, after some digging around I can't really find a reason for Alloy being the colour it is. I've therefore switched Alloy to `#64C800` and left Jupyter as the darker orange. That's all happening over in #2763.

pchaigno:
Thanks @marceloramires!

marceloramires:
@arfon Here it is:
https://gist.github.com/marceloramires/83b14c03542e2bdd8c7b

I've tried using two python libraries (skimage and colormath) to compute the cie2000 delta of colors (the one used to detect difference in colors, that takes into account the aspects of human vision), but both gave me the same result, which differed from the one provided by ruby's library.

So I cheated and used ruby's one calling from python (since I was with my hands in python already). 
It's slower and uglier, but grants 100% fidelity.

I've ran it on the entire range of RGB colors (with an interval) and found a bunchload of usable colors. It seems like bad luck now that I managed to bump into two existing ones here.

Hope it helps.
image

marceloramires:
@pchaigno Thank you!

arfon:
> @arfon Here it is: https://gist.github.com/marceloramires/83b14c03542e2bdd8c7b

That's great. Thanks @marceloramires!

marceloramires:
@kivancmuslu Look at this

PR #2749, opened 2015-11-27
Add language support for MAXScript

Alhadis:
This PR adds support for MAXScript, the built-in scripting language for [3ds Max](https://en.wikipedia.org/wiki/Autodesk_3ds_Max).

Usage
-----

MAXScript isn't exactly kicking JavaScript in terms of popularity, to say the least, but there's enough in-the-wild usage to warrant the language's inclusion on GitHub:

* [extension:ms fn](https://github.com/search?q=extension%3Ams+fn) (~6,610 results)
* [extension:ms function](https://github.com/search?q=extension%3Ams+function) (~13,470 results)

[ScriptSpot](http://www.scriptspot.com/) has been a well-known and popular resource for MAXScripts since at [least 2000](http://www.scriptspot.com/3ds-max/plugins/maxscribe), although the site's maintenance appears to have declined sharply in recent years. It's probably as good a time as any to introduce formal support for MAXScript on GitHub, which may encourage more MAXScripters to upload/share their work.

Heuristic modifications
-----------------------

MAXScript already offers a `move` function, so we can't rely on the current heuristic of identifying GAS by the presence of a `move` command.

My initial approach was to check for Assembly-like statements like the following, since it'd be a syntax error in MAXScript:

```
.LGCOMMAND
	.now
	.return
```

However, because of MAXScript's fluidity, it's unreasonable to assume lengthy statements won't be broken across multiple lines:
```
	value = colossal_plugin_struct_thing \
		.calculate this mumbo:#jumbo

	-- That has the same effect as writing this:
	value = colossal_plugin_struct_thing.calculate this mumbo:#jumbo
```

In other words, simply checking `\n\s*\.[A-Za-z]` isn't good enough; we need to account for things like escaped line-breaks and string literals/comments:

```rb
data.gsub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'/, "").gsub(/\\\s*(?:--.*)?\n/, "\n")
```

This might be slight overkill, but it irons out any potential for misclassification.

Performance
-----------

I took the time to analyse the performance of different heuristics when run against a [repository of considerable size](https://github.com/Alhadis/Linguist-Heuristics-MS) (15,609 files, 1.4 GiBs). I've averaged the speed results below, but the full stats can be [viewed here](http://pastebin.com/EmLwT0xU) for anybody who's curious.

### Original/Current `.ms` Heuristic

```
/((^|\s)move?[. ])|\.(include|globa?l)\s/.match(data)

Average speed (10 runs):
	real    39.8s
	user    38.9869s
	sys     0.8023s
```

### Amended Heuristic

```
/(?<!\S)\.(include|globa?l)\s/.match(data) || 
	/(?<!\/\*)(\A|\n)\s*\.[A-Za-z]/.match(data.gsub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'|\\\s*(?:--.*)?\n/, ""))

Average speed (10 runs):
	real    41.3077s
	user    40.3206s
	sys     0.9744s
```

### Earlier variations:
These are potentially slightly faster, but offer a little less accuracy:

```
/(?<!\S)\.(include|globa?l)\s/.match(data) ||
	/(?<!\/\*)(\A|\n)\s*\.[A-Za-z]/.match(data.sub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'/, ""))

Average speed (10 runs):
	real    40.0s
	user    39.1899s
	sys     0.7947s
```

```
/(?<!\S)\.(include|globa?l)\s/.match(data) || /(?<!\/\*)(\A|\n)\s*\.[A-Za-z]/.match(data)

Average speed (10 runs):
	real    38.9172s
	user    38.1532s
	sys     0.7561s
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -689,3 +689,6 @@
 [submodule "vendor/grammars/FreeMarker.tmbundle"]
 	path = vendor/grammars/FreeMarker.tmbundle
 	url = https://github.com/freemarker/FreeMarker.tmbundle
+[submodule "vendor/grammars/language-maxscript"]
+	path = vendor/grammars/language-maxscript
+	url = https://github.com/Alhadis/language-maxscript
--- a/grammars.yml
+++ b/grammars.yml
@@ -345,6 +345,8 @@ vendor/grammars/language-javascript:
 vendor/grammars/language-jsoniq/:
 - source.jq
 - source.xq
+vendor/grammars/language-maxscript:
+- source.maxscript
 vendor/grammars/language-ncl:
 - source.ncl
 vendor/grammars/language-python:
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -238,8 +238,10 @@ module Linguist
     disambiguate ".ms" do |data|
       if /^[.'][a-z][a-z](\s|$)/i.match(data)
         Language["Groff"]
-      elsif /((^|\s)move?[. ])|\.(include|globa?l)\s/.match(data)
+      elsif /(?<!\S)\.(include|globa?l)\s/.match(data) || /(?<!\/\*)(\A|\n)\s*\.[A-Za-z]/.match(data.gsub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'|\\\s*(?:--.*)?\n/, ""))
         Language["GAS"]
+      else
+        Language["MAXScript"]
       end
     end
 
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1956,6 +1956,15 @@ M:
   tm_scope: source.lisp
   ace_mode: lisp
 
+MAXScript:
+  type: programming
+  color: "#00aaaa"
+  extensions:
+  - .ms
+  - .mcr
+  tm_scope: source.maxscript
+  ace_mode: text
+
 MTML:
   type: markup
   color: "#b7e1f4"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1958,7 +1958,7 @@ M:
 
 MAXScript:
   type: programming
-  color: "#00aaaa"
+  color: "#00a6a6"
   extensions:
   - .ms
   - .mcr
--- a/samples/MAXScript/macro-1.mcr
+++ b/samples/MAXScript/macro-1.mcr
@@ -1,0 +1,29 @@
+-- Taken from an example from Autodesk's MAXScript reference:
+-- http://help.autodesk.com/view/3DSMAX/2016/ENU/?guid=__files_GUID_84E24969_C175_4389_B9A6_3B2699B66785_htm
+
+macroscript MoveToSurface
+    category: "HowTo"
+(
+    fn g_filter o = superclassof o == Geometryclass
+    fn find_intersection z_node node_to_z = (
+        local testRay = ray node_to_z.pos [0,0,-1]
+        local nodeMaxZ = z_node.max.z
+        testRay.pos.z = nodeMaxZ + 0.0001 * abs nodeMaxZ
+        intersectRay z_node testRay
+    )
+    
+    on isEnabled return selection.count > 0
+    
+    on Execute do (
+        target_mesh = pickObject message:"Pick Target Surface:" filter:g_filter
+        
+        if isValidNode target_mesh then (
+            undo "MoveToSurface" on (
+                for i in selection do (
+                    int_point = find_intersection target_mesh i
+                    if int_point != undefined then i.pos = int_point.pos
+                )--end i loop
+            )--end undo
+        )--end if
+    )--end execute
+)--end script
--- a/samples/MAXScript/macro-2.mcr
+++ b/samples/MAXScript/macro-2.mcr
@@ -1,0 +1,53 @@
+-- Taken from an example from Autodesk's MAXScript reference:
+-- http://help.autodesk.com/view/3DSMAX/2016/ENU/?guid=__files_GUID_0876DF46_FAA3_4131_838D_5739A67FF2C1_htm
+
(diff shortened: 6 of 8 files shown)


larsbrinkhoff:
It would be nice if your updated `.ms` heuristics added some matching function for MAXScript, rather than a catch-all `else`.

Alhadis:
Aye, but as I explained in the [notes for the last commit](https://github.com/Alhadis/linguist/commit/06c049b8c0533add03d2412616bef316da770d8e), real world usage of MAXScript often leaves very few distinguishing traits that can be reliably used to determine the language.

In either case, I did note your preferred method for writing heuristics (as explained at the bottom of `heuristics.rb`) but I consciously chose a catch-all approach for the reasons stated in the link above.

Alhadis:
I should also mention that MAXScript runs in global scope, meaning what gets evaluated in one script is accessible to another running after it.

So, one file might look like this:
```
fn doStuff keyword:#value = (
	
)
```

... with another one looking like this:
```
doStuff 20
```

So, yeah, that's one of the reasons I took an all-inclusive approach.

pchaigno:
Impressive work!

> I should also mention that MAXScript runs in global scope, meaning what gets evaluated in one script is accessible to another running after it.

But are you sure that's not the case for GAS and Groff too?
Because the heuristic rules are rarely perfect, catch-all approaches are dangerous. For those cases where none of the heuristic match, relying on the classifier (even if it's not very good) is generally preferred.

It looks like MaxScript has a number of pretty specific constructs and keywords (`--end`, `macroscript`, `local`, `#(`, `^fn [\w\s]*=`). Couldn't we use a set of those?

Alhadis:
> But are you sure that's not the case for GAS and Groff too?

Groff is being evaluated first, and it's being evaluated based on standalone lines like `.TR`. Those are the commands which make up the language, and they'd be invalid declarations in MAXScript.

> Because the heuristic rules are rarely perfect, catch-all approaches are dangerous.

Well, when I ran it against the [repo](https://github.com/Alhadis/Linguist-Heuristics-MS) I mentioned, it came out with 100% accuracy for each language. These files were collected from public search results for the `.ms` file extension.

After removing duplicates (detected based on [checksum](https://github.com/Alhadis/Linguist-Heuristics-MS/blob/master/Makefile#L44)), I sorted and identified each file's language manually, assigning them to a numbered folder (1: Groff, 2: GAS, 3: MAXScript). There *was* some misclassification at first, but I refined the pattern until the results were airtight.

There were some `.ms` files that weren't GAS, Groff *or* MAXScript, but those were being misclassified anyway. I purposefully excluded them from evaluation in the aforementioned test repo; you can see [them here](https://github.com/Alhadis/Linguist-Heuristics-MS/tree/master/fixtures/public/-1).

> It looks like MaxScript has a number of pretty specific constructs and keywords

Actually, I'm afraid it doesn't:

* `--end`: `--` is the beginning of a single-line comment in MAXScript (analogous to `//` in many other languages). The `--end` is just some coder's way of identifying the ending of a particular scope, possibly for readability. I grabbed the sample files from Autodesk's tutorials, which may explain why.

* `macroscript`: Typically, MacroScripts are defined with a different file extension (`.mcr). They're rarely 
defined in `.ms` files. When 3ds Max evaluates a MacroScript clause, it generates an `.mcr` file in the user's configuration directory.

* `local`: Local (as well as global) keywords are optional modifiers that're often included more for clarity than for programmatic reasons. You can read more about the distinction [here](http://docs.autodesk.com/3DSMAX/16/ENU/MAXScript-Help/index.html?url=files/GUID-A6E95530-F419-4837-B10A-E881A62AD5F9.htm,topicNumber=d30e68307), but I won't go too far into specifics.

* `#(`: That defines an array, and not every MAXScript will have/need/use those.

* `fn [\w\s]*=)`: That's a pattern matching a `fn` expression, which may also be written `function [\w\s]*`. `fn` is just provided as a convenient shorthand (which I happen to miss when writing JavaScript).

Assembly language seems pretty well-defined based on statements that begin with `.something`, which is being checked for first *before* MAXScript. I also wrote it so any GAS files which contain `.include` or `.global` at the start of a new line are identified early as Assembly. If not, it moves on to the heuristic modifications I made/mentioned earlier - checking for statements prepended with a full-stop, as [seen here](https://github.com/Alhadis/Linguist-Heuristics-MS/blob/master/fixtures/public/2/NE.1.ms).

If it can't find any statements like those, it's quite reasonable to assume it isn't an Assembly file, or a [GT]N]roff file. And if it isn't MAXScript either, well, it was being misclassified to begin with, and the owner can use `.gitattributes` if they're unhappy with the misclassification. =)

Alhadis:
@pchaigno : The heuristics are limited to affecting only the respective file extensions, correct? So having a catch-all approach for MAXScript in an `.ms` heuristic isn't going to skew Linguist's perception of Groff or GAS files with different extensions... right?

larsbrinkhoff:
We should clarify that neither @pchaigno nor I are part of GitHub staff, so we can't make any decisions on this.

Alhadis:
Just out of honest curiosity, who exactly gives the final go-ahead for merging the PR? Who/what are we waiting on?

Apologies if I come across impatient, I'm just uncertain as to what's happening, that's all...

larsbrinkhoff:
Usually, it's @arfon.

arfon:
Firstly @Alhadis I wanted to say thanks for submitting a really great Pull Request - I wish they were all as thorough as this :+1: 

As @pchaigno and @larsbrinkhoff mentioned, we try if possible to avoid adding 'catch all' statements to our heuristics but in this case it seems l like constructing an reliable heuristic for MAXScript is unfeasible.

I think we should get this merged in as-is.

arfon:
@Alhadis - this will be live in the next release of Linguist (later this week).

Alhadis:
Perfect, thank you so much! :D It's been an honour to contribute on this scale. Seriously.

PR #2766, opened 2015-12-10
Catch `ENOENT` in `clear_cache`

piki:
There's no need for `clear_cache` to fail if the cache doesn't exist,
either because we call `clear_cache` twice or because no cache was
previously written on this particular repo.

/cc @vmg, who wrote `bin/git-linguist`


vmg:
Ugh yes, this was in my TODO list. Sorry!

PR #2772, opened 2015-12-16
Remove copypasta artifact.

ibnfirnas:


pchaigno:
While you're at it, if you have a sample file from a real application it would be better ;)

samples/Mathematica/TestString.mt
```diff
@@ -1 +1 @@
-Test[1"a" <> "b", "ab", TestID -> "Concat \"a\" and \"b\""]
+Test["a" <> "b", "ab", TestID -> "Concat \"a\" and \"b\""]
```
mickey00zz: บ้า

PR #2775, opened 2015-12-18
minor code improvements

humzashah:
Improved an instance of `if/else` conditions and removed usage of `attr` as a variable.


humzashah:
I shall re-open this after fixing a typo and agreeing to the Github CLA.

PR #2776, opened 2015-12-18
minor code improvements

humzashah:
Improved an instance of `if/else` conditions and removed usage of `attr` as a variable.


arfon:
:+1: thanks for keeping us on our toes :smile_cat:

PR #2778, opened 2015-12-20
New extensions for consistency with Markup

pchaigno:
This pull request adds the `.wiki` file extension to Mediawiki.
There are [around 373,000 `.wiki` files on GitHub](https://github.com/search?utf8=%E2%9C%93&q=extension%3Awiki+NOT+dkfjkfdjkd).

These files are [already recognized and rendered](https://github.com/github/markup#markups) as Mediawiki files by github/markup. This change would make Linguist consistent with Markup (in order to [use Linguist as the detection engine for Markup](https://github.com/github/markup/pull/537) in a second time).

EDIT: I added `.rst.txt` and `.rest.txt` for reStructuredText since they follow the same logic, ie. they're already recognized by github/markup.

### DIFF
--- a/HACKING.rst.txt
+++ b/HACKING.rst.txt
@@ -1,0 +1,329 @@
+Contributing to SciPy
+=====================
+
+This document aims to give an overview of how to contribute to SciPy.  It
+tries to answer commonly asked questions, and provide some insight into how the
+community process works in practice.  Readers who are familiar with the SciPy
+community and are experienced Python coders may want to jump straight to the
+`git workflow`_ documentation.
+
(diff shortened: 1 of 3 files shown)


HACKING.rst.txt
```diff
@@ -0,0 +1,329 @@
+Contributing to SciPy
```
arfon: @pchaigno - this file should have gone into `samples` right?

pchaigno: #2844 fixes it - don't know how I missed it :flushed:

PR #2782, opened 2015-12-24
Reorder Perl heuristic rules

### LINKED ISSUE #2781: Perl5 detected as Perl6
https://github.com/drzowie/Alien-Gnuplot/search?l=perl6

File is https://github.com/drzowie/Alien-Gnuplot/blob/0d4659a3687fafd79ba7aabc3a12f633e2d6ec06/Makefile.PL

IMHO, "use strict;" should be enough to classify this as Perl5.

pchaigno:
This pull request fixes #2781 and mitigate the issue of Perl miss-detections.
I reordered heuristic rules by accuracy. Because of the `if/elsif` structure, I think it makes sense to place the rule in which we are most confident first.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -276,20 +276,20 @@ module Linguist
     end
 
     disambiguate ".pl" do |data|
-      if /^(use v6|(my )?class|module)/.match(data)
-        Language["Perl6"]
+      if /^[^#]+:-/.match(data)
+        Language["Prolog"]
       elsif /use strict|use\s+v?5\./.match(data)
         Language["Perl"]
-      elsif /^[^#]+:-/.match(data)
-        Language["Prolog"]
+      elsif /^(use v6|(my )?class|module)/.match(data)
+        Language["Perl6"]
       end
     end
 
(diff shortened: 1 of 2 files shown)


pchaigno:
@arfon The test are failing because [this sample Perl6 file](https://github.com/github/linguist/blob/master/samples/Perl6/hash.t#L15) contains Perl5 code. I'm not sure how I should fix that. Should I remove the Perl5 code (it's in a string) or should I remove this sample file entirely?

chorny:
IMHO, this can be solved by checking for ":lang<perl5>" and "use v6;". If both are present, it is perl6.

arfon:
> @arfon The test are failing because this sample Perl6 file contains Perl5 code. I'm not sure how I should fix that. Should I remove the Perl5 code (it's in a string) or should I remove this sample file entirely?
 
Is [`use strict`](https://github.com/github/linguist/blob/master/samples/Perl6/hash.t#L15) invalid Perl6 code? If so, perhaps we should just delete the line?

pchaigno:
Line deleted; tests are passing!

arfon:
:+1: thanks @pchaigno

PR #2804, opened 2016-01-11
"Protocol Buffer" -> "Protocol Buffers"

haneefmubarak:
According to https://developers.google.com/protocol-buffers/?hl=en, it is indeed Protocol Buffer**s**.


parkr:
:+1: to the name update.

pchaigno:
It's not currently possible to rename languages :/ See [@arfon's comment](https://github.com/github/linguist/pull/2034#issuecomment-71346028) for background.

@arfon Could you add the label to remember to look into this when it becomes possible? ;)

PR #2805, opened 2016-01-11
Add .xproj to list of XML file extensions

jamesqo:
- `.props` is a VS property file that's formatted in XML. [Sample](https://github.com/dotnet/corefx/blob/master/dir.props).
- `.xproj` is the new ASP.NET 5/DNX file that's used to manage a project (similar to csproj). [Sample](https://github.com/ufcpp/UfcppSample/blob/master/Chapters/Resource/ClassOrStruct/ClassOrStruct.xproj).

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3797,6 +3797,7 @@ XML:
   - .xliff
   - .xmi
   - .xml.dist
+  - .xproj
   - .xsd
   - .xul
   - .zcml


pchaigno:
I'm not sure we should add `.props` until we have a *not-a-language* option; half the `.props` files on GitHub are not XML files: [`extension:props NOT xml NOT version`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aprops+NOT+xml+NOT+version&type=Code&ref=searchresults).

jamesqo:
@pchaigno Good point; I've removed `.props` from the list.

arfon:
Thanks @jamesqo. This will be live in the next release of Linguist.

PR #2807, opened 2016-01-11
Support .rs.in as a file extension for Rust files.

mrmonday:
When using syntax extensions in stable or beta Rust channels using the syntex package, it is common to use the file extension .rs.in for the source file, and .rs for the generated file.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3081,6 +3081,7 @@ Rust:
   color: "#dea584"
   extensions:
   - .rs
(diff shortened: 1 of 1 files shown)


pchaigno:
There are only 86 repositories using `.rs.in` among 75 users. At the same time, it seems unlikely it will ever conflict with another language.

arfon:
> At the same time, it seems unlikely it will ever conflict with another language.

Agreed.

mrmonday:
Thanks!

Issue #2801, opened 2016-01-10
Can't clone: no access to vendor/grammars/MagicPython


c-lipka:
Can't create a local clone of the repo (using Atlassian SourceTree), due to vendor/grammars/MagicPython getting in the way, for _some_ reason denying me access to the corresponding repo.

pchaigno:
This is probably happening because you're cloning Linguist with a user who doesn't have an SSH key registered on GitHub. [Some submodules](https://github.com/github/linguist/blob/master/.gitmodules#L694) have an SSH link and will fail during cloning.

I see two solutions here:
- Either we make it an explicit requirement to have an SSH key registered to be able to clone submodules;
- or we change the links to your submodules to use HTTPS links.

With the second solution, we would also need to change the links to the submodules of the submodules if any (see [the Stan grammar submodule](https://github.com/jrnold/stan-language-definitions) for an example). We'd have to open a pull request each time that happen.

/cc @arfon

c-lipka:
Another quality issue with submodules: Some of them include filenames ending in blanks, and one contains a file name with single-quote characters. Once any of these end up on a Windows system, they're a pain to get rid of again.

arfon:
> - or we change the links to your submodules to use HTTPS links.
. With the second solution, we would also need to change the links to the submodules of the submodules if any (see the Stan grammar submodule for an example). We'd have to open a pull request each time that happen. For now, there are only two submodules which have submodules themselves (jrnold/stan-language-definitions and guillermooo/dart-sublime-bundle), only one with an SSH link.

This is probably the best solution here @pchaigno. It looks like there are only three submodules linked to as SSH links? (Although this doesn't address the submodules of submodules issue).

> Another quality issue with submodules: Some of them include filenames ending in blanks, and one contains a file name with single-quote characters. Once any of these end up on a Windows system, they're a pain to get rid of again.

@c-lipka - can you link to an example of this please?

c-lipka:
@arfon - I've removed the repo from my Windows system again by now (including the offending files, after I figured out how to) because I couldn't seem to download it properly anyway; but IIRC one of the offending subprojects was https://github.com/textmate/applescript.tmbundle, with a file in the Commands directory refusing to leave peacefully.

[...]

PR #2808, opened 2016-01-11
HTTPS links for submodules

pchaigno:
This pull request changes three links to submodules that were previously using SSH links.
As reported in #2801, users without an SSH key registered on GitHub will fail to clone submodules with an SSH link.

~~This is still missing a test to make sure we don't add submodules with SSH links again.~~
Working on that.


pchaigno:
@arfon This should be ready ;)

arfon:
:+1: and a test too :zap:

PR #2811, opened 2016-01-16
Heuristic for Pod vs. Perl

pchaigno:
Currently, all `.pod` are rendered via [markup](https://github.com/github/markup) as Pod files. [Most of them](https://github.com/search?utf8=%E2%9C%93&q=extension%3Apod+NOT+dfdgfgfvffg) are, however, (incorrectly) recognized as Perl by Linguist. With github/markup#537 Markup will rely on Linguist to detect the language of files to render and this situation will become problematic.

This pull request adds a new heuristic rule for `.pod` files. If they contain any Pod syntax they are directly classified as Pod files. Files without any Pod syntax are classified as Perl files.
- `.pod` files without any Pod syntax will be highlighted as Perl and counted in statistics.
- Other `.pod` files will be rendered via Markup and won't be counted in statistics.

An alternate solution would be to recognize all `.pod` files as Pod (since they're not supposed to contain executable code but only documentation). I chose to keep `.pod` as a Perl extension because there are still [`.pod` files with only Perl code](https://github.com/search?p=1&q=extension%3Apod+use+strict&type=Code&utf8=%E2%9C%93).


arfon:
Thanks @pchaigno

PR #2812, opened 2016-01-16
Improved vim modeline detection

### LINKED ISSUE #2789: vim modeline regex too restrictive.
The current `vim` modeline regex only captures a narrow subset of legal modeline syntax.
```ruby
 VimModeline = /vim:\s*set.*\s(?:ft|filetype)=(\w+)\s?.*:/i
```
### 1: It requires the "set" token, while vim itself does not.

```text
vim: syntax=perl
vim: set syntax=perl:
```

These are both legal, and equivalent.

> There are two forms of modelines.  The first form:
>	[text]{white}{vi:|vim:|ex:}[white]{options}
...
> The second form (this is compatible with some versions of Vi):
>	[text]{white}{vi:|vim:|Vim:|ex:}[white]se[t] {options}:[text]

Its also worth pointing out the first of these forms does not require the trailing colon either, which your regex currently requires.

### 2: It anchors only on the `ft` and `filetype` settings, which IME are less common than the `syntax` setting. 

And there are copious examples of people using both of these features on github, eg:

```
# vim: syntax=Ruby
```

https://github.com/search?l=ruby&q=vim%3A+syntax%3D&ref=searchresults&type=Code&utf8=%E2%9C%93

chrisarcand:
TLDR: This greatly increases the flexibility of vim modeline detection to manually set the language of a file.

Fixes #2789. In vim there are two forms of modelines:

```
[text]{white}{vi:|vim:|ex:}[white]{options}
examples: 'vim: syntax=perl', 'ex: filetype=ruby'
```
-and-
```
[text]{white}{vi:|vim:|Vim:|ex:}[white]se[t] {options}:[text]
examples: 'vim set syntax=perl:', 'Vim: se ft=ruby:'
```
As you can see, there are many combinations. These changes should allow most combinations to be used. The two most important additions are the use of the keyword 'syntax', as well as the addition of the first form (you now no longer need to use the keyword 'set' with a colon at the end). The use of first form with 'syntax' is [very, very common across GitHub](https://github.com/search?l=ruby&q=vim%3A+syntax%3D&ref=searchresults&type=Code&utf8=%E2%9C%93)


arfon:
:+1: thanks for this @chrisarcand!

arfon:
Please note, these changes won't be live on GitHub until we cut a new Linguist gem (likely next week sometime).

chrisarcand:
No problem, and thanks for the GitHub note! :wave:

PR #2813, opened 2016-01-19
add golang vendor experiment directory

srijs:
For reference: https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -39,6 +39,7 @@
 
 # Go dependencies
 - Godeps/_workspace/
+- vendor/
 
 # Minified JavaScript and CSS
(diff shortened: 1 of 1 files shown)


pchaigno:
`vendor/` is [already there](https://github.com/srijs/linguist/blob/patch-1/lib/linguist/vendor.yml#L72) ;)

srijs:
Oh, how could I have not noticed that!

We could add a comment saying that it's already covered by the general vendor regexes, or just closr this issue then. Your call :)

Bereiche54:
It's already  covered by the general vendor

PR #2825, opened 2016-01-29
Missing Drupal extensions

basvdheijden:
Currently, Drupal .module & .install files are not treated as PHP. I've added these in the PR.

https://github.com/search?utf8=%E2%9C%93&q=extension%3Amodule+NOT+nothack&type=Code&ref=searchresults

https://github.com/search?utf8=%E2%9C%93&q=extension%3Ainstall+NOT+nothack&type=Code&ref=searchresults

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2525,6 +2525,8 @@ PHP:
   - .php5
   - .phps
   - .phpt
+  - .module
+  - .install
   filenames:
   - Phakefile
   interpreters:
--- a/samples/PHP/drupal.install
+++ b/samples/PHP/drupal.install
@@ -1,0 +1,45 @@
+<?php
+
+/**
+ * @file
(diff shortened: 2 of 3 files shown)


arfon:
Thanks for this PR @basvdheijden. As things stand though this change is going to cause issues. Specifically, as `.module` and `.install` aren't defined as language extensions currently, this change will mean that *all* `.module` and `.install` files will be classified as `PHP` which is not correct.

Looking at the breakdown of languages that are being assigned currently to `.module` and `.install` files (I've taken the screen grabs from your search links) it looks like _some_ of these files are already being classified as `PHP`.

screen shot 2016-02-26 at 5 25 48 pm

Do you think you could link to some examples of files that are not being classified correctly?

arfon:

basvdheijden:
You're right I suppose.

PR #2837, opened 2016-02-14
Shell: Add shell-script alias for recognizing Emacs modelines

scop:
Bunch of files that would benefit from this here: https://github.com/scop/bash-completion/tree/master/completions

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3274,6 +3274,7 @@ Shell:
   color: "#89e051"
   aliases:
   - sh
+  - shell-script
   - bash
   - zsh
   extensions:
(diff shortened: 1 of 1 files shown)


arfon:
@scop - thanks for this. It looks like we have a failing test here:

```
  1) Failure:
TestPedantic#test_nonprimary_extensions_are_sorted [/home/travis/build/github/linguist/test/test_pedantic.rb:15]:
.sh.in should come after .bash

```
Would you mind updating this PR to fix the ordering?

scop:
Done. BTW the error `.sh.in should come after .bash` could be improved; I think what it actually should say is `.sh.in should come after .ksh`, i.e. tell the right location right away, not one spelling order error at a time.

arfon:
:+1: thanks @scop! I agree the error messages could be better.

This will be live in the next release of Linguist (next week sometime)

Issue #2841, opened 2016-02-18
DirectX Effect files detected as Flux


kusma:
DirectX Effect files (*.fx) gets detected as Flux. See for example here:
https://github.com/kusma/vlee/search?l=flux

This seems rather common:
https://github.com/search?utf8=%E2%9C%93&q=extension%3Afx+NOT+nothack&type=Code&ref=searchresults

pchaigno:
I'm sure DirectX Effect would be accepted as a new language if it meets the requirements. Please see the [contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language) on how to add a new language ;)

kusma:
I gave adding HLSL/fx a go, [here](https://github.com/kusma/linguist/tree/hlsl)

However, it seems to make the test-suite a bit angry:
```
  1) Error:
TestBlob#test_language:
NoMethodError: undefined method `[]' for nil:NilClass
    /home/ubuntu/workspace/lib/linguist/classifier.rb:131:in `token_probability'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:120:in `block in tokens_probability'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:119:in `each'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:119:in `inject'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:119:in `tokens_probability'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:105:in `block in classify'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:104:in `each'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:104:in `classify'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:78:in `classify'
    /home/ubuntu/workspace/lib/linguist/classifier.rb:20:in `call'
    /home/ubuntu/workspace/lib/linguist/language.rb:116:in `block (3 levels) in detect'
    /home/ubuntu/workspace/lib/linguist.rb:17:in `instrument'
    /home/ubuntu/workspace/lib/linguist/language.rb:115:in `block (2 levels) in detect'
    /home/ubuntu/workspace/lib/linguist/language.rb:113:in `each'
    /home/ubuntu/workspace/lib/linguist/language.rb:113:in `block in detect'
    /home/ubuntu/workspace/lib/linguist.rb:17:in `instrument'
[...]

PR #2842, opened 2016-02-18
Add support for HLSL

kusma:
This adds support for DirectX HLSL language. Fixes issue #2841

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1323,6 +1323,16 @@ HCL:
   ace_mode: ruby
   tm_scope: source.ruby
 
+HLSL:
+  type: programming
+  extensions:
+    - .fx
+    - .fxh
+    - .hlsl
+    - .hlsli
+  ace_mode: text
+  tm_scope: none
+
 HTML:
   type: markup
   tm_scope: text.html.basic
--- a/samples/FLUX/gameserver.fx
+++ b/samples/FLUX/gameserver.fx
@@ -1,0 +1,54 @@
+typedef engine isEngineMessage;
+typedef turn isTurnMessage;
+typedef connect isConnectMessage;
+typedef disconnect isDisconnectMessage;
+
+ClientMessage(char* data) => ();
+ParseMessage(char* data) => (int type, int client, char* data);
+ReadMessage(int type, int client, char* data) => ();
+
+ParseEngine(int type, int client, char* data) => (int client, int direction);
+DoEngine(int client, int direction) => ();
+
+ParseTurn(int type, int client, char* data) => (int client, int direction);
+DoTurn(int client, int direction) => ();
+
+ParseConnect(int type, int client, char* data) 
+	=> (int client, char* host, int port);
+DoConnect(int client, char* host, int port) => ();
+
+ParseDisconnect(int type, int client, char* data) => (int client);
+DoDisconnect(int client) => ();
+
+UpdateBoard(ClientList clients) => (ClientList clients);
+SendData(ClientList clients) => ();
+
+DoUpdate(ClientList clients) => ();
+
+DataTimer() => (ClientList clients);
+
+GetClients() => (ClientList clients);
+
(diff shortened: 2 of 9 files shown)


larsbrinkhoff:
Nice.  I did some work to add HLSL, but never quite finished.  I'll try to piggy-back it on your work.

larsbrinkhoff:
`.fx` is also used by GLSL.  Did you check for conflicts?

larsbrinkhoff:
There's also `.fxh` which seems to be used for HLSL header files.

larsbrinkhoff:
I can't seem to find much FLUX code at all on GitHub.  Is there any?

On the other hand, I found [this JavaFX code](http://github.com/search?q=extension%3Afx+package+import+public+class&).

kusma:
@larsbrinkhoff: Thanks for feedback.

Interesting point about .fx also being used for GLSL, I was not aware... I wonder why, though. Anyway, I don't think this patch series makes the situation any worse in that regard, though. So perhaps adding .fx to GLSL would be OK to be done as a follow-up?

I forgot about .fxh, I'll update the commit to include that. It seems [.hlsli is also used](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ahlsl+%23include+AND+hlsli&type=Code&ref=searchresults)

About the FLUX code, I wasn't even aware of FLUX before I noticed my HLSL were misdiagnosed. And it seems to be an experimental language that only really lived for a very short time, according to https://plasma.cs.umass.edu/emery/flux.1.html (it's marked as unmaintained, and only ever reached v0.02)... So I would be surprised if there were any notable amounts of FLUX code on GitHub. But I don't want to be the one to remove someone else's contribution, so I tried to fix it up best as I could.

People who care about JavaFX can IMO add support for that. I don't know much about it besides it's existence...

So, unless you disagree with my assessments, I think updating HLSL to include .fxh and .hlsli is the way to go...

larsbrinkhoff:
I'm not opposed to stating that I don't disagree!

arfon:
@kusma - would you mind updating this thread with links to search queries demonstrating the usage of each of these extensions?

Thanks!

kusma:
@arfon: Sure, here you go:

- [.fx](https://github.com/search?q=extension%3Afx+NOT+nothack&type=Code&ref=searchresults)
- [.fxh](https://github.com/search?q=extension%3Afxh+NOT+nothack&type=Code&ref=searchresults)
- [.hlsl](https://github.com/search?q=extension%3Ahlsl+NOT+nothack&type=Code&ref=searchresults)
- [.hlsli](https://github.com/search?q=extension%3Ahlsli+NOT+nothack&type=Code&ref=searchresults)

arfon:
:+1: thanks @kusma. Sorry it took me a while to review this.

This will be live in the next release of Linguist (later this week sometime)

kusma:
Cool, thanks! No problem with the timing, I'm not in a hurry ;)

larsbrinkhoff:
Was it intentional to make `.fx` the primary file extension instead of e.g. `.hlsl`?

kusma:
Hm, I started out with .fx, so I guess that's why. But I guess the primary extension should be .hlsl, yeah. So no, not intentional, and probably not ideal.

kusma:
Hm, wait. Didn't I make HLSL the name of the language? I don't think I understand what the primary extension is. I just alphabetized the list of extensions... Was that wrong?

larsbrinkhoff:
It's a well-kept secret that the first item in the extensions list is the primary extension. ;-)

https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L10

arfon:
> https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L10-L12

Worst docs ever... :disappointed:

kusma:
OK, I'll submit a patch to fix this :)

kusma:
#2870 tries to remedy this. Sorry for the mistake!

bleamika:
yes it's really too.. make me dubbing

PR #2843, opened 2016-02-19
Go vendor

bkeepers:
This is an update to https://github.com/github/linguist/pull/2620 that only treats `.go` files in `vendor` as generated.

I'm not a gopher, so one question I have about go 1.6: does it vendor non-`.go` files as well? If it does, then this won't work.

/cc @kr @rakyll @cespare @parkr @arfon


cespare:
The vendoring mechanism itself is not (yet) provided by an official tool as part of the Go distribution.

Some tools might choose to implement only the specific files needed as dependencies (but even this can include `.go`, `.s`, `.c`, `.cc`, `.h`, and other files). Other tools might instead choose to simply `cp` packages (and maybe delete the VCS metadata). Finally, most vendoring tools also write metadata files (typically `.json`).

So yes, typical Go `/vendor` directories will contain non `.go` files.

lib/linguist/generated.rb
```diff
@@ -280,6 +281,16 @@ def node_modules?
       !!name.match(/node_modules\//)
     end
 
+    # Internal: Is the blob part of the Go vendor/ tree,
+    # not meant for humans in pull requests.
+    #
+    # Returns true or false.
+    def go_vendor?
+      return false unless extname == '.go'
```
parkr: I started doing this too until I realized that it's possible `.md` and other files might also be placed here. I don't know if they filter out the dependencies saved to just `.go` files.

bkeepers:
> So yes, typical Go `/vendor` directories will contain non `.go` files.

I was afraid of that.

Ok, any other brilliant ideas? Currently, linguist evaluates each file separately, so there's not a way for it to know that it is looking at a "go project" and should ignore `vendor`.

parkr:
@bkeepers What about looking for a hostname? `vendor/github.com/*`, for example, would satisfy this.

tv42:
Vendoring a cgo-using project means you must vendor `*.c` etc too, regardless of the tool that does it. Consider what happens when you vendor https://github.com/mattn/go-sqlite3

arfon:
> @bkeepers What about looking for a hostname?

@parkr will there always be a hostname in the path?

cespare:
@arfon no

arfon:
> @arfon no

@cespare :disappointed: OK. Unless we can craft a regular expression to match these paths then I'm afraid this PR is a non-starter because of the effect it will have for non-Go projects.

parkr:
> @arfon no

@cespare What if we just include when there _is_ a hostname and if you have vendor/mypkg then we don't consider it generated? Then at least for the vast majority of the time (where you do have a hostname), you don't have to see the diff.

c2h5oh:
Is there a reason not to consider everything in `/vendor` as generated?

arfon:
> @cespare What if we just include when there is a hostname and if you have vendor/mypkg then we don't consider it generated? Then at least for the vast majority of the time (where you do have a hostname), you don't have to see the diff.

@cespare - what do you think about @parkr's suggestion?

cespare:
@parkr @arfon SGTM, that seems like it will have some false positives and false negatives but probably quite few of either.

arfon:
🎏 flagging this a stale 🎏

arfon:
Seems like we're going to need to revisit this. We've recently changed the behaviour of GitHub to hide all paths defined in [`vendor.yml`](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml) which is a good change for Go folks but not for lots of other people.

@parkr's [suggestion above](https://github.com/github/linguist/pull/2843#issuecomment-207592798) is the closest solution I have seen to solving the issues for Go people but not hiding all `vendor` paths by default.

/ cc @nathany

nathany:
@afron I can't think of any quick fix for this, though @parkr's might work, it still feels like a bit of a hack. For those in language communities where they place first-party code under `vendor/`, I don't know if they have paths containing hostnames or not, though [this example seems not to](https://github.com/github/linguist/pull/3262#issuecomment-256217658).

As per my [earlier comment](https://github.com/github/linguist/issues/3234#issuecomment-256034711), the change didn't quite achieve the desired effect, because the full diff still isn't available after collapsing vendor files.

It may be best to revert and come up with a better overall solution. My [suggestion](https://github.com/github/linguist/pull/3262#issuecomment-256232062) is:

> A generalized solution requires some way to review individual directories of code at a time -- as a way to split up large diffs.

> The goal of [this pull request](https://github.com/github/linguist/pull/3262) and [associated issue ](https://github.com/github/linguist/issues/3234) is to help make the GitHub UI more responsive on large diffs, not to prevent people from reviewing third-party code.

nathany:
Would it be helpful if I created an example repository and pull request that used a vendor tool such as [gvt](https://github.com/FiloSottile/gvt) to bring in a few large dependencies as one might do when starting a new Go project?

arfon:
> An existing workaround for the Go community is to review individual commits, with one or more commits for vendoring libraries, and a single commit for everything else.

> This is really the key issue. The error "Sorry, we could not display the entire diff because it was too big" needs to be addressed. The UI shouldn't need an error message like that.

I feel your pain. Thanks for describing this so clearly, this is excellent feedback @nathany.

> It may be best to revert and come up with a better overall solution. 

I think that's what I'm going to have to do unfortunately. In addition, let's patch up this PR (albeit in its somewhat hacky form to match where there's a hostname) to remove some of the pain for some Go developers.

> Would it be helpful if I created an example repository and pull request that used a vendor tool such as gvt to bring in a few large dependencies as one might do when starting a new Go project?

Yes please. That would be really useful to see.

nathany:
Here is an example repository. https://github.com/nathany/gvt-sample/pull/1

arfon:
OK, I'm about to revert the changes on GitHub.com that have hidden the `vendor` path by default. I then plan to cut a new gem based on this PR to improve the situation for Go users.

arfon:
OK these changes are now live on GitHub. Looking at https://github.com/nathany/gvt-sample/pull/1/files it would appear that the Go-specific vendor check is catching a good number of the files that we would want it to.

parkr:
:tada:

c2h5oh:
Now if only we have `Sorry, we could not display the entire diff because too many files (305) changed` fixed with 0 diffs displayed :-)

From UX perspective it would be great idea to have all suppressed changes grouped and collapsed by default. 

Functionally we just need suppressed files not to count towards diff limit.

PR #2844, opened 2016-02-19
Move HACKING.rst.txt to sample directory

pchaigno:
This PR moves the sample file where it belongs :-)

/cc @arfon


arfon:
:zap: thanks!

PR #2848, opened 2016-02-19
Fix ordering of makefile in languages.yml

Dominator008:


Dominator008:
Fixed in master, closing

PR #2851, opened 2016-02-22
Add .yap ext and yap interpreter for Prolog lang

edechter:
Support .yap extension commonly used with the YAP prolog compiler. Corresponding search results [here](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ayap+NOT+notyap&type=Code&ref=searchresults)


arfon:
Thanks @edechter

PR #2857, opened 2016-03-01
Add BUCK filename to Python in languages.yml

Dominator008:
BUCK is the filename for the build files of the Facebook Buck build system. BUCK files are valid Python files.

Eg.:
  https://github.com/GerritCodeReview/gerrit/blob/master/BUCK

Also add a missing sample for Pants / Bazel BUILD files. They are also valid Python files.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -474,7 +474,7 @@ CSS:
   color: "#563d7c"
   extensions:
   - .css
-  
+
 CSV:
   type: data
   ace_mode: text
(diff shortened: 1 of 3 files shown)


arfon:
Thanks for this @Dominator008. Do [these](https://github.com/search?q=name+filename%3ABUCK&ref=searchresults&type=Code&utf8=%E2%9C%93) all look to be `BUCK` files too?

Dominator008:
@afton I skimmed through them and yes they look to be `BUCK` files.

arfon:
> @afton I skimmed through them and yes they look to be `BUCK` files.

OK great. Thanks for checking. @Dominator008 - this will be live on GitHub once we release a new Linguist gem which will be in the next week or so.

Dominator008:
Thanks!

PR #2861, opened 2016-03-04
Recognizing Jenkinsfile as Groovy source

jglick:
`Jenkinsfile` is the conventional name for a Groovy-syntax script containing build instructions for the Jenkins continuous integration server using the Pipeline feature. [Details](https://github.com/jenkinsci/workflow-plugin/blob/7d2004ea482418f84c3f221d3f55dff20e8ada44/TUTORIAL.md#creating-multibranch-projects) and [context](https://issues.jenkins-ci.org/browse/JENKINS-31152).

Look to be about a couple hundred [known files](https://github.com/search?utf8=%E2%9C%93&q=Jenkinsfile+in%3Apath&type=Code&ref=searchresults) out there. To date the trick for getting syntax highlighting has been to use `#!groovy` as [here](https://github.com/jenkinsci/pipeline-examples/blob/0a7aa6180e14fd17901a514c5e6bb356e416f41f/global-library-examples/global-function/Jenkinsfile#L1).

For my colleagues: @reviewbybees esp. @kohsuke who selected this filename.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -302,3 +302,6 @@
 
 # Android Google APIs
 - (^|/)\.google_apis/
+
+# Jenkins Pipeline
+- ^Jenkinsfile$
--- a/samples/Groovy/filenames/Jenkinsfile
+++ b/samples/Groovy/filenames/Jenkinsfile
@@ -1,0 +1,46 @@
+jettyUrl = 'http://localhost:8081/'
+
+def servers
+
+stage 'Dev'
+node {
+    checkout scm
+    servers = load 'servers.groovy'
+    mvn '-o clean package'
+    dir('target') {stash name: 'war', includes: 'x.war'}
+}
+
+stage 'QA'
+parallel(longerTests: {
+    runTests(servers, 30)
+}, quickerTests: {
+    runTests(servers, 20)
+})
+
+stage name: 'Staging', concurrency: 1
+node {
+    servers.deploy 'staging'
+}
+
+input message: "Does ${jettyUrl}staging/ look good?"
+
+stage name: 'Production', concurrency: 1
+node {
+    sh "wget -O - -S ${jettyUrl}staging/"
+    echo 'Production server looks to be alive'
+    servers.deploy 'production'
+    echo "Deployed to ${jettyUrl}production/"
+}
+
+def mvn(args) {
+    sh "${tool 'Maven 3.x'}/bin/mvn ${args}"
+}
+
+def runTests(servers, duration) {
+    node {
+        checkout scm
+        servers.runWithServer {id ->
+            mvn "-o -f sometests test -Durl=${jettyUrl}${id}/ -Dduration=${duration}"
+        }
+    }
+}
--- a/test/test_file_blob.rb
+++ b/test/test_file_blob.rb
@@ -522,6 +522,9 @@ class TestBlob < Minitest::Test
     assert sample_blob("myapp/My Template.xctemplate/___FILEBASENAME___.h").vendored?
     assert sample_blob("myapp/My Images.xcassets/some/stuff.imageset/Contents.json").vendored?
     assert !sample_blob("myapp/MyData.json").vendored?
+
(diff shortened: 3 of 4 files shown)


lib/linguist/vendor.yml
```diff
@@ -302,3 +302,6 @@
 
 # Android Google APIs
 - (^|/)\.google_apis/
+
+# Jenkins Pipeline
+- ^Jenkinsfile$
```
jglick: Typically a `Jenkinsfile` will be at the root of its repository; when used as part of a multibranch Pipeline project, that is required as this exact filename is the indication that a repository/branch is ready to build.

samples/Groovy/filenames/Jenkinsfile
```diff
@@ -0,0 +1,46 @@
+jettyUrl = 'http://localhost:8081/'
```
jglick: Adapted from a [standard demo](https://github.com/jenkinsci/workflow-plugin/blob/7d2004ea482418f84c3f221d3f55dff20e8ada44/demo/repo/Jenkinsfile).

test/test_file_blob.rb
```diff
@@ -522,6 +522,9 @@ def test_vendored
     assert sample_blob("myapp/My Template.xctemplate/___FILEBASENAME___.h").vendored?
     assert sample_blob("myapp/My Images.xcassets/some/stuff.imageset/Contents.json").vendored?
     assert !sample_blob("myapp/MyData.json").vendored?
+
+    # Jenkins
+    assert sample_blob("Jenkinsfile").vendored?
```
jglick: By the way `lib/linguist/vendor.yml` suggests adding test coverage to `test/test_blob.rb#test_vendored` but it seems that almost all of the existing tests are here instead, so I tried to follow suit.

samples/Groovy/filenames/Jenkinsfile
```diff
@@ -0,0 +1,46 @@
...
+
+def servers
+
+stage 'Dev'
+node {
+    checkout scm
+    servers = load 'servers.groovy'
```
jglick: Auxiliary library files are conventionally just named `*.groovy` and so require no special treatment.

amuniz:
Nice. 
:bee: and :+1: !

recena:
:bee: and :+1: undoubtedly!

lib/linguist/vendor.yml
arfon: @jglick - is there a reason this should be considered a `vendored?` file for Linguist - this basically means that it won't contribute to language statistics in the GitHub repository (the coloured bar at the top of the repository page)

test/test_file_blob.rb
arfon: :+1: yeah, this is fine.

arfon:
Thanks for the very nice PR @jglick. I've left a couple of comment inline.

lib/linguist/vendor.yml
jglick: Yes, that was intentional, and by analogy with very domain-specific files like `Vagrantfile` which you would not consider associated with Ruby development. Groovy is used as a host for the EDSL, but examples like [this](https://github.com/jenkinsci/pipeline-examples/blob/0a7aa6180e14fd17901a514c5e6bb356e416f41f/global-library-examples/global-function/Jenkinsfile#L5-L15) show that the actual `Jenkinsfile` can be made to look more like a config file than a program. Furthermore, allowable methods calls are restricted by a sandbox to a whitelist, and execution proceeds via a custom interpreter via continuation-passing transform which imposes various constraints on serializability etc., so the result does not look much like a general-purpose Groovy program.

jglick:
@reviewbybees done

lib/linguist/vendor.yml
arfon: OK great. Thanks for the clarification.

arfon:
Thanks for the contribution @jglick! This will be live in the next release of Linguist (probably another week or so before we cut a new gem)

PR #2863, opened 2016-03-05
Add new strategy using .gitattributes

nojebar:
This tiny patch implements a strategy using the `linguist-language` attribute present in `.gitattributes`, if the file is in a `git` repo.  See #1792.

The implementation is so trivial that I wonder if there are any reasons for not having implemented this before (performance?).

If there is a willingness to accept this PR, I will add tests and clean up the patch.

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -11,6 +11,7 @@ require 'linguist/samples'
 require 'linguist/file_blob'
 require 'linguist/blob_helper'
 require 'linguist/strategy/filename'
+require 'linguist/strategy/gitattributes'
 require 'linguist/strategy/modeline'
 require 'linguist/shebang'
 
@@ -89,6 +90,7 @@ module Linguist
 
     STRATEGIES = [
       Linguist::Strategy::Modeline,
+      Linguist::Strategy::Gitattributes,
       Linguist::Shebang,
       Linguist::Strategy::Filename,
       Linguist::Heuristics,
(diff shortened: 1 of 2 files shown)


arfon:
Good idea @nojebar, so good that `LazyBlob` [already does this](https://github.com/github/linguist/blob/master/lib/linguist/lazy_blob.rb#L62-L70) :-)

nojebar:
But if I am reading this correctly, `LazyBlob` is only used when gathering language statistics, not when deciding the language for syntax highlighting... or I am missing something ?  Thanks!

arfon:
> But if I am reading this correctly, `LazyBlob` is only used when gathering language statistics, not when deciding the language for syntax highlighting... or I am missing something ? Thanks!

Kind of and the answer depends upon whether you're looking to us Linguist yourself or trying to fix something on GitHub.com. 

We currently don't have support for using the `.gitattributes` [overrides](https://github.com/github/linguist#overrides) for syntax highlighting as our front-end notes (application servers) don't actually have access to raw Git repositories (that have the `.gitattributes` data). We typically include the [`BlobHelper`](https://github.com/github/linguist/blob/master/lib/linguist/blob_helper.rb) class in this setting.

When we run the repository language statistics job this is run on our fileservers (that do have Git data) and so we use the `LazyBlob` class in this case.

So basically this PR doesn't really solve the problem (that I think) you're trying to solve here.

nojebar:
Indeed.  Thanks for the explanation! Closing this now.

PR #2866, opened 2016-03-06
Classification of props file

### LINKED ISSUE #2852: Classify props file as XML
In VS land `.props` file are XML format, which contain the project properties and used by msbuild:
https://msdn.microsoft.com/en-us/library/669zx6zc.aspx

Example of such a file is https://github.com/dotnet/corefx/blob/67a17b3/dir.props

Please classify `.props` and `.xml`, so GitHub syntax highlighting can catch up.

cc @mellinoe, @rainersigwald

jasonwilliams200OK:
* Initially treat as XML
* Disambiguate from ini (key-value pair style with `=`)
* If the file is neither XML-style nor INI, classify as SQL

Fix #2852

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -313,6 +313,14 @@ module Linguist
       end
     end
 
+    disambiguate ".props" do |data|
+      if /^(\s*)(<Project|<Import|<Property|<?xml|xmlns)/i.match(data)
+        Language["XML"]
+      elsif /\w+\s*=\s*/i.match(data)
+        Language["INI"]
+      end
+    end
+
     disambiguate ".r" do |data|
       if /\bRebol\b/i.match(data)
         Language["Rebol"]
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3810,6 +3810,7 @@ XML:
   - .osm
   - .plist
   - .pluginspec
+  - .props
   - .ps1xml
   - .psc1
   - .pt
--- a/samples/XML/Default.props
+++ b/samples/XML/Default.props
@@ -1,0 +1,29 @@
+<!--
+***********************************************************************************************
+Default.props
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Import Before -->
+  <Import Condition="Exists('$(MSBuildThisFileDirectory)ImportBefore\Default')" Project="$(MSBuildThisFileDirectory)ImportBefore\Default\*.props" />
+  
+  <PropertyGroup>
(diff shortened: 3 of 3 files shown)


lib/linguist/heuristics.rb
```diff
@@ -313,6 +313,16 @@ def call(data)
       end
     end
 
+    disambiguate ".props" do |data|
+      if /^(\s*)(<Project|<Import|<Property|<\/Project|<\/Import|<\/Property)/i.match(data) 
+        Language["XML"]
+      elsif /\w+\s*=\s*/i.match(data)
+        Language["INI"]
+      else
```
pchaigno: This is a bit dangerous and we try to avoid it. Isn't there a regular expression we could use here for SQL?

pchaigno:
@jasonwilliams200OK Thanks for the pull request! I've added a comment above. You'll also need to add `.props` to SQL and INI for this to work properly ;)

lib/linguist/heuristics.rb
jasonwilliams200OK: Hi, is there a suitable language with this kind of syntax:

```
key1 value1
key2     value2
```
(this is ini/config, but without `=`)

sure I will add a regex for it. :)

jasonwilliams200OK:
Thanks @pchaigno, if there any other suitable language which matches the pattern, I would add a regex and specify that lang.

```sql
# just used sql for this because I couldn't find a better match :(
key1 value1
key2   value2
```

jasonwilliams200OK:
Analogically with INI specifier and `=` sign, the syntax highlighting works:

```ini
key1 =  value1
key2=value2
key3      = value3
```

But some `.props` files (configuration type) lack `=`.

lib/linguist/heuristics.rb
jasonwilliams200OK: Or should I just change the else case to `Text` until we found a better substitute?

jasonwilliams200OK:
@arfon could you review this item please?

lib/linguist/heuristics.rb
```diff
@@ -313,6 +313,16 @@ def call(data)
       end
     end
 
+    disambiguate ".props" do |data|
+      if /^(\s*)(<Project|<Import|<Property|<\/Project|<\/Import|<\/Property)/i.match(data)
```
arfon: Could we simplify this to simply check for a `xmlns=`?

jasonwilliams200OK: @arfon, xmlns is optional as there are over 3k files: https://github.com/search?p=2&q=extension%3Aprops+xml+version+encoding+OR+project+NOT+xmlns&ref=searchresults&type=Code&utf8=%E2%9C%93. However, I think we should add `<?xml` and `xmlns` checks and remove the closing tags patterns. Does it sound good?

arfon: Yeah that sounds reasonable.

jasonwilliams200OK: Done.

jasonwilliams200OK:
@arfon, @pchaigno, I have removed the last case (https://github.com/github/linguist/pull/2866#issuecomment-193003371), which will be classified as text for now. Later when someone will know what is the closest language matching key-value-without-equals-sign, they can add it.

arfon:
:+1: thanks @jasonwilliams200OK. This will be live in the next release of Linguist.

jasonwilliams200OK:
Thanks @arfon! :tada:

PR #2868, opened 2016-03-09
2603 local

arfon:
Replaces #2603


arfon:
:+1: thanks for this @matejak. This will be live in the next release of Linguist (later this week sometime I hope).

Issue #2850, opened 2016-02-22
Add .sublime-syntax to YAML extensions


michaelblyons:
Sublime Text has updated its syntax definitions to use [a YAML `.sublime-syntax` derivative][st-yaml] in addition to the plist-derived `.tmLanguage` they had before. [Lots of people][gh-search] appear to be using this already.

Let's add `.sublime-syntax` to the recognized file extensions for [YAML][].

[st-yaml]: http://www.sublimetext.com/docs/3/syntax.html
[gh-search]: https://github.com/search?utf8=%E2%9C%93&q=extension%3Asublime-syntax+yaml&type=Code&ref=searchresults
[yaml]: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3898

pchaigno:
I count 125 users of the extension with a total of 144 repositories.
@arfon Would that be enough to add support for the extension?

arfon:
> I count 125 users of the extension with a total of 144 repositories. @arfon Would that be enough to add support for the extension?

Seems pretty harmless - let's do it.

pchaigno:
Addressed in #2869.

PR #2869, opened 2016-03-09
Support for YAML .sublime-syntax extension

pchaigno:
This pull request adds support for the YAML extension `.sublime-syntax` as discussed in #2850.


arfon:
:+1: thanks

PR #2870, opened 2016-03-09
make .hlsl the primary extension for HLSL

kusma:
When I submitted the HLSL-language, I accidentally missed that the
first extension was the primary one (as is documented in the source
code, which I unfortunately missed), and instead alphabetized the
whole list.

The primary extension should be .hlsl, so let's remedy this.


arfon:
Perfect. Thanks!

PR #2876, opened 2016-03-11
git-linguist: Print PWD

vmg:
Print the current directory when `git-linguist` is being ran outside of a Git directory.

We'll pick this up on the next release.

cc @arfon


arfon:
:+1: planning on cutting a new get early next week.

Issue #2384, opened 2015-05-04
PLpgSQL files are treated as SQLPL files


vitalyisaev2:
I've written some stored procedures for PostgreSQL. Code contains several PLpgSQL-specific keywords, but for a some reason it's still treated as SQLPL. 

https://github.com/vitalyisaev2/river-oxygen-web/search?l=sqlpl

vitalyisaev2:
Overrided it with a ```.gitattributes```.

pchaigno:
@Shoelace Should we change [the first regular expression for PLpgSQL](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L244) from:
`/^\\i\b|AS \$\$|LANGUAGE '+plpgsql'+/i` to
`/^\\i\b|AS \$\$|LANGUAGE '*plpgsql'*/i`?

It looks like the single quotes are not required (see [@vitalyisaev2's file](https://github.com/vitalyisaev2/river-oxygen-web/blob/f23c55633b4774a12a4fab1d9907f958c4e3cf45/sql/procedures.sql#L13)).

Shoelace:
postgres 8.4 at http://www.postgresql.org/docs/8.4/static/sql-createfunction.html says 

```langname
The name of the language that the function is implemented in. Can be SQL, C, internal, or the name of a user-defined procedural language. For backward compatibility, the name can be enclosed by single quotes.
```

where as 9.4 ( http://www.postgresql.org/docs/9.4/static/sql-createfunction.html) says 

```lang_name
The name of the language that the function is implemented in. It can be sql, c, internal, or the name of a user-defined procedural language, e.g. plpgsql. Enclosing the name in single quotes is deprecated and requires matching case.
```

so to be honest it should probably be
    /^\\i\b|AS \$\$|LANGUAGE '?plpgsql'?/i
as in exactly 0 or 1 single quote..    but if you want it to be more relaxed then * is fine.

pchaigno:
@Shoelace Okay, thanks a lot for the information :+1: 
I'll open a PR tonight to fix that.

arfon:
@pchaigno - did you ever submit a PR to change this regex? I think we still have the [old one defined](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L357)

pchaigno:
@arfon #2877 should fix it ;)

arfon:
> @arfon #2877 should fix it ;)

[...]

PR #2877, opened 2016-03-12
Fix PLpgSQL heuristic rule

pchaigno:
Fixes the PLpgSQL heuristic rule as discussed in #2384.


arfon:
:+1: many thanks @pchaigno

PR #2883, opened 2016-03-17
Fix regex in .rpy disambiguate

williamd1k0:
I wrote the regex wrongly, matching `define` (which is not python) and not only `def`.
I'm sorry for the mistake. :pensive:


arfon:
:+1: thanks for the fix.

PR #2894, opened 2016-03-19
Fix warnings

pchaigno:
This pull request fixes many warnings in tests.
Examples of these warnings can be seen in [recent Travis builds](https://travis-ci.org/github/linguist/jobs/117110750).

There are still a couple warnings I wasn't able to fix. In particular, `warning: setting Encoding.default_external` requires us to pass the external string encoding as an argument when calling Ruby. I'm not sure how to do that in tests.

### DIFF
--- a/lib/linguist/blob.rb
+++ b/lib/linguist/blob.rb
@@ -63,7 +63,7 @@ module Linguist
     #
     # Returns an Array
     def extensions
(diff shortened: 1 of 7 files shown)


pchaigno:
Another warning was fixed via benbalter/licensee#89.

arfon:
👍 you are a good fellow @pchaigno.

PR #2901, opened 2016-03-21
Add MSBuild's .proj to the list of XML files

jamesqo:
In addition to filetypes like `.csproj` and `.fsproj`, plain old `.proj` can also be used for building MSBuild-based projects.

[Here's](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aproj+Project+msbuild&type=Code&ref=searchresults) an example of some code that would benefit from this, which appears to be about 24k files. In contrast, if you look at `.proj` files that [are not XML](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aproj+NOT+Project&type=Code&ref=searchresults), most of them appear to be binary.

cc: @arfon

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3870,6 +3870,7 @@ XML:
   - .osm
   - .plist
   - .pluginspec
+  - .proj
   - .props
   - .ps1xml
   - .psc1


arfon:
@pchaigno - what do you think about this PR? As you noted in [this comment](https://github.com/github/linguist/pull/2805#issuecomment-170502876) we don't have a 'no language' option so usually we wouldn't be able to accept this PR.

That said, if @jamesqo is correct and the rest of these files are being detected as binary then perhaps this is OK as we short-circuit the language detection here: https://github.com/github/linguist/blob/1efd4c83f93ae16b2b67cb7a25008cd591b39aee/lib/linguist.rb#L20

pchaigno:
To get an idea of how many of those files are binaries, I downloaded 2,465 of [the 5,226 files from this query](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aproj+NOT+Project&type=Code&ref=searchresults). 1,333 (~54%) are binary files, the rest (1,132) are text files.
@arfon Your call :wink:

larsbrinkhoff:
I have a work-in-not-so-much-progress for this:
https://github.com/larsbrinkhoff/linguist/commits/msbuild

larsbrinkhoff:
@pchaigno How do you get more than 1000 results from a search?

pchaigno:
@larsbrinkhoff I download with the three sort options. `new` and `old` parameters ;)

arfon:
🎏 flagging this as stale 🎏

PR #2902, opened 2016-03-21
Add App.config + NuGet.config to the XML file list

jamesqo:
#### Changes

- Files named `App.config` and `NuGet.config` have been added to the list of XML files.

#### Rationale

- There are [virtually no](https://github.com/search?utf8=%E2%9C%93&q=filename%3ANuGet.config+NOT+xml+NOT+config+NOT+settings+NOT+configuration+NOT+solution&type=Code&ref=searchresults) files on GitHub named `NuGet.config` that aren't XML.

- There are about [32k](https://github.com/search?utf8=%E2%9C%93&q=filename%3AApp.config+NOT+xml+NOT+version+NOT+configuration&type=Code&ref=searchresults) files named `App.config` that aren't related to .NET projects, and approximately [926k](https://github.com/search?utf8=%E2%9C%93&q=filename%3AApp.config&type=Code&ref=searchresults) that are.

cc: @arfon (sorry if I spammed your inbox haha :smile:)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3916,6 +3916,8 @@ XML:
   filenames:
   - .classpath
   - .project
+  - App.config
+  - NuGet.config
   - Settings.StyleCop
   - Web.Debug.config
   - Web.Release.config
--- a/test/fixtures/Data/app.config
+++ b/test/fixtures/Data/app.config
@@ -1,0 +1,6 @@
+<?xml version="1.0"?>
+<configuration>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
+  </startup>
+</configuration>


arfon:
> There are about 32k files named App.config that aren't related to .NET projects, and approximately 926k that are.

Thanks for this @jamesqo. We'll need to figure out what these other 32k files are before we can add `App.config` as these would be marked as XML automatically by this PR as it stands.

Do you have any ideas what these are?

jamesqo:
@arfon Actually, the number seems to be closer to [2k](https://github.com/search?utf8=%E2%9C%93&q=filename%3AApp.config+extension%3Aconfig+NOT+xml+NOT+version+NOT+configuration&type=Code&ref=searchresults) (I added an `extension:config` parameter to the search query). Of those, most seem to be Erlang files ([example](https://github.com/dominikb1888/csd_tutdev/blob/af9c04d6f3ccd2c86b1223f0967236a0d0edb349/apps/csd_web/priv/app.config)) and have a lowercase filename.

Is there any way to make the filename case-sensitive, so only `App.config` gets picked up and not `app.config`? That would eliminate most of these extraneous results.

pchaigno:
> Is there any way to make the filename case-sensitive, so only `App.config` gets picked up and not `app.config`? That would eliminate most of these extraneous results.

Yes, the filename strategy is case-sensitive.

Maybe we should also add one of the `app.config` files in `test/fixtures/Data/` to make sure that if we make the filename strategy case-insensitive in the future, we get a failing test. (`test/fixtures/Data/` corresponds to test files which shouldn't be recognized by Linguist.)

arfon:
That's a good idea. @jamesqo do you think you could add the lowercase example as @pchaigno suggests?

jamesqo:
@arfon Done: 0c86818

arfon:
Thanks for this contribution @jamesqo :zap:. This will be live in the next release of Linguist (likely sometime in the next week).

PR #2906, opened 2016-03-22
Add .vssettings to the list of XML files

jamesqo:
Yet another PR to add an XML file extension :smile:

Here's an example of a file that would benefit from this in [corefxlab](https://github.com/dotnet/corefxlab/blob/master/corefx.vssettings).

cc @arfon

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3895,6 +3895,7 @@ XML:
   - .ux
   - .vbproj
   - .vcxproj
+  - .vssettings
   - .vxml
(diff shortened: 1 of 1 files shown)


pchaigno:
[Search results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Avssettings+NOT+eeerefgfdefdef&type=Code&ref=searchresults) - All of them seem to be XML files, even though some don't have a `<?xml` header.

arfon:
:+1: thanks.

PR #2920, opened 2016-03-29
Add ".es" to recognised ECMAScript extensions

Alhadis:
The usage of `.es` as a JavaScript file extension has grown in recent times, undoubtedly due to the influx of transpilers. However, GitHub currently misclassifies these files as Erlang scripts; this PR remedies that.

The heuristic
-------------
I don't know a single word of Erlang, but it appears that `.es` files ("Erlang scripts") are an interpreted form of the language which requires an author to define a routine named `main` in the source. From the [official documentation](http://erlang.org/doc/man/escript.html):

> An Erlang script file must always contain the function **main/1**. When the script is run, the `main/1` function will be called with a list of strings representing the arguments given to the script (not changed or interpreted in any way).

I deduced that Erlang scripts could be identified by `main([]) ->`, which is also illegal JavaScript.

I extended the heuristic to check for lines beginning with `%%`. While Erlang uses `%` to initiate a comment, the doubled form is also abundant. It's also the required form of passing arguments to the emulator. To quote the docs I linked to above:

> Such an argument line must start with `%%!` and the rest of the line will interpreted as arguments to the emulator.

While it's unlikely a JavaScript file would have a modulus operator at the start of a line, it isn't invalid syntax. For instance, this is legal JavaScript (albeit really weird):
```js
let a = 2
% 1;
```

Results
----------
Not surprisingly, most search results for `.es` yielded files [localised](https://github.com/Alhadis/Linguist-Heuristics/tree/es/fixtures/Unrelated/Espa%C3%B1ol) in Spanish (esto es un buen lingüista, jajaja). All but two of the ~2,777 files collected from [search results](https://github.com/search?q=extension%3Aes+NOT+nothack&type=Code) were [correctly identified](https://github.com/Alhadis/Linguist-Heuristics/blob/es/results.txt) as either ECMAScript or Erlang:

**[App.2.es:](https://raw.githubusercontent.com/embedthis/ejscript/f85405f5c25a3335a9bf4f4a249e5f3eaf225081/samples/web/mvc-blog/blog/src/App.es)**
```js
module App {
}
```

**[locale.js.es:](https://github.com/delsovalle/prometeus/blob/d79577ff8aee2b14053e3be9f18497363fe37d30/synoman/phpsrc/web/js/locale.js.es)**
```
_={"a":"No se encuentra la página","b":"Lo sentimos, no se encuentra la página que está buscando.","c":"Anterior"};
```

I don't think those two edge cases are really worth the trouble of bloating the heuristic for. For a start, the first one isn't even real JavaScript (but a mutated superset called [Embedthis Ejscript](https://github.com/embedthis/ejscript/)). As for the second, well, I'd be lying if I said I see something like that every day.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -127,6 +127,14 @@ module Linguist
         Language["ECL"]
       end
     end
+    
+    disambiguate ".es" do |data|
+      if /^\s*(?:%%|main\s*\(.*?\)\s*->)/.match(data)
+        Language["Erlang"]
+      elsif /(?:\/\/|("|')use strict\1|export\s+default\s|\/\*.*?\*\/)/m.match(data)
+        Language["JavaScript"]
+      end
+    end
 
     disambiguate ".for", ".f" do |data|
       if /^: /.match(data)
--- a/samples/Erlang/release.es
+++ b/samples/Erlang/release.es
@@ -1,4 +1,3 @@
-#!/usr/bin/env escript
 %%!
 %-*-Mode:erlang;coding:utf-8;tab-width:4;c-basic-offset:4;indent-tabs-mode:()-*-
 % ex: set ft=erlang fenc=utf-8 sts=4 ts=4 sw=4 et:
@@ -119,4 +118,3 @@ execute_overlay([{copy, In, Out} | Rest], Vars, BaseDir, TargetDir) ->
 
 exit_code(ExitCode) ->
     erlang:halt(ExitCode, [{flush, true}]).
-
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1694,6 +1694,7 @@ JavaScript:
   - .js
   - ._js
   - .bones
+  - .es
   - .es6
   - .frag
   - .gs
--- a/samples/Erlang/170-os-daemons.es
+++ b/samples/Erlang/170-os-daemons.es
@@ -1,0 +1,26 @@
+#! /usr/bin/env escript
+
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+loop() ->
+	loop(io:read("")).
+
+loop({ok, _}) ->
+	loop(io:read(""));
+loop(eof) ->
+	stop;
+loop({error, Reason}) ->
+	throw({error, Reason}).
+
+main([]) ->
+	loop().
--- a/samples/Erlang/release
+++ b/samples/Erlang/release
@@ -1,3 +1,4 @@
+#!/usr/bin/env escript
 %%!
 %-*-Mode:erlang;coding:utf-8;tab-width:4;c-basic-offset:4;indent-tabs-mode:()-*-
 % ex: set ft=erlang fenc=utf-8 sts=4 ts=4 sw=4 et:
--- a/samples/Erlang/single-context.es
+++ b/samples/Erlang/single-context.es
@@ -1,0 +1,104 @@
+#! /usr/bin/env escript
+% This file is part of Emonk released under the MIT license. 
+% See the LICENSE file for more information.
+
+main([]) ->
+	start(64, 1000);
+main([N]) ->
+	start(list_to_integer(N), 1000);
+main([N, M]) ->
+	start(list_to_integer(N), list_to_integer(M)).
+
+
+start(N, M) ->
+	code:add_pathz("test"),
+	code:add_pathz("ebin"),
+	{ok, Ctx} = emonk:create_ctx(),
+	{ok, undefined} = emonk:eval(Ctx, js()),
+	run(Ctx, N, M),
+	wait(N).
+
+run(_, 0, _) ->
+	ok;
+run(Ctx, N, M) ->
+	Self = self(),
+	Pid = spawn(fun() -> do_js(Self, Ctx, M) end),
+	io:format("Spawned: ~p~n", [Pid]),
+	run(Ctx, N-1, M).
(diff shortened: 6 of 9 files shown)


Alhadis:
Honourable mention:
================

While looking through Erlang's mailing lists for info on Escript file extensions, I nearly died laughing at the guy [who accidentally submitted a butt-typed e-mail](http://erlang.org/pipermail/erlang-questions/2011-August/060530.html):
> [erlang-questions] Gxffhthdhchchfhdydgsfhchcjchxjx
> Tim Williams <> 
> Sun Aug 7 18:50:31 CEST 2011
>
> Dhdjjffjfffjfjghfnfjdjfjdjjdyeyehrhffjhjfugfj
>
> Sent from my iPod

BRB, rofling again forever.

lib/linguist/heuristics.rb
```diff
@@ -127,6 +127,14 @@ def call(data)
         Language["ECL"]
       end
     end
+    
+    disambiguate ".es" do |data|
+      if /^\s*(?:%%|main\s*\(.*?\)\s*->)/.match(data)
+        Language["Erlang"]
+      elsif /(?:\/\/|("|')use strict\1|export\s+default\s|\/\*.*?\*\/)/m.match(data)
```
pchaigno: For the `\/\*.*?\*\/` part, it won't match if the comment is on several lines. Would `\/\*` be sufficient?
Also, would a `^\s*var\s` make sense? `var` is very common in JavaScript.

samples/Erlang/release.es
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env escript
```
pchaigno: Why did you remove the shebang?

pchaigno:
Awesome pull request! One of the best I've ever seen ;)

samples/Erlang/release.es
Alhadis: @pchaigno I was merely following the instructions laid out in the contribution notes:

> In addition, if your new language defines an extension that's already listed in `languages.yml` ... make sure that example .foo files are present in the samples directory for ***each*** language that uses `.foo`.

The Erlang directory lacked any files with an `.es` extension.

lib/linguist/heuristics.rb
Alhadis: @pchaigno The second expression will (the one you've quoted). I've enabled the `/m` flag so it *will* span multiple lines.

As for `var`, I'm afraid not. Authors using `es` extensions will undoubtedly be using ES6 alternatives to `var`: `let` and `const`. So that's probably less reliable; conversely, most ES6 files are written in "use strict" mode.

samples/Erlang/release.es
Alhadis: Erm, and as for the shebang removal (just reread your question now... ugh, sorry), I didn't want to skew the results by telling Linguist to expect a shebang in the header of any Erlang `.es` files. Not sure if that actually has any bearing on classification, but since not every executable EScript file will have the shebang, I figure it wouldn't hurt to.

pchaigno: Hm. makes sense :-)
I think it would be preferable to add a new EScript samples rather than editing one though.

lib/linguist/heuristics.rb
pchaigno: > I've enabled the /m flag so it will span multiple lines.

I missed that :S 

Could you add a test case for these regexp in [`test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb)?
(I'm worried about regressions if we ever change them in the future.)

samples/Erlang/release.es
Alhadis: Did you want me to amend that?

lib/linguist/heuristics.rb
Alhadis: Sure thing, on it.

samples/Erlang/release.es
pchaigno: I probably should have made that clear before, but I'm not from GitHub staff, I can only give an opinion ;)

Alhadis: All good!

While you're here, erh, would you happen to know if the staff would object to me swapping out a grammar with a more detailed and accurate one?

That was the next PR I was going to submit. The header of [Lightshow](https://github-lightshow.herokuapp.com/) says "If you have an improved grammar we should be using on GitHub, please [file an issue](https://github.com/github/linguist/issues/new)." so I'd vouch the answer is "yes"... but since the grammar in question was written by me, I still feel like a prick. :(

pchaigno: If you can demonstrate that the grammar is objectively better at highlighting, sure. The maturity and the license might also be considered.

Can't you merge your improvements/fixes back into the original grammar?

Alhadis: Well I would, but there'd be too many breaking changes. The author of the [original grammar](https://github.com/StoneCypher/sublime-apl) used subpattern names that don't comply with [TextMate's recommendations on naming conventions](http://manual.macromates.com/en/language_grammars#naming_conventions)... e.g., if I were to somehow integrate my work into his, there'd be just as much (if not more) work involved with migrating the choice of class names to ensure there're no unexpected changes to users of the Sublime package.

[This is the grammar I worked on](https://github.com/Alhadis/language-apl). Aside from ironing out edge cases in numeral handling, it also covers a much wider range of keywords and has [partial support for heredocs](https://github.com/Alhadis/language-apl/blob/master/tests/gnu-web.apl#L127) (I only say "partial" because I wasn't able to set dynamic embedded highlighting based on a secondary parameter... I literally pushed TextMate's engine to its limit).

The license is ISC, functionally identical to the MIT license. I proudly release all my work under the most permissive license as possible, because restrictions are for pussies.

lib/linguist/heuristics.rb
Alhadis: Done! :)

samples/Erlang/release.es
Alhadis: I'm also working on another language package for Atom, which is for Unix manual pages. It'll actually support the entire troff typesetting language, as opposed to just a handful of macros. Biggest improvement to the existing man-page grammar is the existence of markup classes... so hopefully source code on GitHub will be a lot more readable with bold/italic/etc styling applied. =)

arfon:
👍 thanks for this @Alhadis - your Pull Requests are an absolute pleasure.

arfon:
This now live on GitHub.

PR #2921, opened 2016-03-29
Fix protocol of Nix submodule's URL

Alhadis:
While working on #2920, I noticed the [tests were failing](http://pastebin.com/VcwcrgVF) because one of the submodules didn't have a URL that used the HTTPS protocol. I checked the repository's recent commit history, and noticed [this was the culprit](https://github.com/github/linguist/commit/bd0f4f6f78).


arfon:
Oops. Thanks for spotting this!

Alhadis:
No problem! :D I'm working on another PR, so I figure I may as well take care of it.

PR #2929, opened 2016-04-01
Add .BUILD extension to Python languages

jart:
You'll notice that files named `BUILD` are already supported, which is for [Bazel](https://github.com/bazelbuild/bazel). But there are also situations where BUILD files might be named something like [closure_library.BUILD](https://github.com/bazelbuild/rules_closure/blob/master/closure/library/closure_library.BUILD). We'd like to have those syntax highlighted too.

CC: @Dominator008

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2879,6 +2879,7 @@ Python:
   ace_mode: python
   color: "#3572A5"
   extensions:
+  - .BUILD
   - .py
   - .bzl
   - .cgi


jart:
Does Google have a corporate SLA with GitHub? If not, I can't contribute this change.

Dominator008:
@jart I think there's a way you can sign one.

jart:
@Dominator008 If you want to send this pull request yourself, it would probably be better. I can't sign CLAs.

Dominator008:
@jart Sure. Is this extension official? Can you post a link to the related Bazel documentation if it is?

jart:
There are examples of it being used in the Bazel repository. For example: https://github.com/bazelbuild/bazel/blob/master/src/main/tools/jdk.BUILD Also it's the same as the name of the BUILD files.

Dominator008:
https://github.com/github/linguist/pull/2930.

PR #2930, opened 2016-04-01
Add the ".BUILD" extension to Python

Dominator008:
Files named BUILD, which are build files for the Bazel and Pants build
systems, are already highlighted. This commit adds support for files
with a ".BUILD" extension, which are also build files recognized by
Bazel.

Proposed by @jart.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2880,6 +2880,7 @@ Python:
   color: "#3572A5"
   extensions:
   - .py
+  - .BUILD
   - .bzl
   - .cgi
   - .fcgi
--- a/samples/Python/jdk.BUILD
+++ b/samples/Python/jdk.BUILD
@@ -1,0 +1,143 @@
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "jni_header",
+    srcs = ["include/jni.h"],
+)
+
+filegroup(
+    name = "jni_md_header-darwin",
+    srcs = ["include/darwin/jni_md.h"],
+)
+
+filegroup(
+    name = "jni_md_header-linux",
+    srcs = ["include/linux/jni_md.h"],
+)
+
+filegroup(
+    name = "jni_md_header-freebsd",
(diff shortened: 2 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -2879,6 +2879,7 @@ Python:
   ace_mode: python
   color: "#3572A5"
   extensions:
+  - .BUILD
```
nixel2007: Don't put your extension to first place. First extension in a list is a primary extension of files for language.

Dominator008: Oh right. I somehow forgot this magic rule :)

arfon:
Thanks for this @Dominator008. As per our [contributing guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) - would you mind linking to an example GitHub search demonstrating the usage of this extension?

Dominator008:
@afron I tried, but didn't figure out how to do a case sensitive search.

pchaigno:
@Dominator008 Unfortunately, you can't, and file extensions are case-insensitive in Linguist too.

Thus, with this pull request `.build` files would be detected as Python. There seem to be [a good number of Python files](https://github.com/search?p=1&q=extension%3Abuild+def+OR+from+OR+class&ref=searchresults&type=Code&utf8=%E2%9C%93) among `.build` files, but there are [even more non-Python files](https://github.com/search?p=66&q=extension%3Abuild+NOT+def+NOT+from+NOT+class&ref=searchresults&type=Code&utf8=%E2%9C%93) (these are coarse approximations since the search queries return many false positives).

We'd have to recognize those other languages before we can add `.build` to Python.

jart:
unnamed-8

Dominator008:
@pchaigno That's sad to hear. Is there a reason file extensions are case-insensitive? Recognizing non-python `.build` files seems harder than making file extensions case-sensitive.

arfon:
🎏 flagging this as stale 🎏

PR #2934, opened 2016-04-03
Add support for Alpine Abuild (APKBUILD)

jirutka:
APKBUILD is a file containing "recipes" on how something should be built/compiled. It's used by APK, the Alpine Linux package manager.

/cc @ncopa

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -147,6 +147,17 @@ Alloy:
   - .als
   ace_mode: text
 
+Alpine Abuild:
+  type: programming
+  group: Shell
+  aliases:
+  - abuild
+  - apkbuild
+  filenames:
+  - APKBUILD
+  tm_scope: source.shell
+  ace_mode: sh
+
 Ant Build System:
   type: data
   tm_scope: text.xml.ant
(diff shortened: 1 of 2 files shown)


arfon:
Hi @jirutka, thanks for opening this Pull Request :zap:

Please take a look at our [contributing guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) to finish up the additional steps required before review :smile:

jirutka:
Okay, here it is: [Search filename:APKBUILD](https://github.com/search?utf8=%E2%9C%93&q=filename%3AAPKBUILD&type=Code&ref=searchresults) – 42,363 code results

Or did you mean some different steps?

pchaigno:
Shouldn't we add `APKBUILD` to the Shell filenames list instead of adding it as a new language?

jirutka:
I don’t think so. Abuilds are not interpreted (directly) by sh, but apk. Alpine Abuild is very similar to Gentoo Ebuild that is already defined in the linguist as a language (see [here](https://github.com/github/linguist/blob/f28cdc8a15fa1fe7f149b56cc13fbe5527877e1a/lib/linguist/languages.yml#L1157-L1163)).

pchaigno:
> Alpine Abuild is very similar to Gentoo Ebuild that is already defined in the linguist as a language (see here).

I'm afraid that's a mistake. Gentoo Ebuild files are counted and highlighted as Shell, so they should directly be defined as a new extension for Shell. I'll open a pull request to fix that.

jirutka:
See https://github.com/github/linguist/pull/2935#issuecomment-204991645 – where’s the line between a “language” and a DSL? I consider ebuild and abuild as a DSL.

PR #2935, opened 2016-04-03
Add support for OpenRC runscripts

jirutka:


arfon:
Hi @jirutka, thanks for opening this Pull Request :zap:

Please take a look at our [contributing guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) to finish up the additional steps required before review :smile:

jirutka:
Okay, here it is: [Search openrc-run](https://github.com/search?utf8=%E2%9C%93&q=openrc-run&type=Code&ref=searchresults) – 4,660 code results

Or did you mean some different steps?

pchaigno:
Is OpenRC really a language? Or should we just add the `openrc-run` interpreter to Shell?

jirutka:
To be honest, I don’t know where’s the line between “language” and DSL built on top of existing language.

Maybe I should note that OpenRC runscript is _not_ similar to SysV init script – it’s not and usually can’t be run directly using sh. openrc-run specifies many variables and functions for “convention over configuration”; typical runscripts declares just few variables and procedural code (i.e. declarative configuration). So it’s a DSL built on top of sh.

davidzchen:
I agree with @jirutka that the guidelines around what should or should not be considered a separate "language" in Linguist needs to be more well-defined. Another example of this is Gradle, which is syntactically a Groovy DSL but is [marked as a separate language in Linguist](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml). Similarly, there are separate languages defined for [Maven Pom](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L2139) and [Ant](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L150), even though both are syntactically XML files.

To answer this, I think we should consider what the purpose of Linguist is to GitHub's users. To me, Linguist allows me to understand the language makeup of different projects on GitHub and helps me search for source files in specific language. In this case, if I want to search specifically for OpenRC runscripts, Alpine Abuild files, Gradle build files, or Bazel build extensions (.bzl files), it is less useful and less productive if I have to sift through a bunch of regular shell scripts, Groovy source files, and Python source files to find what I am looking for.

As a result, I think there is a lot of value to GitHub's users if files that are specific to a DSL are categorized separately from regular source files written in that language. However, currently, the policy around whether and to what extent DSLs built on top of a language should be categorized separately from regular source files written in that language are not well-defined and should be made more clear.

jirutka:
@arfon ping

arfon:
> Maybe I should note that OpenRC runscript is not similar to SysV init script – it’s not and usually can’t be run directly using sh. openrc-run specifies many variables and functions for “convention over configuration”; typical runscripts declares just few variables and procedural code (i.e. declarative configuration). So it’s more a DSL built on top of sh.

In that case perhaps we should add it as a separate language but put it in the `Shell` group? That would mean that it was searchable on GitHub (i.e. it's a real, differentiated language) but it would be included in the `Shell` category in the language bar on repositories that include OpenRC runscripts

jirutka:
Okay, I’ve added `group: Shell` and rebased.

arfon:
👍 thanks for this @jirutka. This will be live in the next release of Linguist (next week sometime)

jirutka:
@arfon That’s great, thanks! Could you please also include #2934?

PR #2938, opened 2016-04-06
Add "mkfile" as a recognised Makefile name

Alhadis:
Some older Unix-like systems used the name `mkfile` for their system's Makefiles - most notably [Plan 9](https://github.com/codesrxxx/plan9_source/blob/master/sys/src/mkfile). A search result for `mkfile` yields a whooping [+32k results](https://github.com/search?utf8=%E2%9C%93&q=filename%3Amkfile&type=Code&ref=searchresults), many of which are probably copies of distributed source code.

There're also about [~116 matches](https://github.com/search?q=extension%3Amkfile+NOT+nothack&type=Code) for `mkfile` in GitHub's file extension search, so it probably warrants inclusion as a file extension addition too.

People seemed rather laid-back about file extensions in those days...

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2068,6 +2068,7 @@ Makefile:
   - .mak
   - .d
   - .mk
+  - .mkfile
   filenames:
   - GNUmakefile
   - Kbuild
@@ -2076,6 +2077,7 @@ Makefile:
   - Makefile.in
   - Makefile.inc
   - makefile
+  - mkfile
   interpreters:
   - make
   ace_mode: makefile
--- a/samples/Makefile/filenames/mkfile
+++ b/samples/Makefile/filenames/mkfile
@@ -1,0 +1,25 @@
+TYPE=gemstones
+GEM=amethyst bloodstone citrine emerald garnet jade lapis-lazuli moonstone opal perl ruby sapphire topaz unakite variscite wernerite xanthite yttrotitanite zincostaurolite
+
+all:G:
+	cd /sys/log
+	gem=`{date|sed 's/(^[^ ]*) .*/\1/'}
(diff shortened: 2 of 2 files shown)


Alhadis:
**NOTA BENE:** The [test fixture](https://github.com/Alhadis/linguist/blob/mkfile/samples/Makefile/filenames/mkfile) I added for the extension is a mutated form of [this makefile](https://github.com/brho/plan9/blob/89d43d2262ad43eb4b26c2a8d6a27cfeddb33828/sys/lib/sysconfig/auth/boundary/sys.log.mkfile), which appears to have been distributed under a GPLv2 license.

I've swapped out the command names and variables so it's mostly a heap of useless garbage that's still syntactically-shaped in the form of the original. I'm aware of the issues surrounding the GPL license, especially concerning distribution, so please let me know if this is an issue.

arfon:
> I'm aware of the issues surrounding the GPL license (especially concerning distribution), so I've swapped out the command names and variables to make it a heap of useless garbage that's still syntactically-shaped in the form of the original. Please let me know if this is an issue.

Yeah, I'm afraid it is. To be on the safe side would you mind looking around a little more to see if you can fine a `mkfile` with a more permissive license? 🙏

Alhadis:
Sure thing. =)

(Bloody GPL...)

Alhadis:
Sadly, almost every `mkfile` I'm finding seems released under a GPL license, or encumbered by something similar.

However, I noticed patterns in the way each one was written, and used those to write a new `mkfile` from scratch. That should be enough. =)

arfon:
👍 thanks @Alhadis

arfon:
This will be live in the next release of Linguist (likely sometime in the next week).

PR #2945, opened 2016-04-13
Added Django env folder to Vendors

AnimeshSinha1309:
The Python local environment folder stores all the dependencies of the project like django, pillow and other libraries in Reqirements.txt. Since those are 3rd party apps and libraries, they should be excluded from the language statistics. So the python **env folder** has been added to vendors.yml.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -183,7 +183,8 @@
 # Knockout
 - (^|/)knockout-(\d+\.){3}(debug\.)?js$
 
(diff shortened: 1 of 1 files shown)


arfon:
Thanks for this @AnimeshSinha1309. Would you mind adding a test for this new path? You can see examples of how to do this here: https://github.com/github/linguist/blob/master/test/test_file_blob.rb#L266

arfon:
Closing in favour of #2992

Issue #2637, opened 2015-09-16
Add GraphQL syntax highlighting to GFM code fences


steveluscher:
Opening this issue to track the addition of GraphQL syntax highlighting to GFM.

    ```graphql
    query FooQuery {
      field(call:$value) @directive(awesome:true, number:42, strings:"yes") {
        subfield as alias
      }
    }
    ```

Use the syntax highlighter from https://github.com/rmosolgo/language-graphql

[...]

PR #2937, opened 2016-04-06
Switch grammar used for APL highlighting

Alhadis:
I'm not sure if the differences are glaring enough to make switching this a clear-cut decision, but I figure I'd try anyway. I was hesitant to propose switching grammars, because the one being switched to is mine, [which I wrote for Atom](https://atom.io/packages/language-apl).

Anyway... the grammar being used to highlight APL on this site is... okay, but it leaves out a lot. It highlights symbols only, ignoring their context. For example, `⍵` and `⍶` only have significance when used inside the body of a user-defined function or operator, but the grammar highlights them irrespective of where they're used.

Its coverage of the APL alphabet is also incomplete:

* [Current grammar](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=source.apl&grammar_url=&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2FAlhadis%2Flanguage-apl%2Fmaster%2Ftests%2Fsymbols.apl&code=text+%E2%86%90+%E2%8E%95INP+%22END-OF-HTML%22%0D%0A%3Cheader+id%3D%22top%22%3E%E2%8D%9D+Heading+%E2%8D%9D%3C%2Fheader%3E%0D%0Aasfdwger%3D%3D%3D%3D%3D+END-OF-HTML+%3D%3D%3D%3Dq3ewrfd..%3F)
* [New grammar](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=http%3A%2F%2Fpastebin.com%2Fraw%2F5X7kA5D0&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2FAlhadis%2Flanguage-apl%2Fmaster%2Ftests%2Fsymbols.apl&code=text+%E2%86%90+%E2%8E%95INP+%22END-OF-HTML%22%0D%0A%3Cheader+id%3D%22top%22%3E%E2%8D%9D+Heading+%E2%8D%9D%3C%2Fheader%3E%0D%0Aasfdwger%3D%3D%3D%3D%3D+END-OF-HTML+%3D%3D%3D%3Dq3ewrfd..%3F)

**NOTE:** The symbols categorised as "Unused" in the above example are defined by the APL standard, but aren't recognised as functions in many interpreters. They may be implemented as functions by interpreters I'm unfamiliar with, though.

Comparisons
-----------
[...]

PR #2947, opened 2016-04-14
Add support for GraphQL

gjtorikian:
Closes https://github.com/github/linguist/issues/2637.

I noticed @charliesome mentioned wanting support for this. Seems like a good idea!

### DIFF
--- a/vendor/licenses/grammar/language-graphql.txt
+++ b/vendor/licenses/grammar/language-graphql.txt
@@ -1,0 +1,7 @@
+---
+type: grammar
+name: language-graphql
+license: mit
+curated: true
+---
+MIT
--- a/.gitmodules
+++ b/.gitmodules
@@ -707,3 +707,6 @@
 [submodule "vendor/grammars/language-apl"]
 	path = vendor/grammars/language-apl
 	url = https://github.com/Alhadis/language-apl.git
+[submodule "vendor/grammars/language-graphql"]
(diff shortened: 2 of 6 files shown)


kdaigle:
😍 Thank you @gjtorikian! You're my only hope.

charliesome:
Oh my 😻

gjtorikian:
@charliesome Yep! I don't know if it's the preferred way but there's some mechanism that converts the CSON grammar so that you can use the same definitions in Linguist and Atom. For example, here's [a recent PR](https://github.com/github/linguist/pull/2937) that also opts to use a language built for Atom.

vendor/licenses/grammar/language-graphql.txt
```diff
@@ -0,0 +1,7 @@
+---
+type: grammar
+name: language-graphql
+license: mit
+curated: true
+---
+MIT
```
bkeepers: This should have attribution (`Copyright 2016 Author McName`) and the full text of the license.

arfon: Well perhaps Licensed should have picked that up 😛. Will fix shortly.

arfon: Fixed in 5c7aa5406af530d563846ed09d2a3d337eb8fadc

steveluscher:
<3

PR #2963, opened 2016-04-21
Classify .jscsrc and .eslintrc

jasonwilliams200OK:
Note that  supports both JSON and YAML formats (http://eslint.org/docs/user-guide/configuring).
Therefore, I have added a disambiguation rule.


jasonwilliams200OK:
@arfon, I have fixed a test_classfier bug in second commit. Please have a look and let me know if that would cause any (false-positive) issue.

lib/linguist/languages.yml
```diff
@@ -4016,13 +4018,13 @@ YAML:
...
   - yml
   extensions:
-  - .yml
   - .reek
   - .rviz
   - .sublime-syntax
   - .syntax
   - .yaml
   - .yaml-tmlanguage
+  - .yml
```
arfon: This needs to go back to the top of the list of extensions. The first in the list is considered the 'primary' extension for the language.

jasonwilliams200OK: Done.

test/test_classifier.rb
```diff
@@ -50,7 +50,7 @@ def test_classify_ambiguous_languages
       next unless languages.length > 1
 
       results = Classifier.classify(Samples.cache, File.read(sample[:path]), languages)
-      assert_equal language.name, results.first[0], "#{sample[:path]}\n#{results.inspect}"
+      assert results.map{|key,value| key}.include?(language.name), "#{sample[:path]}\n#{results.inspect}"
```
pchaigno: I'm wondering whether it was written that way on purpose... I guess, because we trained the classifier in part with the sample we are now trying to classify, it should be correctly detected and have the strongest score assigned (thus being the first in the list).

lib/linguist/heuristics.rb
```diff
@@ -136,6 +136,14 @@ def call(data)
       end
     end
 
+    disambiguate ".eslintrc" do |data|
+      if /^{/.match(data)
```
pchaigno: Could you add a test in [`test/test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb) for that heuristic rule?

test/test_classifier.rb
jasonwilliams200OK: Can you tell exactly why YAML is first in the list when:

* JSON is passed `samples/JSON/filenames/.eslintrc`
* heuristic is correctly classifying ` samples/JSON/filenames/.eslintrc` as JSON and ` samples/YAML/filenames/.eslintrc` as YAML.

pchaigno: The above line only uses the Bayesian classifier to detect the language. So the [filename](https://github.com/github/linguist/blob/master/lib/linguist/strategy/filename.rb) and [heuristic](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb) strategies are not used here.

lib/linguist/heuristics.rb
jasonwilliams200OK: Done.

test/test_heuristics.rb
```diff
@@ -91,6 +91,13 @@ def test_es_by_heuristics
     })
   end
 
+  def test_eslintrc_by_heuristics
+    assert_heuristics({
+      "JSON" => all_fixtures("JSON", ".eslintrc"),
```
pchaigno: You have to put an asterix before the dot, otherwise it doesn't match anything ;)
In this case, I think you could simply use
```ruby
    "JSON" => all_fixtures("JSON"),
    "YAML" => all_fixtures("YAML"),
```
as the heuristic should work for any JSON or YAML file, irrespective of the extension.

lib/linguist/heuristics.rb
pchaigno: :+1:

test/test_heuristics.rb
jasonwilliams200OK: > you could simply use

That does not work. The heuristic under question is only testing `.eslintrc` file. The rest of the files under samples/JSON dir (for example `*.lock`) are listed in `languages.yml` as extensions, which are not classified by heuristic at all.

Looking at the rest of this file how single files are handled, I have made the change accordingly:

```
"JSON": ["samples/JSON/filenames/.eslintrc"]
```

rebased and repushed

pchaigno: Oo, right. I forgot we made that change :-)

test/test_classifier.rb
jasonwilliams200OK: Ah, didn't knew about Bayesian classifier. Could it be the bug in scoring? For `samples/JSON/filenames/.eslintrc`, the classifier is returning this score:

```
"YAML": -38.629626828197445
"JSON": -56.866265060139305
```

When clearly the contents of that file constitute valid JSON.

arfon: Yeah I don't think we want to make this change as I think the ordering _does_ matter. My best guess for what's going on here is that after tokenizing both of the files the Bayesian classifier isn't able to figure out the difference between the two reliably.

jasonwilliams200OK: @arfon, thanks for the info. Should I investigate and try to fix Bayesian classifier scoring for this particular case as part of this PR? Because, without this second commit, tests are failing (on CI as well).

PR #2964, opened 2016-04-22
Add .app.src as Erlang extension.

thwg:
*.app.src is a convention that modern build tools use as input to describe an Erlang application.

In-the-wild search doesn't pull up any results, I think due to the period in the extension, but it should be popular among Erlang projects.

### DIFF
--- a/samples/Erlang/sample.app.src
+++ b/samples/Erlang/sample.app.src
@@ -1,0 +1,8 @@
+{application, sample,
+ [{description, "sample app"},
+  {vsn, "1.0.0"},
+  {registered, []},
+  {mod, {sample_app, []}},
+  {applications, [kernel, stdlib]},
+  {env, []},
+  {modules, []}]}.
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -948,6 +948,7 @@ Erlang:
   color: "#B83998"
   extensions:
   - .erl
+  - .app.src
   - .es
   - .escript
   - .hrl


samples/Erlang/sample.app.src
```diff
@@ -0,0 +1,8 @@
+{application, sample,
```
pchaigno: That doesn't look like the usual Erlang file... Is there a reason for that?

pchaigno:
A few thousands can be found with the following search: https://github.com/search?p=1&q=filename%3A.app.src+extension%3Asrc&ref=searchresults&type=Code&utf8=%E2%9C%93

samples/Erlang/sample.app.src
thwg: *.app.src is a declarative Erlang data structure, rather than an executable module like *.erl.

[rebar.config](https://github.com/github/linguist/blob/master/samples/Erlang/filenames/rebar.config) is similar example.

pchaigno: Unfortunately, because it's not the same syntax, [it won't be highlighted properly](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fgithub.com%2Ftextmate%2Ferlang.tmbundle%2Fblob%2F39c144ddb7a39bdb967c893a33fda27482beb770%2FSyntaxes%2FHTML%2520%2528Erlang%2529.tmLanguage&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fthwg%2Flinguist%2Fblob%2Fapp-src%2Fsamples%2FErlang%2Fsample.app.src&code=). Do you know a Sublime Text/Atom/TextMate package that could be used to highlight those files?

thwg: Unless I'm missing something, it should highlight correctly.  It's the same syntax.

If a mode-line `-*- mode: erlang; -*-` comment is added to the top of the file, [it works correctly](https://github.com/erlang/rebar3/blob/master/src/rebar.app.src).

pchaigno: My bad. I selected the wrong grammar. It is indeed [highlighted properly](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fgithub.com%2Ftextmate%2Ferlang.tmbundle%2Fblob%2F39c144ddb7a39bdb967c893a33fda27482beb770%2FSyntaxes%2FErlang.plist&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fthwg%2Flinguist%2Fblob%2Fapp-src%2Fsamples%2FErlang%2Fsample.app.src&code=).

arfon:
Thanks for this contribution @thwg :zap:. This will be live in the next release of Linguist (likely sometime in the next week).

Issue #2968, opened 2016-04-23
Javascript in repos


staabm:
The repo https://github.com/krakjoe/bundle doesnt contain a single line javascript, but the github language statistics shows ~36% javascript

Is this intended?

pchaigno:
[This file](https://github.com/krakjoe/bundle/blob/c1d5ed2c2b5fc137b610787b0d9a39e312bbdd11/Makefile.frag) (`Makefile.frag`) is detected as JavaScript because [`.frag` is defined as a potential file extension](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1710). What language should is be detected as?

staabm:
It should be a makefile

pchaigno:
I will add `Makefile.frag` as a [filename for Makefile](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L2083). Can I use your file as a sample (there isn't a license in your repository)?

krakjoe:
License added, thanks :)

pchaigno:
Thanks!

#2969 should fix this ;)

arfon:
Fixed in #2969

PR #2969, opened 2016-04-24
Makefile.frag as a Makefile filename

pchaigno:
This pull request adds `Makefile.frag` as a Makefile filename, as discussed in #2968.
There are [around 3,800 `Makefile.frag` files](https://github.com/search?utf8=%E2%9C%93&q=filename%3AMakefile.frag) on GitHub.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2085,6 +2085,7 @@ Makefile:
   - Kbuild
   - Makefile
   - Makefile.am
+  - Makefile.frag
   - Makefile.in
   - Makefile.inc
   - makefile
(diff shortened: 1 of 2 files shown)


staabm:
per https://github.com/github/linguist/pull/2973 it looks like this should already be  deployed to production...?

stats at https://github.com/krakjoe/bundle still point out that 36% is javascript.

arfon:
@staabm - looks like there was a stale cache somewhere for the page. Things are looking better now:

screen shot 2016-05-03 at 8 09 44 am

staabm:
@arfon confirmed, thx for the fast response!

PR #2971, opened 2016-04-28
Add support for World of Warcraft .toc files

Alhadis:
I'm less interested in [the data format](http://wow.gamepedia.com/TOC_format) than I am fixing its widespread misclassification: searching for [`.toc`](https://github.com/search?q=extension%3Atoc+NOT+nothack&type=Code) files tosses up an impressive number of results: most of them WoW-related.

This misclassification sharply skews TeX's usage statistics, so it's worth addressing. As usual, I collated public `.toc` files from extension search results to test classification accuracy; the [results are here](https://github.com/Alhadis/Linguist-Heuristics/blob/toc/results.log). Since I've classified WoW's addon data as... well, data, it doesn't show up in the breakdown graph. Which I'm guessing is the expected behaviour, really.


pchaigno:
> Since I've classified WoW's addon data as... well, data, it doesn't show up in the breakdown graph. Which I'm guessing is the expected behaviour, really.

You can temporarily change its type to `programming`. Otherwise it won't show up in results and won't be counted in the overall pourcentage ;)

Alhadis:
Heh, yeah, that occurred to me only after I'd scanned the files halfway. I felt that scanning the other half was probably going to be quicker than rebuilding the program.

Alhadis:
Conflicts fixed.

pchaigno:
How did you select the samples for tests? How did you distinguish WoW files from TeX files?

Alhadis:
According to the [format's documentation](http://wow.gamepedia.com/TOC_format), WoW's metadata files provide tags to store auxiliary information to a consuming client. The tags are identified by lines beginning with "## ", with every other line being either a comment "# " or a filepath.

As for TeX, well, that's pretty easy. Files containing lines beginning with `\contentsline`, `\defcounter`, `\beamer` or `\boolfalse` are guaranteed to be TeX instead of WoW-metadata.

The samples were just random crap I wrote. The TeX fixtures were cobbled together using titles of the [German Wikipedia page](https://de.wikipedia.org/wiki/Perl_%28Programmiersprache%29) for Perl.

Alhadis:
@arfon the grammar's repository is MIT-licensed. What do I have to do to get the build to succeed...?

arfon:
Sorry @Alhadis - since you opened this PR we added an additional check for licenses. You need to do this too:

```
Download the license for the grammar: script/licensed. Be careful to only commit the file 
for the new grammar, as this script may update licenses for other grammars as well.
```

Alhadis:
Hahaha! :D Man, talk about crap timing... alright, give me a minute.

arfon:
😄⚡

Alhadis:
Done! =) Sorry that took so long. 😢

arfon:
Thanks @Alhadis. Looks like the current build failure is unrelated to this PR. I'm working on a fix.

arfon:
@Alhadis would you mind cherry picking this commit into your PR? https://github.com/github/linguist/pull/2983/commits/eeae8d4839b92e89b1da77536ad4dd020421839c

It fixes the grammar failures.

Alhadis:
Cherry picked, sir! =)

arfon:
Awesome thanks for this! (And your patience!)

Alhadis:
Haha, any time my friend. :D Always a pleasure to help.

nebularg:
Good to see this! Might be a bit late, but I should point out having "@no-lib-strip@" as part of the heuristic is a bit of a misnomer, it really has nothing to do with the toc format, they're tags used by a popular distribution site for packaging and are used in lua files as well. "^## Interface" would probably be the best test, as it's the only mandatory tag.

Alhadis:
Well, Bayesian analysis handles most of the classification. It's when the classifier can't decide on a file that it consults the heuristics (AFAIU). The heuristics are there mainly to "point" the classifier in the right direction. To the best of my understanding, this is how it "learns" which results are wrong and which are accurate. @pchaigno may be able to offer a more informed explanation.

Anyway, since the heuristic is already based on lines that begin with `## `, I felt a different-looking token would be more helpful to correctly classify WoW-ToC. Incidentally, a WoW that was still being misclassified as TeX was correctly identified once I'd added the `@no-lib-strip@` part.

pchaigno:
> It's when the classifier can't decide on a file that it consults the heuristics (AFAIU).

Actually, it's the other way around :wink: 
When the heuristic strategy fails to determine the language, the classifier is used (since it's the less reliable strategy). You can see the [current order of strategies in `linguist.rb`](https://github.com/github/linguist/blob/9ae19a1f9470a6f15a1ae70f870c044f55e9d8f5/lib/linguist.rb#L61-L65).

Alhadis:
Ah rats, whoops. Guess that makes a lot more sense from a performance standpoint too... thanks!

nebularg:
> Anyway, since the heuristic is already based on lines that begin with ##, I felt a different-looking token would be more helpful to correctly classify WoW-ToC.

My point was that any WoW toc will have "## Interface: XXXXXX" so it's a better test, where as @no-lib-strip@ is pretty arbitrary.

Alhadis:
Hrm, fair enough.

Either way, the heuristics are working perfectly; [short of 3,000 files](https://github.com/Alhadis/Linguist-Heuristics/tree/toc) were correctly classified as WoW or TeX. Only four escaped correct classfication:

- [ChatKeys.toc](https://github.com/Alhadis/Linguist-Heuristics/blob/toc/fixtures/kek/ChatKeys.toc)
- [login.toc](https://github.com/Alhadis/Linguist-Heuristics/blob/toc/fixtures/kek/login.toc)
- [nagrand.toc](https://github.com/Alhadis/Linguist-Heuristics/blob/toc/fixtures/kek/nagrand.toc)
- [test.toc](https://github.com/Alhadis/Linguist-Heuristics/blob/toc/fixtures/kek/test.toc)

There was no hope for the last three to be correctly identified anyway, given their incomplete content.

`ChatKeys` is somewhat unusual though... it *should* be matching, but it isn't... @pchaigno, has Linguist had any difficulties with certain line-endings before? That one file is using old-style `CR` terminators (used on ancient Mac OS systems). Could that have something to do with it?

pchaigno:
> has Linguist had any difficulties with certain line-endings before?

Not that I know.

> That one file is using old-style CR terminators (used on ancient Mac OS systems). Could that have something to do with it?

That's exactly what's happening. I tried on the same file with LF newlines and it worked fine.
I'm not sure how we should proceed...

Alhadis:
I think this is worth opening a new issue for, quite frankly. :| If line-endings have the ability to affect language recognition, then this definitely isn't a trivial issue.

pchaigno:
Agree :-)

Alhadis:
[Filed](https://github.com/github/linguist/issues/2988). =)

PR #2978, opened 2016-05-02
Removed Jade from html group

### LINKED ISSUE #2977: Jade is detected but listed as html anyway
The github language bar shows the files as being HTML however if you click on them it will say that no files are found. Searching for jade then shows the files properly.
For example:
[Main page where language is reported as html](https://github.com/zacps/go-wiki)
[HTML search returning nothing](https://github.com/zacps/go-wiki/search?l=html )
[Jade search returning files properly.](https://github.com/zacps/go-wiki/search?l=jade )

zacps:
Jade should not be included under html because it serves a very different purpose, while it does compile to html it doesn't look like html and is a language for templating  rather than just markup.

Fixes #2977

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1662,7 +1662,6 @@ JSX:
   ace_mode: javascript
 
 Jade:
-  group: HTML
   type: markup
   extensions:
   - .jade


grubern:
GitHub, in its sole discretion, has the right to suspend or terminate your account and refuse any and all current or future use of the Service, or any other GitHub service, for any reason at any time. Such termination of the Service will result in the deactivation or deletion of your Account or your access to your Account, and the forfeiture and relinquishment of all Content in your Account. GitHub reserves the right to refuse service to anyone for any reason at any time.................!

arfon:
Hi @zacps, thanks for opening this PR. 

While I believe I understand your motivations for making this change, we've chosen to implement language groups in Linguist as we believe they represent higher-level logical groupings for collections of languages. For example the `HTML` group includes:

- [Jade](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1665)
- [HTML+Django](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1377)
- [HTML+ERB](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1401)
- [HAML](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1433)
- [Handlebars](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1442)
- ...and many more

Quoting from http://jade-lang.com/reference/

> Jade is a terse language for writing HTML templates.

Which I think makes Jade a good candidate for the `HTML` group as currently defined.

> Jade should not be included under html because it serves a very different purpose, while it does compile to html it doesn't look like html and is a language for templating rather than just markup.

This argument can be made for pretty much every example in the list above too.

My question to you is: is there a usability problem that you're facing here? One thing I do think needs fixing on GitHub is what happens if you click the language bar for a repository and view the search results.

Currently, clicking the language bar on https://github.com/zacps/go-wiki shows you that a bunch of the files are in `HTML`, when this is clicked though...

click

you get sent to this page:

screen_shot_2016-05-06_at_11_42_52_am

which I think is pretty confusing for all. My suggested fix is that we work out how to link these groups properly to the languages in the group (in this case we would show the `Jade` results).

The alternative is that we simply remove the groups from Linguist but I think they serve a useful purpose right now.

zacps:
Yeah that's a fair point, I can't say I would know how to change that behaviour though. I guess wherever the handler for the bar is it's looking at the group name rather than the name itself.

nazar-pc:
There are languages like CoffeeScript/LiveScript, which are shown separately from JavaScript.
I think this is very similar to how Jade/Pug, HAML and others are different from HTML, so I'd also like to see them be separated from HTML group.

PR #2984, opened 2016-05-04
Detect .js files with source maps as generated

jkrems:
This covers compile-to-JavaScript languages other than CoffeeScript. E.g. it come up in one of our projects for files with JSX code that were compiled to "plain JavaScript" using babel. If the file has a source map reference, it's pretty much by definition generated from another file. :)

This also adds an additional check to the other `generated?` tests. Previously it was possible to comment out the `minified_files? ||` and the tests would still happily pass.

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -59,6 +59,7 @@ module Linguist
       godeps? ||
       generated_by_zephir? ||
       minified_files? ||
+      has_source_map? ||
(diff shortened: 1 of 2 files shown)


arfon:
Thanks for this contribution @jkrems :zap:. This will be live in the next release of Linguist (likely sometime in the next week).

jkrems:
Thanks!

PR #2981, opened 2016-05-04
Add support for TLA+

agentultra:
This patch adds support for the TLA+ specification language.

https://github.com/search?utf8=%E2%9C%93&q=MODULE+extension%3Atla&type=Code&ref=searchresults

PR #2990, opened 2016-05-06
TLA revisited

arfon:
Replaces https://github.com/github/linguist/pull/2981 because of duplicate entry for `TLA` in languages.ym


arfon:
Thanks for this @agentultra. This will be live on GitHub next week sometime.

PR #2992, opened 2016-05-06
2945 local

arfon:
Replaces #2945


arfon:
Thanks for this @AnimeshSinha1309 - this will be live in the next release of Linguist (next week sometime)

PR #3001, opened 2016-05-11
Remove .lock from JSON extentions

snh:
#1451 added the `.lock` extension to the JSON language, as a result of an existing test for `composer.lock` added in #818.

`.lock` is used for many purposes, the most common of which being lock files. JSON does not appear to be a common format for `.lock` files.

This PR removes `.lock` from the JSON language, and updates a test and sample for `composer.lock`, which remains JSON (which itself was added in #790)

[This](https://github.com/search?utf8=%E2%9C%93&q=extension%3Alock+lock&type=Code&ref=searchresults) search returns a number of examples of `.lock` files that should not be classified as JSON.


arfon:
:+1: makes sense. Thanks @snh

PR #3007, opened 2016-05-18
Fix incorrect docs on vendored files & restore docs on generated files

russelldavis:
Vendored files *are* displayed in diffs on github.com, so I removed the line that said the opposite.

The **Generated file detection** section is restored from a previous version of the readme (with "unlike vendored and documentation files" added as a clarification). I'm not sure why it was originally removed -- perhaps related to the confusion over diffs on vendored files -- but it's still relevant and useful (e.g., the section is linked to at http://stackoverflow.com/a/24382933/278488, and that link is now broken).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ special-vendored-path/* linguist-vendored
 jquery.js linguist-vendored=false
 ```
 
-Similar to vendored files, Linguist excludes documentation files from your project's language stats. (Unlike vendored files, documentation files are displayed in diffs on github.com.) [lib/linguist/documentation.yml](lib/linguist/documentation.yml) lists common documentation paths and excludes them from the language statistics for your repository.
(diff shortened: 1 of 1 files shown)


russelldavis:
:crickets:

jdennes:
This looks good. Thanks. 👍 

cc: @arfon This fixes documentation in `README.md` which currently implies that vendored files are suppressed in diffs, which is *incorrect*.

Also cc: @donal as this recently confused someone using GitHub Enterprise, which you dealt with.

Also ref: https://github.com/github/linguist/issues/2206#issuecomment-77637869.

donal:
Great change, thanks @russelldavis :+1:

arfon:
👍 thanks @russelldavis. Apologies for missing this PR and thanks for clearing up this confusion.

PR #3010, opened 2016-05-21
Add "README.1ST" to recognised text-file names

Alhadis:
Fairly trivial addition: [there're many readme files](https://github.com/search?q=filename%3AREADME.1st+NOT+nothack&type=Code) using `.1st` as a file extension (to the surprise of nobody). Adding `.1st` as an extension was tempting, but I did find [some](https://github.com/gatesvp/UltimatePsionics/blob/ca037de77cbf16a31f17c6908c734033de10a054/PFRPG_Psionics_Unleashed_Tags.1st) [files](https://github.com/jgshort/jgshort.github.io/blob/be374cfbfad173762d08a73d6d393e3755ccde76/downloads/code/wordptr/fnptrs/test.1st) that obviously weren't pure text, so I've added it as a filename instead.

The contents of the sample are randomly-generated excerpts of [Lorem Ipsum](http://lipsum.com/feed/html) translated into French for the amusement of @pchaigno.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3669,6 +3669,7 @@ Text:
   - INSTALL
   - LICENSE
   - NEWS
+  - README.1ST
   - README.me
   - click.me
   - delete.me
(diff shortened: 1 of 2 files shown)


arfon:
:zap:

PR #3024, opened 2016-05-30
Add ACS (Action Code Script)

agwilt:
ACS is a compiled language for the ZDoom game engine. From the ZDoom wiki:
> ACS (Action Code Script) is the scripting language that was originally created for Hexen by Raven Software and has been greatly expanded by ZDoom.

**Examples on [GitHub](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aacs+NOT+nothack&type=Code&ref=searchresults):**
- alexey-lysiuk/Realm667-AAA
- dptetc/membrance2
- wormt/bcc
- coelckers/gzdoom
- dptetc/PC96-Doom
- csnxs/retromaze
- csnxs/hydrogen
- StrikerMan780/shadow-mavericks-multi-purpose-mod
- neonplayer/Rop
- skylerdavidson/real-guns-hardcore
- agwilt/zdoom-insurrection
- rheit/zdoom
- rheit/acc <-- Well, that's the standard ACS compiler ..
- and so on

**File extension:** .acs

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -36,10 +36,11 @@ ABAP:
 
 ACS:
   type: programming
-  color: "#D70000"
+  color: "#430000"
   extensions:
   - .acs
   ace_mode: c_cpp
+  tm_scope: none
 
 AGS Script:
   type: programming
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -39,8 +39,7 @@ ACS:
   color: "#D70000"
   extensions:
   - .acs
-  tm_scope: none
-  ace_mode: text
+  ace_mode: c_cpp
 
 AGS Script:
   type: programming
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -34,6 +34,14 @@ ABAP:
   - .abap
   ace_mode: abap
 
+ACS:
+  type: programming
+  color: "#D70000"
+  extensions:
+  - .acs
+  tm_scope: none
+  ace_mode: text
+
 AGS Script:
   type: programming
   color: "#B9D9FF"
--- a/samples/ACS/movement.acs
+++ b/samples/ACS/movement.acs
@@ -1,0 +1,75 @@
+#library "movement"
+#include "zcommon.acs"
+
+int leanleft = 0;
+int leanright = 0;
+int x;
+int y;
+int z;
+int PlayerInput;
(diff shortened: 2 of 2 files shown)


arfon:
Thanks for opening this Pull Request @agwilt. I'm trying to figure out if there are any other files using the `.acs` extension - is [this file](https://github.com/fingolfin/gap-osx-binary/blob/b7ea78241904c5eb6d3726fb1ee627b48592e1c0/lib/gap/small/id3/id8.acs) Action Code Script too?

agwilt:
Oh, good point. That doesn't seem to be ACS. But the .acs extension doesn't seem to generally apply to that type of file, as they are all named things like id8.acs, id8.abm, id8.bn, etc.

arfon:
> Oh, good point. That doesn't seem to be ACS. But the .acs extension obviously doesn't generally apply to that type of file, as they are all named things like id8.acs, id8.abm, id8.bn, etc.

The issue here is that as `ACS` is the first language to define `.acs` as a valid extension, that means that as currently implemented in this Pull Request *all* `.acs` files on GitHub will be classified as `ACS` by Linguist (which isn't what we want to do).

This isn't a show-stopper but it does mean that before we can consider merging this Pull Request we're going to have to audit what other languages are using the extension (e.g. I found some examples that were XML) and add `.acs` to those language definitions too.

@agwilt - if you're willing to do this that would make this process go significantly faster than if on of the core maintainers has to do this!

pchaigno:
Using [this github.com search](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aacs+Script&type=Code&ref=searchresults), I found 135 ACS repositories by 90 different users. It's not much for a conflicting file extension. @arfon Should we wait for a greater adoption of ACS on github.com?

arfon:
> Using this github.com search, I found 135 ACS repositories by 90 different users. It's not much for a conflicting file extension. @arfon Should we wait for a greater adoption of ACS on github.com?

I think we should let this one sit for a while. This is lower than the required limits in our contributing guidelines.

agwilt:
OK, that probably makes sense ...

Also, telling linguist that it's C results in sensible code highlighting.

pchaigno:
Great to hear that you found a temporary solution!

I'm closing this pull request. Don't hesitate to reopen (or open a new pull request) when you think ACS has reached the requirement in-the-wild adoption (hundreds of repositories).

PR #3032, opened 2016-06-01
Update grammar for Coq

Zimmi48:
This update is supposed to fix the issue reported here: https://github.com/mkolosick/Sublime-Coq/issues/5


arfon:
Thanks @Zimmi48 - we actually update the grammars automatically on each release of Linguist. I'll be releasing a new gem later today (hopefully).

PR #3036, opened 2016-06-03
Added RobotC

Unionjackjz1:
Sorry if I did this wrong! Please tell me what I did wrong so I can fit it.

Thank you!!!

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3187,6 +3187,16 @@ RobotFramework:
   # - .txt
   tm_scope: text.robot
   ace_mode: text
+  
+RobotC:
+  type: programming
+  color: "#0000cc"
+  extensions:
+  - .c
+  - .cpp
+  - .h
+  - .rbc
+  ace_mode: robotc
 
 Rouge:
   type: programming


Alhadis:
Have you read the [contribution docs](https://github.com/github/linguist/blob/master/CONTRIBUTING.md)?

Unionjackjz1:
I have, but still kind of confused.

Alhadis:
What part doesn't make sense?

Unionjackjz1:
Well, I think it would help if I made a sample directory. What do I put in there?

Alhadis:
Let's start with the more important questions first... what's RobotC? How many people are using it?

Unionjackjz1:
RobotC is a language mainly used in VEX Robotics Competitions (VRC). There are many thousands of teams all around the world, so quite a lot of people use RobotC. RobotC also works with Arduino, lego NXT, and some other micro controllers I can't think of off the top of my head.

Alhadis:
Well, it needs to be proven with a link showing how many results are actually in use on GitHub. [Here're search results](https://github.com/search?q=extension%3Arbc+NOT+nothack&type=Code) for the `.rbc` extension, which is the only non-C extension the language uses. 14,974 results looks promising, but most of the results don't appear to be related. They [look like this](https://github.com/ruting/rideOn.sg/blob/d33eb72866590acb9cddb6f3eb879a11f1b2d2c3/vendor/bundle/gems/ref-1.0.4/test/soft_reference_test.rbc).

Unionjackjz1:
If you search RobotC, a lot comes up that is actually RobotC. People rarely use the .rbc extension, I've never actually seen it used.

larsbrinkhoff:
It seems `.rbc` is also used for some kind of compiled Ruby code.

Alhadis:
I'm afraid searching for a language's name does absolutely nothing to indicate its actual usage. As you'll see, [many results are](https://github.com/search?q=RobotC&type=Code&utf8=%E2%9C%93) readme files that contain the word "RobotC". Which isn't a reliable gauge of a language's use.

If most of RobotC's files use the same extensions as C, that's going to make gauging its actual usage very, very difficult...

Unionjackjz1:
If it helps at all, DriveSystem and SPURFLYS (both on my github) are both done in RobotC.

Alhadis:
I'm afraid it doesn't... particularly since you're the one who needs to demonstrate the language's *widespread* use. To quote [the contributing docs](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language):

> We try only to add languages once they have some usage on GitHub. In most cases we prefer that each new file extension be **in use in hundreds of repositories** before supporting them in Linguist.

On the bright side, having the language classified as C isn't *too* inaccurate. The [language's website](http://robotframework.org/) describes it as a C-based programming language.

Unionjackjz1:
That website link is to RobotFramework, not RobotC. [Here](http://www.robotc.net) is a link to their website. 

Out of the thousands of people that use RobotC, it surprises me that so few use GitHub.

Thank you for being so helpful!

Alhadis:
Ugh, sorry, wrong link. I copied that from my browser's address bar, relying on recent history to give me the right one... yes, that's the one I meant.

larsbrinkhoff:
Some search results.

The vast majority of the 14,974 `.rbc` files are probably not RobotC:

| Search terms | Number of files |
| --- | --- |
| [extension:rbc rbix](https://github.com/search?q=extension%3Arbc+rbix) | 13,983 |

These are probably RobotC, but mostly use the `c` extension:

| Search terms | Number of files |
| --- | --- |
| [extension:rbc task main](https://github.com/search?q=extension%3Arbc+task+main) | 12 |
| [extension:c "task main"](https://github.com/search?q=extension%3Ac+%22task+main%22) | 37,966 |

Alhadis:
@Unionjackjz1 What does `task main` do? Are there any other terms we should be searching for to locate RobotC files? Many of the ones I'm seeing which contain `task main` seem to be blank.

larsbrinkhoff:
More search results.  `#org` or `#globalorg` are found in some kind of Pokemon scripts.  XML files which contain `RBCVersion` are RobotC.

Updated previous post.

Unionjackjz1:
Sorry, I forgot to respond. 

`task main(){

}`

is where you put all of your main code. That is the only function that runs, and you can have other voids and tasks but they have to be called to run.

Alhadis:
So is that all we have to go on in order to identify a RobotC file?

Unionjackjz1:
I think so, because I don't think it is used the same way in other languages. So, from what I know about other languages, yes.

larsbrinkhoff:
I believe the phrase `Code automatically generated by 'ROBOTC' configuration wizard` is probably quite unique to RobotC files.

Unionjackjz1:
That is not in every file, it is in most of them, but you don't need it to get code working.

Unionjackjz1:
But I think everyone uses them, so it could be used to detect RobotC?

arfon:
🎏 flagging this as stale 🎏

PR #3037, opened 2016-06-03
Add syntax highlighting for the Clean language

Alhadis:
A [language package](https://github.com/timjs/atom-language-clean) for the [Clean programming language](http://clean.cs.ru.nl/) was published to Atom.io earlier tonight. The package is in the hands of one of the language's maintainers, which makes it an obvious choice, even if the site already had highlighting for Clean (which it doesn't).


arfon:
:+1: this won't be live for another couple of weeks (when I next cut a gem)

PR #3038, opened 2016-06-04
Switch grammar used for reStructuredText highlighting

Alhadis:
Linguist currently uses [TextMate's bundle for reStructuredText](https://github.com/textmate/restructuredtext.tmbundle), which hasn't been updated in nearly 1.5 years. [The grammar](https://github.com/Lukasa/language-restructuredtext) I've recently submitted a few fixes to was converted from the same bundle, which inherited the same problems I addressed.

Admittedly, this isn't *too* much of a big deal, since GitHub renders reStructuredText files as full HTML. However, the errors are still visible in embedded rST:

```rst
.. code::

.. py:class:: Test
```
That particular example was taken from [an issue](https://github.com/Lukasa/language-restructuredtext/issues/13) submitted to the package's repo.

As for the specific improvements over TextMate's grammar, well, the majority of them can be reviewed in pull requests I submitted:
* Lukasa/language-restructuredtext#41
* Lukasa/language-restructuredtext#38


Alhadis:
@arfon I think something's wrong with Travis...

Again.

pchaigno:
@Alhadis You also need to remove the submodule from `vendor/grammars/restructuredtext.tmbundle` ;)

Alhadis:
The hell? I swear I must've shot that thing like THREE times, and it keeps coming back. :|

pchaigno:
I usually follow [these steps to remove a submodule completely](http://stackoverflow.com/a/1260982).

Alhadis:
Ah thanks! That'll come in handy. :D

Also, I must've missed this line in the build feedback:
```
No submodule mapping found in .gitmodules for path 'vendor/grammars/restructuredtext.tmbundle'
```
My bad. On it.

Alhadis:
Okay, done. It's weird that error didn't show up in the tests I ran locally, though.

I'm wondering if there should be a tool in the `script` path to remove/change grammars easily. Similar to how we have `script/convert-grammars`.

arfon:
> I'm wondering if there should be a tool in the script path to remove/change grammars easily. Similar to how we have `script/convert-grammars`.

That would be pretty great 💖

Alhadis:
Heh, when I think about it, each of the steps for adding a new grammar could be handled with a single tool. It's slightly precarious to pass the name of the grammar file on command-line, and it could be lifted from the repository in question.

arfon:
Yeah that makes sense.

PR #3039, opened 2016-06-05
Update POV-Ray Scene Description Language grammar.

c-lipka:


Alhadis:
This is done automatically with every release of Linguist.

arfon:
> This is done automatically with every release of Linguist.

Yep, we pick these up automatically.

Issue #2940, opened 2016-04-08
Wrongly states repository is Perl6, actual language Turing


MathBunny:
My recent GitHub repository was coded in a Turing file (*.t), but it is marked as Perl6.

Link: https://github.com/MathBunny/connect-four

pchaigno:
You can fix the issue locally, on your repository, with [Linguist overrides](https://github.com/github/linguist#overrides). If it's a general issue on github.com, we could add new heuristic rules to distinguish between Turing, Perl, Perl6 and Terra files.

Alhadis:
@MathBunny The [Turing sample](https://github.com/github/linguist/blob/master/samples/Turing/turing.t) that Linguist uses looks pretty basic. It would probably help the classifier if it had a more substantial fixture to work off.

[Your program](https://github.com/MathBunny/connect-four/blob/master/ConnectFour.T) looks like it'd be a good example, but we can't use it unless it's released under a permissive license (like MIT, ISC, Apache-2, etc).

At the moment, your project's license isn't listed, so I'm not sure how you'd feel about this.

arfon:
Closing this out as #3040 is merged.

Alhadis:
I'll test this locally, and if the repo is still being misclassified, I'll write a heuristic. =)

MathBunny:
@Alhadis I was looking at the file `test_heuristics.rb`, and I couldn't find Turing there, but I saw Perl6 was listing there with the extension `*.t` - is this how it's supposed to work? Shouldn't Turing be listed there?

Alhadis:
Haha, no no. =) Linguist identifies files based on the frequency of certain patterns and tokens, which is good enough for most cases. However, there are situations when it gets stuck or confused, and it needs human help to push it in the right direction. Those are the times when a heuristic is written is help.

[This](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L310-L316) heuristic would've been written to help Linguist tell the difference between Perl and Perl6, since both languages use the `.pm` and `.t` extensions, and both probably have a lot in common (I only know Perl, never been bothered to learn Perl6, and I frankly don't want to).

MathBunny:
Got it! Thanks :)

Alhadis:
No wukkaz!

(That's Australian for *"you're most welcome"*).

Alhadis:
[...]

PR #3040, opened 2016-06-06
Improve Turing sample

MathBunny:
Improved the Turing sample code with commands including graphics, audio, conditionals, if statements, etc. This is a follow up from [here.](https://github.com/github/linguist/issues/2940)


Alhadis:
Hey mate, glad you're willing to share!

Just a note, though: we prefer multiple fixtures, and there's often a mix of both complex and simple files. So the file should be added separately. =)

MathBunny:
Hey. Thanks! I'm going to fix it in a bit and I'll try again :)

Alhadis:
No problem. =)

BTW, there's no syntax highlighting for Turing on GitHub. I searched for an existing grammar, but nobody appears to have written one. Guess that's up to me, haha. I'll try to make time to research the language and write one up. =)

Is there a specification for the language I can use as reference?

MathBunny:
Ah. Yeah :( I wish I could help you. Perhaps check [this link.](http://compsci.ca/holtsoft/doc/)

MathBunny:
Just a question, do I just add the new file? Is that all, or do I have to modify another file for the linguist to know that I added a new sample program for Turing? Thanks!

Alhadis:
Ah, that's perfect! Thanks!

No no, just add the file as you would normally. Linguist uses whatever's kept in the relevant sample directories. =) You should revert the changes to the original file so it becomes this again:

```
% Accepts a number and calculates its factorial

function factorial (n: int) : real
    if n = 0 then
        result 1
    else
        result n * factorial (n - 1)
   end if
end factorial

var n: int
loop
    put "Please input an integer: " ..
    get n
    exit when n >= 0
    put "Input must be a non-negative integer."
end loop

put "The factorial of ", n, " is ", factorial (n)
```

MathBunny:
Awesome, thanks! This was fun. What is the GitHub CLA? Do I have to do that?

Alhadis:
Ah, [details are here](https://cla.github.com/github/linguist/accept/MathBunny). It's just a standard process every contributor needs to agree to... a bit like an EULA, really. =)

MathBunny:
Okay, I accepted. Is this all? :)

Alhadis:
Yep, that's all! =) Now we wait for a maintainer of the project to accept your PR.

I'm just an obsessive contributor who likes to help, I'm not staff or anything, haha.

MathBunny:
Awesome!! Ah, I would've guessed you were staff or something :p Thanks for your help!

Alhadis:
Haha, no problem. ^^

The staff are busy as hell running around constantly, so it's good to get things sorted as much as possible so their todo list doesn't pile up with writing explanations to lingering PRs.

MathBunny:
Makes sense  :) It's great that you're volunteering your time!

arfon:
👍 looks good, thanks @MathBunny. Also, thanks @Alhadis for steering this PR!

@MathBunny - this change to Linguist won't be live on the site for at least another week - we generally release a new gem every couple of weeks.

MathBunny:
Awesome! Thanks everyone :) It was fun

PR #3046, opened 2016-06-08
Write samples out as UTF-8

arfon:
/ cc @tenderlove


arfon:
So it looks like there are some CI failures where we're checking for specific encoding behaviour.

@tenderlove - do you think https://github.com/github/linguist/pull/3046/files#diff-27dd121418d642475a079f07fb7c9f4eR24 on its own should fix your needle here?

tenderlove:
> do you think https://github.com/github/linguist/pull/3046/files#diff-27dd121418d642475a079f07fb7c9f4eR24 on its own should fix your needle here?

Yes, I think so

Issue #3045, opened 2016-06-08
Policy for lock files including npm-shrinkwrap.json


Limess:
Hello,

I was wondering what the perceived policy is for generated `npm-shrinkwrap.json` files. These files, although clearly useful, clog up diffs and github statistics.

The php composer lock file seems to be ignored in https://github.com/github/linguist/blob/baf56666d40fda769b2c1bfacdcc03208a4f72da/lib/linguist/generated.rb, however neither `npm-shrinkwrap.json` nor `Gemfile.lock` dependency locks are so it seems there isn't a set policy.

arfon:
@Limess - looks like there are [lots of these files](https://github.com/search?utf8=%E2%9C%93&q=npm-shrinkwrap.json+in%3Apath&type=Code&ref=searchresults).

There isn't a strict policy here. If you think they are generally considered 'noise' in a Pull Request diff then we could certainly add them to `generated.rb` and see if anyone complains loudly.

PR #3048, opened 2016-06-10
Add npm-shrinkwrap.json to generated files

Limess:
This file is generated by the `npm shrinkwrap` command. It's large, generated, and has a noisy diff. One example of a similar file - php composer lock files, is already ignored as generated.

Resolves https://github.com/github/linguist/issues/3045.


arfon:
:+1: looks good, thanks! This will be live in the next release of Linguist (later this week sometime)

PR #3049, opened 2016-06-11
Add heuristic for Perl6 and Turing

Alhadis:
After running Linguist a few times on Turing repositories in the wild (such as [this one](https://github.com/jbrick/highschool_turing)), I feel it could still benefit from an additional heuristic, even if #3040 greatly improved Linguist's classification of Turing from Perl6.

Turing and Perl6 both use different styles of line-comments (`%` versus `#`, respectively), so the heuristic was easy to add. [Surprisingly](http://i.imgur.com/9Fxm1eT.png), I managed to get Perl6 running to test if `%` was invalid syntax at the start of a line. A glaring error message confirmed that it was indeed invalid syntax:

Figure 1

Turing's heuristic also checks for `var {name} : {type}`-like patterns, where `{name}` and `{type}` are sequences of word-characters (`\w`). These lines are unlikely to manifest in Perl 6, which retains Perl 5's method of initialising variables ([`my $a = accum 5;`](https://github.com/perl6/perl6-examples/blob/master/categories/best-of-rosettacode/accumulator-factory.pl#L84)).

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -391,6 +391,14 @@ module Linguist
       end
     end
     
+    disambiguate ".t" do |data|
+      if /^\s*%|^\s*var\s+\w+\s*:\s*\w+/.match(data)
+        Language["Turing"]
+      elsif /^\s*use\s+v6\s*;/.match(data)
+        Language["Perl6"]
+      end
+    end
+    
     disambiguate ".toc" do |data|
       if /^## |@no-lib-strip@/.match(data)
         Language["World of Warcraft Addon Data"]


pchaigno:
Nice pull request!

Were you able to test it on a few sets of Perl6/Turing files?

Alhadis:
No, actually. Every time I tried installing a locally-built gem, it wound up using some other version that was oblivious to the heuristic I added. I kept giving me the same incorrect results for the [repository I mentioned](https://github.com/jbrick/highschool_turing) as before I added the heuristic.

arfon:
:+1: thanks @Alhadis

Alhadis:
Always a pleasure, mate.

Alhadis:
Always a pleasure, mate.

PR #2440, opened 2015-06-06
adding NCL language

rpavlick:
https://github.com/search?utf8=%E2%9C%93&q=extension%3Ancl+NOT+what

PR #3054, opened 2016-06-15
Add support for SubRip Text files and SRecode Templates

Alhadis:
There are two types of `.srt` files, both totalling almost [103,000 search results](https://github.com/search?p=1&q=extension%3Asrt+NOT+nothack&type=Code) on GitHub:

1. [**SRecode .SRT files**](http://cedet.sourceforge.net/srecode.shtml)<br/> A templating engine of sorts ~~written using~~ *written in* Emacs Lisp with some Handlebars-like template tags. The [sample I've added](https://github.com/Alhadis/linguist/blob/d0370a3b4cfbe18cea0ce17fb58ea03355737a40/samples/Emacs%20Lisp/linguist.srt) was written by hand while referencing a boilerplate file in [CEDET's](https://sourceforge.net/projects/cedet/files/cedet/cedet-1.1.tar.gz/download) tarball - the entire project is GPLed to hell and back, with derived results in the wild being released under a similar license.<br/><br/>**EDIT:** I've moved this to its own language entry, rather than leaving it classified as a type of Emacs Lisp file. See the discussion between @larsbrinkhoff and myself below.<br/><br/>The colour I've picked for SRecode Templates is sampled from the CEDET project's [logotype](http://cedet.sourceforge.net/img-gen/cedet-logo-left.png). I had to bump the luminosity a notch, because the shade of green was too close to C#'s colour.

2. [**SubRip Text**](https://en.wikipedia.org/wiki/SubRip#SubRip_text_file_format)<br/> `.srt` files would appear to be one of the most widely-spread subtitle formats, and while I can't foresee GitHub ever being inundated by them, they did turn up in search results ([this](https://github.com/lgautrot/translations/blob/ce9c42723e3dcc45c289447b3441ab74671e79de/subtitles/cockleshell_heroes-fr.srt) and [this](https://github.com/matthewfu/TheBibleSubChinese/blob/990e175c75884d0c2d9dda6d1ed81c5d735f2cfe/%E8%81%96%E7%B6%93%E5%BD%B1%E9%9B%86.The.Bible.S01E01.1080p.BluRay.x264.cht%26eng.utf8.srt), for example). The [sample I added](https://github.com/Alhadis/linguist/blob/d0370a3b4cfbe18cea0ce17fb58ea03355737a40/samples/SubRip%20Text/Adding.NCL.Language.S01E01.1080p.BluRay.x264.srt) is #2440 transcribed into a cinematic format.


larsbrinkhoff:
To me it seems that the CEDET srt thing can't be said to be Emacs Lisp.

Alhadis:
Think it should be added as a derived language, like Handlebars, then?

larsbrinkhoff:
I would said it's a separate syntax.  Typical files seem to consist of a few directives, and then template snippets mostly in the syntax of the target output, which can be anything.

I suppose Handlebards is mostly used for HTML output?  Which would explain why it's grouped under the HTML language.  CEDET srt seems to be used for many different programming languages.

Alhadis:
Well, true, but it's using Emacs Lisp macros for gluing stuff together (near as I can tell).

larsbrinkhoff:
Can you provide some links to srt files with Lisp inside?

By the way, what's the problem with GPL-licensed sample files?

Alhadis:
> **Can you provide some links to srt files with Lisp inside?**

Well, I'm actually going off the project's own documentation:

> <i>"More complex application specific templates are a mix of Emacs Lisp code and templates. Creating a template based application is a breeze in SRecode where dictionaries can be created and applied to the the template catalog. A single Emacs Lisp application can actually write code to do the same thing in multiple languages based on a strong set core templates for each language"</i>

> ** By the way, what's the problem with GPL-licensed sample files?**

Not an expert on this subject, but to the best of my understanding, Linguist can't distribute content that forbids distribution under a different license. The whole idea of "Copyleft" being that material is free so long as it's distributed and modified under the same terms. Linguist is MIT-licensed, which conflicts with the terms of the GPL... so in order to be included in the software's codebase without attracting "issues", we need to ensure it's released under a permissive license.

Alhadis:
Right, the more I read into this, the more I'm convinced this is a general-purpose templating language, and less a derivative of Emacs Lisp. Probably shouldn't have put so much stock in the project's own description of itself... updating.

larsbrinkhoff:
Nice.  But then I suppose it doesn't make sense to use Lisp syntax highlighting for those `.srt` files?

Alhadis:
Well, Lisp [doesn't do too bad a job](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=source.lisp&grammar_url=&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fabelardojarab%2Femacsfull%2Fblob%2F3d0857b8bde729530988aadcfb239495e04eeb34%2F.emacs.d%2Fcedet%2Fetc%2Fsrecode%2Fsrecode-dot.srt&code=) with highlighting the template files. I chose Lisp mainly to identify comments from non-comments, something which the Handlebars grammar wouldn't be well-suited for.

The *correct* approach would of course be to use a dedicated language grammar. I couldn't find a TextMate-compatible grammar for SRecode Templates, and I'd be surprised if one even exists. The project is, after all, Emacs-centric, and syntax highlighting for said templates is already included in CEDET. I doubt there's been anybody serious about working with SRecode who's had the incentive to port the grammar to other editors (which won't be able to make use of CEDET).

larsbrinkhoff:
You're right, that does look surpisingly good.

larsbrinkhoff:
Thank you, @Alhadis!

I have reviewed the changes, and I think everything looks fine.  Awaiting +1 from a githubber.  @arfon?

| Search | Results |
| --- | --- |
| [extension:srt "set mode"](https://github.com/search?q=extension%3Asrt+"set+mode") | 16000 |
| [extension:srt NOT "set mode"](https://github.com/search?q=extension%3Asrt+NOT+"set+mode") | 86000 |

arfon:
Looks good :+1:

larsbrinkhoff:
I meant to write about this in more detail.

I made a few searches to try to gague how many "SRecode Template" and "SubRip Text" files there are.  Clearly, both meed the criteria for addition.  It's difficult to search for subtitle files, but I made some random sampling of the second search, and I didn't find anything else.

| Search | Results |
| --- | --- |
| [extension:srt "set mode"](https://github.com/search?q=extension%3Asrt+"set+mode") | 16000 |
| [extension:srt NOT "set mode"](https://github.com/search?q=extension%3Asrt+NOT+"set+mode") | 86000 |

Alhadis:
@larsbrinkhoff You probably should've squash-merged the way @arfon does... I keep commits atomic to make PRs easier to review, but this will probably lead to a cluttered revision history. ;)

PR #3055, opened 2016-06-15
Add --json option to bin/linguist

vmarkovtsev:
git-linguist requires the current directory to be a git repo, so it is not an option in my case. Thus adding machine-readable output to bin/linguist

### DIFF
--- a/bin/linguist
+++ b/bin/linguist
@@ -5,6 +5,7 @@
 #
 require 'linguist'
 require 'rugged'
+require 'json'
 
 path = ARGV[0] || Dir.pwd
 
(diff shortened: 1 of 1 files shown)


pchaigno:
@vmarkovtsev Thanks for your interest in Linguist!

Out of curiosity, what's your use case for this?

vmarkovtsev:
I am working at [source{d}](http://sourced.tech) and among other things we are doing kind of GitHub mining. linguist is particularly used as the reference codebase preprocessor.

pchaigno:
So you're calling the Linguist ruby gem from another language?

Alhadis:
To be honest, JSON output would be very useful in cross-language data consumption.

vmarkovtsev:
Correct, I am calling it from Python.

arfon:
@vmarkovtsev - as currently implemented though this still requires Rugged to be installed right?

What happens if Rugged doesn't detect a Git repository on [this line](https://github.com/github/linguist/pull/3055/files#diff-ebb4bd26e9dba157952d6110f5781735L21).

vmarkovtsev:
@arton I guess the same as with --breakdown happens: it is used pretty the same way.

bin/linguist
```diff
@@ -16,14 +17,17 @@ end
 
 ARGV.shift
 breakdown = true if ARGV[0] == "--breakdown"
+json_breakdown = true if ARGV[0] == "--json"
```
pchaigno: Should it be `--breakdown --json` for a full JSON breakdown and `--json` just to get statistics as JSON?

vmarkovtsev: No objections. I designed them to be mutually exclusive because the current command line argument handling is rudimental and I would have to rewrite it with two possible options without knowing Ruby. Actually, the way --breakdown is handled already sucks IMHO.

vmarkovtsev:
Anyone?

brandonblack:
Merged manually. Thanks @vmarkovtsev 👍

vmarkovtsev:
Hurray!

PR #3059, opened 2016-06-17
Detect Javascript files generated by Protocol Buffers

zbrock:
Protobuf added Javascript support in 3.0.0 Beta 2.


lib/linguist/generated.rb
```diff
@@ -274,6 +275,18 @@ def generated_protocol_buffer?
...
 
+    # Internal: Is the blob a Javascript source file generated by the
+    # Protocol Buffer compiler?
+    #
+    # Returns true of false.
+    def generated_javscript_protocol_buffer?
+      return false unless extname == ".js"
+      return false unless lines.count > 6
+
+      return lines[5].include?("GENERATED CODE -- DO NOT EDIT!")
```
pchaigno: Is this text always on the fifth line? Couldn't there be more (or less) annotations such as `@fileoverview` and `@public`, or even no comment lines before?

lib/linguist/generated.rb
```diff
@@ -69,6 +69,7 @@ def generated?
       compiled_cython_file? ||
       generated_go? ||
       generated_protocol_buffer? ||
+      generated_javscript_protocol_buffer? ||
```
pchaigno: Looks like you missed an `a` in the name here ;)

pchaigno:
@zbrock Thanks for the pull request!

lib/linguist/generated.rb
zbrock: hah, oops! fixed.

lib/linguist/generated.rb
zbrock: It seems to be hardcoded in the js generator right now: https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/js/js_generator.cc#L1383

But Javascript support is technically only Alpha right now, so it may change.

pchaigno:
It doesn't seem to be used very much on GitHub: [search results](https://github.com/search?p=1&q=extension%3Ajs+generated+code+fileoverview+enhanceable&ref=searchresults&type=Code&utf8=%E2%9C%93). @arfon Is there a requirement on this for generated files?
Maybe we could turn this into a more generic heuristic rule...?

arfon:
> Maybe we could turn this into a more generic heuristic rule...?

Yeah, I think that might be a good idea.

zbrock:
@pchaigno yeah, it's a newly supported language target in ProtoBufs. Not surprised there aren't many usages. We keep running into it in our private repo. Small PRs to .proto files yield large diffs that aren't meant to be viewed by humans.

zbrock:
@arfon Happy to take a look at improving it either in this PR or in a follow up! What are you thinking?

zbrock:
@pchaigno Anything else I can do to help here?

brandonblack:
Merged manually. Thanks @zbrock 👍

PR #3066, opened 2016-06-19
Remove pragma keyword from PLSQL heuristic rule

### LINKED ISSUE #3063: PHP files being misidentified as PLSQL
Hi,

I recently pushed up a restructure of the https://github.com/BNETDocs/bnetdocs-web repository.

However, I noticed that GitHub now believes it's a 70%+ PLSQL repository, which I had to look up what PLSQL even was after I found out. If I click on filter by PLSQL, nothing shows up.
image

https://github.com/BNETDocs/bnetdocs-web/search?l=plsql
image

Is this a bug? What do I need to do to restore the correct language that is PHP? Also notice how in the above snapshot, GitHub's search shows the majority of the repository being PHP on the left-hand side.

pchaigno:
This pull request removes the `pragma` keyword from the PLSQL heuristic rule. The `pragma` keyword is not specific to PLSQL and [can be found in other SQL languages](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asql+mysql+pragma).

Fixes #3063.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -382,7 +382,7 @@ module Linguist
       elsif /(alter module)|(language sql)|(begin( NOT)+ atomic)/i.match(data)  || /signal SQLSTATE '[0-9]+'/i.match(data)
         #IBM db2
         Language["SQLPL"]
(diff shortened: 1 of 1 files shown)


arfon:
:shipit:

PR #3069, opened 2016-06-21
Add language colour for Roff

Alhadis:
This PR adds a colour for Roff (incorrectly named "Groff" on GitHub):

Figure 1

I chose a colour to match the look of aged paper, which symbolises both the language's longevity and its importance in authoring documentation/manuals. No canonical implementation of Roff exists which has any branding with distinguishable colours... so I fell back on natural sources as my inspiration.

----

This PR also attempted to add a colour for the PostScript language, which was the same [PANTONE red](https://www.pantone.com/color-finder/485-C) used in Adobe's official branding. [Example](https://www.adobe.com/products/postscript/pdfs/ps3datasheet.pdf):

Figure 2

However, the colour-proximity test was blocking every attempt at picking a close enough colour... and even shades that were noticeably different were being rejected. Here's a comparison of the conflicting colours, with the attempted PostScript shades on the right:

Figure 3

This might be attributed to my colour-blindness, but I fail to see how the bottom-right colour could possibly be confused with any of the other shades depicted here.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2875,7 +2875,6 @@ Pony:
 
 PostScript:
   type: markup
-  color: "#da291c"
   extensions:
   - .ps
   - .eps
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1342,6 +1342,7 @@ Graphviz (DOT):
 
 Groff:
   type: markup
+  color: "#ecdebe"
   extensions:
   - .man
   - '.1'
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2875,6 +2875,7 @@ Pony:
 
 PostScript:
   type: markup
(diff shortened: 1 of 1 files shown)


pchaigno:
@Alhadis Neat PR as always! :+1: 

> This might be attributed to my colour-blindness, but I fail to see how the bottom-right colour could possibly be confused with any of the other shades depicted here.

I'm not saying it's the case here, but sometimes it's harder to distinguish colors when they are on the narrow language bar.
In any case, the formula to compute color distances is not perfect and probably never will since that depends on people's views too.

arfon:
Looks good 👍

Alhadis:
Thanks @arfon. =)

> In any case, the formula to compute colour distances is not perfect and probably never will since that depends on people's views too

That's why I personally feel the colour proximity should raise a warning rather than a build error. If two colours differ visually more than they do mathematically, it shouldn't be a deal-breaker.

PR #3076, opened 2016-06-25
Another Arduino sample for better detection

liavt:
Currently, there is only one Arduino sample. This proposal has much more common Arduino functions, like pinMode(), digitalRead(), digitalWrite(), LOW, HIGH, INPUT,OUTPUT, and delay(). This will allow Linguist to more accurately find Arduino files, in case they don't have the .ino extension.


arfon:
👍 thanks @liavt

PR #3088, opened 2016-06-27
Replace GPL licensed gml samples

sahildua2305:
Removed all GPL licensed .gml samples.

Sources of all added files are mentioned in the first line of every file.

@arfon Please have a look :smile:


sahildua2305:
@arfon @pchaigno Please review this :smile:

pchaigno:
Looks like the new files are way smaller than the old files (+447, -2315). Not sure if that could be an issue. @arfon ?

sahildua2305:
@pchaigno I tried to minimize the difference. Couldn't find larger files with desired license. I'd appreciate help here. :)

arfon:
> Looks like the new files are way smaller than the old files (+447, -2315). Not sure if that could be an issue. @arfon ?

If the tests are passing then I think we're OK. There are some issues with the Bayesian classifier and sample sizes but I think we're probably OK here.

sahildua2305:
@arfon Can we then go ahead with this one?

Issue #2730, opened 2015-11-19
Gdb command files


eush77:
[Gdb command files](https://sourceware.org/gdb/onlinedocs/gdb/Command-Files.html) are script files that are loaded and executed by gdb debugger either at startup (in which case they are typically named `.gdbinit`) or at runtime with `source` command (in which case I don't know of any common naming conventions).

Some of these files can be [quite huge](https://github.com/gdbinit/Gdbinit/blob/master/gdbinit).

I know lldb has also some equivalent of these.

pchaigno:
Can you confirm that [these files](https://github.com/search?p=4&q=extension%3Agdbinit+NOT+djfhdkghfjf&type=Code&utf8=%E2%9C%93) are Gdb command files? If they are, then [it meets the requirements for adding a language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language) and you should open a pull request ;)

larsbrinkhoff:
Yes @pchaigno, those are overwhelmingly gdb.

Also, [possibly 17000](https://github.com/search?q=extension%3Agdb+define+OR+set+OR+end+OR+thread) out of [24000](https://github.com/search?q=extension%3Agdb+NOT+slartibartfast) `.gdb` files seem to be gdb scripts.

pchaigno:
Support for GDB was added in #3090.

PR #3090, opened 2016-06-28
Support for GDB

pchaigno:
This pull request adds support for GDB command files, as discussed in #2730.
There are around [31,700 `.gdb`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Agdb+NOT+dkfjdkfdnkd&type=Code&ref=searchresults) and [48,000 `.gdbinit`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Agdbinit+NOT+jfhdkfndnf) files on github.com. GDB seems to be the only language using these file extensions.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -752,3 +752,6 @@ url = https://github.com/austinwagner/sublime-sourcepawn
 [submodule "vendor/grammars/language-blade"]
 	path = vendor/grammars/language-blade
 	url = https://github.com/jawee/language-blade
+[submodule "vendor/grammars/SublimeGDB"]
+	path = vendor/grammars/SublimeGDB
+	url = https://github.com/quarnster/SublimeGDB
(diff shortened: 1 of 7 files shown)


larsbrinkhoff:
I suggest also adding `.gdbinit`, also discussed in #2730.

pchaigno:
Thanks @larsbrinkhoff! I had completely forgotten that file extension (I added as a file extension rather than a filename since there are also many `dot.gdbinit`, etc.).

larsbrinkhoff:
@pchaigno, great!

pchaigno:
@arfon This pull request should be ready now!

I had to add the Zlib license to http://choosealicense.com: github/choosealicense.com#438, then update `licensee`.
Unfortunately, I cannot update `licensed` (which Linguist uses) because it's closed source :/
Thus, I added a development dependency on `licensee` and manually edited `vendor/licenses/grammar/SublimeGDB.txt`. That's only a short term fix; a better fix would be to update `licensed`.

arfon:
> That's only a short term fix; a better fix would be to update licensed

Wilco. FYI I think there are plans to open the source of `licensed`.

Issue #3044, opened 2016-06-07
JSX Syntax highlighting broken? 


cozos:
Hey,

I was redirected here by support@github.com - it seems that JSX Highlighting is broken on Github.

For example: https://gist.github.com/cozos/d6636a93cba29c77e3221f2ca30212b0

``` jsx
import React from 'react';

var ExampleApplication = React.createClass({
  render: function() {
    var elapsed = Math.round(this.props.elapsed  / 100);
    var seconds = elapsed / 10 + (elapsed % 10 ? '' : '.0' );
    var message =
      'React has been successfully running for ' + seconds + ' seconds.';

    return <p>{message}</p>;
  }
});

var start = new Date().getTime();

setInterval(function() {
  ReactDOM.render(
    <ExampleApplication elapsed={new Date().getTime() - start} />,
    document.getElementById('container')
  );
}, 50);
```

If it helps, I'm on the latest build of Google Chrome.

pchaigno:
Linguist only selects the grammar to use for highlighting. For JSX, GitHub uses [gandm/language-babel](https://github.com/gandm/language-babel/tree/bece4f80cebd0ab13d798a684dd28d51b3b854c0). You should open an issue there if there isn't already one. Sorry for the two redirects :S

cozos:
👌

arfon:
Closing as this is an upstream issue.

gandm:
@arfon @pchaigno Author of language-babel here. I've just checked my grammar using the version on linguist - 2.25.1 -  inside Atom and it appears to work so I don't know what is going on. 

e.g.
![image](https://cloud.githubusercontent.com/assets/2313237/15882723/9298536a-2d39-11e6-9f62-fae1e12943f0.png)
 
but on github 
```jsx
/***
  XMLHttpRequest example
 */


import 'js/web.jsx';
import 'console.jsx';

class _Main {
    static function main(args : string[]) : void {
        var xhr = new XMLHttpRequest();
        var path = dom.document.location.pathname;
        xhr.open("GET", path.replace(/\/[^\/]*$/, "") + "/hello.txt");
        xhr.onreadystatechange = function (e) {
            if (xhr.readyState == xhr.DONE) {
                _Main.update(xhr.responseText);
            }
        };
        xhr.onerror = function (e) {
            console.error("XHR error");
        };
        xhr.send();
    }

    static function update(text : string) : void {
        var output = dom.id("output");
        var textNode = dom.document.createTextNode(text);
[...]

PR #3091, opened 2016-06-28
Change language-babel to release v.2.22.0

cozos:
Issue https://github.com/github/linguist/issues/3044

@arfon @gandm

`language-babel` was refactored to use back-reference notation on <s>`v.2.25.x`</s> v2.23.0, which the Linguist regex engine couldn't handle. 

If I understand git submodules correctly, this PR locks `language-babel` to <s>`v2.24.4` or commit `d14329f2d3c0a1154c50f668e0a58122ca97854a`</s> `v2.22.0` or commit `0d093e8ffeaa80602c6f3ae86fade6a3b5f56e4d`, which should be able to handle JSX imports and exports.

<s>Refers to https://github.com/gandm/language-babel/releases/tag/v2.22.4 aka https://github.com/gandm/language-babel/tree/d14329f2d3c0a1154c50f668e0a58122ca97854a</s>
Refers to https://github.com/gandm/language-babel/releases/tag/v2.22.0 aka https://github.com/gandm/language-babel/tree/0d093e8ffeaa80602c6f3ae86fade6a3b5f56e4d


cozos:
@gandm Whoops, I was wrong - it appears that the `literal-module-import` change was done on this commit: https://github.com/gandm/language-babel/commit/b10a055f5d716155e990d2ff37d5f08a2513d1db#diff-84d196d176a806166c5319d65e586325R1425, which seems to correspond to v2.23.0.

I guess I'll have to revert to v2.22.0

gandm:
@cozos yes 2.23.0 was the change that started using `\k` type back referencing in import/exports statements so you need 2.22.0. All a bit of kludge to get this working! There are many other places where I used `\k` references that may impart he parsing of syntax. Shame linguist doesn't use Oniguruma.

cozos:
@pchaigno What do you think?

cozos:
@arfon @pchaigno @larsbrinkhoff 

[This](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fgithub.com%2Fgandm%2Flanguage-babel%2Fblob%2F0d093e8ffeaa80602c6f3ae86fade6a3b5f56e4d%2Fgrammars%2FBabel%2520Language.json&grammar_text=&code_source=from-text&code_url=&code=import+React+from+%27react%27%3B%0D%0A%0D%0Avar+ExampleApplication+%3D+React.createClass%28%7B%0D%0A++render%3A+function%28%29+%7B%0D%0A++++var+elapsed+%3D+Math.round%28this.props.elapsed++%2F+100%29%3B%0D%0A++++var+seconds+%3D+elapsed+%2F+10+%2B+%28elapsed+%25+10+%3F+%27%27+%3A+%27.0%27+%29%3B%0D%0A++++var+message+%3D%0D%0A++++++%27React+has+been+successfully+running+for+%27+%2B+seconds+%2B+%27+seconds.%27%3B%0D%0A%0D%0A++++return+%3Cp%3E%7Bmessage%7D%3C%2Fp%3E%3B%0D%0A++%7D%0D%0A%7D%29%3B%0D%0A%0D%0Avar+start+%3D+new+Date%28%29.getTime%28%29%3B%0D%0A%0D%0AsetInterval%28function%28%29+%7B%0D%0A++ReactDOM.render%28%0D%0A++++%3CExampleApplication+elapsed%3D%7Bnew+Date%28%29.getTime%28%29+-+start%7D+%2F%3E%2C%0D%0A++++document.getElementById%28%27container%27%29%0D%0A++%29%3B%0D%0A%7D%2C+50%29%3B) is the JSX highlighting on Lightshow using the commit that I reverted to. Should be pretty straightforward. Could you please take a look?

cozos:
screen shot 2016-07-07 at 5 54 49 pm

arfon:
👍 looks good to me.

pchaigno:
> language-babel was refactored to use back-reference notation on v2.23.0, which the Linguist regex engine couldn't handle. 

@vmg Could you confirm that the github.com regex engine doesn't support `\k` back-references?

If we continue with this pull request, I think we should add a test with the version hardcoded. Otherwise we could update it again by mistake on the next release :/

(Unless @gandm plans to backport changes to the `2.22.0` version, on the long run, we might have to change the grammar for another...)

Alhadis:
@gandm Why are you using `\k` instead of hardcoded (and portable) backreference numbers?

If it's an issue with maintainability (trying to count and keep track of each capturing group's index), you can always add inline comments the [way I've done here](https://github.com/Alhadis/language-apl/blob/master/grammars/apl.cson#L357). Heck, I even [published an Atom package](https://atom.io/packages/regex-comments) to make regex comments look good in TextMate grammars.

gandm:
I have no plans to remove the use of `\k` from the grammar as it works in Atom (which is my main target) and allows a block to be cut and paste because the references are relative.  A few textmate grammars exist that support Facebook's JSX, so maybe one of these might be better match for linguist.

cozos:
Why is the use of `\k` controversial in the first place? 

I think that it's @gandm's prerogative to do as he wants with his project, but I also want `.jsx` highlighting to work. What's so bad about locking the grammar dependency to a certain release?

vmg:
I'm not sure `\k` is the reason why the JSX grammar is not supported. Regardless, let's pin back the grammar to this old version while I investigate further.

vmg:
@arfon: Good to merge.

arfon:
Thanks for looking into this @vmg.

arfon:
@cozos - I'm planning on releasing a new Linguist gem later this week which will include this update/fix.

cozos:
Thanks 👌

Issue #3084, opened 2016-06-27
.sass files are labeled as markup


controversial:
If I understand correctly, `.sass` files do not show in the top bar of a GitHub repo because [they're labeled as markup](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3385). I don't think this is correct. The [tab-indented sass syntax](http://sass-lang.com/documentation/file.INDENTED_SYNTAX.html) is its own language, which compiles to CSS. It's not even CSS-like, it doesn't use curly brackets, etc.

I agree with the grouping of `scss` files as CSS, but I think `sass` files should be their own language. Could the sass language be marked as not a markup language, but its own language?

pchaigno:
Closing in favor of #3093, which implements this change.

Issue #2933, opened 2016-04-03
Nervana sass files are wrongly identified as CSS


kriskhaira:
Nervana shader assembly files (*.sass) are being wrongly identified as CSS. Examples:

* https://github.com/baidu-research/nervanagpu
* https://github.com/NervanaSystems/maxas

pchaigno:
To fix this, we'd have to recognize Nervana shader assembly files in Linguist and distinguish them from SASS. Is there a simple way to identify a Nervana shader assembly file?

kriskhaira:
I'm not sure. I was just looking for Baidu's web-related open source projects and found this.

But looks like all sass files start with `Kernel: ` and contain `--:-:-:-:`.

pchaigno:
Hm. Not always: https://github.com/woodun/swl_outputs/blob/12c46dd6fad73d9c92b471ecdaa2e1f8b213b6e1/gto/lonestar/sssp/_cuobjdump_1.sass but that may be another dialect :/

pchaigno:
Closing in favor of #3093.

Issue #2650, opened 2015-09-24
SCSS, SASS, LESS, Stylus, etc. = CSS?


jamen:
When doing searches on projects or viewing what the project contains, it's misleading.

Say for instance you only want to find SCSS projects.  When doing a search that can be hard to perform since all CSS, SASS, LESS, Stylus, etc. are all categorized as "CSS" along with the SCSS project you're trying to find.

We should separate them out into separate languages.

auscompgeek:
What's wrong with searching for `extension:scss` or similar? IMO, this is a feature, not a bug.

davidzchen:
IMO it would be better to separate these into separate languages if they meet the requirements.

SASS, LESS, Stylus, etc. _are_ different languages and code cannot be freely shared between them interchangeably.

jamen:
I figured out that GitHub does separate them in searches, however it just doesn't distinguish them in the repository "language bar". 

This is kinda pointless, I was only really concerned about searches.  I don't think the language bar matters too much, but it would be nice if it were distinguishable on the repo's page.

julmot:
:+1:

PR #3093, opened 2016-06-29
Distinguish SASS and CSS

pchaigno:
This pull request makes SASS a language of its own, distinct from CSS.
This change was requested several times in #2933, #3084, #2585 and #2650.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3403,7 +3403,6 @@ SaltStack:
 Sass:
   type: markup
   tm_scope: source.sass
-  group: CSS
   extensions:
   - .sass
   ace_mode: sass


arfon:
...so if we do this then I think we need to consider making the same change for `HAML`, `Handlebars` etc. and many of the other languages that are currently grouped into some higher-level category.

My rationale for *not* doing this is that SASS is almost always used to generate CSS (please correct me if I'm wrong here) and so it makes sense (to me at least!) to have this listed under CSS for the general repository stats.

I would also note that this change doesn't affect search results/[GitHub explore](https://github.com/trending) (trending repos by language). Languages such as SASS are already discoverable there.

Basically we can consider doing this if we really want to but it should be part of a much larger rethink of what Linguist languages groups are.

controversial:
@arfon You're right that `sass` is almost always used to generate CSS. However, it has its own syntax which is quite different from that of CSS, unlike the other language in the sass family, scss. Isn't your argument also true of many compiled languages? CoffeeScript and JavaScript are separate in Github's eyes. SASS is to CSS as CoffeeScript is to JavaScript. It's about the same level of separation.

arfon:
> Isn't your argument also true of many compiled languages? CoffeeScript and JavaScript are separate in Github's eyes. SASS is to CSS as CoffeeScript is to JavaScript. It's about the same level of separation.

It is, and I want to be clear that I'm not completely against this idea, I just think we need to make sure we've thought this through before we start removing `group:` files from `languages.yml`.

@controversial I'd be curious to hear what problem you're trying to solve with this change? Is this a UI concern of the language bar on the repo or something else?

controversial:
@arfon That's basically it, I'd like my repo to show that it's styled with SASS instead of CSS. It's not a big deal, but I think it'd be better. ¯\_(ツ)_/¯

Alhadis:
If CoffeeScript is recognised as its own language, then Sass should too. Otherwise, we may as well be categorising the former as "JavaScript" as well.

controversial:
Please?

arfon:
> @arfon That's basically it, I'd like my repo to show in the top bar that it's styled with SASS instead of CSS. It's not a big deal, but I think it'd be better. ¯_(ツ)_/¯

Sorry for the delayed reply here. I'm not comfortable merging this until we've taken a good look at _all_ grouped languages. It's possible there is UI fix on GitHub that would address this more satisfactorily.

controversial:
Ok. Makes sense, I guess.

nazar-pc:
Basically, the same story as #2978, where Jade/Pug is definitely not HTML

controversial:
Any update? I'm once again back here after discovering that Slim and Jade are classified as HTML, not as separate languages. Seriously think this is a necessary change.

@arfon you say there might be a UI fix, but that sounds like something that's far, far less likely to happen, especially not in the short term. For now, do you still have a problem with separating out similar languages? This issue has been brought up many many times about different languages, and I haven't yet seen a very compelling argument for keeping them together beyond "there might be a UI fix."

You say
> until we've taken a good look at *all* grouped languages

what does this entail? I'd really like to see some change soon.

EmmaRamirez:
I feel like this PR should really be considered. SASS is quite a different extension of CSS.

It's also not particularly consistent--why is Smarty, for example, considered separate from PHP?

ruchern:
Is this going to happen afterall?

yegortimoshenko:
I'm not sure why there's grouping at all in the user interface, GitHub could just show each language as its own. Some grouping can be done if several files in group take very small percentage of repo's source code (as done with "Other"), but it'd be neat to distinct various preprocessors from the language they compile to.

plibither8:
@arfon @Alhadis is it going to happen? It's been more that two years since the PR was opened

pchaigno:
Arfon is not working on this project anymore, but as he said:

> I'm not comfortable merging this until we've taken a good look at _all_ grouped languages

I agree with him on this. But it takes a lot of time to do this kind of work.

plibither8:
This might sound rather curt but I feel that two years is *a lot of time*.

lildude:
> This might sound rather curt but I feel that two years is a lot of time. 😄 I'd love to help, if I can in any way.

You're right it has been a long time, but sadly it's likely to a while longer as Linguist is currently not a priority for any team within GitHub…  it is essentially running as a community maintained project with final approval from GitHub (required because it integrates directly into GitHub.com).

A change like degrouping needs to thought through thoroughly as it has been in place for a very long time and could have consequences elsewhere.

Alhadis:
We shouldn't have to reconsider *all* grouped languages. Only the very obvious cases. Sass is no closer to CSS than CoffeeScript is to JavaScript.

pchaigno:
@Alhadis You can if you want to start. I'll close this PR in favor of that issue in that case.

Alhadis:
I've opened #4291 so we can discuss this matter in depth.

Let the bikeshedding begin!

PR #3095, opened 2016-07-01
GCC machine descriptions

larsbrinkhoff:
The `.md` file extension is used not only for markdown files, but also GCC machine descriptions.  There are about [111000](https://github.com/search?q=extension%3Amd+insn) of those, so this language should qualify for addition.

The syntax is a Lisp-like variation of RTL, Register Transfer Language.

My intention is to open this issue for discussion, and later turn it into a PR if there are no objections.


larsbrinkhoff:
@arfon @pchaigno Please review.

The Bayesian classifier doesn't do well with a set of 1000 `.md` samples.  I added a heuristic: if a line begins with `;;` or `(define_`, it's almost certainly a GCC machine description.  Almost anything else is a Markdown file.

lib/linguist/languages.yml
```diff
@@ -1182,6 +1182,15 @@ GAS:
   tm_scope: source.assembly
   ace_mode: assembly_x86
 
+GCC machine description:
```
pchaigno: Should we title case the name? (We can't change it after; so whatever we do, we better be sure about it :stuck_out_tongue: )

pchaigno:
> The Bayesian classifier doesn't do well with a set of 1000 `.md` samples.

Did you get a chance to test the classifier + heuristic on those samples?

lib/linguist/languages.yml
Alhadis: Please do, holy shit. :|

larsbrinkhoff:
> Did you get a chance to test the classifier + heuristic on those samples?

Of course.  There were only Markdown files in that set, so I tuned the heuristic until there were no false positives for GCC *M*achine *D*escriptions.

lib/linguist/languages.yml
larsbrinkhoff: Ok, history rewritten.

Alhadis: Close call. Every OCD coder like myself would've gone nuts over the discrepancy.

larsbrinkhoff: Read these and weep: [IRC log](https://github.com/github/linguist/blob/5066f66dcd7be1f8d68cecd1f226e2b4c517963b/lib/linguist/languages.yml#L1612), [OpenRC runscript](https://github.com/github/linguist/blob/5066f66dcd7be1f8d68cecd1f226e2b4c517963b/lib/linguist/languages.yml#L2618), [Python traceback](https://github.com/github/linguist/blob/5066f66dcd7be1f8d68cecd1f226e2b4c517963b/lib/linguist/languages.yml#L3031), [Raw token data](https://github.com/github/linguist/blob/5066f66dcd7be1f8d68cecd1f226e2b4c517963b/lib/linguist/languages.yml#L3146).

pchaigno:
Ooh, 100% then, nice! :tada:

larsbrinkhoff:
Well, some unusual Markdown files fall through the heuristic and lands in the Bayesian classifier, which doesn't always get it right.

larsbrinkhoff:
Ok, here are the results from a set ot 999 `.md` files.  (Don't know where number 1000 went.)

| Language | Detected | Actual |
| --- | --- | --- | --- |
| Markdown | 993 | 996 |
| GCC Machine Description | 6 | 0 |
| Unknown| 0 | [3](https://github.com/lisp/de.setf.asdf.x/tree/master/website/source/resources) |

arfon:
@larsbrinkhoff - is this good to :ship:?

Alhadis:
ship-it

larsbrinkhoff:
@arfon Yes.

PR #3116, opened 2016-07-13
Bumping to v4.8.8

arfon:
Updates since last release: https://github.com/github/linguist/compare/v4.8.7...5076539df51a752d048ab22ff81e586420782dd1

### DIFF
--- a/lib/linguist/version.rb
+++ b/lib/linguist/version.rb
@@ -1,3 +1,3 @@
 module Linguist
-  VERSION = "4.8.7"
+  VERSION = "4.8.8"
 end


Alhadis:
🚀 BLAST-OFF 🚀

jbrooksuk:
🎉

arfon:
@chrislgarry - support for `Apollo Guidance Computer` is included in this release. I think you might need to rename the files in https://github.com/chrislgarry/Apollo-11 to have `.agc` extensions now...

Alhadis:
@arfon I took the liberty of [taking care of that in advance](https://github.com/chrislgarry/Apollo-11/pull/123).

@chrislgarry Safe to merge now! =)

jbrooksuk:
How long does GitHub take to reflect these changes? :)

Alhadis:
I'm pretending it takes the same amount of time for a spacecraft to leave the earth's atmosphere. Ssssh.

(Real answer: make a cuppa, come back in an hour)

arfon:
@jbrooksuk - in theory immediately but the language bar is cached for 24 hours. If you want to see an immediate change then you'll need to push a change to the repo as the commit sha is included in the cache key too.

jbrooksuk:
@arfon should we also immediately see highlighting for files then?

Alhadis:
Yes. :)

jbrooksuk:
In which case, Blade doesn't appear to work; https://github.com/jbrooksuk/Sublime-Blade/blob/master/test.blade.php

jbrooksuk:
It highlights the HTML and PHP, but not the Blade syntax itself.

Alhadis:
Make a change to the file, mate. The highlighting is probably still cached.

jbrooksuk:
Just pushed a change :)

Alhadis:
Hrm, Blade isn't listed in the languages dropdown (for either GitHub or Lightshow), so maybe something's still processing somewhere...

Then again, I'm just firing blank guesses here, so maybe I should shaddap.

jbrooksuk:
Another example of Blade not highlighting, just in case, https://github.com/CachetHQ/Cachet/blob/2.3/resources/views/index.blade.php

Alhadis:
[S'okay, must be taking its time then](https://gist.github.com/Alhadis/d4bf4f04a5134eccca8ab68c4b96532c). AGC isn't showing up either yet.

I think we should all settle down so Arfon can actually read the non-spammy notifications in his feed. =) lel

arfon:
@jbrooksuk not exactly sure what's going on here. One issue is that some of the `.php` files that are Blade aren't being detected as such (i.e. the Bayesian classifier is doing a bad job):

https://gist.github.com/arfon/6ca677f3b9aefcb8cb1de3c1aba26e10

You can see how Linguist is detecting the language by looking at the DOM:

screen shot 2016-07-14 at 10 39 32 am

arfon:
> S'okay, must be taking its time then. AGC isn't showing up either yet.

Ugh. Not sure what's going on here. I'm going to try and debug this for a while.

arfon:
OK, this should be fixed now. Somehow managed to deploy GitHub without actually including the new Gem 👊 

@jbrooksuk - as you can see here (https://gist.github.com/arfon/2bb03617c65c0741f8cf125fadb20318) some files are being correctly detected as `Blade` but this won't always be the case for `.php` files as the Bayesian classifier isn't very good.

We might want to look into expanding [this heuristic](https://github.com/github/linguist/blob/3f4b8368e81430e3353cb5ad8b781cd044697347/lib/linguist/heuristics.rb#L300-L306) to help with this.

jbrooksuk:
Yay!

PR #3122, opened 2016-07-16
Add .jet and .jetl as Kotlin file extensions

FarbodSalamat-Zadeh:
Although these file extensions are deprecated, they are still Kotlin file extensions:

https://discuss.kotlinlang.org/t/whats-the-deference-between-kt-and-kts-files-jet-and-jetl-files/1058

### DIFF
--- a/samples/Kotlin/Foo.jetl
+++ b/samples/Kotlin/Foo.jetl
@@ -1,0 +1,47 @@
+package addressbook
+
+class Contact(
+  val name : String,
+  val emails : List<EmailAddress>,
+  val addresses : List<PostalAddress>,
+  val phonenums : List<PhoneNumber>
+)
+
+class EmailAddress(
+  val user : String,
+  val host : String
+)
+
+class PostalAddress(
+  val streetAddress : String,
+  val city : String,
+  val zip : String,
+  val state : USState?,
+  val country : Country
+) {
+   assert {(state == null) xor (country == Countries["US"]) }
+}
+
+class PhoneNumber(
+  val country : Country,
+  val areaCode : Int,
+  val number : Long
+)
+
+object Countries {
+  fun get(id : CountryID) : Country = countryTable[id]
+  
+  private var table : Map<String, Country>? = null
+  private val countryTable : Map<String, Country>
+    get() {
+      if (table == null) {
+        table = HashMap()
+        for (line in TextFile("countries.txt").lines(stripWhiteSpace = true)) {
+          table[line] = Country(line)
+        }
+      }
+      return table
+    }
+}
+
+class Country(val name : String)


Alhadis:
`.jet` and `.jetl` need to be listed after `.kt`, and before `.ktm`. That's what's causing the build error.

FarbodSalamat-Zadeh:
@Alhadis 

Thank you for your note. I will push a commit to fix this. 

Could you explain why they have to be listed in that order? Just interested.

Alhadis:
I actually don't know. One of the project's maintainers can probably answer that question better than I.

pchaigno:
@FarbodSalamat-Zadeh Thanks for the pull request! Please see [the contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language) in order to complete the necessary steps to add an extension. In particular, you'll need to link to some example of in-the-wild usage via a github.com search.

> Could you explain why they have to be listed in that order? Just interested. 

I don't think there's a particular reason. It just makes things easier when there's a long list of file extensions for a language or when several contributor are adding extensions to the same language.

Alhadis:
That was always my theory, but the matter about capitalisation affecting alphabetisation made me wonder if the reasons were performance-related. =) E.g., UPPERCASE codepoints coming before lowercase, and whatnot...

FarbodSalamat-Zadeh:
I've reordered the extensions, and found some `.jetl` files from the initial commit to the JetBrains/kotlin repository:

https://github.com/JetBrains/kotlin/tree/369b1974782b821e44b7aa6cd68e2e41eb2ba036/examples/src 

Do I need to create a new file for the samples directory, as there currently does not seem to be sample files for the `.ktm` and `.kts` extensions already listed in the `languages.yml` file?

pchaigno:
> I've reordered the extensions

The two new extensions need to be before `.ktm`. The first extension `.kt` is the primary extension and the following extensions must be alphabetically sorted.

> Do I need to create a new file for the samples directory, as there currently does not seem to be sample files for the `.ktm` and `.kts` extensions already listed in the `languages.yml` file?

You only need to add samples for the two extensions you added in `samples/Kotlin/`.

FarbodSalamat-Zadeh:
I've added samples now.

pchaigno:
> @FarbodSalamat-Zadeh Thanks for the pull request! Please see the contribution guidelines in order to complete the necessary steps to add an extension. In particular, you'll need to link to some example of in-the-wild usage via a github.com search.

Were you able to find good github.com search queries for these file extensions? Do you need help with that?

FarbodSalamat-Zadeh:
@pchaigno

I found `.jetl` files from the first commits to the JetBrains/kotlin repository:

https://github.com/JetBrains/kotlin/tree/369b1974782b821e44b7aa6cd68e2e41eb2ba036/examples/src

Apart from that, I could find much, as apparently, the `.jet` and `.jetl` extensions are deprecated. 

Will this be sufficient?

pchaigno:
I found [only 23 files for `.jetl`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ajetl+class+val&type=Code&ref=searchresults). Similarly, I found [only 53 repositories for the `.jet` file extension](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ajet+class+val&type=Code&ref=searchresults). Unless we can make better search queries, I'm afraid that won't be enough :/

FarbodSalamat-Zadeh:
@pchaigno

Okay. How many should we be aiming for?

Also, is there a particular reason why we need to have a certain amount of usages of these extensions. Surely the fact that they're extensions for the Kotlin language should be enough. 

Thanks.

pchaigno:
We should be aiming for hundreds of repositories as per the contribution guidelines (which usually translates in more than a thousand files). For the `.jet` extension we may also need to identify the other languages that use it if we want to avoid conflicts.

First, each file extension we add has an impact on performance as a lot of new files need to go through the Linguist detection pipeline.
There's also the issue of file extension conflicts. For instance, if we add the `.jet` file extension for Kotlin, all `.jet` files on github.com will be recognized as Kotlin. If the file extension is used by other languages, it will generate lots of false positives. Thus, we have to add the file extension to all languages that use it and make sure our language detection strategies work well. That's a lot of work.
Even if there are no conflicts when we add the file extension, some might appear in the future. So for each new file extension there is also a maintenance cost to take into account.
I hope that helps explaining why we have this sort of rules of thumb in place :-)

For extensions that don't meet the requirements yet, there are always the [Linguist overrides](https://github.com/github/linguist#overrides). They can help you fix the results on a per-repository basis.

FarbodSalamat-Zadeh:
@pchaigno

Thank you for the explanation, you have a very valid point. 

I have yet to find more repositories with these extensions, but I will try to update this conversation if I find more.

PR #3110, opened 2016-07-12
https://github.com/search?utf8=✓&q=extension%3Arexx+NOT+nothack&type=Code&ref=searchresults

idrougge:

PR #3124, opened 2016-07-17
Add REXX.

larsbrinkhoff:
Fixes and supercedes #3110.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3131,6 +3131,17 @@ REALbasic:
   tm_scope: source.vbnet
   ace_mode: text
 
+REXX:
+  type: programming
+  aliases:
+  - arexx
+  extensions:
+  - .rexx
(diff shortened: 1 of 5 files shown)


larsbrinkhoff:
@idrougge I had to remove syntax highlighting, because http://github.com/mblocker/rexx-sublime/ doesn't have a license.

arfon:
👍 looks good to me.

idrougge:
@larsbrinkhoff @arfon Is this supposed to work now or must it be further integrated in Github's backend?

Alhadis:
It'll take effect the next time there's a new release of Linguist, which usually happens every couple of weeks or so. =)

larsbrinkhoff:
The Sublime grammar added a license.  In case someone wants to give it a stab.

idrougge:
MIT licence, so that is compatible with Linguist, too.

PR #2139, opened 2015-02-20
Improve repository language statistics

aroben:
This PR fixes a number of bugs that have been pointed out since #2097 landed.

* Reclassifies XML as `data`, since in most cases it is merely a serialization format for a tool (e.g., a project file for an IDE), not something essential to the repository that contains it.
* Adds a few more documentation and vendor patterns based on some repositories that are getting misclassified as HTML.

I'm still looking for more misclassified repos and will update this PR with more fixes.

Fixes #2124 #2134 #2138 

/cc @github/linguist

PR #3125, opened 2016-07-19
Make XML a markup language

FarbodSalamat-Zadeh:
XML is, by definition, a markup language, so it should be considered one, in my opinion, by GitHub's linguist. The first sentence of the [Wikipedia page for XML](https://en.wikipedia.org/wiki/XML) is this:

>Extensible Markup Language (XML) is a markup language that defines a set of rules for encoding documents in a format that is both human-readable and machine-readable.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4032,7 +4032,7 @@ XC:
   ace_mode: c_cpp
 
 XML:
-  type: data
+  type: markup
   ace_mode: xml
   aliases:
   - rss


larsbrinkhoff:
I believe your're technically correct (the best kind of correct), but keep in mind it would change the way repositories are classified.  As of now, XML - being data - is not considered in language statistics. Since XML files are so common, I'm afraid there's a risk many repositories would change from being classified according to their most prevalent programming language, to XML.

But then, this happened before with HTML and CSS, and I haven't seen many complaints about that.

Alhadis:
Thing is, XML is actually rarely used for marking up documents. That's more an XSLT thing, and that *is* counted in language statistics.

Let's be realistic: when you see an XML file, you expect it'll be holding data instead of something that can be used to present a formatted document to a reader.

FarbodSalamat-Zadeh:
In response to @Alhadis 's point, I know XML files are the most common way of positioning widgets in a layout in Android, similar to how HTML organises things on a webpage.

Alhadis:
HTML doesn't exist to "organise things on a webpage". HTML exists to present information in a structured, semantically sound way. XML exists for entirely different reasons.

Alhadis:
BTW, we may as well be classing JSON as a markup language too, since it can also be used for arranging elements for an application.

larsbrinkhoff:
@Alhadis That seems very strange to me.  What does "markup language" mean to you?

Alhadis:
I was sarcastically saying that XML isn't a markup language, despite its name.

FarbodSalamat-Zadeh:
Anyone else? Thoughts?

Alhadis:
Just noticed that XML [*used* to be classed](https://github.com/github/linguist/pull/2139) as markup on the site, but a GitHub staff member fixed it, citing obvious reasons echoed above.

To be blunt, I don't think this PR has a very bright future, mate...

FarbodSalamat-Zadeh:
Fair enough. 

Although it could add a little complication, I was thinking that there could be a file that you could create or modify in a GitHub repository to include or exclude particular languages from the language bar. 

In this case, the language bar would show languages based on the clarifications in `languages.yml` by default, but if the owner wanted to include XML, for example, they could add a line to the file like so:

```
xml
```

This would be most useful with GitHub repositories that do not want to have HTML in the language bar (I understand this has been an issue for a long time and still is for some repositories), as ideally, they would be able to use this line to remove it:

```
-html
```

(Note the `-` for exclude). 

I think the main advantage with this is that repositories could alter which languages should be shown based on the defaults from `languages.yml`, as sometimes the defaults don't make sense for some repositories. Using the HTML example, it would make sense for open source websites to include HTML in the language bar, but perhaps not a C project using HTML for data.

Alhadis:
Erm, they can already use [`.gitattributes` overrides](https://github.com/github/linguist#troubleshooting) to filter specific files from language classification. However, that still won't do anything for authors wanting to display data formats in their repository's language bar.

There'd probably be too many complications involved with "making an exception" for data files in some repositories, and not others. Recall that these statistics are also used by GitHub to monitor the language's use site-wide, so it's not simply a case of what's depicted on the author's repo. For the added effort this would involve from the site's maintenance, there'd be very little gain.

FarbodSalamat-Zadeh:
@Alhadis

>they can already use .gitattributes overrides

As silly as it sounds, I was not aware of that :)

>Recall that these statistics are also used by GitHub to monitor the language's use site-wide, so it's not simply a case of what's depicted on the author's repo.

So the same statistics that are displayed on the language bar are used for determining how much languages are used throughout all GitHub repositories? I'm guessing then that languages like XML aren't monitored at all?

>I'd say the code for monitoring and evaluating statistics is already written on grounds that the underlying classifications will behave stably. What you're proposing would probably involve a hell of a lot of refactoring.

If this is the case, then I completely would agree that it is not worth implementing this idea.

Alhadis:
> As silly as it sounds, I was not aware of that :)

I'm starting to think this repository needs a [PR template](https://github.com/blog/2111-issue-and-pull-request-templates) to address FAQs...

FarbodSalamat-Zadeh:
@Alhadis Sorry about that - it was just an idea that came to mind (not originally part of the PR).

I guess if XML won't be changed to a markup language, I should close this...

Alhadis:
No hard feelings, mate. These things get brought up quite often - it never hurts to run a search before submitting a PR/issue, even if you think it's unlikely to have been brought up before.

FarbodSalamat-Zadeh:
Thanks @Alhadis - I'll keep that it mind for next time.

PR #3132, opened 2016-07-25
An example comparison for docs

zyuhu:
test


Alhadis:
... why does this keep happening...? :( Why us?

zyuhu:
@Alhadis Sorry about this, just jump here from the sort of man page of github, thought it's official test, didn't realize it's personal.

pchaigno:
@zyuhu Do you have the link to that man page? We're having this issue a lot, would be good if we could change the example to something more appropriate...

zyuhu:
@pchaigno I follow this path to here:
1. the first jump is from github help site: https://help.github.com/
2. search "Comparing commits across time", click it.
3. then went to https://github.com/github/linguist/compare/
4. there you would see the title " an-example-comparison-for-...", and that's where I came from.

Hope this will help.

pchaigno:
Thanks @zyuhu! I opened #3135 to see if we can improve this.

PR #3141, opened 2016-08-02
Make PostScript red again

Alhadis:
Second crack at what I tried doing a [couple months ago](https://github.com/github/linguist/pull/3069). This PR sets the colour of PostScript to [PANTONE 485 C](https://www.pantone.com/color-finder/485-C), as per the language's branding. And if there's ANY language whose colours deserve to retain their tonal integrity, it's the language that pretty much pioneered the whole world of desktop publishing.

The conflicting colours (Scala and, uhm, "Red") were pushed apart somewhat. Red was made redder, and Scala was given a desaturated darker shade of red. Scala's change is a bit more noticeable than Red's, but it was [too close to HTML's colour](https://github.com/guardian/membership-frontend) anyway.

I'd show you a "before-and-after", but my Internet connection decided to die as I was looking for a repository that'd show a Red/HTML comparison. I've had to copy+paste this PR message to my phone to post it using my data. Whoever said Australia is the "lucky country" clearly wasn't talking about our fucking Internet.

PR #3143, opened 2016-08-04
Classify PostScript as a programming language

Alhadis:
No idea how I missed this the other night in [my other PS-related PR](https://github.com/github/linguist/pull/3141), but I can tell you now, PostScript is definitely not markup. It's a full-blown Turing-complete programming language which just happens to be used primarily for graphical purposes. Wikipedia's entry on [DSLs](https://en.wikipedia.org/wiki/Domain-specific_language) even uses PostScript to describe why the term "Domain-specific language" can be nebulous:

> For example, Perl was originally developed as a text-processing and glue language, for the same domain as AWK and shell scripts, but was mostly used as a general-purpose programming language later on. **By contrast, PostScript is a Turing complete language, and in principle can be used for any task, but in practice is narrowly used as a page description language.**

I don't know if there's any difference between `markup` and `programming` as far as Linguist's classifications are concerned, but I felt impelled to correct this nonetheless. [Here's an example of a program](https://github.com/bwipp/postscriptbarcode) written in PostScript that's written like, uh, a program.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2930,7 +2930,7 @@ Pony:
   ace_mode: text
 
 PostScript:
-  type: markup
+  type: programming
   color: "#da291c"
   extensions:
   - .ps


Crissov:
Yes, Postscript is a full-blown programming language. The question is whether actual `.ps` *files on Github* are more likely to be document files which should be categorized as either `data` (≈ automatically generated code, may be human-readable) or `markup` (≈ edited manually, must be human-readable) – if I understand the purpose of Linguist correctly. I would assume that most `.ps` files of this latter kind would be in the scope of `.gitignore` files and therefore Github is indeed one of the few places where it makes sense to treat Postscript files as source code (like in the example repository), hence `programming`.

Files in some other languages are in a similar ambiguous situation or gray area, e.g. hand-authored SVG documents (`markup`) vs. exported SVG images (`data`).

Alhadis:
@pchaigno Any thoughts on this, or is this PR slightly too pedantic? =)

pchaigno:
I'm inclined to agree with @Crissov (unless we can demonstrate that most PostScript files are actually programs).

Alhadis:
Does this actually have any bearing on how files are classified on GitHub? I assumed this was a purely semantic change, but it sounds like there's more to it than that.

pchaigno:
I don't think this has any impact on the language statistics currently, but that could change in the future. For instance, it had been considered in [#2139 (Comment)](https://github.com/github/linguist/pull/2139#issuecomment-75292264).

Not sure what to do here. @arfon What's your opinion?

larsbrinkhoff:
Here is my take on this.  PostScript is obviously a complete programming langauge.  **However**, that may not make it appropriate to have GitHub classify it as a programming language!  In my opinion, of course GitHub staff makes these decisions.

Why do I think so?  Because I believe GitHub users want their repositories classified according to the "most important" (programming) language.  And I believe users to regard most PostScript files as documents, not programs.

Alhadis:
It just occurred to me that generated PostScript files already contain header data that mark them as auto-generated. Maybe we could use that to filter them from statistics in future.

Issue #3146, opened 2016-08-07
Haskell highlighting: class definition with no "where"


Fedjmike:
When a class definition doesn't have a "where" body (which is legal Haskell) the hightlighter gets stuck and doesn't highlight the rest of the code correctly. Examples:

```haskell
-- Highlighted as comment
class X y
-- Not highlighted correctly
```

https://github.com/Fedjmike/ngen/blob/07296373ff6dc9a7933652f6fa1bc373f0c3d06f/language.hs#L12

Alhadis:
This is an issue with the [language-grammar](https://github.com/textmate/haskell.tmbundle) Linguist uses to provide syntax highlighting. You're better off submitting a bug report there. =)

PR #3147, opened 2016-08-07
Change grammar used for Haskell highlighting

Alhadis:
#3146 reminded me the [Atom package](https://github.com/atom-haskell/language-haskell/issues/80) for Haskell support is frequently updated and well-maintained.

Surely enough, it fixes [the highlighting issue](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fgithub.com%2Fatom-haskell%2Flanguage-haskell%2Fblob%2Fmaster%2Fgrammars%2Fhaskell.cson&grammar_text=&code_source=from-text&code_url=&code=--+Highlighted+as+comment%0D%0Aclass+X+y%0D%0A--+Not+highlighted+correctly) and probably others we don't know about, too.


Alhadis:
inb4 CI reports an error with my attempt at submodule removal

lierdakil:
Ouch. I don't think you want to use all grammars defined in language-haskell, some of those are for internal use (like limited subset for type tooltips highlighting, and a grammar for highlighting compiler output). Long story short, ignore `hint.*` grammars.

pchaigno:
@lierdakil That's fine, with the pull request as is, we'll only use the `source.haskell` scope. Thanks for reviewing :+1:

Alhadis:
@lierdakil That reminds me, I noticed these hints get listed in Atom's grammar-selector menu:

Figure 1

You might be surprised to know a grammar's `name` property is actually optional; only the `scopeName` seems to be required. Not specifying a `name` hides the grammar from being listed, which might be a better way of including "internal" grammars. =)

lierdakil:
@Alhadis, right, that's my bad, I honestly didn't think twice about that when adding those ^_^"

Alhadis:
Conflict fixed!

We need notifications for when a pending PR suddenly has conflicts... :(

arfon:
:+1: :shipit:

PR #3148, opened 2016-08-09
Added syntax highlighting for ASN.1

ajLangley12:
https://github.com/search?utf8=✓&q=ASN.1+extension%3Aasn+extension%3Aasn1&type=Code&ref=advsearch&l=&l=

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -755,3 +755,6 @@ url = https://github.com/austinwagner/sublime-sourcepawn
 [submodule "vendor/grammars/language-roff"]
 	path = vendor/grammars/language-roff
 	url = https://github.com/Alhadis/language-roff
+[submodule "vendor/grammars/language-asn1"]
+	path = vendor/grammars/language-asn1
+	url = https://github.com/ajLangley12/language-asn1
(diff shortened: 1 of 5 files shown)


Alhadis:
What the hell is ASN.1?

ajLangley12:
https://en.wikipedia.org/wiki/Abstract_Syntax_Notation_One

It's a language used to send instructions between two different machines that don't necessarily share the same implementation language, in a really compressed format (it gets encoded into binary when sent.

ajLangley12:
Sorry for the crappy pull request, I was having difficulties with the instruction for downloading the license.

Alhadis:
Hrm, never heard of it, but it sounds like it's less of a language, and more of an established notation. Is that correct...? As in, it's as much of a language as ABNF grammars are...

ajLangley12:
I believe you are correct.

ajLangley12:
Please let me know if I need to change anything to reflect the difference in terminology.

ajLangley12:
Looks like the travis build is hanging. Anything to do about this on my end?

PR #3162, opened 2016-08-23
Grammar for REXX from Sublime Text package

pchaigno:
REXX support was added recently in #3124, but the grammar didn't have a permissive license at the time. It now has an MIT license, so we can use it.


larsbrinkhoff:
Build error:

    No submodule mapping found in .gitmodules for path 'vendor/grammars/haskell.tmbundle'

arfon:
:+1:

PR #3175, opened 2016-08-28
Classify CSON as data

Alhadis:
**CSON** stands for **C**offee**S**cript **O**bject **N**otation. From that name alone, you'd correctly guess that it holds the same relation to CoffeeScript as JSON does to JavaScript. [It's a data format](https://github.com/bevry/cson#what-is-cson) that uses the same syntax and semantics as its namesake, but without the ability to include arbitrary commands.

From [`cson-parser`'s description](https://www.npmjs.com/package/cson-parser#why-not-just-use-coffeescript-directly):
> **Why not just use CoffeeScript directly?**
You don't want data files being able to run arbitrary code. Even when ran in a proper sandbox, `while(true)` is still possible.

My [Emacs Lisp repo](https://github.com/Alhadis/language-emacs-lisp) is a good representation of why data formats are filtered from the stats-bar; the only chunk of programming code is a JavaScript file, and it's being misrepresented as 5.1% of the codebase.

On a lighter note, I actually couldn't find any CSON files in the wild that *weren't* Atom-related, so it's tempting to stick "Atom JavaScript" in as an alias for CoffeeScript (in reference to Emacs Lisp...).

**Origin of the samples:**
- [`base.cson`](https://github.com/atom/atom/blob/master/keymaps/base.cson): Atom
- [`config.cson`](https://github.com/DanBrooker/file-icons/blob/v2/config.cson): Truncated version of File-Icons v2's config file
* [`ff-sfd.cson`](https://github.com/Alhadis/language-fontforge): Taken from a grammar repo I slapped together last night for a DSL
* [`wercker-status.cson`](https://github.com/felipefdl/wercker-status/blob/c0e6f8332e247f04637633b578676558fbc8df5d/menus/wercker-status.cson): Some dude's package, [it's MIT-licensed](https://github.com/felipefdl/wercker-status/blob/c0e6f8332e247f04637633b578676558fbc8df5d/package.json#L7)


Alhadis:
Done! Once merged, I'll update the [Pic PR](https://github.com/github/linguist/pull/3277/files) ("Pic request"?) to include an incremented ID.

PR #3177, opened 2016-08-29
Add support for LotusScript

dev-ng:
I propose simple solution to highlight syntax in LotusScript lss files. It can be treated as Visual Basic.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4007,6 +4007,7 @@ Visual Basic:
   - .cls
   - .frm
   - .frx
+  - .lss
   - .vba
   - .vbhtml
   - .vbs


pchaigno:
Please see the [guidelines to add support for a new extension to Linguist](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language). In particular, notice the requirement on in-the-wild usage.

dev-ng:
Do you mean I need to add this: [github_search](https://github.com/search?utf8=%E2%9C%93&q=in+lotusscript&type=Code&ref=searchresults)

LotusScript is very similar to ViasualBasic [wikipedia](https://en.wikipedia.org/wiki/LotusScript)
However there are some differencies: [ibm](https://www.ibm.com/support/knowledgecenter/SSVRGU_9.0.1/com.ibm.designer.domino.main.doc/H_ACCESSING_THE_DOMINO_OBJECTS_THROUGH_COM_LANGUAGE.html)

Actually, according to these differencies, it should be then a separate definition. The reason I committed initial request is that I'm going to upload some framework libraries to share with the community and I would like to see it highlighted and to keep it simple if possible. One of the options is just to rename files to .bas, but this could confuse people. I would like to focus on the project I want to share and not on the tools like highlighting.

brandonblack:
Echoing @pchaigno's feedback, I only see 62 repositories and 38 users for this. It unfortunately doesn't meet our minimum requirements for in-the-wild adoption.

From the [Contributor's Guide](https://github.com/github/linguist/blob/master/CONTRIBUTING.md):
>We try only to add languages once they have some usage on GitHub. In most cases we prefer that each new file extension be in use in hundreds of repositories before supporting them in Linguist.

For your specific use case, you should [use .gitattributes](https://github.com/github/linguist/blob/master/README.md#using-gitattributes) to override linguist behavior.

PR #3180, opened 2016-08-30
Update GraphQL language support

rmosolgo:
The new version of the Atom grammar for GraphQL contains a couple of improvements:

- Better highlighting for directives: previously, only lowercase letters were allowed. Now,
  any GraphQL identifier will be recognized as a directive (as the spec describes)
- Better support for the &quot;Schema language&quot;: highlighting for some of the constructs
  in the schema description language. Sadly, there&#39;s an outstanding (unsolveable!?!?)
  issue with field highlighting, see https://github.com/rmosolgo/language-graphql/issues/3</textarea>


pchaigno:
Grammars referenced by Linguist are updated with every release automatically. Thus, this grammar will be updated in the next couple weeks.

rmosolgo:
Cool, thanks!

PR #3182, opened 2016-08-31
Exclude "dist" directory from language statistics

adius:
It is listed in a lot of .gitignore files, if this is any indicator that it should not be checked: https://github.com/search?utf8=%E2%9C%93&q=dist+filename%3Agitignore&type=Code&ref=searchresults

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -15,6 +15,9 @@
 # Dependencies
 - ^[Dd]ependencies/
 
+# Distributions
+- (^|/)dist/
+
 # C deps
 #  https://github.com/joyent/node
 - ^deps/


pchaigno:
@adius Thanks for the pull request! Do you have a few examples of `dist` directories on github.com? I've looked up a few of the repositories from the search results, but none of them seem to actually have a `dist` directory.

adius:
🙈 Yeah of course none of them have a `dist` directory, because having dist in your gitignore does exactly mean this: "Don't upload the directory to GitHub" 😂

So what I'm saying is, that if somebody uploads a `dist` directory it was probably accidentally or they explicitly wanted to bundle 3rd party files. In any way it should not be checked for language statistics!

pchaigno:
> 🙈 Yeah of course none of them have a `dist` directory, because having dist in your gitignore does exactly mean this: "Don't upload the directory to GitHub" 😂

Hm, right :flushed: Don't know where my head's at.

I went through [a few of these directories](https://github.com/search?o=desc&q=path%3Adist+if&s=indexed&type=Code&utf8=%E2%9C%93) and, although the usage seems to vary a bit, I couldn't find a case where we wouldn't want to vendor it. So :+1: for me.

Alhadis:
[Relevant](https://github.com/Alhadis/Accordion/blob/master/.gitattributes).

arfon:
👍 thanks @adius

PR #3183, opened 2016-08-31
Add support for Schematron XML validation files (.sch)

akhtars:
Update `languages.yml` and add three public domain sample documents.

["In-the-wild" examples](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asch+schematron). The `.sch` extension is currently shared in Linguist by Eagle and KiCad, both of which already have example files.

If a heuristic is necessary, some potential options:

1. Find "schematron", which is likely (but not guaranteed) to appear in an XML namespace declaration.
2. Find the top-level element `<schema>` and/or a subset of other common (but optional) elements (e.g. `<rule>`, `<assert>`, `<pattern>`, `<phase>`, `<report>`), each of which may appear with various attributes and namespace prefixes.

Thanks for maintaining this valuable project.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4139,6 +4139,7 @@ XML:
   - .rdf
   - .resx
   - .rss
+  - .sch
   - .scxml
   - .sfproj
   - .srdf
--- a/samples/XML/HITSP_C32.sch
+++ b/samples/XML/HITSP_C32.sch
@@ -1,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+
+<!--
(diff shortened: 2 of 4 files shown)


arfon:
> Update languages.yml and add three public domain sample documents.

Thanks for thinking of this @akhtars.

@mlinksva - is public domain OK for Linguist or should we be looking for example files with explicitly-stated licenses?

mlinksva:
Same as https://github.com/github/linguist/pull/3137 -- public domain by virtue of being work of US federal government employee is not _ideal_ because in theory it could be subject to copyright outside the US -- but this hasn't proven to be a real problem as far as I know. Go ahead and use.

arfon:
> Go ahead and use.

Thanks @mlinksva.

arfon:
> Go ahead and use.

Thanks @mlinksva.

Issue #3186, opened 2016-09-02
Shell: Incorrect syntax highlighting for command substitution using backticks with comment inside


Herst:
All these would output `(foo)()(bar)`, which is not what the syntax highlighting implies:

``` sh
printf '(%s)' foo "`# '`" bar
```

``` sh
printf '(%s)' foo "`# "`" bar
```

``` sh
printf '(%s)' foo "`# '`" \
  bar
```

``` sh
printf '(%s)' foo "`# "`" \
  bar
```

Alhadis:
Syntax highlighting on GitHub is provided by TextMate-compatible grammars, which Linguist selects to provide code colouring. For shell-scripts, the highlighting is provided by [atom/language-shellscript](https://github.com/atom/language-shellscript). You should report this issue there.

Herst:
Oh, OK. It was https://help.github.com/articles/creating-and-highlighting-code-blocks/ ("We use Linguist to perform language detection and syntax highlighting.") that lead me here.

pchaigno:
> It was https://help.github.com/articles/creating-and-highlighting-code-blocks/ ("We use Linguist to perform language detection and syntax highlighting.") that lead me here.

@arfon Maybe this could be slightly rephrased to explain that Linguist only *selects* the grammars for syntax highlighting? I'm not sure what's the best way to redirect users to those grammars though :/

arfon:
> @arfon Maybe this could be slightly rephrased to explain that Linguist only selects the grammars for syntax highlighting?

Yeah, that's probably worth updating to make things a little clearer. 

>  I'm not sure what's the best way to redirect users to those grammars though :/

I mean, [this list](https://github.com/github/linguist/tree/master/vendor/grammars) is the source of truth here but the idea of sending people to a folder of Git submodules is literally the least user-friendly thing I think I've ever thought of...

Alhadis:
We could add a `README.md` file to the [`vendor/` directory](https://github.com/github/linguist/tree/master/vendor) that lists what each language uses for syntax highlighting. Something like this, perhaps:

> - **1C Enterprise:** [xDrivenDevelopment/atom-language-1c-bsl](https://github.com/xDrivenDevelopment/atom-language-1c-bsl)
> - **AGS Script:** [textmate/c.tmbundle](https://github.com/textmate/c.tmbundle)
> - **AMPL:** [ampl/sublime-ampl](https://github.com/ampl/sublime-ampl)
[...]

PR #3189, opened 2016-09-06
Replace Batchfile grammar and cleanup submodule list

Alhadis:
After starting [work on this](https://github.com/github/linguist/issues/3186#issuecomment-244612496), I noticed some oddities in the grammar/submodule lists:

* There're two grammars that were accidentally commited to this repository after they were partially removed in favour of other grammars:
	* [`less.tmbundle`](https://github.com/github/linguist/tree/master/vendor/grammars/less.tmbundle)
	* [`Sublime-Inform`](https://github.com/github/linguist/tree/master/vendor/grammars/Sublime-Inform)

	<br/>Evidently, the submodules were deregistered without their corresponding contents removed from the working tree. So they were erroneously included in the next commit.

* Batchfile highlighting is provided by an [abandoned bundle](https://github.com/ericzou/ebundles) of grammars that hasn't been updated in 8 years. The grammar is buggy and unmaintained, and has [highlighting issues](https://github.com/github/linguist/issues/2748) that'll never be noticed by the author. I've found a [fresh, healthy replacement](https://github.com/mmims/language-batchfile) that [fixes the issues](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fgithub.com%2Fmmims%2Flanguage-batchfile%2Fblob%2Fmaster%2Fgrammars%2Fbatchfile.cson&grammar_text=&code_source=from-text&code_url=&code=%40echo+off%0D%0Arem+this+text+should+be+green%0D%0Aecho+hello+world) previously reported.

Other than the egregious cases, I did some assorted tidy-up that would've been too trivial to warrant a separate pull request.

PR #3192, opened 2016-09-06
Add script to generate a reader-friendly grammar list

Alhadis:
Users who encounter syntax highlighting errors on GitHub are apt to report them here, unaware that Linguist isn't directly responsible for the actual highlighting. This may be partly due to the fact the [site's own help articles](https://help.github.com/articles/creating-and-highlighting-code-blocks/) point to Linguist as its *"source of language recognition and syntax highlighting"*.

Currently, we lack anything better to direct users to who want to report syntax highlighting errors. @arfon [put it best](https://github.com/github/linguist/issues/3186#issuecomment-244610844):

> I mean, [this list](https://github.com/github/linguist/tree/master/vendor/grammars) is the source of truth here but the idea of sending people to a folder of Git submodules is literally the least user-friendly thing I think I've ever thought of...

My suggestion was to maintain a list of repository links in the form of a `README.md` file, which would be programmatically updated whenever grammars are added or removed. This PR is an implementation of that suggestion.

Some things that're missing:
* **We need to tie this into a build step:**  
I'm not familiar with Bundler or Rake, so I don't know where or how to go about having this script called when `bundle exec rake` is called. Somebody help, please? =)
* **A better preamble for the readme:**  
I only added it as a placeholder to verify that rerunning the script wouldn't chop it off each time it executed. Feel free to replace the wording with something you feel is more appropriate.

**NOTE:** ~~I've yet to investigate why some of the grammars aren't being included (such as Perl). Will look into this soon.~~ Resolved by 3247d46.


pchaigno:
> I'm not familiar with Bundler or Rake, so I don't know where or how to go about having this script called when bundle exec rake is called. Somebody help, please? =)

Wouldn't it be sufficient to tie this to `convert-grammar --add`?
It might also be time we make a script to remove grammars (to remove the submodule, the lines in `grammars.yml`, and update this list).

Alhadis:
I could easily do that, but... I kinda thought that's what `prune-grammars.rb` does...

pchaigno:
`prune-grammars.rb` only removes grammar from `grammars.yml` when their scopes are not used. It could actually benefit from a script that *entirely* removes the grammar. We can make the script to remove grammars in a second time/PR though.

Alhadis:
Agreed. Right now, we could have something to divert syntax-related queries to, heh. =)

Apropos of annoying questions, [thoughts on this](https://github.com/github/linguist/pull/3149)? It's beginning to get mouldy.

Alhadis:
@pchaigno The reason I thought to leave it to a `bundle rake` step was so the list would only be updated for the public once a release was cut. So there's no risk of users reporting issues to the wrong repositories if a grammar's been swapped out, but not showing results on GitHub yet. =)

pchaigno:
Good thinking! I think the easiest solution will be to add it to [the instructions to release a new version](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#releasing).

Alhadis:
Sounds good. =) I'll leave that to you or @arfon to take care of, haha.

arfon:
Sorry for taking a frikin' eternity to look at this @Alhadis. This is looking pretty good and agreed we should hook this into some other part of the build/docs for contributing so people know to run this when adding a new grammar.

I just made my own branch with this PR and current `master` and it looks like it's tripping up on [L80](https://github.com/github/linguist/pull/3192/files#diff-db68bd961ab0ce17baa056c59024741dR80) when  scope is `text.groff`. The error is 
```
./script/list-grammars:83:in `block in to_markdown': undefined method `chomp' for nil:NilClass (NoMethodError)```

It's not immediately clear what's going on here.

Adding a couple of lines to print out some debug information:

```
      else
        puts scope if @sources[scope].nil?
        next if @sources[scope].nil?
        submodule = @submodules[@sources[scope].chomp("/")]
        next unless submodule
        short_url = submodule[:short]
        long_url  = submodule[:url]
      end```

it looks like it's these two grammars causing issues:

```
source.dosbatch
text.groff
```

Any ideas what might be going on here? 😕

Alhadis:
Ah, it might've been because it was waiting [on this](https://github.com/github/linguist/pull/3189/files). The `dosbatch` grammar was weird in that it was using a targeted subdirectory... I'm not sure what the hell's the deal with `text.groff` though (perhaps a lingering vestige of the old manpage grammar?)

I've synced with master. Could you try again?

Alhadis:
Can't reproduce; the list updated perfectly fine. Must've been because of the issue I mentioned, then. =)

I should mention that choosing grammars with non-standard grammar locations will cause issues, as was the case with the DOSBatch grammar. Such cases will need to be [explicitly handled here](https://github.com/github/linguist/pull/3192/files#diff-db68bd961ab0ce17baa056c59024741dR62), but in the case of DOSBatch, I knew the grammar needed replacing... so I purposefully left it out.

arfon:
> Can't reproduce; the list updated perfectly fine. Must've been because of the issue I mentioned, then. =)

Re-cloned the repo and checked out this PR and I can't reproduce it either. Go figure. I've made a new PR with these changes and some small additions over in #3307

Alhadis:
I blame submodules. Always blame submodules.

arfon:
> I blame submodules. Always blame submodules.

Thinkin' the same thing.

PR #3193, opened 2016-09-06
Add script to add or replace grammars

Alhadis:
The current procedure for adding or replacing grammars is fiddly and error-prone, and definitely not contributor-friendly. So I wrote a script to automate the process:

```bash
script/add-grammar https://github.com/JaneSmith/MyGrammar
script/add-grammar --replace old.tmbundle https://github.com/JaneSmith/NewBundle
```

The script attempts to be idiot-proof: it accepts a variety of URL syntaxes, resolving them all to an HTTPS link:

```bash
script/add-grammar https://Alhadis@bitbucket.org/Alhadis/language-roff
script/add-grammar git@github.com:Alhadis/language-roff.git
script/add-grammar Alhadis/language-roff
script/add-grammar http://github.com/Alhadis/language-roff.git
script/add-grammar github.com/Alhadis/language-roff.git
```

To replace an existing grammar, the argument for `--replace` is the name of the unwanted submodule, without or without a leading `vendor/grammars/` attached:

```bash
script/add-grammar --replace vendor/grammars/old.tmbundle
script/add-grammar --replace old.tmbundle
```
Both resolve to the same thing. If the resolved name doesn't match a path in the `vendor/grammars` directory, it barks at the user and exits.


larsbrinkhoff:
Automation is good!  Until our robot overlord smashes us into a pulp.

Alhadis:
Relax, it'll probably be coded in Java.

script/add-grammar
```diff
@@ -0,0 +1,92 @@
...
+end
+
+# Print debugging feedback to STDOUT if running with --verbose
+def log(msg)
+  puts msg if $verbose
+end
+
+
+usage = """Usage:
+  #{$0} [--replace grammar] url
```
pchaigno: Could you document the `-v` option here too?

Alhadis: Oh! That was only for debugging/development... I was kinda learning Ruby as I was writing this, so I had to have something to see what the hell I was doin'. :D Spent half the night screaming "WHY ISN'T THIS JAVASCRIPT?!"

`#drama-queen`

Alhadis: + okay, hang on

script/add-grammar
```diff
@@ -0,0 +1,92 @@
...
+repo_old = parse_submodule($replace) if $replace
+
+if repo_old
+  log "Deregistering: #{repo_old}"
+  `git submodule deinit #{repo_old}`
+  `git rm -rf #{repo_old}`
+end
+
+log "Registering new submodule: #{repo_new}"
+`git submodule add -f #{https} #{repo_new}`
```
pchaigno: We should check the output code here before continuing execution. If the directory already exists for instance, it will error out and add a second line in `grammars.yml` anyway (because the next line which runs whatever happens).

Alhadis: Shit, good catch. Will fix.

script/add-grammar
pchaigno: Yeah, it looks like the OptionParser does much less than in Python :/
Seems weird to have to write a usage message by hand...

pchaigno:
Looks good to me :+1:

script/add-grammar
Alhadis: It does? :D Dude, you're talking to a guy who writes Unix manual pages *by head*. Who even [inflicted this on himself](https://github.com/Alhadis/.files/commit/496993a1d1f39ffd004aa251631b6f5558ce345a) because nobody gives a crap about poor old `man` anymore.

pchaigno: Yes [it does](https://docs.python.org/3/library/argparse.html#example) :smile:

Alhadis:
Damn, didn't notice there was a conflict. Fixed!

Alhadis:
... conflicts fixed again.

arfon:
This is great. Thanks @Alhadis

Alhadis:
My very first Ruby script, on GitHub. <3 What an honour.

PR #3196, opened 2016-09-09
Fix issue when no candidates for the Classifer strategy

mwpastore:
There is [logic](https://github.com/github/linguist/blob/983ff20d3cee56a8d1625fcd6347372d765b8c57/lib/linguist/classifier.rb#L77) in the Classifier strategy to use all the languages in the training set if there are no language candidates at call time (a nil guard). However, this logic never fires, because `.call` passes [an empty array](https://github.com/github/linguist/blob/983ff20d3cee56a8d1625fcd6347372d765b8c57/lib/linguist/classifier.rb#L19) in this case (which doesn't trip the nil guard). This simple fix restores the intended behavior and allows the Classifier strategy to do its thing when there are no language candidates at call time.

### DIFF
--- a/lib/linguist/classifier.rb
+++ b/lib/linguist/classifier.rb
@@ -73,8 +73,8 @@ module Linguist
     #
     # Returns sorted Array of result pairs. Each pair contains the
     # String language name and a Float score.
-    def self.classify(db, tokens, languages = nil)
-      languages ||= db['languages'].keys
+    def self.classify(db, tokens, languages = [])
+      languages = db['languages'].keys if languages.empty?
       new(db).classify(tokens, languages)
     end
 
--- a/test/test_classifier.rb
+++ b/test/test_classifier.rb
@@ -55,6 +55,6 @@ class TestClassifier < Minitest::Test
   end
 
   def test_classify_empty_languages
-    assert_equal [], Classifier.classify(Samples.cache, fixture("Ruby/foo.rb"), [])
+    refute_equal [], Classifier.classify(Samples.cache, fixture("Ruby/foo.rb"), [])
   end
 end


mwpastore:
I could use some help figuring out how to resolve these test cases.

* Under 2.2, test/fixtures/Data/README is being detected as [Creole](http://www.wikicreole.org), which is being treated as an error, but it sure looks like it could be Creole (or Markdown?) to me. 
* Under 2.3, test/fixtures/Data/app.config is being detected as [Maven POM](https://maven.apache.org/pom.html), which is being treated as an error, but it sure looks like it could be Maven POM (or XML?) to me. 

Is it okay that these languages are being detected for this features? If not, why not? If so, what's the best way to modify the testing suite to allow them, and how can I prevent conflicts between 2.2 and 2.3?

lildude:
Giving this a nudge as it's been sitting here patiently for a while.  @mwpastore do you want to merge in master to see how things go and then we can drive this across the line?

mwpastore:
@lildude I'd love to, but it's not really my call. 😁

lildude:
Oops sorry for the confusion @mwpastore, I meant do you want to update your branch with the comments of linguist's master branch. This will bring your branch up to date with the only differences between it and master being the changes you're implementing here.

mwpastore:
@lildude Oh, I see. Sure, I can rebase.

lildude:
Ooooof. I hate to say it, but I think we've been relying on this "feature" 😉 since it was implemented five years ago in as much as it's hoped that one or more of the previous strategies have added at least one possible language to the list before getting to the classifier, with the classifier doing the final elimination based on the samples we've got.

I'm going to hold off merging this as it needs a lot more thought and I think we need to reassess how the classifier should behave in such situations.

veox:
@lildude Would this also be good time to implement `not-a-language` detection?

I've detailed the need for this feature as good as I can in PR #3560.

In short, the need arises for files with extensions that are used arbitrarily. There could be other use cases,  too. I haven't yet created a separate issue for that.

A [hack proposed in a comment there](https://github.com/github/linguist/pull/3560#issuecomment-293069399) uses `[nil]` to indicate that the heuristic did have a positive hit (return array), and that the syntax is known not to be a language (`nil`). Possibly, that could be wrapped into `Language[nil]`, or something similar.

Anyway, just wanted to mention that whatever a solution to this PR's blocker might be, it would be nice if it didn't forbid `not-a-language` indication.

lildude:
> @lildude Would this also be good time to implement not-a-language detection?

It might be. I've not looked into implementing such functionality and don't have the time right now, but I have seen it's come up a few times.

/cc @github/data-engineering for consideration as something you might have the resources for or thoughts on implementing in future.

veox:
OK. I've moved my thoughts to issue #3589.

lildude:
Closing this as "not a bug" as per @pchaigno's review.

mwpastore:
Does the modeline strategy need a file extension?

Does the shebang strategy need a file extension?

Does the heuristics strategy need a file extension?

> If you're trying to classify a snippet of code without knowing the file extension, then I don't think using Linguist is the right approach.

So Linguist only serves to disambiguate cases where the file extension is present and it could be more than one language? 

Can anyone answer: what is the original purpose of [this](https://github.com/github/linguist/blob/983ff20d3cee56a8d1625fcd6347372d765b8c57/lib/linguist/classifier.rb#L77) nil guard?

pchaigno:
> Does the modeline strategy need a file extension?
Does the shebang strategy need a file extension?
Does the heuristics strategy need a file extension?

No, no, kinda (since it needs a short list of candidate languages).

> So Linguist only serves to disambiguate cases where the filename is present and it could be more than one language?

Yes, that's the targeted use case.

> Can anyone answer: what is the original purpose of [this](https://github.com/github/linguist/blob/983ff20d3cee56a8d1625fcd6347372d765b8c57/lib/linguist/classifier.rb#L77) nil guard?

It dates from the beginning of the Linguist project (2012). I don't know if it had a purpose then, but it doesn't have one know (as far as I can see). We could probably remove it.

PR #3197, opened 2016-09-09
Revise patterns for Vim modeline detection

Alhadis:
I rewrote the Vim expressions from scratch after [recent testing](https://github.com/atom/language-perl/pull/76) of a pull request exposed some errors with their handling of certain cases. Specifically, filetypes wouldn't be matched after a boolean option:

    vim: noexpandtab: ft=javascript:

[Version-specific modelines](http://vimdoc.sourceforge.net/htmldoc/options.html#modeline-version) were also unaccounted for:

    vim600: set foldmethod=marker ft=javascript:   # >= Vim 6.0
    vim<600: set ft=javascript:                    # <  Vim 6.0

I found backslashes weren't considered either, though this is probably a non-issue. I fixed it anyway:

    vim: titlestring=foo\ ft=notperl ft=javascript:

Preceding whitespace isn't being checked, either. Vim's [documentation states](http://vimdoc.sourceforge.net/htmldoc/options.html#modeline) that a preceding blank is required before `vim` or `vi`:

> The white space before `{vi:|vim:|ex:}` is required.  This minimises the chance that a normal word like `lex:` is caught.  There is one exception: `vi:` and `vim:` can also be at the start of the line (for compatibility with version 3.0). Using `ex:` at the start of the line will be ignored (this could be short for `example:`).

Which means something like this isn't a valid modeline, yet Linguist detects it as one:

    novim: ft=perl

The extent of my research/testing can be [seen here](https://github.com/atom/language-perl/pull/76/files#diff-8825cccf1f3f704520d28628a83f2ec8R688).

@chrisarcand You were the last one to modify the previous expressions, and you're probably more familiar with Vim than I am. It'd be nice if you could review this and tell me if there's anything I've missed. :)


pchaigno:
@Alhadis Could you use the free-spacing mode to add comments to your regular expression? Otherwise I'm afraid that it's going to be unmaintainable by anyone else than you :/

Alhadis:
[Will these suffice](https://github.com/Alhadis/Snippets/commit/4e370fb00474c9e63388ea5fdfa56b5e3ff94ace), or would you prefer comments too?

Alhadis:
Would [this](https://github.com/Alhadis/Snippets/blob/master/regex/modeline-vim.regexp) be enough on its own, or do you still prefer I include comments?

pchaigno:
Yep, that would be better :smiley: Maybe with a few comments for the not-so-obvious parts (version-specifics and whitespace)?

Alhadis:
... heh, trying to explain my expressions is harder than actually writing them. Give me a moment...

Alhadis:
Okay, before I permit this rambling garbage into the codebase, could you tell me if this is adequate, s'il vous plaît? =)

```rb
VIM_MODELINE   = /
  
  # Start modeline. Could be `vim:`, `vi:` or `ex:`
  (?:
    (?:\s|^)
    vi
    (?:m[<=>]?\d+|m)? # Version-specific modeline
    |
    (?!^)\s
    ex
  )
  
  # If the option-list begins with `set ` or `se `, it indicates an alternative
  # modeline syntax partially compatible with older versions of Vi. Here, the colon
  # serves as a terminator for an option sequence, delimited by whitespace.
  (?=
    # So we have to ensure the modeline ends with a colon
    : (?=\s* set? \s [^\n:]+ :) |
    
    # Otherwise, it isn't valid syntax and should be ignored
    : (?!\s* set? \s)
  )

  # Possible (unrelated) `option=value` pairs to skip past
  (?:
    # Option separator. Vim uses whitespace or colons to separate options (except if
    # the alternate "vim: set " form is used, where only whitespace is used)
    (?:
      \s
      |
      \s* : \s* # Note that whitespace around colons is accepted too:
    )           # vim: noai :  ft=ruby:noexpandtab
    
    # Option's name. All recognised Vim options have an alphanumeric form.
    \w*
    
    # Possible value. Not every option takes an argument.
    (?:
      # Whitespace between name/value is allowed: `vim: ft   =ruby`
      \s*=
      
      # Option's value, where the value itself might be blank (`vim: ft= ` says "use no filetype")
      (?:
        [^\n\\\s] # Beware of escaped characters: titlestring=\ ft=ruby
        |         # will be read by Vim as { titlestring: " ft=ruby" }.
        \\.
      )*
    )?
  )*

  # The actual filetype declaration
  [\s:] (?:filetype|ft|syntax) \s*=
  
  # Language's name
  (\w+)
  
  # Ensure it's followed by a legal separator
  (?=\s|:|$)
/xi
```

`L337 FRENCH`

pchaigno:
*Parfait !* :+1:

Alhadis:
Je l'ai fait, mate. I force-pushed an amended commit to preserve the original commit message.

chrisarcand:
@Alhadis This is excellent. Nice work! Your background work is fantastic, I learned some things about Vim modelines I had no idea existed ;)

I guess maybe I'd add 1-2 more tests for other behavior you've added (Versioning, maybe?) but other than that the pattern looks great to me. 👍

Alhadis:
Thank you! :)

> I guess maybe I'd add 1-2 more tests for other behavior you've added (Versioning, maybe?)

Not a bad idea, actually. Added some tests for version-specific modelines, and also [covered](https://github.com/github/linguist/pull/3197/commits/e73a4ecd0ece2a6599625e28aace649f5fa0eba3) an edge-case that's probably never going to be encountered.

Alhadis:
@pchaigno Is it worth adding fixtures for modelines that *aren't* valid? When I was stress-testing my expressions, I wrote several that were designed *not* to match. I caught a few bugs that way I hadn't anticipated.

None of these are valid, for example:
```
ex: se filetype=perl:
_vi: se filetype=perl:
 vi: se filetype=perl
# vim set ft=perlo
# vim: soft=perl
# vim: hairy-syntax=perl:
# vim set ft=perl:
# vim: setft=perl:
# vim: se ft=perl backupdir=tmp
# vim: set ft=perl set cmdheight=1
# vim:noexpandtab sts:4 ft:perl ts:4
# vim:noexpandtab titlestring=hi\\|there\\ ft=perl ts=4
# vim:noexpandtab titlestring=hi\\|there\\\\\\ ft=perl ts=4

/* --*perl-*- */
/* -*-- perl -*-
/* -*- -- perl -*-
/* -*- perl -;- -*-
// -*- iPERL -*-
// -*- perl-stuff -*-
/* -*- model:perl -*-
/* -*- indent-mode:perl -*-
// -*- font:mode;Perl -*-
// -*- mode: -*- Perl
// -*- mode: grok-with-perl -*-
// -*-font:mode;mode:perl--*-
```

pchaigno:
> Is it worth adding fixtures for modelines that aren't valid?

I don't think so. If someone write an invalid modeline, they can't expect it to work.

Alhadis:
Fair enough... I thought I'd ask, because Emacs modelines have a very unassuming syntax. It's possible for an author to unknowingly include one as a subheading or callout, for example:

```
-*- WARNING: Restart with "mode: perl" enabled in XYZ's config -*-
```

I've already tested it to hell and back, so it's no big deal, but I thought I'd ask nonetheless. :)

arfon:
> It's possible for an author to unknowingly include one as a subheading or callout, for example:

...and this has for sure happened at some point in the past (can't quite remember when).

👏 @Alhadis this is great. Apologies for forgetting about this PR when cutting two new Linguist releases in as many days... This will be in the next release.

arfon:
Also, [this line](https://github.com/Alhadis/linguist/blob/abf7bee4645b392a35e49c1ab4505c30fd2c4333/lib/linguist/strategy/modeline.rb#L69) helps get rid of most accidental modeline detections.

Alhadis:
Thank you! :D No it's okay, I understand that you have duties aplenty to attend to, so I'm patient. Though I submitted a rather trivial PR over a month ago that's had zero feedback, and it should only warrant a short review. I'm legitimately wondering if something went wrong in the submission process and nobody else but me is able to see it. :s

I'm going to bullet-proof the Emacs modeline next. :) I wrote a pattern a while ago (before starting the Emacs Lisp grammar), but I very recently learned I was incorrect of my understanding of the syntax. Gonna make it airtight too (if only for the sake of writing more regex, which I have an unhealthy obsession with...)

arfon:
🎉 from now on I'm going to ask you all my gnarly regex questions 😆

chrisarcand:
Thanks @alhadis!

Alhadis:
> from now on I'm going to ask you all my gnarly regex questions

Heh, I'm used to it. Everyone always comes to me for help with regular expressions, haha.

PR #3198, opened 2016-09-09
Add Filebench WML

larsbrinkhoff:
Every now and then a [new repo pops up that's misclassified as Forth](https://github.com/search?q=language%3Aforth+fxmark).  After some research, I found out that these `.f` files are written in the [Filebench Workload Model Language](https://github.com/filebench/filebench/wiki/Workload-model-language).

There seem to be at least [1700 files](https://github.com/search?q=extension%3Af+flowop+OR+fileset).  (Recent changes in GitHub code search makes it harder to estimate number of repositories.)

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -144,10 +144,22 @@ module Linguist
       end
     end
 
-    disambiguate ".for", ".f" do |data|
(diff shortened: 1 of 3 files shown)


arfon:
:+1:

Issue #2956, opened 2016-04-18
Jade has renamed itself to Pug


TimothyGu:
... and with the name change file extension detection should now be changed from `.jade` to `.pug`.

See pugjs/pug#2184.

abdulhannanali:
`.jade` should also be highlighted meanwhile, because there's plenty of templates on github that are in .jade. so .pug should be added in the file detection. Suggestion

arfon:
> .pug should be added in the file detection. 

We'd welcome a pull request with this change.

Alhadis:
@arfon This issue can be closed now. =)

abdulhannanali:
:smile:

OutsourcedGuru:
I'm finding that after converting an Express-generated Node.js app from jade to pug, suddenly github is incorrectly shifting the project from being a JavaScript-type to an HTML.

As a workaround, I added a **.gitattributes** file to my project and it seems to correctly shift things back to JavaScript.

```
node_modules/* liguist-vendored
README.md linguist-documentation
views/*.pug linguist-documentation
LICENSE linguist-documtation
package.json linguist-documentation
```

PR #3205, opened 2016-09-13
Language ids

arfon:
Currently it's not possible to change the name of a language on GitHub 😞 

This PR adds:
- A `language_id` field to every language known to Linguist (all 422 of them)
- A script (`script/set-language-ids`) to add (and update) `language_id` fields
- Some tests to raise hell if anyone ever tries to change these.

**Rationale**

Currently we can't update a language name as the language name (e.g. `Ruby`) is actually _in_ the GitHub search index. This PR adds a `language_id` field to all of the languages defined in `languages.yml` so that this field can be used in the search indexes instead. We'll then add some view-layer transformations to turn the `language_id` back into a language (with human-readable name) on GitHub.com.

Please note, once this is merged (and shipped on GitHub), modifying these `language_id` fields would be very bad news unless we were planning on re-indexing all of the code on GitHub. As such, I've added some tests for the expected `language_id` value for a few known languages. There's also a `script/set-language-ids --update` which is what contributors adding a new language should use to set the `language_id` value for their new language.

I'd like to get this merged :soon: as we're looking at doing a major re-index of GitHub shortly. That said, a review from any/all of @pchaigno @larsbrinkhoff @Alhadis would be very welcome.

/ cc @TwP

### DIFF
--- a/script/set-language-ids
+++ b/script/set-language-ids
@@ -1,0 +1,82 @@
+#!/usr/bin/env ruby
+require 'yaml'
+require 'pry'
+
+header = <<-EOF
+# Defines all Languages known to GitHub.
+#
+# type              - Either data, programming, markup, prose, or nil
+# aliases           - An Array of additional aliases (implicitly
+#                     includes name.downcase)
+# ace_mode          - A String name of the Ace Mode used for highlighting whenever
+#                     a file is edited. This must match one of the filenames in http://git.io/3XO_Cg.
+#                     Use "text" if a mode does not exist.
+# wrap              - Boolean wrap to enable line wrapping (default: false)
+# extensions        - An Array of associated extensions (the first one is
+#                     considered the primary extension, the others should be
+#                     listed alphabetically)
+# interpreters      - An Array of associated interpreters
+# searchable        - Boolean flag to enable searching (defaults to true)
+# search_term       - Deprecated: Some languages may be indexed under a
+#                     different alias. Avoid defining new exceptions.
+# language_id       - Integer used as a language-name-independent indexed field so that we can rename
+#                     languages in Linguist without reindexing all the code on GitHub. Must not be 
+#                     changed for existing languages without the explicit permission of GitHub staff.
+# color             - CSS hex color to represent the language.
+# tm_scope          - The TextMate scope that represents this programming
+#                     language. This should match one of the scopes listed in
+#                     the grammars.yml file. Use "none" if there is no grammar
+#                     for this language.
+# group             - Name of the parent language. Languages in a group are counted
+#                     in the statistics as the parent language.
+#
+# Any additions or modifications (even trivial) should have corresponding
+# test changes in `test/test_blob.rb`.
+#
+# Please keep this list alphabetized. Capitalization comes before lowercase.
+
+EOF
+
+generated = true if ARGV[0] == "--force"
+update = true if ARGV[0] == "--update"
+
+if generated
+  puts "You're regenerating all of the language_id attributes for all Linguist "
+  puts "languages defined in languages.yml. This is almost certainly NOT what"
+  puts "you meant to do!"
+
+  language_index = 0
+
+  languages = YAML.load(File.read("lib/linguist/languages.yml"))
+  languages.each do |name, vals|
+    vals.merge!('language_id' => language_index)
+    language_index += 1
+  end
+
+  File.write("lib/linguist/languages.yml", header + YAML.dump(languages))
+elsif update
+  puts "Adding new language_id attributes to languages.yml that don't have one set"
+  languages = YAML.load(File.read("lib/linguist/languages.yml"))
+
+  # First grab the maximum language_id
+  language_ids = []
+  languages.each { |name, vals| language_ids << vals['language_id'] if vals.has_key?('language_id')}
+  max_language_id = language_ids.max
+  puts "Current maximum language_id is #{max_language_id}"
+
+  missing_count = 0
+  language_index = max_language_id
+
+  languages.each do |name, vals|
+    unless vals.has_key?('language_id')
+      language_index += 1
+      missing_count += 1
+      vals.merge!('language_id' => language_index)
+    end
+  end
+
+  File.write("lib/linguist/languages.yml", header + YAML.dump(languages))
+  puts "Updated language_id attributes for #{missing_count} languages"
+else
+  puts "Whatever you want me to do, I can't figure it out. Giving up..."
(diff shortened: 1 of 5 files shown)


arfon:
@TwP - we probably need a way to look up a language by it's `language_id` right e.g. in a similar fashion to how we do by alias [here](https://github.com/github/linguist/blob/1fc4c9fdc65994f66060fb7aff9bcfd82f0de0f3/lib/linguist/language.rb#L133-L136)

Alhadis:
Now we're talking!

Personally, I'd shorten the field's name to just `id`. But otherwise, definitely a step in the right direction. 👍

script/set-language-ids
```diff
@@ -0,0 +1,82 @@
...
+      missing_count += 1
+      vals.merge!('language_id' => language_index)
+    end
+  end
+
+  File.write("lib/linguist/languages.yml", header + YAML.dump(languages))
+  puts "Updated language_id attributes for #{missing_count} languages"
+else
+  puts "Whatever you want me to do, I can't figure it out. Giving up..."
+end
```
larsbrinkhoff: Perhaps check that the input YAML doesn't have any duplicates?

Alhadis:
No big deal, but I noticed `languages.yml`'s output has each extension in quote-marks.

Short of `s/^( +- )"(\.[-\w.+]+)"/$1$2/`, there any clean way to get rid of that?

pchaigno:
What if we remove the language with the highest id and add a new language some time later (thus reusing the id). Would that be an issue? The only way I see to prevent that from happening is to store the language id somewhere (which would also remove the need for a lookup).

Alhadis:
Honestly, my approach would be to infer an internal ID from the language's human-readable name, and permit a different name to be displayed if an optional parameter is added:

```yaml
VimL:
  type: programming
  color: "#199f4b"
  search_term: vim
  full_name: "Vimscript"
```

So if `full_name` is absent, the site uses whatever the language's usual key is. :) Seems a little less obvious, but a more stable and readable approach.

script/set-language-ids
```diff
@@ -0,0 +1,82 @@
...
+
+  languages.each do |name, vals|
+    unless vals.has_key?('language_id')
+      language_index += 1
+      missing_count += 1
+      vals.merge!('language_id' => language_index)
+    end
+  end
+
+  File.write("lib/linguist/languages.yml", header + YAML.dump(languages))
```
pchaigno: Any way we could keep the additional newline between each entry in `languages.yml`? IMHO, it makes the file more readable.

TwP:
> we probably need a way to look up a language by it's language_id right e.g. in a similar fashion to how we do by alias here?

@arfon that is correct. A `find_by_id` method would be needed to map language metrics keyed by language ID back to the human readable language name.

Alhadis:
@TwP Would [my suggestion](https://github.com/github/linguist/pull/3205#issuecomment-246631325) be feasible? (I'm tipping there's a pretty good reason a solution that easy hasn't been implemented yet...)

TwP:
> my approach would be to infer an internal ID from the language's human-readable name, and permit a different name to be displayed if an optional parameter is added

@Alhadis one concern with this approach is that it obscures the intent of an immutable language ID. 

For example, the change of language name from [Jade to Pug](https://github.com/github/linguist/issues/2956) would require the key in the `languages.yml` file to remain `Jade` forever. But the `full_name` would change to `Pug`. The temptation would be to change the key from `Jade` to `Pug` as well. Everyone who touches the `languages.yml` file would need to know about the this implicit language ID generated from the key name. This would lead to very confusing behavior if the key name were changed, and it would be difficult to debug.

In this case, being explicit with the language ID has fewer unintended consequences.

Alhadis:
Ah, I see. Well, what about naming it `internal_name` instead? :) That communicates the intent a lot clearer, although I understand what you're saying. And it'd be easy for a test to raise a warning if a diff is detected to an existing `internal_name` property (assuming that an explicit warning in the `languages.yml` file's header wasn't enough to inform contributors...)

It just feels slightly excessive to introduce an extra layer of complexity when the good majority of languages have names that won't ever need to be changed...

(I'm not protesting, just offering my thoughts... I'm all for anything that'll allow languages to be freed from stone)

arfon:
> What if we remove the language with the highest id and add a new language some time later (thus reusing the id). Would that be an issue? The only way I see to prevent that from happening is to store the language id somewhere (which would also remove the need for a lookup).

Huh, yeah, that's a good point @pchaigno. That would be bad too. There's some urgency to get a partial version of this shipped (with at least a `language_id` field present) as we're about to kick off a full re-index of GitHub (this takes ~30 days). I think we'll need to come up with some approaches to make this more robust in the future (i.e. not allowing for ids to be reused).

script/set-language-ids
arfon: Not sure. I'm just doing `YAML.dump` currently.

test/test_language.rb
```diff
@@ -386,6 +402,22 @@ def test_all_languages_have_type
...
+    assert missing.empty?, message
+  end
+
+  def test_all_language_id_are_unique
+    duplicates = Language.all.group_by{ |language| language.language_id }.select { |k, v| v.size > 1 }.map(&:first)
+
+    message = "The following language_id are used several times in languages.yml. Please use script/set-language-ids --update as per the contribution guidelines.\n"
+    duplicates.each { |language_id| message << "#{language_id}\n" }
+    assert duplicates.empty?, message
+  end
```
TwP: :clap: glad to see this duplicates test case

arfon:
> Ah, I see. Well, what about naming it `internal_name` instead? :) That communicates the intent a lot clearer, although I understand what you're saying.

@Alhadis I think renaming the key in the YAML may be a good idea, especially if it helps communicate intent. Also, this PR has made the YAML a whole lot uglier (`"` everywhere and lack of line breaks between language blocks). All that said, I'm going to merge this _as is_ at this point so we can cut a new gem for the folks in the search team - Language renaming is on the horizon folks! 🎉

TwP:
@arfon this looks good. Regarding language ID reuse, keep a `MAX_LANGUAGE_ID` constant somewhere in the codebase would help ensure that language IDs are not reused in the scenario laid out by @pchaigno. The requirement being that all new language IDs must be larger than `MAX_LANGUAGE_ID`, and then increment `MAX_LANGUAGE_ID` after the new language is added.

arfon:

👍 Tracking this in https://github.com/github/linguist/issues/3211

PR #3207, opened 2016-09-13
Test the uniqueness of language ids

pchaigno:
A new test for #3205 to check that language ids are unique. With this test I'm hoping to catch some of the incorrect additions of `language_id`s without using `script/set-language-ids`.


arfon:
😻 thanks!

PR #3079, opened 2016-06-26
YAML grammar files

pchaigno:
This pull request partially fixes the issue at #3070.

The scopes for `vendor/grammars/NSIS` and `vendor/grammars/awk-sublime` were not recognized because the Sublime Text XML files were recently removed. The authors only kept the YAML files from which the XML files are generated.

Thus, this pull request adds support for YAML syntax definitions in `script/convert-grammars`.
A few new scopes were detected via `.YAML-tmLanguage` files. I added them in `grammars.yml`.

:warning: Of course, for this to work, GitHub backend for syntax highlighting must also recognize YAML grammar files.

PR #3208, opened 2016-09-13
Lock grammar versions

### LINKED ISSUE #3070: Issue with some grammars
There are a few grammars that I can't update because we fail to detect any scopes:

```
vendor/grammars/NSIS
vendor/grammars/sourcepawn
vendor/grammars/awk-sublime
```

To be honest I've not had a chance to figure out what's going on here but we get errors like this when running the tests (with the updated grammars):

```
3) Failure:
TestGrammars#test_local_scopes_are_in_sync [/Users/arfon/github/linguist/test/test_grammars.rb:70]:
The scopes listed for vendor/grammars/awk-sublime in grammars.yml don't match the scopes found in that repository.
Expected: ["source.awk"]
Actual: nil
```

This isn't a huge issue right now, except that we can't update grammars that throw errors like this. It's also a pain when trying to run `git submodule update --remote` as part of the release process for Linguist.

If anyone fancies trying to debug this then please be my guest. Otherwise I'll try and get to this at some point.

pchaigno:
Fixes #3070 by locking the version of the three grammars in forks:
- github-linguist/awk-sublime
- github-linguist/NSIS
- github-linguist/sublime-sourcepawn

For all three grammars, the issue is that the authors removed one of the files we were using for parsing (usually `.tmLanguages` files).

#3079 should have fixed the issue with two of the three grammars. It allows `convert-grammars` to parse YAML files. It unfortunately does not work (see [`.nsi` files](https://github.com/search?p=5&q=extension%3Ansi+if&ref=searchresults&type=Code&utf8=%E2%9C%93) and [`.awk` files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aawk+if&type=Code&ref=searchresults)). I'll open another issue to explain why I think it fails and possible solutions.

On the shorter term, I think the best would be to lock the versions.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -667,7 +667,7 @@
 	url = https://github.com/goblindegook/sublime-text-pig-latin
 [submodule "vendor/grammars/sourcepawn"]
 	path = vendor/grammars/sourcepawn
-	url = https://github.com/austinwagner/sublime-sourcepawn
+	url = https://github.com/github-linguist/sublime-sourcepawn
 [submodule "vendor/grammars/gdscript"]
 	path = vendor/grammars/gdscript
 	url = https://github.com/beefsack/GDScript-sublime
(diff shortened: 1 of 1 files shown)


arfon:
:+1: awesome. Thanks @pchaigno

PR #3219, opened 2016-09-15
Add .gnus, .viper and Project.ede as Emacs Lisp extensions

Alhadis:
While digging through Emacs's source, I noticed Emacs Lisp mode is activated by [three filenames](https://github.com/emacs-mirror/emacs/blob/ec12c6705f91106c4f4c53e76447dafa21e12f9f/lisp/files.el#L2529-L2532) I'd not seen before. Searching for them on GitHub shows a non-trivial number of each:

* [`.gnus`](https://github.com/search?q=filename%3A.gnus+NOT+nothack&type=Code) (23,212 results)
* [`.viper`](https://github.com/search?q=filename%3A.viper+NOT+nothack&type=Code) (92,058 results)
* [`Project.ede`](https://github.com/search?utf8=%E2%9C%93&q=filename%3AProject.ede+NOT+nothack&type=Code&ref=searchresults) (15,983 results)

@larsbrinkhoff Regarding the `.eld` extension, I don't think [there's enough](https://github.com/search?q=extension%3Aeld+NOT+nothack&type=Code) usage on GitHub to warrant listing it. The extension also seems to be an auto-generated filetype for storing program data (`.eld`: *"Emacs Lisp Data"*...?)

## Questions
1. [`abbrev_defs`](https://github.com/search?q=filename%3A.abbrev_defs+NOT+nothack&type=Code) is another auto-generated file used by Emacs to store user abbreviations (which is like an autocomplete feature, I'm guessing). ~1,342 results probably aren't enough to add the filename to Linguist, I'm guessing, but these files are most definitely auto-generated. ~~Should it be added as such?~~<br>**UPDATE:** Added as both `.abbrev_defs` and `abbrev_defs`.

2. Emacs has a package manager called **ELPA** ([Emacs Lisp Package Archive](https://www.emacswiki.org/emacs/ELPA_)), which installs Lisp packages to one's user directory: `~/.emacs.d/elpa/`. Users who host their dotfiles on GitHub [mightn't always](https://github.com/haipengliu/my.emacs.d/tree/796abad212e127f99a3f6fddc1e1873681ea859c/elpa) remember to add `elpa/` to `.gitignore`, so this should be marked as a vendored path.<br><br>I don't know how many users on GitHub *have* added these packages to their repository, however, because I don't know how to search files by their containing directory's name, as opposed to their filenames only. Somebody help?

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1039,7 +1039,10 @@ Emacs Lisp:
   filenames:
   - ".emacs"
   - ".emacs.desktop"
+  - ".gnus"
   - ".spacemacs"
+  - ".viper"
+  - "Project.ede"
   extensions:
   - ".el"
   - ".emacs"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1037,12 +1037,15 @@ Emacs Lisp:
   - elisp
   - emacs
   filenames:
+  - ".abbrev_defs"
   - ".emacs"
   - ".emacs.desktop"
   - ".gnus"
   - ".spacemacs"
   - ".viper"
   - "Project.ede"
+  - "_emacs"
+  - "abbrev_defs"
   extensions:
   - ".el"
   - ".emacs"
--- a/samples/Emacs Lisp/filenames/.abbrev_defs
+++ b/samples/Emacs Lisp/filenames/.abbrev_defs
@@ -1,0 +1,6 @@
+(define-abbrev-table 'c-mode-abbrev-table '(
+	))
+(define-abbrev-table 'fundamental-mode-abbrev-table '(
+	("TM" "™" nil 0)
+	("(R)" "®" nil 0)
+	("C=" "€" nil 0)))
--- a/samples/Emacs Lisp/filenames/.gnus
+++ b/samples/Emacs Lisp/filenames/.gnus
@@ -1,0 +1,20 @@
+(setq user-full-name "Alhadis")
+(setq user-mail-address "fake.account@gmail.com")
+
+(auto-image-file-mode)
+(setq mm-inline-large-images t)
+(add-to-list 'mm-attachment-override-types "image/*")
+
+(setq gnus-select-method 
+	  '(nnimap "gmail"
+		(nnimap-address "imap.gmail.com")
+		(nnimap-server-port 777)
+		(nnimap-stream ssl)))
+
+(setq message-send-mail-function 'smtpmail-send-it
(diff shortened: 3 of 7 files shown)


larsbrinkhoff:
Thanks for checking out `.eld`.  Yes, you guessed right; they're syntactically Emacs Lisp files, but sematically only data.  (Well, Lisp data is code is data, etc.)

Alhadis:
HAHA, yeah, let's not get sidetracked about how functions and data are one-and-the-same in Lisp. :D

So, should it be added? If so, what about `abbrev_defs` files? Both should be highlighted as Emacs Lisp, but neither really commands widespread usage on GitHub. I'm also not 100% sure `.eld` files *are* universally data-only, and Googling the extension didn't enlighten me either.

larsbrinkhoff:
About `.eld`: There seems to be about [260](https://github.com/search?q=extension%3Aeld+emacs+OR+gnus+OR+slime) Emacs Lisp files  (out of a total of 750 files) search results, and there seems to be only one file per repository, so technically it would probably meet the requirement "hundreds of repositories".  On the other hand, there are ~500 mystery `.eld` files which would be misclassified.

It'd say it's not important enough to add unless someone complains.

larsbrinkhoff:
As for `abbrev_defs` (aka `.abbrev_defs` aka `abbrev-defs` etc).  There are more files than `.eld`, and seemingly very little risk for misclassification.  I think it would be ok to add.

larsbrinkhoff:
What about the 200k `_emacs` files?

Alhadis:
Shit, I missed that. \*sigh* This is like herding cats.

larsbrinkhoff:
Crazy cats that like to name their files whatever they damn well like.

Alhadis:
Okay, added. The `_emacs` file is my own Emacs [`init.el`](https://github.com/Alhadis/.files/blob/master/.emacs.d/init.el) file.

larsbrinkhoff:
As an avid Emacs user, I'd say that's probably enough for now.

arfon:
@Alhadis - can you confirm these sample files come from an acceptable source (i.e. with a permissive license)

Alhadis:
Yep, I wrote them from scratch after failing to find any that *weren't* released under the GPL (or an obviously permissive license). The `_emacs` file is just my own [`init.el`](https://github.com/Alhadis/.files/blob/master/.emacs.d/init.el) file.

arfon:
> Yep, I wrote them from scratch after failing to find any that weren't released under the GPL (or an obviously permissive license). The _emacs file is just my own init.el file.

👍 thanks.

PR #3220, opened 2016-09-15
Update Clean grammar

camilstaps:


Alhadis:
This isn't necessary. ;-) Linguist grammars are updated automatically with every release, and changes won't show on the site until then. So even if we accepted this now, there'd still be no visible effect. =)

camilstaps:
Ah, I see, thanks :)

PR #3221, opened 2016-09-15
Allow golang as an alias for Go code fences

jredville:
Goal is for

``` golang
func AFunction() error {
  return nil
}
```

to work like

``` go
func AFunction() error {
  return nil
}
```

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1343,6 +1343,8 @@ Gnuplot:
 Go:
   type: programming
   color: "#375eab"
+  aliases:
+  - golang
   extensions:
   - ".go"
   ace_mode: golang


Alhadis:
Are you trying to display embedded Go code with `\`\`\`golang`?

If so, you should open an issue at the [language-package GitHub uses](https://github.com/atom/language-gfm) to highlight Markdown. :) It's responsible for highlighting embedded code blocks.

jredville:
Will do, I was following the link from https://help.github.com/articles/creating-and-highlighting-code-blocks/#syntax-highlighting which (at least to me) made it sound like this is where this patch would be needed. Are both needed for consistency?

Alhadis:
No no, Linguist only selects the grammar to be used for syntax highlighting. :) In this case, it's the Markdown package that's fully responsible for colouring the rendered output.

jredville:
Out of curiosity, what does the `aliases` construct get used for then? Also, it seems that help article shouldn't point to here if that's not the right place to look for the language tag :)

Alhadis:
Yeah, that was [brought up quite](https://github.com/github/linguist/issues/3186#issuecomment-244378301) recently, and we realised there's nowhere better to send users to who encounter syntax highlighting issues. :S I've [submitted a PR](https://github.com/github/linguist/pull/3192) that should hopefully ameliorate this matter.

Alhadis:
Oh, and as for the `aliases` thing, I believe it's for using [`.gitattributes`](https://github.com/github/linguist#using-gitattributes) so users can override languages using more than one known name. I could be wrong, though. :)

jredville:
Awesome, thanks

pchaigno:
Actually, this is controlled by the `aliases` list as @jredville guessed. #3224 should fix this.

jredville:
This should be good to close since #3224 has gone in

pchaigno:
@jredville No it hasn't, I just closed it :smile_cat:

jredville:
Got it 😆 

I'll take a look at the failures

pchaigno:
We're already working on it in #3225. They do not appear to have anything to do with your proposed changes but I'd prefer to fix them before we merge any new pull request.

PR #3242, opened 2016-09-23
Set interpreters for Julia

JoshCheek:
Eg [this](https://github.com/JoshCheek/language-sampler-for-fullpath/blob/b766dcdbd249ec63516f491390a75315e78cba95/julia/fullpath#L1) file is not currently highlighted.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1994,6 +1994,8 @@ Julia:
   type: programming
   extensions:
   - ".jl"
+  interpreters:
+  - julia
   color: "#a270ba"
   ace_mode: julia
   codemirror_mode: julia
--- a/samples/Julia/julia
+++ b/samples/Julia/julia
@@ -1,0 +1,60 @@
+#!/usr/bin/env julia
+
+# From https://github.com/JoshCheek/language-sampler-for-fullpath/blob/b766dcdbd249ec63516f491390a75315e78cba95/julia/fullpath
+help_screen = """
+usage: fullpath *[relative-paths] [-c]
+
+  Prints the fullpath of the paths
+  If no paths are given as args, it will read them from stdin
+
+  If there is only one path, the trailing newline is omitted
+
+  The -c flag will copy the results into your pasteboard
+"""
+
+help  = false
+copy  = false
+dir   = pwd()
+paths = []
+
+for arg = ARGS
+  if arg == "-h" || arg == "--help"
+    help = true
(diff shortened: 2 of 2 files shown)


Alhadis:
You'll need to accept the [GitHub Contributor License Agreement](https://cla.github.com/github/linguist/accept/JoshCheek) before a pull request can be accepted, mate. :-) Think of it as like an EULA.

pchaigno:
Thanks for the pull request @JoshCheek!

Could you add the file you linked to as a sample in `samples/Julia/filenames/`?

JoshCheek:
@pchaigno does it go under filenames? That looks like it matches up with the filenames key. Eg [these](https://github.com/github/linguist/tree/39ea9be5f8aa96e4dab10567feb26b799a35af0a/samples/Ruby/filenames) are the entries for Ruby, and [these](https://github.com/github/linguist/blob/39ea9be5f8aa96e4dab10567feb26b799a35af0a/lib/linguist/languages.yml#L3508-L3526) are Ruby's `filenames` entries. Ruby has entries in its main samples dir for the [`ruby` shebang](https://github.com/github/linguist/blob/39ea9be5f8aa96e4dab10567feb26b799a35af0a/samples/Ruby/ruby#L1), the [`jruby` shebang](https://github.com/github/linguist/blob/39ea9be5f8aa96e4dab10567feb26b799a35af0a/samples/Ruby/shoes-swt#L1), the [`macruby` shebang](https://github.com/github/linguist/blob/39ea9be5f8aa96e4dab10567feb26b799a35af0a/samples/Ruby/macruby#L1), and others.

So, based on their naming, it seems like I should put it in something like `samples/Julia/julia`, or a name that would make more sense to me as someone w/o project familiarity would be `samples/Julia/julia-shebang` or `samples/Julia/julia-interpreter`.

Also, do you want me to put the entirety of the file in there?

pchaigno:
> So, based on their naming, it seems like I should put it in something like `samples/Julia/julia`, or a name that would make more sense to me as someone w/o project familiarity would be `samples/Julia/julia-shebang` or `samples/Julia/julia-interpreter`.

Yes, you're absolutely right, my mistake.

> Also, do you want me to put the entirety of the file in there?

Yes. These files are used to train the Bayesian classifier as well as for certain tests. We try to keep the files as-is as much as possible.

pchaigno:
@arfon I think this is ready to :ship:

PR #3243, opened 2016-09-23
Convert from mode names to mimetypes for better usage.

tberman:
Turns out the mode name alone isn't really enough, convert to using mimetypes.


lib/linguist/language.rb
```diff
@@ -398,13 +398,13 @@ def initialize(attributes = {})
     # Returns a String name or nil
     attr_reader :ace_mode
 
-    # Public: Get Codemirror mode
+    # Public: Get Codemirror mode (as expressed by a mimetype)
```
Alhadis: Should this not be `CodeMirror` instead of `Codemirror`...?

PR #3246, opened 2016-09-24
Improve Mathematica's heuristic rule

pchaigno:
This pull request fixes the incorrect recognition of [this file](https://github.com/glfw/glfw/search?l=mathematica)'s language, reported recently in #2197.

This file is currently recognized as Mathematica because of our Mathematica heuristic (`^\s*\(\*`), which matches [this line](https://github.com/glfw/glfw/blob/3d62c9ed9d1ba4c6c285248d839c17bc5c5028f7/src/cocoa_monitor.m#L334). This is not such an uncommon construct in Objective-C/C/C++ languages. I think an easy fix is to use the closing of Mathematica comments (`*)`) instead of the opening (`(*`).

This pull request implements this, as well as other changes in tests, in three commits:
- First commit adds tests for our `.m` heuristic rules.
- Second commit enables testing the absence of a heuristic result with `assert_heuristics()`, which I use in the third commit to implement the test for the `cocoa_monitor.m` file.
- Third commit contains the new heuristic rule.

/cc @revolter @elmindreda @hrothgar @tbodt

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -244,7 +244,7 @@ module Linguist
         Language["MUF"]
       elsif /^\s*;/.match(data)
         Language["M"]
-      elsif /^\s*\(\*/.match(data)
+      elsif /\*\)$/.match(data)
         Language["Mathematica"]
       elsif /^\s*%/.match(data)
         Language["Matlab"]
--- a/samples/Objective-C/cocoa_monitor.m
+++ b/samples/Objective-C/cocoa_monitor.m
@@ -1,0 +1,413 @@
+//========================================================================
+// GLFW 3.3 OS X - www.glfw.org
+//------------------------------------------------------------------------
+// Copyright (c) 2002-2006 Marcus Geelnard
+// Copyright (c) 2006-2016 Camilla Berglund <elmindreda@glfw.org>
+//
+// This software is provided 'as-is', without any express or implied
+// warranty. In no event will the authors be held liable for any damages
+// arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it
+// freely, subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented; you must not
+//    claim that you wrote the original software. If you use this software
+//    in a product, an acknowledgment in the product documentation would
+//    be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such, and must not
+//    be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source
+//    distribution.
+//
+//========================================================================
+
+#include "internal.h"
+
+#include <stdlib.h>
+#include <limits.h>
+
+#include <IOKit/graphics/IOGraphicsLib.h>
+#include <CoreVideo/CVBase.h>
+#include <CoreVideo/CVDisplayLink.h>
+#include <ApplicationServices/ApplicationServices.h>
+
+
+// Get the name of the specified display
+//
+static char* getDisplayName(CGDirectDisplayID displayID)
+{
+    char* name;
+    CFDictionaryRef info, names;
+    CFStringRef value;
+    CFIndex size;
+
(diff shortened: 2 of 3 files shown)


revolter:
Thanks for the cc, and hope that this PR solves it. I'm using a Sublime Text package that uses linguist, which failed for an Objective-C file. I'll search for it these days and let the owner know after this PR gets merged, so he updates the dependency if needed and see if the problem is gone.

pchaigno:
> which failed for an Objective-C file

If the file is open sourced, I can tell you if this pull request will fix it.

> I'm using a Sublime Text package that uses linguist

Out of curiosity, what's the Sublime Text package's name? (It's always interesting to know who uses Linguist aside from GitHub.)

revolter:
> If the file is open sourced, I can tell you if this pull request will fix it.

As I said in a previous comment on the original issue, it's not :( Is there an easy way to test it myself? Could you point me to some instructions for running tests against it? I would happily test it myself.

> Out of curiosity, what's the Sublime Text package's name?

That was what I wanted to search for, as I often forget what package did what 😆  But I found it quite fast this time: [ApplySyntax][1]. Though I saw not that [it's only a regex][2] found in linguist, so I think that that regex needs updating.

  [1]: https://github.com/facelessuser/ApplySyntax
  [2]: https://github.com/facelessuser/ApplySyntax/blob/b6bd437686b18464056bae596ed8a4f2110d1044/ApplySyntax.sublime-settings#L187-L196

pchaigno:
Yes, they're only using one of LInguist's heuristic rules, so the issue is in their own code. They would probably need to strengthen their own heuristic rules.

revolter:
Actually, [this regex][1] (which is used [here][2] too) was causing problems for my file that contains some code like this:

	[NSString stringWithFormat:
		@"SELECT column FROM (\
			%@ FROM table1 UNION\
			%@ FROM table2 UNION\
		)",

		table1Select, table2Select
	];

So this was caused by my weird indentation. Could the Matlab regex be improved somehow?

  [1]: https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L249
  [2]: https://github.com/facelessuser/ApplySyntax/blob/b6bd437686b18464056bae596ed8a4f2110d1044/ApplySyntax.sublime-settings#L204

Alhadis:
> Could the Matlab regex be improved somehow?

Try `/(?<!\\)(?:\n|\A)[ \t]*%/`.

You're welcome.

revolter:
@Alhadis, should you open a PR with that change?

Alhadis:
In which repo? Linguist, or the Sublime-related one?

In the latter case, it'd require me to install Sublime, which I refuse to do outta principle. =)

pchaigno:
I'm not sure it's worth it for Linguist. It looks like a corner case and the regular expression is sensibly more complicated...

Alhadis:
> the regular expression is sensibly more complicated...

Hrm? All it's doing is checking that the previous line *doesn't* end in a backslash.

revolter:
@pchaigno, I thought that too, but nevertheless, that code is Objective-C, and is matched by Matlab regex too, so it looks like a mistake.
@Alhadis, that `\A` is completely new for me too. Other than that, yes, it's simple to understand.

Alhadis:
> that `\A` is completely new for me too

Be forewarned, the `\A` (start-of-string) anchor isn't recognised by all regex engines. It's understood by Oniguruma/Ruby, but I don't know what engine Sublime uses, so... YMMV.

Alhadis:
I should also point out the `\A` is only necessary because Ruby's regular expressions are multiline by default (...?!), meaning that the usual `^` can't be trusted to mean the start of the entire string.

pchaigno:
> I thought that too, but nevertheless, that code is Objective-C, and is matched by Matlab regex too, so it looks like a mistake.

I agree. However, heuristic rules aren't meant to provide a perfect detection.

In this case, we're making the regex more complicated to handle a single case. What if tomorrow someone finds a file were the previous line starts a multiline comment, or a string? Should we change the regular expression every time?

Unless there are more files which would be fixed by this new regex, I think Linguist overrides would be a better fit in this case.

PR #3247, opened 2016-09-24
Auto increment language_id to avoid duplicates

### LINKED ISSUE #3211: Keep a MAX_LANGUAGE_ID somewhere to prevent language_id dups
As @TwP wrote in https://github.com/github/linguist/pull/3205#issuecomment-246774874, we need to:

Adding this here to remind me to do this.

pchaigno:
Fixes #3211.

Anyone has a better idea on where/how to store the `NEXT_LANGUAGE_ID`? The current way (in `languages.yml`) requires a few modifications to `language.rb` and tests, which I don't really like.

/cc @arfon @TwP

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -561,6 +561,8 @@ module Linguist
   end
 
   languages.each do |name, options|
+    next if name == 'NEXT_LANGUAGE_ID'
+
     options['extensions'] ||= []
     options['interpreters'] ||= []
     options['filenames'] ||= []
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -33,6 +33,7 @@
 # Please keep this list alphabetized. Capitalization comes before lowercase.
 
 ---
+NEXT_LANGUAGE_ID: 422
 1C Enterprise:
   type: programming
   color: "#814CCC"
--- a/script/set-language-ids
+++ b/script/set-language-ids
@@ -11,6 +11,8 @@ header = <<-EOF
 # ace_mode          - A String name of the Ace Mode used for highlighting whenever
 #                     a file is edited. This must match one of the filenames in http://git.io/3XO_Cg.
 #                     Use "text" if a mode does not exist.
+# codemirror_mode   - A String name of the CodeMirror Mode used for highlighting whenever a file is edited.
+#                     This must match a mode from https://git.io/vi9Fx
 # wrap              - Boolean wrap to enable line wrapping (default: false)
 # extensions        - An Array of associated extensions (the first one is
 #                     considered the primary extension, the others should be
@@ -20,7 +22,7 @@ header = <<-EOF
 # search_term       - Deprecated: Some languages may be indexed under a
 #                     different alias. Avoid defining new exceptions.
 # language_id       - Integer used as a language-name-independent indexed field so that we can rename
-#                     languages in Linguist without reindexing all the code on GitHub. Must not be 
+#                     languages in Linguist without reindexing all the code on GitHub. Must not be
 #                     changed for existing languages without the explicit permission of GitHub staff.
 # color             - CSS hex color to represent the language.
 # tm_scope          - The TextMate scope that represents this programming
--- a/script/set-language-ids
+++ b/script/set-language-ids
@@ -49,8 +49,12 @@ if generated
 
   languages = YAML.load(File.read("lib/linguist/languages.yml"))
   languages.each do |name, vals|
-    vals.merge!('language_id' => language_index)
-    language_index += 1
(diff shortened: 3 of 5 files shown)


arfon:
> Anyone has a better idea on where/how to store the `NEXT_LANGUAGE_ID`?

What about putting it in https://github.com/github/linguist/blob/master/lib/linguist/language.rb?

pchaigno:
It might be an appropriate place to have this variable but I'm not sure it's a good place to have a generated value. `set-language-ids` will actually need to update the value in the file every time a language is added. This might be off-putting for users. I'm worried we'd end up with many pull requests where the `NEXT_LANGUAGE_ID` value isn't updated, either because the user didn't commit it or because they made the id change manually.

I don't have a strong opinion one way or the other, both solutions seem non-ideal to me. I'll go with the one that seems best to you.

arfon:
> This might be off-putting for users. I'm worried we'd end up with many pull requests where the `NEXT_LANGUAGE_ID` value isn't updated, either because the user didn't commit it or because they made the id change manually.

So I actually think this is probably a good thing. It's a crappy thing from a contributor perspective but a good thing for maintainers as we want to be really careful about edits to these ids. Seeing the tests fail and then asking folks to update `NEXT_LANGUAGE_ID` seems like an OK workflow?

pchaigno:
I'm not sure I explained myself properly. With both options, currently, if the contributors don't update `NEXT_LANGUAGE_ID` we won't see any test failing. We can/should add a test for that, but then this is something we can do whatever is the option we choose. Basically, I'm afraid with option 2 we see more often the tests failing because the contributors forgot to update the value...

arfon:
> We can/should add a test for that, but then this is something we can do whatever is the option we choose.

Right, sorry, missed this.

> Basically, I'm afraid with option 2 we see more often the tests failing because the contributors forgot to update the value...

Yeah, although I really don't like all the exception logic that we currently have in this PR for skipping the `NEXT_LANGUAGE_ID: 422` key.

@bkeepers - do you have any suggestions here for what might be a better approach?

bkeepers:
> @bkeepers - do you have any suggestions here for what might be a better approach?

I haven't been following along here very closely lately, so sorry if this has been discussed. One idea would be to avoid storing `NEXT_LANGUAGE_ID` at all and just randomly generating the ID when adding a new language. With a big enough integer, collisions should rarely happen. Code should have assertions that the ID is unique so it will fail if there ever is a collision.

Another variation of this would be to generate the ID off a hash of the original language name. 

```ruby

>> "Ruby".to_i(36)
=> 1299022
>> "Python".to_i(36)
=> 1570137287

"Ruby".hash
=> -1156080005327119894
"Python".hash
=> -2846398625042276484
```

I'm not sure the best method for hashing a string into an integer with the least likelihood of collisions, but I'm sure there's something straight forward.

Alhadis:
> I'm not sure the best method for hashing a string into an integer with the least likelihood of collisions, but I'm sure there's something straight forward.

What about using the contents of *the entire file*? A script can still check for collisions, but their likelihood is drastically reduced.

arfon:
Thanks for the suggestions @bkeepers and @Alhadis. Changing any of these language ids is contingent on us _not_ having used these ids yet in our search indexes. @TwP has the grant reindexing started/happened yet?

bkeepers:
@arfon the existing ones could stay the same. But going forward it could generate and save random IDs instead of incrementing IDs.

arfon:
> @arfon the existing ones could stay the same. But going forward it could generate and save random IDs instead of incrementing IDs.

Good point. Hadn't thought of that 😁

pchaigno:
I don't know why, I was convinced we discussed this solution and dismissed it before :/

A hash or random id would certainly help reduce the number of collisions between pull requests.

> _Edit:_ you'd need to call `.abs` on those last two results and maybe modulo them to a smaller integer.

@bkeepers Is there a limit on the maximum value for the id??

TwP:
> @bkeepers Is there a limit on the maximum value for the id??

The maximum value is a signed int32, and we should keep the values positive ... so it is `2^31 - 1`.

arfon:
> What about using the contents of the entire `languages.yml`? A script can still check for collisions, but their likelihood is drastically reduced.

Yeah, that's a neat idea. What do you think @pchaigno?

> @arfon the existing ones could stay the same. But going forward it could generate and save random IDs instead of incrementing IDs.

I checked with @TwP today and the currently `language_id` values are already in our search indexes and so we're only talking about generating new ids here.

@pchaigno would you be able to update this PR with @Alhadis' suggestion (or a similar approach)?

pchaigno:
I implemented @bkeepers' suggestion at #3284. Not sure why, but I as convinced we had already discussed and ruled out this idea in the previous PR :/

PR #3223, opened 2016-09-15
Swap grammar used for Vimscript highlighting

Alhadis:
[You guys know the drill](https://github.com/Alhadis/linguist/commit/dad3331600ccc6f72faf64dcc33053db44d01eba).

This grammar wasn't mine to begin with, BTW... it was transferred to me by its author, who originally ported it from a TextMate bundle (which is also what the current grammar was derived from, I believe).

I also [wrote a script](https://github.com/Alhadis/language-viml/blob/master/tools/update-keywords.js) to keep the grammar updated with keywords pulled from Vim's [own syntax file](https://github.com/vim/vim/blob/master/runtime/syntax/vim.vim) for Vimscript. So it'll stay nice and updated. =)

**Files depicted in preview:**
- [Images 1-2](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=http%3A%2F%2Fpastebin.com%2Fraw%2FW5nEtzWm&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fvim%2Fvim%2Fblob%2Fmaster%2Fruntime%2Fsyntax%2Fvim.vim&code=)
- [Image 3](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=http%3A%2F%2Fpastebin.com%2Fraw%2FW5nEtzWm&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Faltercation%2Fsolarized%2Fblob%2Fmaster%2Fvim-colors-solarized%2Fautoload%2Ftogglebg.vim&code=)

PR #3248, opened 2016-09-25
Remove VimL grammar from whitelist

pchaigno:
Since @Alhadis changed the VimL grammar for a new grammar with an MIT license (#3223), the "whitelisting" is not necessary anymore.


pchaigno:
This change was included in #3251.

PR #3251, opened 2016-09-26
Strengthen license detection

pchaigno:
This pull request implements several changes to strengthen our mechanism to detect/recognize licenses. I broke it down into three commits:

0. Enable the recognition of licenses in READMEs using Licensee with `detect_readme: true`.
0. Use hashes to uniquely identify unrecognized licenses. This way we will be notified for any change in a license text.
0. Remove our own license classification code and rely solely on Licensee's. Our classification code sometimes recognizes the wrong license (e.g., [atom/language-csharp](https://github.com/atom/language-csharp/blob/db4468545a8a1a5e76518e0ae2f6697dc21bf686/LICENSE.md) is detected as GPL), which is something we clearly do not want to happen.

/cc @arfon @aroben @benbalter

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -26,6 +26,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'yajl-ruby'
   s.add_development_dependency 'color-proximity', '~> 0.2.1'
   s.add_development_dependency 'licensed'
-  s.add_development_dependency 'licensee', '>= 8.3.0'
+  s.add_development_dependency 'licensee', '>= 8.6.0'
 
 end
--- a/test/test_grammars.rb
+++ b/test/test_grammars.rb
@@ -5,14 +5,47 @@ class TestGrammars < Minitest::Test
 
   # List of projects that are allowed without licenses
   PROJECT_WHITELIST = [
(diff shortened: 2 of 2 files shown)


aroben:
Using Licensee for this is a great idea.

benbalter:
> Use hashes to uniquely identify unrecognized licenses. 

One point of clarification, that shouldn't affect this PR, the hash is based off the normalized licensed body, which is the license body with whitespace *and the copyright notice* removed (in most cases), meaning it is possible for two projects to return the same hash. Put another way, the hash is deterministic of the normalized license text, not the project itself, so two projects may share a hash (but that doesn't seem to be a problem here).

pchaigno:
> Put another way, the hash is deterministic of the normalized license text, not the project itself, so two projects may share a hash (but that doesn't seem to be a problem here).

Actually, I even think it's a good thing in our case. The `bc12b3b4917eab9aedb87ec1305c2a4376e34fd1` hash (first of the list) identifies nearly all TextMate bundle licenses (more than half unrecognized licenses). Our list would be much longer if hashes identified projects.

It's actually the TextMate license example that first motivated me to ask for such a feature in Licensee (benbalter/licensee#67) :smile_cat:

PR #3253, opened 2016-09-27
Tests to ensure the whitelists are up-to-date

pchaigno:
This pull request adds two tests to make sure the hash and project whitelists remain up-to-date.

I also updated submodules during my tests and was able to remove 1 entry from the hash whitelist and 2 from the project whitelist.

### DIFF
--- a/test/test_grammars.rb
+++ b/test/test_grammars.rb
@@ -134,6 +134,21 @@ class TestGrammars < Minitest::Test
     assert_equal [], unapproved, message
   end
 
+  def test_whitelisted_submodules_dont_have_licenses
+    licensed = submodule_licenses.reject { |k,v| v.nil? }.select { |k,v| PROJECT_WHITELIST.include?(k) }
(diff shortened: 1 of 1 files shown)


vendor/grammars/PHP-Twig.tmbundle
```diff
@@ -1 +1 @@
-Subproject commit 9e802d525ed87a90ec7ff842f44c0c992fd56521
+Subproject commit 77def406d70b90dff33d006478198b729e23d22c
```
arfon: Sorry for being a pain but one of the PHP bundles causes test failures in our HTML pipeline tests on GitHub. I'm afraid I've not had a chance to investigate this yet but for now I've been skipping updating`PHP-Twig.tmbundle` and `php.tmbundle` for now. Any chance we can revert these two?

pchaigno: Sure, done.

arfon: 👍 thanks.

arfon:
This looks good to me.

PR #3258, opened 2016-10-04
Added support for TI Code

scottmangiapane:
"TI Program" is an umbrella name for programs/apps written for the Texas Instruments 83+/84 graphing calculator, which has multiple sublanguages such as TI-BASIC, Axe, and Assembly, all possessing the same file extension. This pull request enables recognition of these previously unknown programs.

[Example Programs](https://github.com/search?utf8=%E2%9C%93&q=extension%3A8xp+NOT+nothack&type=Code&ref=searchresults)


scottmangiapane:
Does anybody know why the Travis CI build is failing?

Alhadis:
Linguist doesn't process binary files, only plain-text formats. Those binaries should be removed, and they're also what're causing the build to fail.

scottmangiapane:
Okay, thanks!

Alhadis:
Good grief, I wrote that comment without even noticing your question OR your response... talk about timing.

Now, concerning the in-the-wild usage... I only found [52 results](https://github.com/search?q=extension%3A8xk+NOT+nothack&type=Code) for the `8xk` extension, and nothing for `8xk.txt`. Are these extensions really common enough to be used?

You should also put `8xp` first to identify it as a primary extension (that's an exception we make to alphabetisation; the most used extension for a language comes first).

scottmangiapane:
Haha, I thought so. Your response appeared almost immediately after I clicked "Comment".

As for the in-the-wild usage, that's a bit messy... These programs are written on the  [TI-84](https://education.ti.com/en/us/products/calculators/graphing-calculators/ti-84-plus/tabs/overview) graphing calculator rather than a regular computer. But the TI-84 stores them as binary files rather than plain text, so when you go to read them on the computer you just get gibberish. You can open them up and read/edit the source on the calculator itself but if you send the programs to a computer or put them on GitHub you can't read the source. To fix this, some people use a tool like [this one](https://www.cemetech.net/sc) to convert the binary file to plain text so that it's readable on GitHub. They then upload the plain text to their repository alongside their regular program. There's no standardized extensions for the text files, so `.8xp.txt` and `.8xk.txt` are some extensions I thought up hoping they would catch on eventually. Most people seem to use `.txt`.

My main concern though is just that `.8xp` and `.8xk` programs are recognized, not so much the `.8xp.txt` and `.8xk.txt`. Would you like me to remove them?

Alhadis:
> Would you like me to remove them?

Ah yes, that'd probably be best. Definitely remove the `.8xk.txt` extension. If its use *does* ever catch on, it can be added in future once there's a sizeable amount of ITW usage.

However, this is the part that concerns me:

> Most people seem to use `.txt`

This would imply that `.txt` should be added as the most common extension for these files. That's going to make a serious mess of classifications unless there's a very sturdy [heuristic](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb) involved. Even then, I'm not sure I'd be comfortable having a format as generic as ***plain text*** risk being classified as a programming language... :S

Is there anything we can use to search for TI Code programs stored in `.txt` files?

larsbrinkhoff:
`.txt` is also common for Forth and Forth-like languages:  
https://github.com/search?q=extension%3Atxt+dup+swap

Alhadis:
Ah, right.

In that case, let's keep it at the `8x`-like extensions, and steer clear of anything too generic. I guess there's no harm in keeping the `.8xp.txt` extension, even if it doesn't really command much usage.

Regarding the name... I know you said "TI PGRM" is an umbrella term, but I couldn't find much use of the term when googling it. I noticed you more recently renamed it to "TI Code"... is there anything more definitive, or is that the closest we have?

[TI-BASIC](https://en.wikipedia.org/wiki/TI-BASIC) seems to be the most commonly-used name for it, though I know you said there were other (unrelated) languages that're using the same file extension... how unrelated are they, exactly?

Alhadis:
What about the name **Texas Instruments Program**...? If there's no name for these languages (which is probably the case), I'd say that's more descriptive than "TI Code" (which sounds pretty normative).

We can always add "TI-BASIC" as an alias to help with people who want to search for programs in this language.

scottmangiapane:
That's what I was worried about too with the `.txt` extensions. It is possible to identify which `.txt` files are for the TI calculator, but since it's such a small percentage of total `.txt` files it's probably not worth the risk. I think you're right and it's best to just ignore the .txt files, and let GitHub think they're just plain text especially since they're just an optional addition and not required to run the program.

I also like your thought process for the name! **Texas Instruments Program** is more descriptive than just **TI Code**, and since there's not an official term it's best to be descriptive. What if we abbreviated it so that it's **TI Program**? My reason for this is because when you're looking on someone's [repository page](https://github.com/scottmangiapane?tab=repositories), **Texas Instruments Program** is too long and get's wrapped to three lines.

Example of the line wrap problem:
demo

Alhadis:
It [wouldn't be the first](https://github.com/ajaypr55?tab=repositories) language to do something like that, don't worry. It *is* long, but it also spares a reader from having to google "TI" to find out what it stands for.

I feel that's a point open to discussion, though. @pchaigno, @arfon... what're your thoughts? :)

scottmangiapane:
I'd rather have nothing than have **Texas Instruments Program**, to be honest. I realize that sounds demanding, its just that it looks too cluttered with three lines of text. After all, more common languages such as HTML and CSS are abbreviated. But sure, let's hear some second opinions!

Alhadis:
Don't get too attached to what you see on GitHub. Sporadic layout changes happen all the time. Sure, it might look like this today:

Figure 1

... until you wake up one day and find it looking like this:

Figure 2

The take-home point is that language names have more permanency on GitHub than its layout does. It's not right to justify a choice of name based on how it might be presented at any one point in time.

I feel I should stress I'm actually not GitHub staff, just a regular contributor to this repository. The actual maintainer is very busy, so contributors go out of their way to help answer questions in advance.

arfon:
> The take-home point is that language names have more permanency on GitHub than its layout does. It's not right to justify a choice of name based on how it might be presented at any one point in time.

Absolutely. I don't think we should make a decision on the Linguist language name based upon GitHub's layout.

@scottmangiapane - I think we should either go with `TIP` or `Texas Instruments Program`.

scottmangiapane:
What if we compromised and did `TI Prgm` / `TI-Prgm`? It's similar to `TIP`, except that it's also an actual acronym used in the TI programming community so people will recognize it. Plus it yields relevant results on Google.

Alhadis:
I can only find [133 results](https://www.google.com.au/search?q=%22TI+Prgm%22&oq=%22TI+Prgm%22&aqs=chrome..69i57.1279j0j7&sourceid=chrome&ie=UTF-8) on Google for `TI Prgm`...

scottmangiapane:
`Texas Instruments Program` is actually worse, there are more results sure but literally all of the results are off topic (usually about something called a "Texas Instruments Program Recorder"). And `TIP` is unrelated as well.

Alhadis:
Erh, googling [Texas Instruments Program](https://www.google.com.au/search?q=%22Texas+Instruments%22+Program&oq=%22Texas+Instruments%22+Program&aqs=chrome..69i57.3152j0j7&sourceid=chrome&ie=UTF-8) seems to yield pretty relevant results. Am I missing something here?

This would all be a lot easier if Texas Instruments bothered to give their calculator-languages a proper name. ;p

Alhadis:
Also, you *are* aware that by "search results", we're mostly talking about GitHub's language search... right?

Look, there's a name even bigger than Texas*...
topkek

pchaigno:
What's the issue with the current *TI Program*? That's how I used to call it and it yields relevant results on Google...

arfon:
> What's the issue with the current TI Program? That's how I used to call it and it yields relevant results on Google...

Potentially none `¯\_(ツ)_/¯`. I was under the impression that this was a somewhat arbitrary name that was decided in this thread but if it's the normal/standard name I don't have any strong objections.

scottmangiapane:
Thank you! I think that Alhadis is opposed to it because it's not as descriptive as `Texas Instruments Program`. But `TI Program` is the most standardized in my opinion, it's more normal in the TI programming community to abbreviate it.

arfon:
> If you approve @arfon, the code is all ready to be merged :)

OK thanks @scottmangiapane. Can you just confirm the origin of the sample files included in this PR? They need to have a permissive license (MIT, BST, Apache 2.0 etc.) to be acceptable to Linguist.

scottmangiapane:
Sure! They're all [my own code](https://github.com/scottmangiapane/alphacs), and they're licensed under the MIT license.

arfon:
⚡️ thanks for this @scottmangiapane. This will be live in the next release of Linguist (likely next week sometime). 

@Alhadis - thanks for helping us explore all possible combinations of names here 😁

scottmangiapane:
Thank you @arfon and @Alhadis! I appreciate it.

scottmangiapane:
Hey @arfon / @Alhadis, how come [my repositories](https://github.com/scottmangiapane?tab=repositories) still aren't being recognized? Did I do something wrong in my pull request or is the new release just taking a while to circulate?

Alhadis:
Ah, that'll probably be because the results are still cached from before the release was cut. =)

Just push some changes to your repos, and it should bust the cache and force a reindexing.

scottmangiapane:
No luck :/ I tried updating a file, removing everything, restoring everything, adding a new file, and deleting the new file.

I'm looking through the contribution guidelines though and I might've found it. The readme says that I have to run the 'set-language-ids script' to update the language ID, but I don't think it said that a couple weeks ago so I had just added one manually instead of running the script. Maybe that broke it?

Alhadis:
I've run Linguist locally on your repo, and it's because the `.8xp` files are binary. Linguist will skip anything that contains stuff like `NUL` bytes and kin.

Try changing the file extension of the text file to `.8xp.txt` and try again?

scottmangiapane:
You're right, that fixed it! Thanks :)

PR #3259, opened 2016-10-04
Guard against unused colour definitions

Alhadis:
This slipped in unnoticed. Since ASN.1 is classified as "data", there's no reason for it to have a colour assigned.

Should we add a test for this to stop this happening again...?

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -19,7 +19,7 @@
 # language_id       - Integer used as a language-name-independent indexed field so that we can rename
 #                     languages in Linguist without reindexing all the code on GitHub. Must not be
 #                     changed for existing languages without the explicit permission of GitHub staff.
-# color             - CSS hex color to represent the language.
+# color             - CSS hex color to represent the language. Only used if type is "programming" or "prose"
 # tm_scope          - The TextMate scope that represents this programming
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
Automated tests are good!

arfon:
> Should we add a test for this to stop this happening again...?

:+1: why not.

Alhadis:
Working on it. :) Will revise this PR.

Alhadis:
Done! Sorry, that took longer than I thought it would... couldn't figure out why the license-checks were failing, and it ended up being an issue with local submodules (as always).

@arfon I updated the ASN.1 submodule because of, well, [this](https://github.com/ajLangley12/language-asn1/pull/1).

Maybe we should guard against adding grammars with `.DS_Store` files too? 😂

arfon:
> Maybe we should guard against adding grammars with .DS_Store files too? 😂

Lol. Bane. Of. My. ~~OS X~~ macOS. Life.

larsbrinkhoff:
I think we could go even further.  If a languages is `group`ed under another language, its colour will never show in the bar.

Alhadis:
Hadn't thought of that, actually. I admit I forgot that "grouped" languages weren't shown...

larsbrinkhoff:
I'm on it.

PR #3262, opened 2016-10-04
Treat vendored code as generated

### LINKED ISSUE #3234: collapse vendor folder in diffs
The `vendor/` folder is used by Go 1.6+ and other tools/languages for third-party code. While it may be good to review all the code being used in a production app, the GitHub UI still doesn't handle large amounts of code that well. 

> Sorry, we could not display the entire diff because it was too big.

(It's possible to pick commits or files, but the initial load is still an issue)

Happy to open a pull request modifying `generated.rb` to hide vendor/** in diffs.
ref: https://github.com/github/linguist/pull/1549

nathany:
resolves #3234

I'm only really interested in collapsing `vendor/` for use with Go, but I've added the other directory names defined in vendor.yml.


DocMerlin:
THANK YOU!

nathany:
thanks to @afron for making this happen

jordoh:
Is there a way to override this in `.gitattributes` like there is for `linguist-vendored`? It now appears that all diffs of `vendor/` paths get suppressed, even if `vendor linguist-vendored=false` is present in `.gitattributes`. This is a major annoyance for rails projects with `vendor/gems` and `vendor/plugins` directories that are not 3rd party code, and where diffs should not be suppressed.

arfon:
> Is there a way to override this in .gitattributes like there is for linguist-vendored?

Unfortunately not 😞 . 

> This is a major annoyance for rails projects with vendor/gems and vendor/plugins directories that are not 3rd party code, and where diffs should not be suppressed.

Apologies for any inconvenience caused here @jordoh. Unfortunately it seems like there's no perfect 'generalized' solution here. 

The proper fix is to support these overrides as you suggest / cc @brianmario

nathany:
I think the generalized solution requires some way to review individual directories of code at a time -- as a way to split up large diffs.

After all, the goal of this pull request and associated issue is to help make the GitHub UI more responsive on large diffs, not to prevent people from reviewing third-party code.

arfon:
@nathany - we're getting a bunch of complaints about this change and I think we're going to have to revert it in the short term until we have support for manually setting this in `.gitattributes` data 😞 

In https://github.com/github/linguist/pull/2843 we attempted to craft a Go-specific matcher for hiding these files but failed to come up with a solution. Do you have any ideas? Hiding `vendor` paths by default is the wrong behaviour for lots of people.

pchaigno:
@arfon I've seen issues too, but mostly because the scope of `vendor.yml` is too broad (`tools/` probably shouldn't be vendored by default). Could removing some entries from `vendor.yml` be the solution (or making them more specific)?

arfon:
> @arfon I've seen issues too, but mostly because the scope of vendor.yml is too broad (tools/ probably shouldn't be vendored by default). Could removing some entries from vendor.yml be the solution (or making them more specific)?

Yeah, I think that's part of the issue. Unfortunately, `vendor` is the path that seems to be causing most of the pain here. Go folks really don't want to see all those changes, Ruby (amongst others) usually do :-\

I think probably the best we can do right now is pursue a Go-specific `generated?` check in #2843 that looks for hostnames in the paths. This looks to be doable although finding an efficient reliable regular expression for this might be a little trick.

_\*\*Calling @Alhadis for his regex super-powers\*\*_ 😁

Alhadis:
Can somebody give me an example of the sort of paths we need to vendor? :)

I admit I haven't been paying attention to this matter, probably because I know nothing about Go (other than they use tabs, which immediately earns my respect. :p)

arfon:
> Can somebody give me an example of the sort of paths we need to vendor? :)

I think this is an example: https://github.com/github/linguist/pull/2843#issuecomment-186001697 `vendor/github.com/*` or any other hostname.

> I admit I haven't been paying attention to this matter, probably because I know nothing about Go (other than they use tabs, which immediately earns my respect. :p)

tabs vs spaces - imgur

Alhadis:
Oh! [Piece of piss](https://en.wiktionary.org/wiki/piece_of_piss), hang on.

Link added for the benefit of those residing outside the Commonwealth.

arfon:
> Link added for the benefit of those residing outside the Commonwealth.

Lol. Many thanks for internationalizing your GitHub comments.

Alhadis:
How strict do you want me to make this?

The "correct" way would be to check for a proper TLD at the end of the hostname, but [there's a lot of 'em](https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains)... which would result in one messy regex. I could use a common subset, I guess...

Alhadis:
Otherwise, it'd just be something like this:

```
vendor/((?!-)[-\w]+(?<!-)\.)+(com|net|org|gov|etc)/*
```

arfon:
> The "correct" way would be to check for a proper TLD at the end of the hostname, but there's a lot of 'em... which would result in one messy regex. I could use a common subset, I guess...

Right. So I *believe* these hostnames are going to mostly be of places where you may vendor Go code from (e.g. GitHub, Bitbucket etc). @nathany - could you confirm?

nathany:
@arfon here are some hostnames from my ~src folder:

* go.googlesource.com
* golang.org
* github.com
* bitbucket.org
* launchpad.net
* gopkg.in

gopkg.in is fairly common, so adding `.in` is a good idea

Alhadis:
`vendor/((?!-)[-0-9A-Za-z]+(?<!-)\.)+(com|edu|gov|in|me|net|org|fm|io)/` should cover it then. =)

nathany:
@jordoh I'm just curious what the story is behind using vendor this way. I was a Rubyist for 8 years, and I never put first-party code in `vendor/`, usually opting for the `lib/` folder.

As per [my previous comment](https://github.com/github/linguist/pull/3262#issuecomment-256232062):

> "the goal of this pull request and associated issue is to help make the GitHub UI more responsive on large diffs, not to prevent people from reviewing third-party code."

So I do think it should be possible to review the code and license in vendor/ -- and expect this change will be reverted.

I've made a sample repository and pull request to demonstrate the pain developers experience with the GitHub UI when reviewing a large diff. https://github.com/nathany/gvt-sample/pull/1

jordoh:
@nathany - Rails has historically provided some automatic integration of code in `vendor/plugins`, `vendor/engines`, and `vendor/gems` - depending on the version of Rails - making vendor sub-folders useful for modularly developed first-party code.

Agreed on the need for improved review of large PR diffs. I've experienced similar difficulties, particularly once a large (or even medium) sized PR starts to get a significant number of comments (try adding 100 line comments to your sample PR to see just how slow the page can get ;) ).

arfon:
> `vendor/((?!-)[-0-9A-Za-z]+(?<!-)\.)+(com|edu|gov|in|me|net|org|fm|io)/` should cover it then. =)

⚡️ 💖 thanks @Alhadis

Issue #3256, opened 2016-09-30
No language statistics shown for repository


apsdehal:
Linguist doesn't show any statistics for my repo https://github.com/apsdehal/SPOJ

Alhadis:
That's odd... did you just upload this recently? If so, the site might have yet to refresh its cache or something. But the files [are definitely being indexed](https://github.com/search?utf8=%E2%9C%93&q=language%3Acpp+repo%3Aapsdehal%2FSPOJ&type=Code&ref=searchresults).

apsdehal:
No, repo is 2 years old. I just updated it and saw that it didn't had language statistics.

Alhadis:
I'm actually at a loss to explain this one, then. Running `linguist` on the repository locally confirms that the files are being processed correctly. :S 

```
94.29%  C++
5.19%   Python
0.16%   C
0.15%   JavaScript
0.11%   Ada
0.10%   PHP

C++:
ABA12D-8697864-src.cpp
ABA12D-8698276-src.cpp
ABSYS-9541469-src.cpp
ABSYS-9541484-src.cpp
ACPC10A-8589556-src.cpp
ACPC11B-8624882-src.cpp
ACPC11B-8624955-src.cpp
ACPC11B-8624957-src.cpp
ACPC11B-8624961-src.cpp
ACPC11B-8624966-src.cpp
ADDREV-8566659-src.cpp
AFS-9553991-src.cpp
ALLIZWEL-9001177-src.cpp
AMR10C-8907340-src.cpp
AP2-8598373-src.cpp
ARCTAN-8643093-src.cpp
ARCTAN-8643161-src.cpp
ARCTAN-8645145-src.cpp
ARITH2-9541691-src.cpp
BEENUMS-8666492-src.cpp
BINSTIRL-9531350-src.cpp
BUSYMAN-8707897-src.cpp
BYTESM2-8903915-src.cpp
CANDY-8994598-src.cpp
CANDY-8994606-src.cpp
CANDY3-8995504-src.cpp
CANDY3-8995525-src.cpp
CANTON-8653630-src.cpp
CCHESS-9528856-src.cpp
CODESPTB-8998414-src.cpp
COINS-9506080-src.cpp
COINS-9506086-src.cpp
COMDIV-8605641-src.cpp
CPRMT-8682733-src.cpp
CPRMT-8682821-src.cpp
CYLINDER-9027700-src.cpp
DANGER-8994106-src.cpp
[...]

PR #3264, opened 2016-10-05
Add REXX interpreters

Alhadis:
There are at least [7 interpreters](http://www.rexxinfo.org/html/rexxinfo2.html#Download-Rexx) for running REXX, 3 of which run on POSIX systems:

* [BRexx](https://github.com/vlachoudis/brexx/blob/master/doc/running.html)
* [Rexx/imc](http://www.cs.ox.ac.uk/people/ian.collier/Rexx/rexximc.html)
* [Regina](http://regina-rexx.sourceforge.net/)

[BRexx](https://github.com/vlachoudis/brexx/blob/master/doc/running.html#L22) and [Rexx/imc](http://web.archive.org/web/20060107071407/http://users.comlab.ox.ac.uk/ian.collier/distribution/rexx-imc-1.76/rexx.ref) both use `rexx` as the interpreter name. [Regina](https://foicica.com/lists/code/201510/2725.html) uses `regina`, but I couldn't find any other possible variations.

I don't know how different Rexx dialects like [Object Rexx](http://www.oorexx.org/) are from plain Rexx, so I've purposefully omitted any interpreters that're dialect-specific.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3447,6 +3447,9 @@ REXX:
   - ".rexx"
   - ".pprx"
   - ".rex"
+  interpreters:
+  - regina
+  - rexx
   tm_scope: source.rexx
   ace_mode: text
   language_id: 311


Alhadis:
@idrougge You're the one who [initially proposed](https://github.com/github/linguist/pull/3110) the language's addition, so I'm paging you for help... I couldn't help noticing a [few `.r` files](https://github.com/search?p=1&q=Brexx&type=Code&utf8=%E2%9C%93) popping up while searching for `brexx`. Do you know if this extension is commonly used for storing Rexx source files?

idrougge:
@Alhadis I have no idea. Best to ask @vlachoudis at https://github.com/vlachoudis/brexx since it's his interpreter.

BTW, I agree that OORexx and NetRexx should be considered separate languages.

vlachoudis:
brexx is not bound to any specific extension. you can use anything. I used the .r almost 20years ago when I wrote the first version for MSDOS and I simply continued to use this on

Alhadis:
Ah right, thanks for the answer! :-) Do you know if many Rexx files use this?

I'm curious if it's worth including `.r` as a common Rexx extension. I saw Rexx source being misclassified as R on GitHub, which screwed up the syntax highlighting.

Since I know neither R nor Rexx, it's hard for me to determine the extension's usage by [searching `.r` files alone](https://github.com/search?q=extension%3Ar+NOT+nothack&type=Code).

vlachoudis:
brexx is not bound to any specific extension. you can use anything. I used the .r almost 20years ago when I wrote the first version for MSDOS and I simply continued to use this on

vlachoudis:
what I've seen in some editors they check the first lines of the file and if they detect a comment starting with /* they assume it is rexx

idrougge:
Indeed a starting comment is required by most REXX interpreters. On IBM systems, it's even mandatory that the first line is /* REXX */ and nothing else.
http://www.ibmsystemsmag.com/mainframe/tipstechniques/applicationdevelopment/A-Beginner-s-Guide-to-the-REXX-Programming-Languag/?page=2

Alhadis:
Perfect, that's exactly what I needed to know! Thanks!

I'll snoop around more before amending to this PR. I need to better gauge the extension's usage, and it's rather difficult to search for (since `/* */` is [dropped from search results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ar+%2F*+*%2F&type=Code&ref=searchresults)). Do you know of any common phrases or terms that would help locate repositories with REXX `.r` files?

(Sorry if I'm being annoying, but you're both extremely helpful, haha)

idrougge:
The three most common REXX keywords are probably DO, SAY and END.

idrougge:
You also have YAXX: https://github.com/Jaxo/yaxx

Alhadis:
> You also have YAXX: https://github.com/Jaxo/yaxx

At a glance, that appears to be an *interpreter builder*, rather than an interpreter. The `interpreters` field is for matching the names of executables found in hashbangs. So something like `#!/usr/bin/env rexx` will mark the file as REXX, regardless of its file extension.

> BTW, is this why my REXX repositories currently aren't tagged with any language?

Sorry! I missed this question.

Honestly, no. Your repositories *should* be showing REXX in their statistics bars. Their files are being recognised as REXX:
* [idrougge/sha1-rexx](https://github.com/idrougge/sha1-rexx/search?l=rexx)
* [idrougge/BatchRemapBrushes](https://github.com/idrougge/BatchRemapBrushes/search?l=rexx)
* [idrougge/ShapesInfo](https://github.com/idrougge/ShapesInfo/search?l=rexx)

There've been a couple other issues reported recently where users aren't seeing any languages in their project's stat-bar (#3256). The files *do* appear to be indexed correctly (as the lists above will show), so I think this is an internal issue on GitHub's end.

arfon:
> There've been a couple other issues reported recently where users aren't seeing any languages in their project's stat-bar (#3256). The files do appear to be indexed correctly (as the lists above will show), so I think this is an internal issue on GitHub's end.

Yeah, this is happening sometimes ([caching expiration being hard to get right](http://martinfowler.com/bliki/TwoHardThings.html)). I _believe_ the (undocumented) fix here is to update the code (push any change) and the language bar will be refreshed < 24 hours later.

Alhadis:
Do you know if the site's reindexing will bust those caches too?

> There are only two hard things in Computer Science: cache invalidation and naming things.

WHY IS THIS SO TRUE

arfon:
> Do you know if the site's reindexing will bust those caches too?

I don't believe it will but it would probably be a good idea to do this.

Issue #2662, opened 2015-10-02
sublime-* files with JSON like content detected as JavaScript


aviaryan:
`.sublime-build` , `.sublime-completions` and `.sublime-settings` are detected as JavaScript

All of those 3 files have JSON content.

https://github.com/aviaryan/AutoHotkey/search?l=javascript

deathbeam:
They does not have valid JSON content. In JSON you cannot use comments.

arfon:
So we can close this now?

aviaryan:
Never thought linguist looks beyond file extensions. If that's the case, then it's all right.

keplersj:
JSON5 supports comments, perhaps that is a better fit for these files?

Alhadis:
@keplersj Does Sublime permit you to run arbitrary functions in those files?

If so, it's JavaScript.

JSON5 is a different beast altogether, though... it supports a lot more than just comments (primitives, unquoted keys, etc)

keplersj:
@Alhadis I don't think Sublime allows abstract functions in these, I believe they are just JSON files with comments.

Alhadis:
I think it's probably safe to leave them classified as JSON. It'd surprise me if Sublime were the only program out there which permits comments in JSON data.

keplersj:
They're currently classified as JavaScript because of these comments though

Alhadis:
Whoops, yeah, sorry... I got them switched in my head.

I've only ever used Sublime a handful of times, so I've no idea how its config files are used. But from a cursory glance, it'd seem this issue should be revisited.

Given Sublime's mass popularity, it wouldn't strike me as unusual if the stats for JavaScript for this year's [GitHub Octoverse](http://www.i-programmer.info/news/136-open-source/10126-github-octoverse-reveals-the-state-of-open-source.html) were unfairly inflated:

<img src="http://www.i-programmer.info/images/stories/News/2016/Sept/B/octovers2.JPG" />

I think this should be reopened, honestly.

keplersj:
I agree, it seems unfair to count these configuration files as JavaScript.

keplersj:
SublimeText docs have more information on the files, and they are not JavaScript.

Format: JSON (with comments)
- [.sublime-build](http://docs.sublimetext.info/en/latest/reference/build_systems/basics.html#file-format-of-sublime-build-files)
- [.sublime-commands](http://docs.sublimetext.info/en/latest/reference/command_palette.html#file-format-of-sublime-commands-files)
[...]

PR #3267, opened 2016-10-06
Mark Sublime Text config files as JSON

keplersj:
I don't think any of these actually count as JavaScript from what I've seen.


keplersj:
Nevermind, I just saw #2662.

PR #3272, opened 2016-10-09
Improve documentation on the language_id

pchaigno:
This pull request changes the documentation on how to add a new language to explicitly say that the `language_id` should be set by `./script/set-language-ids` and not manually.


arfon:
👍 thanks

PR #3280, opened 2016-10-12
Remove unused color attributes

larsbrinkhoff:
This removes the color attribute from all languages with a group attribute.  Also updates the test which checks for unused color attributes.

Sorry @Alhadis, the Apollo Guidance System is one of the causualities.

Bonus perk: Removes the redundant `ColdFusion` group from the `ColdFusion` language.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -712,7 +712,6 @@ CoffeeScript:
   language_id: 63
 ColdFusion:
   type: programming
-  group: ColdFusion
   ace_mode: coldfusion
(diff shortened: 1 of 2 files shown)


Alhadis:
> Sorry @Alhadis, the Apollo Guidance System is one of the causualities.

It's ironic, because that's what started this all. We wanted NASA blue as AGC's colour, but it was taken by Markdown. We realised Markdown didn't need it, so we dropped it to allow AGC to have it. Then AGC's colour is later removed for the same reason.

PR #3284, opened 2016-10-17
Generate language_id

### LINKED ISSUE #3211: Keep a MAX_LANGUAGE_ID somewhere to prevent language_id dups

pchaigno:
This pull request implements [@bkeepers' idea](https://github.com/github/linguist/pull/3247#issuecomment-252083158), generating the `language_id` from the hash of the language's name. I also added a test to check that all language ids are unsigned 32bit integers.
Supersedes #3247. Fixes #3211.

The `language_id` is the SHA256 hash of the language name modulo-ed to an unsigned 32bit integer:

``` ruby
Digest::SHA256.hexdigest(language).to_i(16) % (2**30 - 1)
```

Therefore, as long as the language names are different, the probability of collision is of 0.009561175% if we reach 1,000 languages [1]. This is important as we might not always be able to detect a collision. For instance, if a language entry is removed from Linguist and we later add a new language id that conflicts with the older removed language, the tests won't help detecting it.

The choice to generate the language id from the language name is based on the idea that we probably want two entries with the same name to have the same id (if we remove a language and add it back later). If we do not want that, I could simply change the pull request to generate random ids, which would give us the same collision probability.

I think it's important that we discuss this thoroughly, as we probably want to avoid changing the strategy to generate language ids too often (it's already our second system :stuck_out_tongue_closed_eyes: ).
/cc @bkeepers @arfon @Alhadis @TwP 

[1] - ql_15e041ca4f6190549ca56982ef8053e1_l3 - Details at [preshing.com](http://preshing.com/20110504/hash-collision-probabilities/). I took into consideration the fact that we already have 422 fixed language ids.

### DIFF
--- a/test/test_language.rb
+++ b/test/test_language.rb
@@ -426,6 +426,14 @@ class TestLanguage < Minitest::Test
     assert missing.empty?, message
   end
 
+  def test_all_languages_have_a_valid_id
+    invalid = Language.all.select { |language| language.language_id < 0 || language.language_id >= (2**31 - 1) }
+
+    message = "The following languages do not have a valid language_id. Please use script/set-language-ids --update as per the contribution guidelines.\n"
+    invalid.each { |language| message << "#{language.name}\n" }
+    assert invalid.empty?, message
+  end
+
   def test_all_language_id_are_unique
     duplicates = Language.all.group_by{ |language| language.language_id }.select { |k, v| v.size > 1 }.map(&:first)
 
--- a/script/set-language-ids
+++ b/script/set-language-ids
@@ -36,21 +36,23 @@ header = <<-EOF
 # Please keep this list alphabetized. Capitalization comes before lowercase.
 
 EOF
+require 'digest'
 
 generated = true if ARGV[0] == "--force"
 update = true if ARGV[0] == "--update"
+
+def generate_language_id(language)
+  Digest::SHA256.hexdigest(language).to_i(16) % (2**30 - 1)
+end
 
 if generated
   puts "You're regenerating all of the language_id attributes for all Linguist "
   puts "languages defined in languages.yml. This is almost certainly NOT what"
   puts "you meant to do!"
 
-  language_index = 0
-
   languages = YAML.load(File.read("lib/linguist/languages.yml"))
   languages.each do |name, vals|
-    vals.merge!('language_id' => language_index)
-    language_index += 1
+    vals.merge!('language_id' => generate_language_id(name))
   end
 
   File.write("lib/linguist/languages.yml", header + YAML.dump(languages))
@@ -58,20 +60,12 @@ elsif update
   puts "Adding new language_id attributes to languages.yml that don't have one set"
   languages = YAML.load(File.read("lib/linguist/languages.yml"))
 
-  # First grab the maximum language_id
-  language_ids = []
-  languages.each { |name, vals| language_ids << vals['language_id'] if vals.has_key?('language_id')}
-  max_language_id = language_ids.max
-  puts "Current maximum language_id is #{max_language_id}"
-
   missing_count = 0
-  language_index = max_language_id
 
   languages.each do |name, vals|
     unless vals.has_key?('language_id')
-      language_index += 1
       missing_count += 1
-      vals.merge!('language_id' => language_index)
+      vals.merge!('language_id' => generate_language_id(name))
     end
   end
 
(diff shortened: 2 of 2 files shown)


test/test_language.rb
```diff
@@ -426,6 +426,14 @@ def test_all_languages_have_a_language_id_set
...
   end
 
+  def test_all_languages_have_a_valid_id
+    invalid = Language.all.select { |language| language.language_id < 0 || language.language_id >= (2**31 - 1) }
+
+    message = "The following languages do not have a valid language_id. Please use script/set-language-ids --update as per the contribution guidelines.\n"
+    invalid.each { |language| message << "#{language.name}\n" }
+    assert invalid.empty?, message
+  end
+
```
bkeepers: It'd probably also be worth testing that all the IDs are unique. A collision should be rare, but not impossible.

Alhadis:
Weren't we going to use the digest of the entire `languages.yml` file before the ID is calculated? That's guaranteed to be unique every time, because the hash is being generated based on the entire file's data *before* the `language_id` field is inserted.

For instance, let's assume I'm in the middle of adding a new entry:
```yml
0x00:
  type: programming
  extensions:
  - ".00"
  tm_scope: none
  ace_mode: text
```
Prepending that to the [current state of the file](https://github.com/github/linguist/blob/ffd984bb7e6dec6f82f359308a9a1ea542247191/lib/linguist/languages.yml) yields a checksum of `055d502e8b8a4bd1ea4367772cd1e86e`:

```sh
λ GitHub-Linguist (master): md5 lib/linguist/languages.yml
055d502e8b8a4bd1ea4367772cd1e86e
```

Now, if I add *that* checksum to the entry...

```yml
0x00:
  type: programming
  extensions:
  - ".00"
  tm_scope: none
  ace_mode: text
  language_id: 055d502e8b8a4bd1ea4367772cd1e86e
```

... suddenly, the checksum is different:

```
λ GitHub-Linguist (master): md5 lib/linguist/languages.yml
5f2732b95e6ba0ed1f3d66c97d4be45c
`

Theoretically, that should make it impossible for there to ever be a collision, should it not? The only conflict I can foresee is if two people are adding the same language at the same time, from the same copy of the file. =)

test/test_language.rb
pchaigno: Unless you're thinking of something different, [there's already a test for that](https://github.com/github/linguist/blob/ffd984bb7e6dec6f82f359308a9a1ea542247191/test/test_language.rb#L429-L435).

pchaigno:
> Weren't we going to use the digest of the entire languages.yml file before the ID is calculated? That's guaranteed to be unique every time

No, it's only guaranteed to be different with a probability that depends on the size of the hash function output (output values are guaranteed to be uniformly distributed). That should be the probability I computed above.
If we do not want the `language_id`s to be the same for two entries with the same language name, then we should simply generate a completely random id. It is simpler and provides us with the same collision probability.

Alhadis:
Well then, perhaps just hashing a concatenation of the name and a random ID?

pchaigno:
@Alhadis That would give us the same result as generating directly a random id ;)

You seem to think that we should generate a different id every time, even if the languages' names are the same. Why is that? Do you see a case where my proposed method would be an issue?

Alhadis:
No, my approach to dealing with numbers I can't grasp is to blast it with as much OTT syrup as possible.

I'll, erm, let you drive... pardon me while I sit in the back seat with all my regular expressions...

brandonblack:
@pchaigno I'm happy with the size / possibility space for generating these hashes. If we start seeing too many collisions we can add some entropy into into how we generate these but I don't think it'll be necessary.

On GitHub's side, this change won't have much impact since this field in our database and search it set for 32-bit integers and this change stays within that. However, I'm not familiar enough with linguist's usage outside for GitHub to really say the same for everyone else.

How much of a disturbance will this change be to others? Are there any special considerations we should think about in releasing this change?

CC: @LizzHale @lildude @TwP

pchaigno:
> If we start seeing too many collisions we can add some entropy into into how we generate these but I don't think it'll be necessary.

Unfortunately there is no way to reduce the probability of collisions other than extending the id bit size since SHA-256 already achieves uniform distribution.

> However, I'm not familiar enough with linguist's usage outside for GitHub to really say the same for everyone else.

I don't know of any major project (outside of github.com) which uses Linguist.

> How much of a disturbance will this change be to others? Are there any special considerations we should think about in releasing this change?

This change only affects the way ids are generated so I don't think there will be any disturbance; it does not affect the API.

PR #3285, opened 2016-10-18
Dev test exclusion

pierrickrogermele:
Exclude 'test' dans 'tests' directories from language statistics. Add the possibilities to flag files or directories as test inside the .gitattributes file.

I've met the issue with my project [r-biodb](https://github.com/pierrickrogermele/r-biodb) which is tagged as HTML instead of R. This is because in the 'test' directory are stored some big HTML files.
This issue could also have been solved by excluding the 'res' subdirectory, which is the directory containing resources. However I've thought that, more generally, the language used in the tests is not necessarily the one used in the application.

We could also make the 'test.yml' empty and let repository owners decide to add or not test directories in the .gitattributes file.

### DIFF
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -251,6 +251,21 @@ module Linguist
       path =~ DocumentationRegexp ? true : false
     end
 
+    test_paths = YAML.load_file(File.expand_path("../test.yml", __FILE__))
+    TestRegexp = Regexp.new(test_paths.join('|'))
+
+    # Public: Is the blob in a test directory?
+    #
+    # Test files are ignored by language statistics.
+    #
+    # See "test.yml" for a list of test conventions that match
+    # this pattern.
+    #
+    # Return true or false
+    def test?
+      path =~ TestRegexp ? true : false
+    end
+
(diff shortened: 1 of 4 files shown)


Alhadis:
We generally use `linguist-vendored` or `linguist-documentation` for that.

The names are misleading, but they fulfil the same role as what this PR is doing.

pierrickrogermele:
I agree they fulfil the same role. However as you've said it's misleading, that's why I'm proposing this pull request. I think that would be clearer for everybody.
In addition I propose to exclude by default 'test' and 'tests' directories.

arfon:
Thanks for this PR @pierrickrogermele.

> In addition I propose to exclude by default 'test' and 'tests' directories.

I certainly wouldn't want this change on my projects or any of our GitHub Ruby or JavaScript projects (note we already exclude fixtures data [here](https://github.com/github/linguist/blob/67ea35094b50f22864f23b0573c5b6166910ff9e/lib/linguist/vendor.yml#L287-L289)). 

I would encourage you to look at the `linguist-vendored=true` flag that @Alhadis is pointing you at as this will accomplish the same thing.

PR #3287, opened 2016-10-22
Update language-haskell grammar

lierdakil:
A couple annoying bugs fixed upstream, so I figured I should update here as well.


Alhadis:
No need. Grammars are always updated automatically with every new release of Linguist. =)

Besides, even if we were to merge this, the changes wouldn't be visible on the site until the next release is cut.

lierdakil:
Wow, that is a fast response! Okay, feel free to disregard this then.

PR #3306, opened 2016-10-29
Update Swift grammar to fix hexadecimal numbers

jtbandes:
In the recently updated Swift grammar, we used `\h` for hexadecimal digits, which isn't supported by GitHub's highlighter.


Alhadis:
Grammars are already updated automatically with every new release of Linguist. Which is also when changes become visible on the site... so even if we were to merge this, it wouldn't have any effect until the next release. =)

jtbandes:
OK, well...feel free to close or merge as you see fit then! Looking forward to when
```swift
0x1
```
isn't broken 🙂

Alhadis:
I guess you could call that...  
<pre>•_•)  
( •_•)>⌐■-■  
(⌐■_■)  </pre>
... a Swift reply...

jtbandes:
```swift
if case let ಠ_ಠ = @Alhadis {
    // this thread
}
```

huh, highlighting there isn't great either 😉

jtbandes:
In seriousness, though, maybe GitHub should check for `\h` in `.tmLanguage` grammars and somehow convert it to `[0-9a-fA-F]` — TextMate uses Oniguruma where `\h` is a hex digit, but apparently `\h` matches horizontal whitespace in PCRE... I wonder how many other grammars are currently broken.

Alhadis:
I'm on-and-off [experimenting](https://github.com/Alhadis/TMCG-Spec) with ideas for a canonical grammar format... if/when it solidifies, I'll churn out a Node app for converting grammars between different engines/implementations. Which will render this issue moot.

Because I've had a gutful of the PCRE/Oniguruma incompatibilities myself.

jtbandes:
@Alhadis to hijack my own thread for another purpose... perhaps you can advise on how scopes are merged to produce styles in GitHub? In this example:

```swift
func foo(bar: T)
func foo(bar baz: T)
```

On the first line, `bar` is *both* `entity.name.function` and `variable.parameter.function` (whereas on the second line, `bar` is part of the function name and `baz` is the parameter name). This is intentional, and I'd hope that `bar` in the first line would be colored the same way as `foo`, but GitHub doesn't do that. Is there a good reason it doesn't pick up the `entity.name.function` color here?

TM for comparison:

Alhadis:
At a glance, it seems the `bar` on the first line is being tokenised with a class-name of `.pl-smi` (which maps to the [`storage.modifier.import` scope](https://gist.github.com/Alhadis/f742feb8f5c871b449dff39bb3dd6f5d#file-scope-names-js-L36)). That's apparently nested inside an `entity` scope, according to my dev-tools:

Figure 1

But you mentioned `variable.parameter.function`... and, well, I'm not an expert on how GitHub interprets TextMate highlighting, but I know that it probably only uses a subset of scope-names, dropping whatever it doesn't recognise.

jtbandes:
Odd. `storage.modifier.import` shouldn't appear anywhere in the scopes.

image

Alhadis:
At a loss to explain this one then. My only remaining advice would be to swap the scope order so that `entity.name.function` is nested inside `variable.parameter.function`.

Is there any reason they need to be scoped as both? All I know about Swift is that it made every developer piss themselves like an excited dog the year it was released, but nothing else.

jtbandes:
Good call, swapping them produces better results. (They are scoped as both because in `func foo(bar: T)`, `bar` is the argument name inside the function, but it's also required as a label when calling the function as in `foo(bar: 1)`.)

Alhadis:
Alright, well, if we're done here, you can feel free to close this. ;)

(I can't because I'm not added as a project collaborator or anything)

PR #3307, opened 2016-10-29
3192  part deux

arfon:
Replaces #3192


arfon:
✨ @Alhadis. Thanks for this.

Alhadis:
No problem! :D Though it probably wouldn't hurt to have a link to the list from the site's help docs. The page that links here could also include a link to the grammars, stating where to file a report if one encounters an error with syntax highlighting... etc.

arfon:
>No problem! :D Though it probably wouldn't hurt to have a link to the list from the site's help docs. The page that links here could also include a link to the grammars, stating where to file a report if one encounters an error with syntax highlighting... etc.

Good call. I added it to the add grammar script but forgot about that addition. Will work on that now.

arfon:
Hang on, do you mean the docs on help.github.com?

Alhadis:
Yup, the part of GitHub that's untouchable by pull requests.

Alhadis:
I don't know if that's within your domain or not, but I figure I should point it out all the time, since that's where a lot of confused visitors came from...

arfon:
> Yup, the part of GitHub that's untouchable by pull requests.

😸 understood.

So here's the thing, we actually deliberately made the help docs on `help.github.com` just point to this repository as most of the issues that people were having with Linguist required some assistance from people working on Linguist (and not our support team). That's why the README here reads as it does (i.e. as a troubleshooting guide). I'm going to take a stab at adding some additional docs to the README and in the longer term we should think about creating an issue template for this repository to help people open more useful issues.

arfon:
OK, opened something up here: https://github.com/github/linguist/pull/3308

PR #3311, opened 2016-10-31
Add Support for ABNF

sanssecours:
Hi,

this commit adds support for [Augmented Backus–Naur form][], a metalanguage used to specify language grammars. The language does not seem to be [that popular on GitHub](https://github.com/search?utf8=✓&q=extension%3Aabnf+NOT+nothack&type=Code&ref=searchresults), but important enough to be used in [popular repositories](https://github.com/toml-lang/toml/blob/abnf/toml.abnf). A RFC document, describing the current version of the language, is available [here](https://tools.ietf.org/html/rfc5234).

If I should change anything in this pull request, then please just comment below. Thank you.

Kind regards,
  René

[Augmented Backus–Naur form]: https://en.wikipedia.org/wiki/Augmented_Backus–Naur_form

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -49,6 +49,13 @@ ABAP:
   - ".abap"
   ace_mode: abap
   language_id: 1
+ABNF:
+  type: data
+  ace_mode: text
+  extensions:
+  - ".abnf"
+  tm_scope: source.abnf
+  language_id: 429
 AGS Script:
   type: programming
   color: "#B9D9FF"
--- a/.gitmodules
+++ b/.gitmodules
@@ -797,3 +797,6 @@
 [submodule "vendor/grammars/actionscript3-tmbundle"]
 	path = vendor/grammars/actionscript3-tmbundle
 	url = https://github.com/simongregory/actionscript3-tmbundle
+[submodule "vendor/grammars/ABNF.tmbundle"]
+	path = vendor/grammars/ABNF.tmbundle
+	url = https://github.com/sanssecours/ABNF.tmbundle
--- a/grammars.yml
+++ b/grammars.yml
@@ -4,6 +4,8 @@ http://svn.edgewall.org/repos/genshi/contrib/textmate/Genshi.tmbundle/Syntaxes/M
 https://bitbucket.org/Clams/sublimesystemverilog/get/default.tar.gz:
 - source.systemverilog
 - source.ucfconstraints
+vendor/grammars/ABNF.tmbundle:
+- source.abnf
 vendor/grammars/Agda.tmbundle:
 - source.agda
 vendor/grammars/Alloy.tmbundle:
--- a/samples/ABNF/toml.abnf
+++ b/samples/ABNF/toml.abnf
@@ -1,0 +1,190 @@
+; Source:  https://github.com/toml-lang/toml
+; License: MIT
+
+;; This is an attempt to define TOML in ABNF according to the grammar defined
+;; in RFC 4234 (http://www.ietf.org/rfc/rfc4234.txt).
+
+;; TOML
+
+toml = expression *( newline expression )
+expression = (
+  ws /
+  ws comment /
+  ws keyval ws [ comment ] /
+  ws table ws [ comment ]
+)
+
+;; Newline
+
+newline = (
+  %x0A /              ; LF
+  %x0D.0A             ; CRLF
+)
+
+newlines = 1*newline
+
+;; Whitespace
+
+ws = *(
+  %x20 /              ; Space
+  %x09                ; Horizontal tab
+)
+
+;; Comment
+
+comment-start-symbol = %x23 ; #
+non-eol = %x09 / %x20-10FFFF
+comment = comment-start-symbol *non-eol
+
+;; Key-Value pairs
+
+keyval-sep = ws %x3D ws ; =
+keyval = key keyval-sep val
+
+key = unquoted-key / quoted-key
+unquoted-key = 1*( ALPHA / DIGIT / %x2D / %x5F ) ; A-Z / a-z / 0-9 / - / _
+quoted-key = quotation-mark 1*basic-char quotation-mark ; See Basic Strings
+
+val = integer / float / string / boolean / date-time / array / inline-table
+
+;; Table
+
+table = std-table / array-table
+
+;; Standard Table
+
+std-table-open  = %x5B ws     ; [ Left square bracket
+std-table-close = ws %x5D     ; ] Right square bracket
+table-key-sep   = ws %x2E ws  ; . Period
+
+std-table = std-table-open key *( table-key-sep key) std-table-close
+
+;; Array Table
+
+array-table-open  = %x5B.5B ws  ; [[ Double left square bracket
+array-table-close = ws %x5D.5D  ; ]] Double right square bracket
+
+array-table = array-table-open key *( table-key-sep key) array-table-close
+
+;; Integer
+
+integer = [ minus / plus ] int
+minus = %x2D                       ; -
+plus = %x2B                        ; +
+digit1-9 = %x31-39                 ; 1-9
+underscore = %x5F                  ; _
+int = DIGIT / digit1-9 1*( DIGIT / underscore DIGIT )
+
+;; Float
+
+float = integer ( frac / frac exp / exp )
+zero-prefixable-int = DIGIT *( DIGIT / underscore DIGIT )
+frac = decimal-point zero-prefixable-int
+decimal-point = %x2E               ; .
+exp = e integer
+e = %x65 / %x45                    ; e E
+
+;; String
+
+string = basic-string / ml-basic-string / literal-string / ml-literal-string
+
+;; Basic String
+
+basic-string = quotation-mark *basic-char quotation-mark
+
+quotation-mark = %x22            ; "
+
+basic-char = basic-unescaped / escaped
+escaped = escape ( %x22 /          ; "    quotation mark  U+0022
+                   %x5C /          ; \    reverse solidus U+005C
+                   %x2F /          ; /    solidus         U+002F
+                   %x62 /          ; b    backspace       U+0008
+                   %x66 /          ; f    form feed       U+000C
+                   %x6E /          ; n    line feed       U+000A
+                   %x72 /          ; r    carriage return U+000D
+                   %x74 /          ; t    tab             U+0009
+                   %x75 4HEXDIG /  ; uXXXX                U+XXXX
+                   %x55 8HEXDIG )  ; UXXXXXXXX            U+XXXXXXXX
+
+basic-unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
+
+escape = %x5C                    ; \
+
+;; Multiline Basic String
+
+ml-basic-string-delim = quotation-mark quotation-mark quotation-mark
+ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
+ml-basic-body = *( ml-basic-char / newline / ( escape newline ))
+
+ml-basic-char = ml-basic-unescaped / escaped
+ml-basic-unescaped = %x20-5B / %x5D-10FFFF
+
+;; Literal String
+
+literal-string = apostraphe *literal-char apostraphe
+
+apostraphe = %x27 ; ' Apostrophe
+
+literal-char = %x09 / %x20-26 / %x28-10FFFF
+
+;; Multiline Literal String
+
+ml-literal-string-delim = apostraphe apostraphe apostraphe
+ml-literal-string = ml-literal-string-delim ml-literal-body ml-literal-string-delim
+
+ml-literal-body = *( ml-literal-char / newline )
+ml-literal-char = %x09 / %x20-10FFFF
+
+;; Boolean
+
+boolean = true / false
+true    = %x74.72.75.65     ; true
+false   = %x66.61.6C.73.65  ; false
+
+;; Datetime (as defined in RFC 3339)
+
+date-fullyear  = 4DIGIT
+date-month     = 2DIGIT  ; 01-12
+date-mday      = 2DIGIT  ; 01-28, 01-29, 01-30, 01-31 based on month/year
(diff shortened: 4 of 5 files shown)


Alhadis:
👋  Hey @sanssecours, thanks for the pull requests. I'll address them both in this post, since they're so closely related.

I'd say there're enough results across GitHub for both languages for them to warrant addition:

| Language | Files | Repos | Users |
|----------|-------|-------|-------|
| ABNF | 2,308 |  828  |  706  |
| EBNF | 2,011 |  997  |  886  |

I wouldn't mind hearing @pchaigno and @arfon's thoughts too, though. :)

Also, a few other things:

* You'll need to change the language-type of each entry to `data`, not `programming`. It's a notation for describing a language, as we both know, and not a real language per se.
* You might find [this grammar](https://github.com/mackoj/language-bnf) to be useful; it provides highlighting for both ABNF and EBNF.

arfon:
> I wouldn't mind hearing @pchaigno and @arfon's thoughts too, though. :)

Agreed. These look like they meet the required usage guidelines.

> - You'll need to change the language-type of each entry to data, not programming. It's a notation for describing a language, as we both know, and not a real language per se.
> - You might find this grammar to be useful; it provides highlighting for both ABNF and EBNF.

👍

Alhadis:
Great! That settles it then. =)

One more thing: you'll also need to show where you acquired those samples, as we require them to be released under a permissive license. Specifically, [one of these](https://github.com/github/linguist/blob/master/vendor/licenses/config.yml#L1-L11).

Lastly, I'd also ask you to remove the `color` fields for each language. As they fall under the `data` category, they won't be listed in language-statistic bars in repositories... which means the colours won't ever be seen by users.

sanssecours:
> - You'll need to change the language-type of each entry to `data`, not `programming`. It's a notation for describing a language, as we both know, and not a real language per se.

Thanks for the information. I just changed the group. May I ask why [ANTLR files](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L74-L80) are listed under `programming` then? As far as I know `.gs4` files also describe grammars. Here is an example of an ANTLR file/grammar from the [official website](http://www.antlr.org) (`Expr.g4`):

```ANTLR
grammar Expr;		
prog:	(expr NEWLINE)* ;
expr:	expr ('*'|'/') expr
    |	expr ('+'|'-') expr
    |	INT
    |	'(' expr ')'
    ;
NEWLINE : [\r\n]+ ;
INT     : [0-9]+ ;
```

> You might find [this grammar](https://github.com/mackoj/language-bnf) to be useful; it provides highlighting for both ABNF and EBNF.

Thanks for the info. I think I will incorporate the [core rules](https://github.com/mackoj/language-bnf/blob/master/grammars/abnf.cson#L56) in my version of the ABNF TextMate grammar too. The EBNF grammar seems to be a direct translation of [Arne's grammar](https://github.com/arne-schroppe/ebnf-textmate-bundle) :o) so we should be covered there.

> One more thing: you'll also need to show where you acquired those samples, as we require them to be released under a permissive license. Specifically, [one of these](https://github.com/github/linguist/blob/master/vendor/licenses/config.yml#L1-L11).

All of the examples are from the English Wikipedia pages for [ABNF](https://en.wikipedia.org/wiki/Augmented_Backus–Naur_form) and [EBNF](https://en.wikipedia.org/wiki/Extended_Backus–Naur_form). I added a link to the respective articles at the top of each sample file. The files are covered under [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0), which sound permissive to me, although I think I also need to provide a link to the license and the name of the authors. Finding the authors of the code sounds like a complicated task, since so many people contribute to Wikipedia articles.

Should I just search for files that are covered by one of the licenses you linked to instead? If you think this option is more appropriate, is a comment that contains a link to the original source inside the files enough?

Alhadis:
> May I ask why [ANTLR](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L74-L80) files are listed under `programming` then?

I'm not sure, as I'm unfamiliar with ANTLR; @larsbrinkhoff might know more about it than I do. But if it's really just another *BNF-like notation, its classification should definitely be changed. =)

> All of the examples are from the English Wikipedia pages for [ABNF](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form) and [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form).

@pchaigno Do you know if the CC BY-SA license is appropriate? I've forgotten what the stance on Wikipedia-sourced samples is.

> If you think this option is more appropriate, is a comment that contains a link to the original source inside the files enough?

No, I'm afraid not. The samples aren't for human readers: they're for the classifier. Linguist uses them as a basis for Bayesian classification, helping it identify languages based on the frequency of keywords and what-have-you. The bigger the reference material, the better.

> Should I just search for files that are covered by one of the licenses you linked to instead?

You could, or you could write your own. Sometimes the latter is easier, particularly for data formats. :)

sanssecours:
> No, I'm afraid not. The samples aren't for human readers: they're for the classifier. Linguist uses them as a basis for Bayesian classification, helping it identify languages based on the frequency of keywords and what-have-you. The bigger the reference material, the better.

Sorry for the confusing description above 😀. I meant, is it enough to just specify the source at the top of a sample file? For example, if I include the file [`toml.abnf`](https://github.com/toml-lang/toml/blob/abnf/toml.abnf) in the sample directory should I add the comment 

```
; Source: https://github.com/toml-lang/toml/blob/abnf/toml.abnf
```

at the top of the file? Where exactly should I add the information about the source/license of a sample file?

> You could, or you could write your own. Sometimes the latter is easier, particularly for data formats. :)

You are right. Writing a grammar that is valid according to the syntax of EBNF/ABNF is not [that hard](https://github.com/sanssecours/ABNF.tmbundle/blob/master/Tests/Example.abnf). Writing a grammar that also make sense semantically is not trivial though, especially if it should include a lot of the features of the metalanguage.

Alhadis:
> Sorry for the confusing description above 😀. I meant, is it enough to just specify the source at the top of a sample file?

Ah, I see what you mean. No, it doesn't work that way. The material will need to be *explicitly* released under a permissive license.

> Where exactly should I add the information about the source/license of a sample file?

Here will do. =) It doesn't have to be in the codebase; just as long as we have a record of where/when it was acquired.

> Writing a grammar that also make sense semantically is not trivial though, especially if it should include a lot of the features of the metalanguage

Well there's not a lot that can go wrong with *BNF, thankfully. ;)

pchaigno:
> You'll need to change the language-type of each entry to `data`, not `programming`. It's a notation for describing a language, as we both know, and not a real language per se.

@Alhadis In the end, `programming` is just a label. IMHO, what matters is what it means for a language to be a `programming` language on github.com. In this case, it means it will be in the statistics bar and a project can be tagged as a ABNF project.

@sanssecours Would you expect to see git repositories with only ABNF files or are they always included with some other files (as some kind of documentation/specification for instance)?

sanssecours:
> Here will do. =) It doesn't have to be in the codebase; just as long as we have a record of where/when it was acquired.

That sounds easy enough. Thank you for the information.

> @sanssecours Would you expect to see git repositories with only ABNF files or are they always included with some other files (as some kind of documentation/specification for instance)?

I would expect that most repositories use ABNF/EBNF very sparsely, maybe as source of documentation for some programming/configuration language. There are however repositories, like [this one](https://github.com/joelself/toml-abnf), where the main file seems to be an ABNF grammar.

larsbrinkhoff:
> May I ask why ANTLR files are listed under `programming` then?

My impression is that ANTLR is similar to Yacc.  They include both grammar rules and program code.

sanssecours:
I replaced the sample files from Wikipedia with samples from GitHub. I added the source and the license at the top of the files. I hope that is okay.

The table below shows the sources and licenses of the files too.

| Files         | Source        | License |
| ------------- | ------------- | ------------- |
| toml.abnf  | https://github.com/toml-lang/toml  | MIT  |
| tranquil.abnf  | https://github.com/fjolnir/Tranquil  | BSD 3 Clause License  |
| grammar.ebnf  | https://github.com/sunjay/lion  | MIT  |
| material.ebnf, object.ebnf, types.ebnf | https://github.com/io7m/jsom0  | ISC  |

If there is anything left for me to do, then please just comment below. Thank you.

lib/linguist/languages.yml
```diff
@@ -49,6 +49,16 @@ ABAP:
   - ".abap"
   ace_mode: abap
   language_id: 1
+ABNF:
+  type: data
+  ace_mode: text
+  extensions:
+  - ".abnf"
+  interpreters:
+  - ex_abnf
```
Alhadis: This field is for languages that can be identified from interpreter directives. =)

For instance, a file with `#!/usr/bin/env node` in its first line can be identified as JavaScript by listing `node` as an interpreter.

Since that doesn't apply here, it's safe to leave this out.

lib/linguist/languages.yml
```diff
@@ -49,6 +49,16 @@ ABAP:
...
+  tm_scope: source.abnf
+  group: ABNF
```
Alhadis: The `group` field here is redundant; we only use this to let one language affect a parent language's usage statistics. For example, [`GAS`](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1326) falls under the category of `Assembly`, so we want to have it count in the usage stats of Assembly languages across GitHub.

The name is slightly misleading; a more accurate name for this field might be `parent_language`.

In short, it's logistically impossible for anything to be its own parent. ;)

samples/ABNF/tranquil.abnf
```diff
@@ -0,0 +1,61 @@
+; Source:  https://github.com/fjolnir/Tranquil
+; License: BSD 3 Clause License
+; Modified for standard compatibility by René Schwaiger
```
pchaigno: What did you need to modify here? We generally try to keep the files as-is (with the eventual mistakes, etc.) as they are used for tests and training.

sanssecours: The [original file](https://github.com/fjolnir/Tranquil/blob/master/Docs/UnusedOld/tranquil.abnf) uses single quotes, which are not part of the [standard](https://tools.ietf.org/pdf/rfc5234.pdf) as far as I can tell.

Alhadis: There was no need to have done that. This is the sort of in-the-wild discrepancy that's both natural and expected; and also part of the reason we need samples in the first place.

lib/linguist/languages.yml
sanssecours: Thank you for the explanation. This should be fixed now.

lib/linguist/languages.yml
sanssecours: Thank you for the comment. I just removed the group fields.

samples/ABNF/tranquil.abnf
sanssecours: > This is the sort of in-the-wild discrepancy that's both natural and expected; and also part of the reason we need samples in the first place.

I do not think that it makes that much sense to add incorrect files. Otherwise we could just add the following “ABNF code” as (part of) a sample file 😀:

```
Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute
irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
deserunt mollit anim id est laborum.
```

The [ABNF TextMate grammar](https://github.com/sanssecours/ABNF.tmbundle/blob/master/Syntaxes/ABNF.tmLanguage) will also (correctly) mark parts of the original `tranquil.abnf` grammar as incorrect:

tranquil

pchaigno: > I do not think that it makes that much sense to add incorrect files.

It does when it is what people use in-the-wild. If you think this file is not representative of the usage of ABNF on github.com you can replace it ;)

Alhadis: > The [ABNF TextMate grammar](https://github.com/sanssecours/ABNF.tmbundle/blob/master/Syntaxes/ABNF.tmLanguage) will also (correctly) mark parts of the original `tranquil.abnf` grammar as incorrect.

Judging by what the grammar marks as invalid, I'd say it's highlighting anything that isn't a whitelisted character (which excludes single-quotes).

The point of the samples is to give a realistic indication of what to expect. And, as you said, `.abnf` files are intended primarily for documentation-specific purposes, which means we can't (and shouldn't) expect every author to feel beholden by what the standard recommends.

Alhadis: [Here's the relevant rule](https://github.com/sanssecours/ABNF.tmbundle/blob/master/Syntaxes/ABNF.tmLanguage#L552-L568) which I believe is applying the error highlighting:

```
[a-zA-Z][a-zA-Z0-9\-]*|(?<invalid>\S)
```

So it's basically saying "highlight anything that's not whitespace, and that isn't an ASCII word character or dash that commences with an alphabetic ASCII character.

sanssecours: May I add a comment that `tranquil.abnf` it is a non-standard-compliant file?

Alhadis: Why?

Alhadis: Recall that these files are for the classifier: 99% of the time, they remain unread by human beings.

sanssecours: > Why?

Because it is a non-standard-compliant file 😀. I would not want someone to think that it represents a correct ABNF grammar.

However, I just removed `tranquil.abnf`. I hope that is a compromise that works for everyone.

Alhadis:
Looks good enough to me then, I guess. @arfon?

arfon:
👍 looks great. Thanks @sanssecours.

This will be live in the next release of Linguist (later this week)

pchaigno:
Thanks for your patience @sanssecours!

PR #3320, opened 2016-11-07
Exclude "_examples" directory from language statistics.

kataras:
Exclude GoLang "_examples" directory from language statistics.

At Go Programming Language we mostly put our examples in the `_examples` directory, because `_` is Excluded from `go build` tool which provides us the binary, executable program file.

For example if someone develop a template parser in Go and the project has `_examples/` directory with tons of html examples, then the project will act as `HTML` project and not `Go`

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -50,6 +50,9 @@
 # Go dependencies
 - Godeps/_workspace/
 
+# Examples directory with underscore, usually in Go projects
+- _examples/
+
 # Minified JavaScript and CSS
 - (\.|-)min\.(js|css)$
 


kataras:
Note: I know we can put `..gitattributes` and that's we do, but I think its better to have that folder as standard language ignored folder

brandonblack:
What happens for non-Go related projects making use of this same folder name?
What happens for other Go projects that do have same-language examples in that folder or want the contents of that folder included?

While I think ignoring `_*` folders _might_ be ok if it could be scoped just to Go projects (as-is it would apply to everything) even then it seems like a heavy handed approach that could potentially conflict with many types of projects.

kataras:
@brandonblack Do you use root directory `_examples/` to store your source code that delivers your app/library ? Who does that with which language? Please answer me I'm developer for more than a decade and I worked with all known languages and I never seen a software, web application or any library author writes library-relative source code in the `root _examples` folder

brandonblack:
@kataras I guess my point is that just because it makes sense in your case, doesn't mean this makes sense for other Go projects or non-Go projects both of which this proposed change would impact. It's a prescriptive change that I'm not sure will work well for everyone. Some developers may prefer to gather language statistics from example folders.

kataras:
Thanks, ignore it and good luck analyzing the examples folder...

PR #3325, opened 2016-11-12
Add py3 to Python's extensions list

### LINKED ISSUE #3323: Python (3) code files with .py3 extensions are not detected or highlighted
Problem: the repo contains several code files in both Python2 and Python3. To distinguish between them we use `.py` extension for py2 and `.py3` extension for py3. But GitHub doesn't highlight syntax in `.py3` files and doesn't count them as code in python language (or any language in fact). This should be fixed.

danilaml:
Closes #3323 (check it for a discussion)
According to @Alhadis,
> The 1895 py3 files on GitHub are distributed among 392 repositories by 366 users

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3372,12 +3372,12 @@ Python:
   color: "#3572A5"
   extensions:
   - ".py"
-  - ".py3"
   - ".bzl"
   - ".cgi"
   - ".fcgi"
   - ".gyp"
   - ".lmi"
+  - ".py3"
   - ".pyde"
   - ".pyp"
   - ".pyt"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3372,6 +3372,7 @@ Python:
   color: "#3572A5"
   extensions:
   - ".py"
+  - ".py3"
   - ".bzl"
   - ".cgi"
   - ".fcgi"
--- a/samples/Python/py3.py3
+++ b/samples/Python/py3.py3
@@ -1,0 +1,1 @@
+print("Hello, world!")
--- a/samples/Python/py3.py3
+++ b/samples/Python/py3.py3
@@ -1,1 +1,20 @@
-print("Hello, world!")
+import random
+
+guesses = 0
+
+number = random.randint(1, 20)
+
+print("Guess the number between 1 and 20! You have 6 tries.")
+while guesses < 6:
+    guess = int(input("Is it... "))
+
+    if guess == number:
+        print("Hooray! You guessed it right!")
+        break
+    elif guess < number:
+        print("It's bigger...")
+    elif guess > number:
+        print("It's not so big.")
+    guesses += 1
+if guesses == 6:
+    print("You've ran out of tries.")


lib/linguist/languages.yml
```diff
@@ -3372,6 +3372,7 @@ Python:
   color: "#3572A5"
   extensions:
   - ".py"
+  - ".py3"
```
Alhadis: This needs to be listed after `".lmi"` instead. The `extensions` list must be alphabetised, with the exception of the language's most common extension (which is listed first).

Otherwise, the tests won't pass and the build will fail.

samples/Python/py3.py3
```diff
@@ -0,0 +1 @@
+print("Hello, world!")
```
pchaigno: @danilaml We prefer samples from real applications. These samples are used to train the classifier and for our tests.

lib/linguist/languages.yml
danilaml: Done. Maybe worth mentioning it in the md. I was confused by the `py` at the top.

Alhadis:
@danilaml Could you please confirm the origin of the sample file you added? We require samples to be released under a permissive license (specifically, [one of these](https://github.com/github/linguist/blob/master/vendor/licenses/config.yml#L1-L11)). Aside from that, I'd say this is good to go. 👍

danilaml:
@Alhadis My bad, I've traced the sample origin and it looks like it's CC BY-NC-SA. I'm not certain, but I've changed it with one that is in public domain, just to be sure (I squashed it with the previous commit).

lib/linguist/languages.yml
danilaml: Yeah, but not in https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language

Alhadis:
Looks good to me. Should we ⚓️ this?

@brandonblack, you're first in alphabetical order of newly-assigned staff members, so I'm pinging you first for feedback. =)

danilaml:
@Alhadis btw, should I accept the Github CLA? I see the check in the PR, but no mention of or link to it in the CONTRIBUTING.md.

Alhadis:
Ah, yes please! That's actually the first thing every contributor should do... once it's done, you'll never have to worry about it ever again. =) It's just a legal requirement, like agreeing to an EULA when installing a program. There's no hidden agenda or anything.

... as for the CI status, I'm not sure why it's stalled. :S It might be confused by the lack of CLA agreement (though it usually fails if that's the case).

danilaml:
@Alhadis I had to go read [git-lfs readme](https://github.com/github/git-lfs/blob/master/CONTRIBUTING.md#submitting-a-pull-request) to find out how to acquire proper link for this repo's CLA. Is there any information on it in this repo and I've just missed it?

Alhadis:
Okay, looks good! CI's lovely and green. 💚 Incidentally, my favourite colour.

Well, when CI fails due to a missing CLA signature, it's supposed to display a notice which includes links to the relevant pages. In this case though, it's behaving weirdly, so of course you weren't notified about it.

The GitHub CLA actually isn't specific to this project: instead, it's specific to every open source project which GitHub is legally responsible for. So including links to the CLA pages for every affected repository would be slightly spammy.

brandonblack:
This looks good. Thanks @danilaml and @Alhadis.

PR #3332, opened 2016-11-17
Add Reason, a variant of OCaml

jlongster:
Reason is a new variant of OCaml. For now, we can just highlight it as OCaml code. In the future we may have better support for more specific features.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2835,6 +2835,7 @@ OCaml:
   - ".mli"
   - ".mll"
   - ".mly"
+  - ".re"
   interpreters:
   - ocaml
   - ocamlrun


Alhadis:
Hi James, thanks for the pull request! 👋 

First of all, you'll need to [accept the GitHub CLA](https://cla.github.com/github/linguist/accept/jlongster) before we can consider your PR. It's a necessary first step for every contributor. =)

Secondly, it appears as though Reason is a separate language to OCaml... is this correct? In which case, it should ideally be added as a new language in Linguist, provided its in-the-wild usage is widespread enough. Since it's from Facebook, I doubt distribution won't be a problem (I haven't checked yet how many Reason files there are on GitHub).

Please make sure you've also read the contribution notes on [adding a new language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language).

jlongster:
Thanks, I will sign the CLA soon. Reason basically is OCaml but with several small syntactic changes. I was thinking they were small enough that this should work, but I just tried it on my own projects and there is a syntactic change that doesn't work with the OCaml highlighting. So this won't work.

I'll open a new PR at some point to make it a separate language. Thanks for the links to the docs, and for the quick reply!

PR #3335, opened 2016-11-18
change http to https

Lidilidian:
fix the little flaws found on http protocol that is used, one of the web using the http protocol which is already supporting more secure protocol, which is https


Alhadis:
Hi @Lidilidian,

I see Edgewall's HTTP link is redirecting to its HTTPS version, so it looks safe to change.

However, there's [another file](https://github.com/github/linguist/blob/master/grammars.yml#L2) with a URL you have to update in order to do this.

brandonblack:
Hello @Lidilidian 👋  

Thanks for the contribution.

If you can update your pull request with the latest from `master` it should resolve the Ant related issue in the failing build. If you can get the build green and address the other Edgewall URL @Alhadis mentioned we'll take another 👀  at this and try to get it merged.

Lidilidian:
okay, done at #3357

PR #3337, opened 2016-11-18
Remove Ant from grammar whitelist

pchaigno:
The last version of Licensee ([`v8.7.0`](https://github.com/benbalter/licensee/releases/tag/v8.7.0)) can recognize underlined license headers in READMEs (see benbalter/licensee#138). ~~Thus, the license of the Ant grammar is correctly recognized and whitelisting it is not necessary anymore.~~ EDIT: Unfortunately the Ant grammar doesn't have exactly the usual TextMate bundle license, so the license hash is different and I had to add it to the hash whitelist (meaning we just switched one whitelist entry for another :/). 

Fixes the master branch Travis build.

### DIFF
--- a/test/test_grammars.rb
+++ b/test/test_grammars.rb
@@ -6,7 +6,6 @@ class TestGrammars < Minitest::Test
   # List of projects that are allowed without licenses
   PROJECT_WHITELIST = [
     "vendor/grammars/Sublime-Lasso",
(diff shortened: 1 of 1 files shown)


Alhadis:
So, Linguist had an Ant problem?

Thankfully it wasn't termites.

`#self-high-five`

brandonblack:
Thanks for this @pchaigno 👍

PR #3340, opened 2016-11-22
Treat bcsymbolmap as binary

kkodev:
There are reasons why people need to keep these in the repo, so at least lets treat these files as binary.

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -52,6 +52,7 @@ module Linguist
     # Return true or false
     def generated?
       xcode_file? ||
+      bitcode_symbolmap? ||
       generated_net_designer_file? ||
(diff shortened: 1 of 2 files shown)


pchaigno:
Thanks for the pull request!

What's the result you're trying to achieve here? These files are currently not included in statistics so I'm not sure why marking them generated is required.

kkodev:
I was hoping to get Github to collapse these files in diffs and seems like modifying linguist is one way to do it.

brandonblack:
This actually won't accomplish what you're hoping for. 😞 

Files on GitHub get flagged as binary and collapsed according in diffs based on attempts to detect their encoding type. If a binary encoding type is detected or if we fail to detect any encoding type, we treat the file as a binary.

These are auto-generated, but definitely not binary.

PR #3341, opened 2016-11-22
Add Nunjucks highlighting

SimenB:
Search results: https://github.com/search?utf8=%E2%9C%93&q=extension%3Anjk+NOT+nothack&type=Code&ref=searchresults

Nunjucks docs on file extension: https://mozilla.github.io/nunjucks/templating.html#file-extensions

> # File extensions
> Although you are free to use any file extension you wish for your Nunjucks template files, the Nunjucks community has adopted `.njk`.
> If you are developing tools or editor syntax helpers for Nunjucks, please include recognition of the `.njk` extension.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1659,13 +1659,16 @@ HTML+Django:
   tm_scope: text.html.django
   group: HTML
   extensions:
-  - ".mustache"
   - ".jinja"
+  - ".mustache"
+  - ".njk"
   aliases:
   - django
   - html+django/jinja
   - html+jinja
   - htmldjango
+  - njk
+  - nunjucks
   ace_mode: django
   codemirror_mode: django
   codemirror_mime_type: text/x-django
--- a/samples/HTML+Django/nunjucks.njk
+++ b/samples/HTML+Django/nunjucks.njk
@@ -1,0 +1,48 @@
+{% from "forms.html" import label as description %}
+
+
+{% macro field(name, value='', type='text') %}
+    <div class="field">
+        <input type="{{ type }}" name="{{ name }}"
+                value="{{ value | escape }}" />
+    </div>
+{% endmacro %}
+
+<html>
+<head>
+    {% extends "head.html" %}
+</head>
+<body>
+{% if horse %}
+    Chuck Norris once kicked a horse in the chin. Its descendants are known today as Giraffes.
+{% elif optimus %}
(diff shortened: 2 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -1659,13 +1659,16 @@ HTML+Django:
   tm_scope: text.html.django
   group: HTML
   extensions:
-  - ".mustache"
   - ".jinja"
+  - ".mustache"
```
pchaigno: Why did you move the `.mustache` extension?

samples/HTML+Django/nunjucks.njk
```diff
@@ -0,0 +1,9 @@
+{% extends "parent.html" %}
```
pchaigno: Is this from a "real" application? Do you have a more realistic sample file? Sample files are used to train our Bayesian classifier so it's important they reflect in-the-wild syntaxes and keywords.

lib/linguist/languages.yml
SimenB: To make it alphabetical. The tests told me off since i put `njk` before `jinja`, so i just sorted the whole thing 

https://travis-ci.org/github/linguist/jobs/178098254 (line 1108)

samples/HTML+Django/nunjucks.njk
SimenB: No, just a random snippet. I can add a more substantial example, though!

lib/linguist/languages.yml
Alhadis: Ah, please don't. We make an exception for the first extension in an extensions list, because that's what marks the "primary" (most common) extension. So what you've done here is make `.jinja` the primary extension of `HTML+Django`.

SimenB: Well, that's true, isn't it? Mustache is primarily a js templating language. And the django just says "django templating language" (no mention of Mustache at all), and calls out jinja2 as an alternative. 

https://docs.djangoproject.com/en/1.10/topics/templates/

samples/HTML+Django/nunjucks.njk
pchaigno: That'd be great!

SimenB:
@pchaigno Added example with `extends`, `macro`, `if`, `block, `for`, comments, a filter, and `import`. Mostly cut from https://mozilla.github.io/nunjucks/templating.html

SimenB:
@pchaigno Is this GTG?

brandonblack:
Thanks @SimenB! 🎉

PR #3354, opened 2016-11-30
Fix add-grammar and convert-grammars

meganemura:
Since I want to replace grammar, I ran
`./script/add-grammar --replace ruby-haml.tmbundle https://github.com/ezekg/language-haml`.
As a result, I noticed that it was strange behavior.

1. Added new `language-haml` to `grammars.yml`, but there is also old `ruby-haml.tmbundle` in `grammars.yml`

2. The old `ruby-haml.tmbundle` line was deleted from `vendor/README.md`, but the new `language-haml` was not added to `vendor/README.md`.

This PR fixes those behaviors by updating `grammars.yml`. This is why running `script/convert-grammars` after removing old one in `add-grammar`.
And the other commit makes `script/convert-grammars` working well with the directory of the path entried in `grammars.yml` doesn't exist.


meganemura:
license files were not updated in #3355. I try to investigate it later.

meganemura:
I finally straighten it out.
`add-grammar --replace` does not update license files.
But it will do now with these commits.

PR #3361, opened 2016-12-04
Mark .indent.pro files as vendored

### LINKED ISSUE #3202: GNU Indent profile mistaken as Prolog source
Linguist mistakes [GNU indent](https://www.gnu.org/software/indent/manual/indent.html) profile files (usually named `.indent.pro`) as being Prolog source files. Here's an [example](https://github.com/AelitaStyles/libirctree/blob/master/.indent.pro) of this happening.

pchaigno:
This pull request marks [`.indent.pro` files](https://github.com/search?o=desc&q=filename%3A.indent.pro&s=indexed&type=Code&utf8=%E2%9C%93) (GNU Indent profiles) as vendored.

Fixes #3202.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -50,6 +50,9 @@
 # Go dependencies
 - Godeps/_workspace/
 
+# GNU indent profiles
+- .indent.pro
+
 # Minified JavaScript and CSS
 - (\.|-)min\.(js|css)$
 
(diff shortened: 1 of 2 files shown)


Alhadis:
Probably the best option we have, all things considered. 👍

Not widespread enough to warrant its own language, but still widespread enough to be a classification issue. Nasty.

PR #3366, opened 2016-12-07
Add .pod6 filetype to Perl6

samcv:
`.pod6` is only used for Perl 6 so there should not be any issues with adding this. Thanks!

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -175,7 +175,7 @@ module Linguist
         Language["ECL"]
       end
     end
-    
+
     disambiguate ".es" do |data|
       if /^\s*(?:%%|main\s*\(.*?\)\s*->)/.match(data)
         Language["Erlang"]
@@ -380,6 +380,22 @@ module Linguist
       end
     end
 
+    disambiguate ".pod" do |data|
+      if /^\s*=\w+$/.match(data)
+        if /^=pod|=cut/.match(data)
+          Language["Pod"]
+        elsif /^\s*=begin pod/.match(data)
+          Language["Perl 6"]
+        else
+          Language["Pod"]
+        end
+      elsif Perl6Regex.match(data)
+        Language["Perl 6"]
+      elsif /^\s*\/\* XPM \*\//.match(data)
+        Language["XPM"]
+      end
+    end
+
     disambiguate ".pro" do |data|
       if /^[^\[#]+:-/.match(data)
         Language["Prolog"]
@@ -455,13 +471,13 @@ module Linguist
         Language["SQL"]
       end
     end
-    
+
     disambiguate ".srt" do |data|
       if /^(\d{2}:\d{2}:\d{2},\d{3})\s*(-->)\s*(\d{2}:\d{2}:\d{2},\d{3})$/.match(data)
         Language["SubRip Text"]
       end
     end
-    
+
     disambiguate ".t" do |data|
       if Perl5Regex.match(data)
         Language["Perl"]
@@ -471,7 +487,7 @@ module Linguist
         Language["Turing"]
       end
     end
-    
+
     disambiguate ".toc" do |data|
       if /^## |@no-lib-strip@/.match(data)
         Language["World of Warcraft Addon Data"]
@@ -504,7 +520,7 @@ module Linguist
         Language["XML"]
       end
     end
-  
+
     disambiguate ".w" do |data|
       if (data.include?("&ANALYZE-SUSPEND _UIB-CODE-BLOCK _CUSTOM _DEFINITIONS"))
         Language["OpenEdge ABL"]
@@ -512,7 +528,7 @@ module Linguist
         Language["CWeb"]
       end
     end
-  
+
     disambiguate ".x" do |data|
       if /\b(program|version)\s+\w+\s*{|\bunion\s+\w+\s+switch\s*\(/.match(data)
         Language["RPC"]
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3401,6 +3401,8 @@ Perl 6:
   - ".pl6"
   - ".pm"
   - ".pm6"
+  - ".pod"
+  - ".pod6"
   - ".t"
   filenames:
   - Rexfile
--- a/samples/Perl 6/S15-unicode.pod
+++ b/samples/Perl 6/S15-unicode.pod
@@ -1,0 +1,738 @@
+=comment Oh, this is in Pod6 format. Just so you know.
+
+=begin pod
+
+=TITLE Synopsis 15: Unicode [DRAFT]
+
+=VERSION
+=for table
+    Created          2 December 2013
+    Last Modified    10 October 2014
+    Version          8
+
+This document describes how Unicode and Perl 6 work together. Needless to say,
+it would be good for your chosen reader to support various Unicode characters.
+
+=head1 String Base Units
+
+A Unicode string can be looked at in any of four ways. It could be seen in terms
+of its graphemes, its codepoints, its encoding's code units, or the bytes that
+make up the encoding.
+
+For example, consider a string containing the Devanagari syllable नि, which is
+comprised of the codepoints
+
+    U+0928 DEVANAGARI LETTER NA
+    U+093F DEVANAGARI VOWEL SIGN I
+
+There are a variety of ways in which to perceive the length of this string. For
+reference, here is how the syllable gets encoded in each of UTF-8, UTF-16BE, and
+UTF-32BE.
+
+=for table
+    UTF-8       E0 A4 A8 E0 A4 BF
+    UTF-16BE    0928 093F
+    UTF-32BE    00000928 0000093F
+
+And depending on if you desire to count by graphemes, codepoints, code units, or
+bytes, the perceived length of the string differs:
+
+=for table
+    |------------+-------+--------+--------|
+    |  count by  | UTF-8 | UTF-16 | UTF-32 |
+    |============+=======+========+========|
+    | bytes      | 6     | 4      | 8      |
+    | code units | 6     | 2      | 2      |
+    | codepoints | 2     | 2      | 2      |
+    | graphemes  | 1     | 1      | 1      |
+    |------------+-------+--------+--------|
+
+Perl 6 offers various mechanisms to count by each of these "base units" of a
+string.
+
+Perl 6 by default operates on graphemes, so counting by graphemes involves:
+
+    "string".chars
+
+To count by codepoints, conversion of a string to one of NFC, NFD, NFKC, or NFKD
+is needed:
+
+    "string".NFC.chars
+    "string".NFKD.codes
+
+To count by code units, you can convert to the appropriate buffer type.
+
+    "string".encode("UTF-32LE").elems
+    "string".encode("utf-8").elems
+
+And finally, counting by bytes simply involves converting that buffer to a
+C<buf8> object:
+
+    "string".encode("UTF-16BE").buf8.elems
+
+Note that C<utf8> already stores by bytes, so the count for bytes and code units
+is always the same.
+
+=head1 Normalization Forms
+
+=head2 NFG
+
+Perl 6, by default, stores all strings given to it in NFG form, Normalization
+Form Grapheme. It's a Perl 6–invented character representation, designed to deal
+with un-precomposed graphemes properly.
+
+Formally Perl 6 graphemes are defined exactly according to Unicode Grapheme Cluster Boundaries
+at level "extended" (in contrast to "tailored" or "legacy"),
+see Unicode Standard Annex #29 UNICODE TEXT SEGMENTATION
+L<3 Grapheme Cluster Boundaries|http://www.unicode.org/reports/tr29/tr29-25.html#Grapheme_Cluster_Boundaries>>.
+This is the same as the Perl 5 character class
+
+  \X   Match Unicode "eXtended grapheme cluster"
+
+With NFG, strings start by being run through the normal NFC process, compressing
+any given character sequences into precomposed characters.
+
+Any graphemes remaining without precomposed characters, such as ậ or नि, are
+given their own internal designation to refer to them, at least 32 bits in
+length, in such a way that they avoid clashing with any potential future
+changes to Unicode.  The mapping between these internal designations and
+graphemes in this form is not guaranteed constant, even between strings in
+the same process.
+
+The Perl 6 C<Str> type, and more generally the C<Stringy> role, deals
+exclusively in NFG form.
+
+=head2 NFC and NFD
+
+The NFC and NFD normalization forms are a defined part of the Unicode
+standard. NFD takes precomposed characters and separates them into their
+constituent parts, with a specific ordering of those pieces. NFC tries to
+replace characters sequences into singular precomposed characters whenever
+possible, after first running it through the NFD process.
+
+These two Normalization Forms are similar to NFG, except that graphemes without
+precomposed versions exist as multiple codepoints.
+
+NFC is the form Perl 6 uses whenever NFG is not viable, such as printing the
+string to stdout or passing it to a C<{ use v5; }> section.
+
+=head2 NFKC and NFKD
+
+These forms are known as compatibility forms (denoted by a K to avoid confusion
+with C for Composition). They are similar to their canonical counterparts, but
+may transform various characters (such as ﬁ or ſ) to perform better with the
+software.
+
+All four of NFC, NFD, NFKC, and NFKD can be considered valid "codepoint views",
+though each differ in their exact formulation of the contents of a string:
+
+    say "ẛ̣".chars;               # OUTPUT: 1 (NFG, ẛ̣)
+    say "ẛ̣".NFC.chars;           # OUTPUT: 2 (NFC, ẛ + ̣)
+    say "ẛ̣".NFD.chars;           # OUTPUT: 3 (NFD, ſ + ̣+ ̇)
+    say "ẛ̣".NFKC.chars;          # OUTPUT: 1 (NFKC, ṩ)
+    say "ẛ̣".NFKD.chars;          # OUTPUT: 3 (NFKD, s + ̣+ ̇)
+
+Those who wish to operate with strings on the codepoint level may wish to use
+NFC, as it is the least different from NFG, as well as Perl 6's default form for
+NFG-less contexts.
+
+All of C<Uni>, C<NFC>, C<NFD>, C<NFKC>, and C<NFKD>, and more generally the
+C<Unicodey> role, deal with the various codepoint-based compositions.
+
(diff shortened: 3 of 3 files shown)


Alhadis:
Yes, this actually should be classed as Pod. So in other words, `$_` (what he said).

@samcv Could you please amend? =)

samcv:
Well Perl 6 pod has different syntax than Perl pod. And the grammar we use for Perl 6 supports Perl 6 pod. I could add a sample file though.

Alhadis:
[Still looks like Plain Old Documentation](https://github.com/Tux/CSV/blob/5974d7129fa350320369f979d86b41d75104aeeb/doc/Text/CSV.pod6) to me. Am I missing something here?

I'm only familiar with Perl 5, sorry. [Blame Camelia](https://perl6.org/):

> The only requirement is that you know how to be nice to all kinds of people

Yeah fuck that.

samcv:
Here is a sample of how it currently renders pod:
https://gist.github.com/samcv/6ca5fbeac9482d52145b48f8872be3aa

Maybe I should wait until v1.9 of the grammar gets released. I already have it ready to go, but waiting on the only person who has commit access to commit a ton of changes (a few of which are for pod)

Alhadis:
Oooh... right. Indented headings, got it. Alright, well... the correct thing to do would be to actually add a new language entry for it. If it's syntactically distinct from Pod, but still related, we can group it using `language_group: Pod` (so the two languages share the same usage statistics).

Where's this grammar you speak of?

samcv:
This is where the grammar is: https://github.com/MadcapJake/language-perl6fe
[v1.9](https://github.com/MadcapJake/language-perl6fe/pull/43) fixes these:
### Pod
* Pod comments now highlight properly working when there is leading whitespace.
* Make sure pod after `=para` and `=for` immediately stop that block. Make sure formatting doesn't run on further.
* Make sure pod after all other abbreviated forms like `=head` highlight as comments as well.

The documentation on Perl 6 Pod is here: https://design.perl6.org/S26.html
if you are curious.

Alhadis:
Well... I see [I'm not alone with sending people fruit-baskets of bug fixes](https://github.com/atom/language-css/pull/99). :D Which reminds me, I indirectly promised I'd redeem that awful Perl grammar Atom uses, but I've had my hands full with `file-icons`. :c

Alhadis:
Alright, the results of the `.pod6` file search have finished downloading. All-in-all, we're only looking at 52 unique repositories, distributed between 28 unique users... so I'm afraid the usage is too thin to warrant addition as a new language, methinks. I'm a little concerned about how the syntactic differences might skew the classifier, though...

@pchaigno, any input?

samcv:
Here is an example of a file that is in Perl 6 pod not loading because it thinks it is Perl pod:
https://github.com/perl6/specs/blob/master/S15-unicode.pod

There are lots of Perl 6 pod that are not classified properly because of the `.pod` extension.

I think the easiest way to distinguish the two is:
* Perl 6 Pod: has `=begin pod` towards the top and contains `=end pod`
* Perl 5 Pod: has `=pod` and uses `=cut` to end the pod section

Running `perldoc` on a file that had `=begin pod` will show no information at all, since in perl POD that is meant for things like `=begin html` and isn't used in the documentation. 

I have added a change to the heuristics which correctly classifies my added sample and the Perl POD samples. I'm almost totally sure this will not classify any Perl 5 POD incorrectly.

samcv:
@Alhadis let me know what a search using these changes comes up. Thanks!

Alhadis:
Erm, what do you mean? Are you talking about how many results are returned for the `.pod6` extension?

Well, it'll be no different to last time, indexing of recent `.pod6` files notwithstanding.

pchaigno:
> There are lots of Perl 6 pod that are not classified properly because of the `.pod` extension.

Do you mean that Perl 6 pods also use the `.pod` file extension?

> Well Perl 6 pod has different syntax than Perl pod.

Do you have any Perl 6 pod files where the current Pod highlighting doesn't work correctly?

samcv:
@pchaigno yes, I linked one yesterday but here it is again:

https://github.com/perl6/specs/blob/master/S15-unicode.pod

It doesn't even render anything at all. Totally failing.

samcv:
And yeah I meant ones that will match with the new changes. Is that possible? I want to make sure the changes to the heuristics don't improperly flag any Perl 5 Pod. But as I said I am almost completely certain they won't, and err on the side of Perl 5 Pod

pchaigno:
> @pchaigno yes, I linked one yesterday but here it is again:

@samcv Sorry I missed that.

Files on GitHub are rendered (vs. highlighted) using [github/markup](https://github.com/github/markup) and the rendering engines it links to. Markup doesn't seem to support Perl6 pod files yet.

However, it uses Linguist to detect the language of files. Thus, we might be able to make it correctly detect such files as Perl6 pod files. At least, they wouldn't be incorrectly rendered and we could highlight them (display the source code instead of trying to convert it to HTML).

In the meantime, you can use [Linguist overrides](https://github.com/github/linguist#overrides) to disable the rendering (by marking these files as Perl6 for instance). 
<br>

@Alhadis Would you be able to tell how many users and repositories there are for [Perl6 pod files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Apod+%22begin+pod%22+%22end+pod%22&type=Code&ref=searchresults)? (I haven't had the time to test your new scraping script yet :/)

Alhadis:
[103 repositories by 74 users.](https://github.com/Alhadis/Linguist-Silo/blob/misc/pod-results.log)

b2gills:
That list shows a lot of `perltoc.pod` files. (original POD)  
Which will only exist in an installation of Perl 5 because it is a generated file.  
( That is those repositories have a full install of Perl 5 in them, binaries and all )

All of the ones with SPPM-Web are also the original POD

podtreeparser is in POD6  
All of them that have perl6 in the pathname seem to be POD6  
/S26-documentation.pod is POD6  
All of them in perl6/specs that showed up in the list are POD6

pchaigno:
So take that into account, it's now only 53 repositories among 31 users. Unless we missed many Perl6 pod files, I don't think it will be enough to warrant support.
The best course of action here is to use Linguist overrides to mark these files as Perl6. An alternative would be to add support for highlighting Perl6 pod to `Pod::Simple::HTML`, which [github.com seems to be using](https://github.com/github/markup#markups).

b2gills:
[Pod6]: https://design.perl6.org/S26.html
[Pod5]: http://perldoc.perl.org/perlpod.html
[Pod::Simple::HTML]: https://p3rl.org/Pod::Simple::HTML
[Inline::Perl6]: https://p3rl.org/Inline::Perl6

Adding support for [Pod6][] to [Pod::Simple::HTML][] is not going to be worth the effort, as it would be effectivily a complete re-write.  
About the best thing it could do is if it finds something Pod6-like it could try to use [Inline::Perl6][] to defer to a Pod6 parser.

Adding support to a Pod6 parser for Pod5 would be much easier.

When Perl 6 broke compatibility to expand its repertoire, it also did so with Pod.  
So it has such things as nested blocks, or adding configuration options.  
That is Pod6 is mostly a superset of Pos5.

It would also be easier because a Pod6 parser is more likely to be written with the Perl 6 grammar system.

---

One of the main reasons that Pod6 isn't being used that much is missing support in common tools.  
For esample most of the documentation for the only current Perl 6 implementation [Rakudo](https://github.com/rakudo/rakudo), and [the original design docs (superseded)](https://github.com/rakudo/rakudo) are written in a common subset of both Pod5 and Pod6.

GitHub is now a common tool  
almost everything Perl 6 related is on here  
most Perl 5 modules are here

samcv:
I tested the branch in the PR and it says the Perlito and the perltoc.pod are both Pod, not Perl6, so I'm not sure about that. It would probably be fine to mark it as Perl 6 until there is more usage

samcv:
@pchaigno what's the current status of this?

brandonblack:
🏴 Flagging this as stale 🏴

samcv:
Is there anything blocking this being merged? @pchaigno

pchaigno:
The adoption on github.com I'd say (unless @brandonblack or someone else from GitHub staff thinks we should add it anyway...):

> So take that into account, it's now only 53 repositories among 31 users. Unless we missed many Perl6 pod files, I don't think it will be enough to warrant support.
The best course of action here is to use Linguist overrides to mark these files as Perl6.

JJ:
@pchaigno if popularity is the only thing stopping from merging this (plus, I guess, conflicts now), can you please re-check? There's also this github/markup#1173 which could be helped with this. In fact, pod6 files are not recognized as Perl6 now

lildude:
So we find ourselves in a position of two PRs (this one and https://github.com/github/linguist/pull/4083 spawned out of an early comment in this PR) essentially implementing the same thing, but in slightly different ways in terms of how it will appear to the end user.  As this is the oldest PR, I'm going to use this PR to discuss which approach is best to take.

First, lets address the popularity thing... `.pod6` as an extension, is still _not_ sufficiently popular for inclusion. [This search](https://github.com/search?q=extension%3Apod6+begin+pod&type=Code) returns over 1700 repos which on the face of things, suggests this is popular enough, but if you then look at how these are spread out, it's surprising to discover a mere 22 repos hold all of these files.

A [similar search using `.pod`](https://github.com/search?q=extension%3Apod+begin+pod&type=Code) returns waaaay more results and these are definitely spread across hundreds of repos clearly indicating they're Perl 6 related, so there is definitely a need for classification, but not for the `.pod6` extension. 

So whichever option we choose, the `.pod6` extension will need to be removed from that PR - it can always be added later when popularity grows, if it does.

Now we come to how are we to present this to users?

I'm far from a Perl expert, and certainly don't know the differences between Perl 5 and Perl 6 and Pod (5) and Pod 6, but from what I've read between the two PRs, Pod 6 is quite different from Pod (5) and is considered by the Perl 6 interpreter itself as Perl 6 and not something else like Perl 5/Pod (5).

If my understanding is correct, then I think we should go the route of classifying these files as Perl 6 (ie use this PR) rather than "Pod 6" (ie https://github.com/github/linguist/pull/4083)...

... buuuuuuut, and this is an important but ...

Is this what the Perl community as a whole expect, document or even want? 

So sort of written mandate/preference/guideline would definitely help in this matter too.

lildude:
> A [similar search using `.pod`](https://github.com/search?q=extension%3Apod+begin+pod&type=Code) returns waaaay more results and these are definitely spread across hundreds of repos clearly indicating they're Perl 6 related, so there is definitely a need for classification, but not for the `.pod6` extension.

**Update:** I'll need to double-check this ☝️. I've just run my local script and I've come up with only 32 repos. @Alhadis does your harvester return different numbers for the same searches?

The discussion of _how_ we present this to users still stands.

JJ:
I'm not the whole community (obviously), but yes, I think pod 6 *is* perl 6. It is interpreted by the language itself, and it's available as a data structure (`$=pod`) for the program; if you want to interpret an external pod 6 file, you run it through the perl 6 interpreter and access that particular variable, which is not simply the text, but the actual structure composed of [`Pod::Block`s](https://docs.perl6.org/type/Pod::Block). In fact, #4066 (my very own PR) took that particular approach, although now references to Perl 6 have been removed in favor of accepting this one.

Alhadis:
> @Alhadis does your harvester return different numbers for the same searches?

😂 I dig that I spent hours polishing formatting, wording, and structure so Harvester's documentation was as clear and concise as possible... and the next user who sees it goes *"Alhadis, can you run this for me?"* 😢 Really breaks my heart as a Linguist contributor... I don't think this relationship is working out. @pchaigno, our marriage is off.

Kidding, obvs. You'll need to wait for this thing to finish. Speed is throttled because loading anything too quickly causes your server to bark at me and tell me I'm abusing the site. All that hate when I'm just trying to help you guys. 💔

JJ:
2018-04-09 16:54 GMT+02:00 John Gardner <notifications@github.com>:

> @Alhadis <https://github.com/Alhadis> does your harvester return
> different numbers for the same searches?
>
> 😂 I dig that I spent hours polishing formatting, wording, and structure
> so Harvester's documentation was as clear and concise as possible... and
> the next user who sees it goes *"Alhadis, can you run this for me?"* 😢
> Really breaks my heart as a Linguist contributor... I don't think this
> relationship is working out. @pchaigno <https://github.com/pchaigno>, our
> marriage is off.
>

:-) :-) Happens all the time.

> Kidding, obvs. You'll need to wait for this thing to finish. Speed is
> throttled because loading anything too quickly causes your server to bark
> at me and tell me I'm abusing the site. All that hate when I'm just trying
> to help you guys. 💔
>

Anyway, I think this is really relevant if we are going to keep Pod 6 and
Perl 6 separate (which we should not). As a matter of fact, pod is right
now like 99 to 1 in relation with pod6, but I seem to remember @alhadis
mentioned a way to make them apart in, well, the other PR (and if we keep
jumping from one PR to another and _not_ paying attention to my very own
PR, _my_ marriage to you both is off)

lildude:
> 😂 I dig that I spent hours polishing formatting, wording, and structure so Harvester's documentation was as clear and concise as possible... and the next user who sees it goes "Alhadis, can you run this for me?" 😢 Really breaks my heart as a Linguist contributor... I don't think this relationship is working out. @pchaigno, our marriage is off.

I'm 💔 that you mixed me up with your other bit-on-the-side 😆 🤣.

I was just being lazy as I've seen but not looked at your repo and code. I do really appreciate you taking the time to make it public and document it. I've got a bit of time now so will take a peek. 🙇

Alhadis:
Ugh, sorry that took so long. 😞 Results for [`extension:pod begin pod` are in](https://github.com/Alhadis/Linguist-Silo/tree/pod) @lildude, I haven't deduped any of the files as I normally would. Will leave it to you to decide how to sort them.

Kicking off a search for `extension:pod6 begin pod` right now...

Alhadis:
@lildude Alright, the `.pod6` files are [uploaded here](https://github.com/Alhadis/Linguist-Silo/tree/pod6). Let me know if there's anything else you need. =)

nige123:
Personally I think locking in anything with "6" in the name is not a good move in the long term. I would suggest avoiding "pod6" and using something different - IMHO something like ".dod", ".dop" or ".zod" - would be better.

b2gills:
To explain the situation

Let's say we have a file that is made up of both Perl5 and POD. When it has an extension of `.pod` or `.POD` it should be seen as POD, and otherwise be seen as Perl5.

Similarly a Perl6/POD6 file should be seen as POD6 if it has an extension of `.pod` `.pod6` `.POD` `.POD6`, and otherwise as Perl6.

Ideally it would be nice to view a Perl5/Perl6 file also as POD/POD6. I can see why GitHub wouldn't support that, as most programming languages don't have a full fledged documentation format embedded in them. (POD has been used for writing a book)

---

Ignoring the other extensions, it would be nice if there was some way to always distinguish between POD and POD6 as there can be large semantic differences.

If there was something that is always in POD, and never in POD6, or vice versa, it could be used to determine which to parse it as. (No such luck)

The problem is 99% of POD syntax is valid POD6, and many files could be successfully parsed as either but again with major semantic differences.

A way to fairly reliably guess is to assume neither version and build up a heuristic by looking at which features are used. There are some features in POD that aren't in POD6, and many in POD6 that aren't in POD. If the file starts with `=pod` or has `=cut` it is definitely POD, if it starts with `=begin pod` it is probably POD6. If there is Perl code outside (or inside) of the POD it could also be added to the heuristic. If you see something like `=for table  :caption<Table of Contents>` it is POD6. If you see an indented `=foobar` it is probably POD6.

A less reliable way is to assume POD and switch to POD6 if there is a problem parsing it. (Note that a POD6 file starting with `=begin pod` may successfully result in an empty document if POD is used.)

---

This is the reason it was requested that `.pod6` be added. There would be no confusion with a file that had it as an extension.

Grinnz:
As Perl and POD are currently classified separately, I think it would make sense for Perl6 and POD6 to be classified separately. But I have no stake in this other than to make sure that Perl POD files are still correctly classified and rendered. To that end, I just want to make clear that `=pod` and `=cut` are not required to be present in a `.pod` file for it to be valid Perl 5 POD (or even in pod embedded in a Perl file) -- any directive starting with `=` will activate the POD parser. I seem to recall that github may not render `.pod` files without a `=pod` present so that might be a different issue.

JJ:
2018-04-10 1:34 GMT+02:00 Dan Book <notifications@github.com>:

> As Perl and POD are currently classified separately, I think it would make
> sense for Perl6 and POD6 to be classified separately. But I have no stake
> in this other than to make sure that Perl POD files are still correctly
> classified and rendered. To that end, I just want to make clear that =pod
> and =cut are not required to be present in a .pod file for it to be valid
> Perl 5 POD (or even in pod embedded in a Perl file) -- any directive
> starting with = will activate the POD parser. I seem to recall that
> github may not render .pod files without a =pod present so that might be
> a different issue.
>

As I have explained several times, they are a different kettle of fish.
Adding to the fact that Pod 6 is interpreted by the Perl 6 interpreter
itself is the fact that it can include Perl 6 as a part of it, for instance
as configuration information
https://docs.perl6.org/language/pod#Configuration_information

Grinnz:
Yes, but the question is, if it's recognized as a Perl6 file and not a separate type, will github render it as POD6, while still leaving regular Perl6 files just source highlighted?

JJ:
That's again a completely different kettle of fish. In fact, there's this
PR waiting in perl6/markup https://github.com/github/markup/pull/1173 which
in fact recognizes Pod 6 as Perl 6

samcv:
Well the totally ideal situation would probably be rendering Pod6 files like how we do Pod. And only rendering it as html if it is named .pod or .pod6. I think if the end goal is to eventually render Pod6 as html, it should probably be its own filetype. Even though Pod6 *is* valid Perl6, if it is named as .pod or .pod6 the creator is designating that this is *documentation* and not code. So even though the Perl 6 interpreter is used to render Pod6, that doesn't mean .pod and .pod6 shouldn't be considered documentation.

lildude:
Much appreciated @Alhadis. Whilst your figures don't match mine, they are very similar and confirm `.pod6` as an extension is definitely not popular enough yet.

@samcv you make a very good point and I think you've got to the crux of the matter: this is really about code vs prose. Each are treated differently by Linguist. Code counts towards the repo language stats, prose does _not_. Accordingly Perl repos with `.pod` files will not show `Pod` in the language stats and a repo containing _only_ Pod files will not show Perl or Pod in the language stats. 

I think we should do the same for Pod 6 thus making https://github.com/github/linguist/pull/4083 the preferred PR of the two.

pchaigno:
> I don't think this relationship is working out. @pchaigno, our marriage is off.

I'm keeping the house, you take the boat.

lildude:
Closing as we've gone for the approach taken in https://github.com/github/linguist/pull/4083 which has just been merged. Thanks everyone. 🙇

PR #3367, opened 2016-12-07
Add .php_cs and .php_cs.dist

zonuexe:
`.php_cs` and `.php_cs.dist` is [PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) configuration file in PHP.

GitHub search result: https://github.com/search?utf8=%E2%9C%93&q=filename%3A.php_cs&type=Code&ref=searchresults


samples/PHP/.php_cs
```diff
@@ -0,0 +1,37 @@
+<?php
```
pchaigno: You'll need to put these in the `filenames/` subdirectory.

Alhadis:
Looks good to me. Certainly enough usage, that's for sure.

PR #3372, opened 2016-12-08
Fix the R sample that is incompatible with roxygen2.

yutannihilation:
I got the error bellow when try to generate R document from this sample file: https://github.com/github/linguist/blob/master/samples/R/import.r#L195-L196

```r
@TODO [hello.R#196]: unknown tag 
```

In R, lines that starts with `#'` are for [roxygen2](https://cran.r-project.org/web/packages/roxygen2/index.html). So we should use `#` for usual code comments.

### DIFF
--- a/samples/R/import.r
+++ b/samples/R/import.r
@@ -192,8 +192,8 @@ reload = function (module) {
     module_path = module_path(module)
     module_name = module_name(module)
     rm(list = module_path, envir = .loaded_modules)
-    #' @TODO Once we have `attach`, need also to take care of the search path
-    #' and whatnot.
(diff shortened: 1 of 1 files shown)


pchaigno:
I added the sample in #1142 from https://github.com/klmr/modules.
It's okay if sample files have a few errors. They are here to represent the in-the-wild usage of the language, **with** the occasional syntactic errors. We use then to train the Bayesian classifier and for our own tests. Syntactic errors are something we should be able to handle.

yutannihilation:
Thanks for quick response! I got it 👍

PR #3373, opened 2016-12-08
Ignore roxygen2 generated files

yutannihilation:
Disclaimer: I'm a newbie to Ruby.

In R, `roxygen2` is a de facto standard of document generation. The generated documents are typically starts with this line:

```r
% Generated by roxygen2: do not edit by hand
```

For a real example, please see this one: https://github.com/tidyverse/ggplot2/blob/f6f9f9de41c48382c70cbccc253db198e3cdc128/man/layer.Rd#L1

This pull request make linguist to ignore these generated documents.


pchaigno:
It's orthogonal to this pull request, but it looks like `.Rd` files do not actually have a R syntax. Is that TeX?

yutannihilation:
Wow, thanks!

Yes `.Rd` files are not R files, but a kind of TeX. They are purely used for documentation purpose. As this syntax is very different from R syntax and we (I believe I'm not the only one...), R people, are too lazy to learn this, it is common practice to auto-generate `.Rd` file from comments embedded in `R` files.

For example, [this](https://github.com/github/linguist/blob/master/samples/R/import.r):

```r
#' Import a module into the current scope
#'
#' \code{module = import('module')} imports a specified module and makes its
#' code available via the environment-like object it returns.
#'
#' @param module an identifier specifying the full module path
#' @param attach if \code{TRUE}, attach the newly loaded module to the object
#'      search path (see \code{Details})
#' @param attach_operators if \code{TRUE}, attach operators of module to the
#'      object search path, even if \code{attach} is \code{FALSE}
#' @return the loaded module environment (invisible)
...snip...
```

will be converted into [this](https://github.com/yutannihilation/linguist/blob/0aebcd1572f0e82b684ea8751c0ca06e2ed2180e/samples/R/import.Rd) by roxygen2:

```tex
% Generated by roxygen2: do not edit by hand
% Please edit documentation in R/hello.R
\name{import}
\alias{import}
\title{Import a module into the current scope}
\usage{
import(module, attach, attach_operators = TRUE)
}
\arguments{
\item{module}{an identifier specifying the full module path}

\item{attach}{if \code{TRUE}, attach the newly loaded module to the object
search path (see \code{Details})}

\item{attach_operators}{if \code{TRUE}, attach operators of module to the
object search path, even if \code{attach} is \code{FALSE}}
}
\value{
the loaded module environment (invisible)
}
```

So, the diffs of the documentation part will be shown both in `.R` file and the corresponding `.Rd` file redundantly. I believe we can safely ignore `.Rd` :)

PR #3383, opened 2016-12-14
update languages.yml

simongarland:
add definitions for languages k an q

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2037,6 +2037,17 @@ Jupyter Notebook:
   aliases:
   - IPython Notebook
   language_id: 185
+k:
+  type: programming
+  color: "#353C41"
+  extensions:
+  - ".k"
+  interpreters:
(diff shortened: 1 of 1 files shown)


brandonblack:
👋  @simongarland thanks for your contribution. 

I'm going to close these q and k related PRs so that you can reopen them and [follow our contribution guidelines for adding support for a new language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language). 

Ideally, we prefer your create one PR per language you're adding. Also please make sure to include search results for these languages so we can evaluate if its ready to be added to linguist.

>We try only to add languages once they have some usage on GitHub. In most cases we prefer that each new file extension be in use in hundreds of repositories before supporting them in Linguist.

simongarland:
Brandon,

thanks, yes, I’ll resubmit shortly

simongarland:
resubmitted, I hope it’s ok this time

PR #3386, opened 2016-12-14
Update languages.yml

simongarland:
add definitions for languages k and q, fix travis complaints 
(lowercase is common usage)
sample k and q code in sample/k and sample/q 
existing repositories see kxsystems.github.io 

manually added language_ids 422 and 423 (next two free) as the set-.. script crashes for me


brandonblack:
https://github.com/github/linguist/pull/3383#issuecomment-267115696

PR #3388, opened 2016-12-14
Release v5.0.0

brandonblack:
In this release:
- Language Renaming 🎉 
- Removing Deprecated Code Paths
- Separation of Detection by Filename vs. Extension
- Updated Grammars

#
CC: @github/linguist

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2894,7 +2894,7 @@ OpenSCAD:
   - ".scad"
   tm_scope: none
   ace_mode: scad
-  language_id: 431
+  language_id: 266
 OpenType Feature File:
   type: data
   aliases:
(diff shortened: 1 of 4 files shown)


lib/linguist/languages.yml
```diff
@@ -2903,7 +2903,7 @@ OpenType Feature File:
   - ".fea"
   tm_scope: source.opentype
   ace_mode: text
-  language_id: 266
+  language_id: 374317347
```
TwP: :+1: is this the hash generated language ID?

brandonblack: That's correct. This is set automagically by `script/set-language-ids` and its the result of hashing the name.

brandonblack: > 4. Add a language_id for your language using script/set-language-ids. You should only ever need to run script/set-language-ids --update. Anything other than this risks breaking GitHub search.

https://github.com/github/linguist/blob/faaa4470af8ef0ab7c5c53dc2231f5d9a0233f74/CONTRIBUTING.md#adding-a-language

PR #3396, opened 2016-12-19
Add Genie programming language

astavale:
Genie was introduced in 2008 as part of the GNOME project:
https://wiki.gnome.org/Projects/Genie

It is a programming language that uses the Vala compiler to
produce native binaries. It has good bindings to C libraries
especially those that are part of the GObject world such as
Gtk+3 and GStreamer

At present the .gs extension on GitHub shows as Javascript or Gosu.
A GitHub search shows confusing results:
https://github.com/search?p=15&q=extension%3Ags+-language%3Ajavascript+-language%3Agosu+NOT+nothack&ref=searchresults&type=Code&utf8=%E2%9C%93

For some example projects see:
https://github.com/darkoverlordofdata/shmupwarz-vala
https://github.com/tliron/khovsgol/

The shmupwarz-vala project adds a .gitattributes file containing:
*.gs linguist-language=Genie
This pull request should at least enable Genie to be shown when projects use a relevant .gitattributes file.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1389,7 +1389,7 @@ Genie:
   ace_mode: text
   extensions:
   - ".gs"
-  color: "#fba53d"
+  color: "#fb855d"
   tm_scope: none
   language_id: 792408528
 Genshi:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1384,6 +1384,14 @@ Game Maker Language:
   codemirror_mode: clike
   codemirror_mime_type: text/x-c++src
   language_id: 125
+Genie:
+  type: programming
+  ace_mode: text
+  extensions:
+  - ".gs"
+  color: "#fba53d"
+  tm_scope: none
+  language_id: 792408528
 Genshi:
   type: programming
   extensions:
--- a/samples/Genie/Class.gs
+++ b/samples/Genie/Class.gs
@@ -1,0 +1,12 @@
+init
+	new Demo( "Demonstration class" ).run()
+
+class Demo
+	_message:string = ""
+
+	construct ( message:string = "Optional argument - no message passed in constructor" )
+		_message = message
+
+	def run()
+		print( _message )
+		
--- a/samples/Genie/Hello.gs
+++ b/samples/Genie/Hello.gs
@@ -1,0 +1,2 @@
+init
+	print( "Hello, World!" )


pchaigno:
Thanks for the pull request!

Are there keywords, specific to Genie, that we could use for the search on github.com?

astavale:
> Are there keywords, specific to Genie, that we could use for the search on github.com?

Genie has significant white space and other similarities with Python. By default Genie uses tabs for indentation but this can be changed to a given number of spaces, so a number of projects will use the [indent] attribute before code in the file, e.g.:
`[indent = 4]`
or 
`[indent=2]`
This is optional though.

The `main ()` function in Genie is `init`, which is probably unique to Genie. For programs an `init` function is required, but not for libraries. So looking for `init` or `init:int` followed by indentation may identify a Genie program. For example:

```
init
    print( "simple program" )
```

Functions and the no-op keyword are similar to Python:
```
def my_function ()
    pass
```
but Genie is strongly typed so function parameters must also specify the type. If there is a return value then the return type must also be given:
```
def my_function( a:int ):string
    return a.to_string()
```
Functions can also have type parameters:
```
def my_function( a:T ) of T
    pass
```

Similarly with classes. The Python class keyword is used. Genie can also inherit and implement interfaces:
```
class Test:ParentClass implements TestInterface
    def my_method()
        pass
```

Namespaces are brought in to scope using the `uses` keyword. The identifier can be on the same line:
```
uses Gtk
```
or as an indentation block:
```
uses
    Gtk
    Gee
    Sqlite
```
These are common constructs and keywords so I hope that's enough of a sample to give you some ideas.

pchaigno:
The following search query returns [more than 10k files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ags+indent+OR+def+OR+implements+OR+uses+NOT+construct&type=Code&ref=searchresults). It didn't seem to return to many false positives (non Genie files). I checked the first and last page of results for each sort option but I'm not particularly familiar with either Gosu or Genie.

Am I correct to assume that `construct` is not a Genie keyword?

astavale:
`construct` is a keyword used as the constructor in classes:
```
class Example
    _backing_field:string
    construct( a:string = "default value" )
        _backing_field = a
```
After a look at the first couple of results pages from your query one thing does stand out. `package` is not a Genie keyword.

pchaigno:
Great! That helps. I updated [the search query](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ags+indent+OR+def+OR+uses+OR+construct+NOT+var+NOT+package&type=Code&ref=searchresults).
I'm guessing `var` is not a Genie keyword...?

astavale:
`var` is a keyword, It is used for type inference, e.g. `var a = "a string"`

If it's any help keywords can be found from the Genie tokens file:
https://git.gnome.org/browse/vala/tree/vala/valagenietokentype.vala

From the new query some result could be eliminated because the keywords are in comments, e.g. 
```
 *     format) used to construct the inline citation text for insertion.
```
Also the keywords should be surrounded by white space, so that would eliminate examples like:
```
var active_uses = up.getProperty(USER_PROP_AUTOGRADE_ACTIVE_USES);
if (active_uses == null)
```

Also Genie won't have braces, {}, appearing immediately after `construct ()` whereas Gosu appears to. `construct ()` in Genie won't have a return type, so it will always have a white space indentation on the next line.

pchaigno:
*Zut !* With `var`, I'm getting [many JavaScript false positives](https://github.com/search?p=99&q=extension%3Ags+indent+OR+def+OR+uses+OR+construct+NOT+package&ref=searchresults&type=Code&utf8=%E2%9C%93). Looks like `function` [could be used to remove those](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ags+indent+OR+def+OR+uses+OR+construct+NOT+package+NOT+function&type=Code&ref=searchresults) (~900 results).

Unfortunately, github.com's search engine doesn't allow us to use any special character, including braces, white spaces, or asterisks.

astavale:
Looks mostly OK with the latest query. I get the first negative result on page 46 of 93. Later on some negative results could be excluded which have `class!` instead of `class`. Also `def...end`. Genie doesn't have an `end` keyword, but it looks as though GitHub search is limited to five AND/OR/NOT operators.

PR #3397, opened 2016-12-20
add stat.q to sample/q

simongarland:
example source file for language q


lildude:
Looks like this may be related to https://github.com/github/linguist/pull/3398. Closing this PR in favour of you completing that PR which is where this sample file should be added.

PR #3398, opened 2016-12-20
add language q to languages.yml

simongarland:
sample search for in the wild repositories:
user:AquaQAnalytics extension:q

more examples see:
http://kxsystems.github.io/

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4856,6 +4856,16 @@ ooc:
   - ".ooc"
   ace_mode: text
   language_id: 418
+q:
+  type: programming
+  color: "#0000CD"
+  extensions:
+  - ".q"
+  interpreters:
+  - q
+  tm_scope: none
+  ace_mode: text
+  language_id: 970539067
 reStructuredText:
   type: prose
   wrap: true


lildude:
Marking as stale as this appears to have been WIP since inception.

If you're like to continue this PR, please review [Adding a Language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language).

simongarland:
ok, yes, it’s been hanging for a while - sorry about that

I’ve read the adding a language doc, but it seems very very complicated to
add a language that isn’t in
the whole textmate grammar compatible world. Is it possible for me to ask a
couple of questions of someone
on the linguist team to sort out what a minimum profile would be to get the
language added?

Alhadis:
Hey @simongarland,

For a language to be formally recognised on GitHub, it needs to meet the usage criteria. I'll highlight the relevant snippet:

> We try only to add new extensions once they have some usage on GitHub. In most cases **we prefer that extensions be in use in hundreds of repositories** before supporting them in Linguist.

Part of adding a new language is proving that it's being used by enough users for it to warrant official recognition.

simongarland:
yep, the hundreds of repositories isn’t the problem - we can drag that up
for you

it’s the requirement for a textmate grammar/syntax file, as these languages
are in the vim/emacs corner
and nobody’s needed to get these things set up and blessed/included by the
various editor vendors

Alhadis:
Those are optional. There's no requirement to include a language-grammar. =) Languages can still gain formal recognition on GitHub without highlighting.

simongarland:
ah great, then I must have misunderstood / mis-submitted the request

I’ll retry and let you know when it’s done - if you’d have the time to
quickly look over
what I’ve done in case it’s easily corrected that’d be very much
appreciated.

simongarland:
so, just to check that I understand what a minimal submission includes,
please correct if I still haven’t got it:

1) I need to submit an amended <languages.yml> with an entry for Kdb:

 Kdb:
   type: programming
   color: "#3F3F3F"
   extensions:
   - ".q"
   - ".k"
   interpreters:
   - q
   tm_scope: none
   ace_mode: text
   language_id: NNN

where I generate the NNN and fill it in using set-language-ids

2) I should add a couple of samples of .q and .k files

3) I should point you at a big enough list of existing repositories that
you can accept that there is a
community of users who’d benefit from having <Kdb> recognised as a language

that’s it..

lildude:
@simongarland Yup, that's pretty much it.  You may also want to substantiate your choice of colour too just for clarity.

simongarland:
should be all ready now ..

________________________________

PR #3402, opened 2016-12-24
Add XCompose language and highlighter

samcv:
Hopefully this is considered a language? I find 39k results searching for files which are XCompose:

https://github.com/search?p=10&q=%3CMulti_key%3E&ref=searchresults&type=Code&utf8=%E2%9C%93

Not all of them would catch in this filter, but most are named so they would match.


pchaigno:
The search engine doesn't consider `<` and `>`, so this is actually searching for `Multi_key` only. What are the usual file extensions for this language? Are there other keywords we could use?

It looks like we could add `Compose`, `.XCompose`, and `XCompose` as filenames for this language.

samcv:
Yeah, I've already added all of those extensions/filenames
https://github.com/github/linguist/pull/3402/files#diff-3552b1a64ad2071983c4d91349075c75R4590

pchaigno:
Hm, right. Don't know why I though this was an issue, even though the two commits were starring right at me :/

samcv:
I should probably remove xmodmap though. Not all of them are XCompose type (about split between that and another config type). Or I could add heuristics, but there are not enough of them to matter I think.

pchaigno:
> Not all of them are XCompose type (about split between that and another config type).

What's the other config type? Any way to detect it?

> Or I could add heuristics, but there are not enough of them to matter I think.

I found only about [767 `.xmodmap` files](https://github.com/search?p=1&q=extension%3A.xmodmap+NOT+dhvdhgdkfn&ref=searchresults&type=Code&utf8=%E2%9C%93) so if half of them aren't actually XCompose, I'm afraid it won't be enough.

I found only [33 `.XCompose` files](https://github.com/search?utf8=%E2%9C%93&q=extension%3A.xcompose+NOT+dhvdhgdkfn&type=Code&ref=searchresults). That won't be enough to add this extension.

I found [186 files with name `.XCompose`](https://github.com/search?p=19&q=filename%3A.xcompose+NOT+dhvdhgdkfn&ref=searchresults&type=Code&utf8=%E2%9C%93) and [160 files with name `XCompose`](https://github.com/search?p=33&q=filename%3A.xcompose+NOT+dhvdhgdkfn&ref=searchresults&type=Code&utf8=%E2%9C%93), but they're well distributed among users and repositories (1 per repository) and the risk of conflict is very low, so that might be enough to add it as a filename.

lib/linguist/languages.yml
```diff
@@ -4580,6 +4580,19 @@ XC:
...
   codemirror_mime_type: text/x-csrc
   language_id: 398
+XCompose:
+  type: data
+  extensions:
+  - ".XCompose"
+  - ".xcompose"
+  filenames:
+  - ".XCompose"
+  - ".XCompose.In"
```
pchaigno: I could find any file with this filename: https://github.com/search?utf8=%E2%9C%93&q=filename%3A.xcompose.in+NOT+dhvdhgdkfn&type=Code&ref=searchresults

samples/XCompose/_.XCompose
```diff
@@ -0,0 +1,1297 @@
+# for Emacs: -*- coding: utf-8 -*-
```
pchaigno: You'll need to remove this sample file and add one for filenames (in `samples/XCompose/filenames/`).

samcv:
@pchaigno done

PR #3403, opened 2016-12-24
Added C sample

DavidAylaian:
The file was incorrectly being displayed as C++ when it is actually written in C.


DavidAylaian:
I decided to not agree to the Github CLA.

PR #3416, opened 2017-01-04
Add Cask file as Emacs Lisp

zonuexe:
`Cask` file is Emacs [Cask](https://github.com/cask/cask) configuration file.
The file is included in [user initialization file](https://www.gnu.org/software/emacs/manual/html_node/emacs/Init-File.html) directory (`.emacs.d`) and [Emacs Lisp package](https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging.html#Packaging).

GitHub search result: https://github.com/search?utf8=%E2%9C%93&q=filename%3ACask&type=Code&ref=searchresults

### DIFF
--- a/samples/Emacs Lisp/filenames/Cask
+++ b/samples/Emacs Lisp/filenames/Cask
@@ -1,8 +1,9 @@
-(package "cask-example" "1.2.3" "Example package for Emacs Cask")
-(source gnu)
+(package "composer" "0.0.7" "Interface to PHP Composer")
 (source "melpa" "https://melpa.org/packages/")
 
-(package-file "cask-example.el")
+(package-file "composer.el")
 
-(development
- (depends-on "dash"))
+(depends-on "f")
+(depends-on "s")
+(depends-on "request")
+(depends-on "seq")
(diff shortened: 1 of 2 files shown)


samples/Emacs Lisp/filenames/Cask
```diff
@@ -0,0 +1,8 @@
+(package "cask-example" "1.2.3" "Example package for Emacs Cask")
```
pchaigno: Because sample files are used for training the Bayesian classifier and for tests, we prefer to extract them from real applications (with the occasional inconsistencies, etc.). Do you know of such a file you could use (under permissive license)?

zonuexe: @pchaigno I can provide [my project Cask file](https://github.com/zonuexe/composer.el/blob/master/Cask).
(I avoided to use [original project Cask file](https://github.com/cask/cask/blob/master/Cask) that is licensed under GPLv3.)

pchaigno: That'd be great!

PR #3423, opened 2017-01-09
Removing references to @arfon

arfon:
👋  folks. It looks like things are going pretty well over here so I'm just updating the README to reflect reality 😄


arfon:
❤️  godspeed y'all.

PR #3424, opened 2017-01-10
Allow for split(",") returning nil

kivikakk:
If you try to look up a non-existent language, you get `nil`. Unless you look up the string `","`. Then you get an exception.

(Fixes it so you don't get an exception.)


kivikakk:
@Alhadis: thank you! 🙇‍♀️

Alhadis:
Pole tanu vaart! ;)

Issue #3430, opened 2017-01-20
could it take yarn.lock as generated?


sunderls:
yarn.lock is really very large, maybe a good idea to suppress the diff, by checking filename?

example: https://github.com/yarnpkg/yarn/blob/867cf4f2e4fefee5b682b1b61fc360502834d70c/yarn.lock

pchaigno:
Looks like it's a generated file. You could add it to [`generated.rb`](https://github.com/github/linguist/blob/master/lib/linguist/generated.rb); it will be recognized as generated and the diff will be hidden.

pchaigno:
Never mind, that's what you did in #3431. Closing in favor of #3431 :smile:

sunderls:
@pchaigno ah yes,  thank you~

PR #3431, opened 2017-01-20
add yarn.lock in generated detection

sunderls:
https://github.com/github/linguist/issues/3430


pchaigno:
I'd prefer to match on [the first line, which identifies the file as generated](https://github.com/yarnpkg/yarn/blob/867cf4f2e4fefee5b682b1b61fc360502834d70c/yarn.lock#L1).

sunderls:
sorry for the failure, I'll repull

PR #3438, opened 2017-01-24
Be able to configure the number of threads for git submodules update

javierhonduco:
This PR enables, via the `LINGUIST_SUBMODULE_THREADS ` env variable the number of threads that will be used for the submodule update operation.

This will let users to set this value at will in case they have beefier (or less powerful) machines.


pchaigno:
Shouldn't this be an argument to the script instead of an environment variable?

javierhonduco:
I'm ok with that change! :)

javierhonduco:
Thanks, folks! :tada:

PR #3444, opened 2017-01-31
Fixing null reference in yarn.lock check

brandonblack:
This check for generated yarn.lock files lacked an adequate guard.

#
Resolves: github/github#68104
CC: @mikrobi


mikrobi:
Awesome thanks for fixing that! 🎉

Issue #3451, opened 2017-02-02
Markdown is mistaken for GCC Machine Description


HarryBeadle:
As shown here: 
https://github.com/Microsoft/azure-docs/search?l=gcc-machine-description

As an additional note, for Markdown-heavy repos shouldn't linguist start displaying Markdown as a language?

larsbrinkhoff:
I suppose this is my responsibility, because I added the GCC Machine Description language.

From the search results, I'll pick one sample at random:

    > [!div class="op_single_selector" title1="Target environments:"]
    > * [Ingest data overview](../articles/machine-learning/machine-learning-data-science-ingest-data.md)
    > * [Azure Blob Storage](../articles/machine-learning/machine-learning-data-science-move-azure-blob.md)
    > * [SQL Server on Azure VM](../articles/machine-learning/machine-learning-data-science-move-sql-server-virtual-machine.md)
    > * [Azure SQL database](../articles/machine-learning/machine-learning-data-science-move-sql-azure.md)
    > * [HDInsight(Hadoop on Azure)](../articles/machine-learning/machine-learning-data-science-move-hive-tables.md)

Indeed this doesn't match the heuristic regexp.  I see no problem adding `>` to the list of characters that can start a markdown line.

larsbrinkhoff:
> As an additional note, for Markdown-heavy repos shouldn't linguist start displaying Markdown as a language?

This is by design.  The primary repository language must be a programming or marup langauge, but Markdown is classified as "prose".

larsbrinkhoff:
@HarryBeadle I'm trying to fix the heuristic, but it would also be good to add more sample files to Linguist.  They would help when the heuristic doesn't know what to do.

larsbrinkhoff:
Two of the problematic files look very much like scripts:
- includes/site-recovery-sccm-linux-script.md
- includes/site-recovery-sccm-windows-script.md

I'm not sure we can or should make Linguist handle those.  There is still the option to add a `linguist-language` directive to `.gitattributes`.

wopian:
Also noticed the Apache [LICENSE.md][0] file in one of my old repositories is also [mistaken for GCC][1]

[0]:https://github.com/wopian/skylark/blob/master/LICENSE.md
[1]:https://github.com/wopian/skylark/search?l=GCC+Machine+Description

HarryBeadle:
@wopian Looks like that should be fixed in @larsbrinkhoff's latest pull request.

[...]

PR #3453, opened 2017-02-03
Accomodate Markdown lines which begin with '`' and '"'.

larsbrinkhoff:
With this fix in place, all but two files in #3451 are properly detected as Markdown.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -260,7 +260,7 @@ module Linguist
     end
 
     disambiguate ".md" do |data|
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
@Alhadis @brandonblack @lildude @lizzhale @mikemcquaid @pchaigno  
Pingeroo.

pchaigno:
Should there be a few sample Markdown files to test this?

larsbrinkhoff:
Yes, that would be good!  I'm busy, so I'm not going to search for samples at this time.

PR #3454, opened 2017-02-03
Add .lkml to LookML

stoe:
Add `.lkml` extension to LookML syntax added in https://github.com/github/linguist/pull/1408

- https://github.com/llooker/okcupidtrial/blob/master/profile.view.lkml
- [A search for `extension:lkmll` on GitHub](https://github.com/search?o=desc&q=extension%3Alkml+NOT+nothack)


Alhadis:
Summary of [search results](https://github.com/Alhadis/Linguist-Silo/blob/misc/lkml-results.log) if you need it:

**Repositories:** 111
**Users:** 74

pchaigno:
Should we restrict the extensions to `.view.lkml` and `.model.lkml`? I'm not sure there's any risk of false positives with `.lkml`, but if `.view.lkml` and `.model.lkml` are the only two extensions in use, then it's probably worth it...

Alhadis:
Looks good. Just need to fix the order of extensions to please CI:

~~~
  1) Failure:
TestPedantic#test_nonprimary_extensions_are_sorted [/home/travis/build/github/linguist/test/test_pedantic.rb:15]:
.view.lkml should come after .model.lkml
~~~

PR #3457, opened 2017-02-03
Rakefile and irbrc ruby matching (and remove Gemfile.lock)

Dorian:
Interestingly it has tests to match `Rakefile` and has a `Rakefile` sample but it's not in the ruby's filenames list. And `.irbrc` (like `.pryrc`) is better suited as a filename than a file extension

- Examples of `.irbrc`: https://github.com/search?q=irbrc&ref=searchresults&type=Commits&utf8=%E2%9C%93
- Examples of `.pryrc`: https://github.com/search?utf8=%E2%9C%93&q=pryrc&type=Commits&ref=searchresults
- Gemfile.lock has its own format https://github.com/search?utf8=%E2%9C%93&q=Gemfile.lock&type=Commits&ref=searchresults

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3707,7 +3707,6 @@ Ruby:
   - ".fcgi"
   - ".gemspec"
   - ".god"
-  - ".irbrc"
   - ".jbuilder"
   - ".mspec"
   - ".pluginspec"
@@ -3729,6 +3728,7 @@ Ruby:
   - jruby
   - rbx
   filenames:
+  - ".irbrc"
   - ".pryrc"
   - Appraisals
   - Berksfile
@@ -3744,6 +3744,7 @@ Ruby:
   - Mavenfile
   - Podfile
   - Puppetfile
+  - Rakefile
   - Snapfile
   - Thorfile
   - Vagrantfile
--- a/samples/Ruby/filenames/.irbrc
+++ b/samples/Ruby/filenames/.irbrc
@@ -1,0 +1,2 @@
+require "pp"
+IRB.conf[:AUTO_INDENT] = true


Dorian:
@arfon @pchaigno What do you think?

lib/linguist/languages.yml
```diff
@@ -3738,12 +3738,12 @@ Ruby:
   - Deliverfile
   - Fastfile
   - Gemfile
-  - Gemfile.lock
```
brandonblack: I think it makes sense to leave Gemfile.lock alone here, but otherwise I'd ✅  the rest of these changes related to `Rakefile` and `.irbrc`.

Dorian: Why? It's not a ruby file? It's obviously gonna be part of a ruby project, but the file itself is not a ruby file (I used this list for rubocop).

I guess it's good for statistics about the repo but not for actual ruby files detection?

(See `Gemfile.lock` file search: https://github.com/search?utf8=%E2%9C%93&q=Gemfile.lock&type=Commits&ref=searchresults)

brandonblack: `Gemfile.lock` isn't it's own format, that's an incorrect characterization. The lockfile is a generated file type that's an artifact of the `Gemfile` which is definitely Ruby (or at least it can be). It doesn't make any sense at all to classify these two independently from each other.

I'll 👍 the rest of this PR if you pull out this change.

Dorian: @brandonblack I don't follow your reasoning, in `File.write('hello.txt', 'Hello World')`, `hello.txt` is not a ruby file, files in `tmp/` or `log/` are not ruby files, `config/secrets.yml` is not a ruby, but all are associated with a rails project for instance.

e.g. in my use cases I use linguist as `Dir['*'].select { |file| is_ruby?(file) }.map { "do stuff on the ruby files" }` so I would like to avoid non-ruby files (e.g. `ruby Gemfile.lock` is a `syntax error` basically)

PR #3459, opened 2017-02-06
add support for detecting bazel WORKSPACE files

softprops:
this adds support for detecting [Bazel](https://bazel.build/) [WORKSPACE](https://bazel.build/versions/master/docs/getting-started.html#using-a-workspace) files which should be parsed as [python](https://bazel.build/versions/master/docs/build-ref.html#core_build_language). Not bazel also defines BUILD files which seem to already be covered. 

Here is an example [search](https://github.com/search?utf8=%E2%9C%93&q=filename%3AWORKSPACE+not+python&type=Code&ref=searchresults) I'm not sure if it's 100% correct way to search but the results seem to bring up the files I'm targeting.


softprops:
I agreed to the Github CLA and think I followed the other directions. Looking for further direction.

PR #3468, opened 2017-02-11
Add support for rich text format (RTF) as data

lildude:
We've found a few large rich text format (rtf) files are causing timeouts with almost all of the time being spent in linguist:

```
Seconds |   Count | Name
 19.694 |       9 | linguist.strategy
 19.691 |       1 | linguist.detection
[... redacted ...]
  0.000 |       1 | linguist.detected
```

... specifically attempting to unnecessarily tokenise the file:

```
/data/github/870366e/vendor/gems/2.1.7/ruby/2.1.0/gems/github-linguist-4.8.12/lib/linguist/tokenizer.rb:91:in `extract_tokens'
/data/github/870366e/vendor/gems/2.1.7/ruby/2.1.0/gems/github-linguist-4.8.12/lib/linguist/tokenizer.rb:16:in `tokenize'
/data/github/870366e/vendor/gems/2.1.7/ruby/2.1.0/gems/github-linguist-4.8.12/lib/linguist/classifier.rb:99:in `classify'
/data/github/870366e/vendor/gems/2.1.7/ruby/2.1.0/gems/github-linguist-4.8.12/lib/linguist/classifier.rb:78:in `classify'
/data/github/870366e/vendor/gems/2.1.7/ruby/2.1.0/gems/github-linguist-4.8.12/lib/linguist/classifier.rb:20:in `call'
/data/github/870366e/vendor/gems/2.1.7/ruby/2.1.0/gems/github-linguist-4.8.12/lib/linguist.rb:30:in `block (3 levels) in detect'
```

The `samples/RTF/complicated.rtf` file in this PR is an example of such a file.

This is taken from GitHub Enterprise, hence the slightly old version, but github.com exhibits the problem too as you'll be able to see if you attempt to [view the file](https://github.com/lildude/linguist/blob/60253b026b97c0088dd5816e1b518abc33b84324/samples/RTF/complicated.rtf) in this PR.

This PR addresses this problem by explicitly adding support for RTF files and marking them as data so no attempts are made to tokenise the files when rendering.

As for usage: only [434,582 results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Artf+NOT+nothack&type=Code&ref=searchresults) for `.rtf` 😜

I know I'm technically a maintainer of linguist, but this is my first PR for it, so go easy on me 😉

PR #3471, opened 2017-02-13
Return early in Classifier.classify if no languages supplied

lildude:
In an attempt to resolve a problem with rendering large, complex RTF files, I added support for RTF in https://github.com/github/linguist/pull/3468, however this wasn't the best approach as @pchaigno pointed out in https://github.com/github/linguist/pull/3468#issuecomment-279247233

On closer inspection, it makes more sense to return early in `Classifier.classify` if no languages have been supplied as by not doing so, we unnecessarily tokenise the data that won't be used when it comes to scoring further down in the method.

This makes a significant difference in the time it takes to perform against the problematic file, before the early return:

```
$ time bin/linguist ~/tmp/trash/linguist-timeout/License.rtf --breakdown
License.rtf: 7507 lines (7457 sloc)
  type:      Text
  mime type: application/rtf
  language:
bin/linguist ~/tmp/trash/linguist-timeout/License.rtf --breakdown  13.56s user 0.13s system 96% cpu 14.156 total
$
```

... versus _with_ the early return:

```
$ time bin/linguist ~/tmp/trash/linguist-timeout/License.rtf --breakdown
License.rtf: 7507 lines (7457 sloc)
  type:      Text
  mime type: application/rtf
  language:
bin/linguist ~/tmp/trash/linguist-timeout/License.rtf --breakdown  0.42s user 0.07s system 90% cpu 0.545 total
$
```

Hopefully I'm not missing anything obvious with this, but please speak up if you think I am.

### DIFF
--- a/lib/linguist/classifier.rb
+++ b/lib/linguist/classifier.rb
@@ -95,7 +95,7 @@ module Linguist
     # Returns sorted Array of result pairs. Each pair contains the
     # String language name and a Float score.
     def classify(tokens, languages)
-      return [] if tokens.nil?
+      return [] if tokens.nil? || languages.empty?
       tokens = Tokenizer.tokenize(tokens) if tokens.is_a?(String)
       scores = {}
 
--- a/test/test_classifier.rb
+++ b/test/test_classifier.rb
@@ -53,4 +53,8 @@ class TestClassifier < Minitest::Test
       assert_equal language.name, results.first[0], "#{sample[:path]}\n#{results.inspect}"
     end
   end
+
+  def test_classify_empty_languages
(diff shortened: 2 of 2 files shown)


lildude:
> Any idea on how we could test this fix/new behavior?

I do. Test coming up.

brandonblack:
@lildude LGTM. We're due for a small gem bump as well. Let me know if you need help releasing this and bumping the gem on dotcom.

lildude:
@brandonblack If you've got the cycles to do this today (:us: time), then please feel free, else I'll take a stab at it tomorrow morning :gb: time.  I already know we're going to bump into a clash with the rugged versions so that'll slow me down, hence not doing it tonight.

rjkmurray:
Hmm.. that's me

PR #3472, opened 2017-02-13
Add .x as XDR/RPCGEN

dylex:
XDR/RPC language as documented in RFC5531, RFC4506.

Also adding random Logos.x samples to disambiguate, taken from random github search results -- hopefully this is okay license-wise.

Vim has [syntax highlighting for rpcgen](http://ftp.vim.org/vim/runtime/syntax/rpcgen.vim) but I haven't found any other syntax highlighting.  I can write one if necessary.

Many in-the-wild [examples on github](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ax+struct&type=Code&ref=searchresults).

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -513,5 +513,13 @@ module Linguist
       end
     end
   
+    disambiguate ".x" do |data|
+      if /\b(program|version)\s+\w+\s*{|\bunion\s+\w+\s+switch\s*\(/.match(data)
+        Language["RPC"]
+      elsif /^%(end|ctor|hook|group)\b/.match(data)
+        Language["Logos"]
+      end
+    end
+
   end
 end
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3753,6 +3753,17 @@ RMarkdown:
   - ".rmd"
   tm_scope: source.gfm
   language_id: 313
+RPC:
+  type: programming
+  aliases:
+  - rpcgen
+  - oncrpc
+  - xdr
+  ace_mode: c_cpp
+  extensions:
+  - ".x"
+  tm_scope: source.c
+  language_id: 1031374237
 RPM Spec:
   type: data
   tm_scope: source.rpm-spec
--- a/samples/RPC/rusers.x
+++ b/samples/RPC/rusers.x
@@ -1,0 +1,228 @@
+/*
+ * Copyright (c) 2010, Oracle America, Inc.
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ *       copyright notice, this list of conditions and the following
+ *       disclaimer in the documentation and/or other materials
+ *       provided with the distribution.
+ *     * Neither the name of the "Oracle America, Inc." nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ *   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *   "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *   FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *   COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ *   INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ *   DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE
+ *   GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *   INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ *   WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ *   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ *   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+%/*
+% * Find out about remote users
+% */
+
+const RUSERS_MAXUSERLEN = 32;
+const RUSERS_MAXLINELEN = 32;
+const RUSERS_MAXHOSTLEN = 257;
+
+struct rusers_utmp {
+	string ut_user<RUSERS_MAXUSERLEN>;	/* aka ut_name */
+	string ut_line<RUSERS_MAXLINELEN>;	/* device */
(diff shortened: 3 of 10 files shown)


lildude:
@dylex Can you please provide references to the samples you've included. There's no indication within them what their license is.

dylex:
I presume you mean for the Logos samples?  The RPC samples should all be documented I think.  I've added links to their original source, though the licenses are not necessarily clear there either.  I don't really what Logos is or what the included samples are for, but they seem to be the right language.  My goal was just to provide sufficient disambiguation for RPC.

lildude:
Thanks @dylex.  Unfortunately, we can't include any sample files if it's not clear exactly what license the code has been released under.

dylex:
Understood.  Feel free to take only the first commit aad8ab9f3138c247dc8b2fffbb1e92239c8a6214 that adds RPC in that case, since it does not contain any questionably licensed files.

lildude:
Ooops, completely forgot about this.  Thanks for the update.

If you're still interested in this, we're going to need:

- a Logos sample file else all those `.x` files will now become identified as RPC - https://github.com/JonasGessner/NoCarrier/blob/master/NoCarrier.x looks good as it's clearly MIT licensed
- a heuristic to disambiguate between the two languages as you originally provided
- info on where the samples have come from and the license they're licensed under - this doesn't need to be in the files themselves; a comment in this PR is sufficient.

Can you please also remove the nfs.x file - it's quite large and contains a lot of duplication in the "structure" of the content.

dylex:
The commit messages should contain all that information already.

lildude:
Please state it in the PR as a comment. Our legal peeps would prefer it if such info is easy to find. Commits are also squashed when the PR is merged.

dylex:
Sample licenses and sources:

https://github.com/JonasGessner/NoCarrier/blob/master/NoCarrier.x - MIT
https://github.com/basicthinker/PTMC/blob/cf31f4e46654cf967d2b3731f16fdb570be26ba0/llvm-gcc-R3/gcc/testsuite/objc/execute/string1.x - GPL2
https://github.com/fantastxmw/MobileShadowSocks/blob/f6415578fa00e607491c01106be83bba99008e1b/perapp-plugin/Tweak.x - GPL3
https://github.com/hbang/DailyPaper/blob/d1b3e838889c9e4754116fc5bf07fdc370d893b2/NCHax.x - Apache
rpc.x - RFC5531
rusers.x, yp.x - glibc, MIT3

lib/linguist/heuristics.rb
```diff
@@ -513,5 +513,13 @@ def call(data)
       end
     end
   
+    disambiguate ".x" do |data|
+      if /\<(program|version)\s+\w+\s*{|\<union\s+\w+\s+switch\s*\(/.match(data)
```
Alhadis: Why is `\<` escaped?

lib/linguist/languages.yml
```diff
@@ -3752,6 +3752,19 @@ RMarkdown:
...
+  aliases:
+  - rpcgen
+  - oncrpc
+  - xdr
+  ace_mode: c_cpp
+  extensions:
+  - ".x"
+  interpreters:
+  - rpcgen
+  tm_scope: none
```
Alhadis: > *Vim has syntax highlighting for rpcgen but I haven't found any other syntax highlighting. I can write one if necessary.*

You could use Objective-C's grammar for highlighting, as the language appears to be similar enough in syntax. Otherwise, a `tm_scope: none` will cause all RPC files to be shown on GitHub without colour.

dylex:
Oops, `\<` is a pcre-ism that I forgot doesn't exist in ruby.  Changed to `\b`.

Changed `tm_scope` to `source.c` since it's C-like (or at least C-based).

pchaigno:
@dylex Are [these files](https://github.com/search?p=99&q=extension%3Ax+NOT+struct&type=Code&utf8=%E2%9C%93) RPC as well?

samples/RPC/rusers.x
```diff
@@ -0,0 +1,228 @@
...
+		RUSERSPROC_NAMES(void) = 2;
+
+		utmp_array
+		RUSERSPROC_ALLNAMES(void) = 3;
+	} = 3;
+
+} = 100002;
+
+#ifdef RPC_HDR
+%
```
pchaigno: All following lines start with `%`. Does that indicate a comment line?

dylex:
@pchaigno hm, no those look like ld linker scripts, which lingust already has registered as `.ld`.

samples/RPC/rusers.x
dylex: No, any `%` lines are passed through directly to the generated C file by rpcgen, but that does mean that the lines are not in the RPC language.  Since many RPC files include such inline C code, it seemed to make sense to leave one in a sample. (More [RPC docs here]( https://docs.oracle.com/cd/E19683-01/816-1435/rpcgenpguide-21470/index.html).)

pchaigno: > Since many RPC files include such inline C code, it seemed to make sense to leave one in a sample.

👍

lib/linguist/languages.yml
```diff
@@ -3753,6 +3753,19 @@ RMarkdown:
...
+  aliases:
+  - rpcgen
+  - oncrpc
+  - xdr
+  ace_mode: c_cpp
+  extensions:
+  - ".x"
+  interpreters:
+  - rpcgen
+  tm_scope: source.c
```
pchaigno: I found a [grammar for RPC on GitHub](https://github.com/shakeelmohamed/sublime-rpcgen). Could you check if it does a better job than the C grammar, especially regarding the `%` lines mentioned previously? You can use [Lightshow](https://github-lightshow.herokuapp.com/) to test it without installing Sublime Text.

lib/linguist/languages.yml
```diff
@@ -3753,6 +3753,19 @@ RMarkdown:
...
+  type: programming
+  aliases:
+  - rpcgen
+  - oncrpc
+  - xdr
+  ace_mode: c_cpp
+  extensions:
+  - ".x"
+  interpreters:
+  - rpcgen
```
pchaigno: Are there actually RPC files using a `rpcgen` shebang? I [couldn't find any with GitHub searches](https://github.com/search?p=1&q=%22bin%2Frpcgen%22&type=Code&utf8=%E2%9C%93), but it's not particularly easy to search for strings with special characters.

lib/linguist/languages.yml
dylex: It looks pretty much just like C, and doesn't seem to add much useful (like `program` or `string` keywords or special `%` processing). I'd say it's probably not worth it.

lib/linguist/languages.yml
dylex: Not that I've ever seen, but rpcgen is the only program I know of that could.

Actually, looks like my rpcgen gets confused about the `#!` line anyway, so perhaps there's no point.

lib/linguist/languages.yml
pchaigno: Ok. Thanks for checking.

lib/linguist/languages.yml
pchaigno: Hm. Yes, it's probably better to remove it if it's not used.

lildude:
@pchaigno are you happy with this PR now?

pchaigno:
I think we should also add the `.x` file extension to [the Linker Script entry](https://github.com/github/linguist/blob/8da6ddf9d97ee1cd1d7119f3e8a5249df7d1590a/lib/linguist/languages.yml#L2362-L2371). But since these files are already misclassified as Logo, maybe we can defer this to a subsequent PR...?

lildude:
> I think we should also add the .x file extension to the Linker Script entry. But since these files are already misclassified as Logo, maybe we can defer this to a subsequent PR...?

Yeah, I think we can defer this to a subsequent PR as a quick bit of digging suggests there are several other `.x` variants for linker scripts according to the comments at https://github.com/pathscale/binutils/blob/5c2c133020e41fc4aadd80a99156d2cea4754b96/ld/genscripts.sh#L55-L66

PR #3473, opened 2017-02-14
Release v5.0.5

lildude:
Quite a few changes: https://github.com/github/linguist/compare/v5.0.4...lildude:cut-release-v5.0.5

With this release I've also updated the grammars and relaxed the `rugged` dependency as we're using 0.26.0b3 on dotcom and we probably shouldn't be shipping a release of Linguist that depends on a beta version of rugged.

@brandonblack as this is my first release, I'm asking for your review of this PR before I go ahead and merge, push etc, just to be safe.

Local testing with dotcom doesn't reveal any problems.

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'charlock_holmes', '~> 0.7.3'
   s.add_dependency 'escape_utils',    '~> 1.1.0'
   s.add_dependency 'mime-types',      '>= 1.19'
-  s.add_dependency 'rugged',          '0.25.1.1'
+  s.add_dependency 'rugged',          '0.26.0b3'
 
   s.add_development_dependency 'minitest', '>= 5.0'
   s.add_development_dependency 'mocha'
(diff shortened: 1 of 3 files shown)


brandonblack:
@lildude looks good overall, but we need to test the grammar updates in dotcom. I had some trouble with MagicPython in particular last time and had to roll back to the previous version. 

I've build the gem locally from this branch and opened a PR against dotcom for this.

github-linguist.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'charlock_holmes', '~> 0.7.3'
   s.add_dependency 'escape_utils',    '~> 1.1.0'
   s.add_dependency 'mime-types',      '>= 1.19'
-  s.add_dependency 'rugged',          '0.26.0b3'
```
brandonblack: Loosening this version is fine provided it remains compatible with dotcom (this was a problem in previous releases).

However, can you confirm your branch here is build from the latest? We actually bumped this a while ago and it looks like you were building off something old here.

https://github.com/github/linguist/blob/master/github-linguist.gemspec#L19

lildude: The loosening from 0.2.6.0b3 was just me reversing the bump I did in https://github.com/github/linguist/pull/3473/commits/0b9eb0c1033a8af1afef35bc9d8f8cae1b113a11 (in this PR) and instead opting for a loosening of the version.  The overall change in this PR is from `= 0.25.1.1` to `>= 0.25.1.1`.

lildude:
All sorted with no changes to the php.tmbundle submodule.  Gonna merge and :ship:.

PR #3469, opened 2017-02-11
Add grammar for HLSL (High Level Shading Language)

tgjones:
Linguist already has a language defined for HLSL, but it doesn't have an associated grammar, so HLSL files are not highlighted on GitHub.

This PR adds a TextMate grammar for HLSL, to enable syntax highlighting on GitHub.

PR #3490, opened 2017-02-22
Add ShaderLab language

tgjones:
This PR adds a new language definition for ShaderLab (used to write shaders in the Unity game engine). It uses the grammar that was already added as part of #3469.

I've moved the `.shader` extension from the GLSL language definition (where it was incorrect, but better than nothing) to the new ShaderLab language definition.

[ShaderLab search result on GitHub.com](https://github.com/search?utf8=✓&q=extension%3Ashader+NOT+nothack&type=Code&ref=searchresults)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1358,6 +1358,7 @@ GLSL:
   - ".geom"
   - ".glslv"
   - ".gshader"
+  - ".shader"
   - ".vert"
   - ".vrx"
   - ".vsh"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1358,7 +1358,6 @@ GLSL:
   - ".geom"
   - ".glslv"
   - ".gshader"
-  - ".shader"
   - ".vert"
   - ".vrx"
   - ".vsh"
@@ -3996,6 +3995,13 @@ Self:
   tm_scope: none
   ace_mode: text
   language_id: 345
+ShaderLab:
+  type: programming
+  extensions:
+  - ".shader"
+  ace_mode: text
+  tm_scope: source.shaderlab
+  language_id: 664257356
 Shell:
   type: programming
   color: "#89e051"
--- a/samples/GLSL/SyLens.shader
+++ b/samples/GLSL/SyLens.shader
@@ -1,0 +1,161 @@
+#version 120
+
+/*
+  Original Lens Distortion Algorithm from SSontech (Syntheyes)
+  http://www.ssontech.com/content/lensalg.htm
+  
+  r2 is radius squared.
+  
+  r2 = image_aspect*image_aspect*u*u + v*v
+  f = 1 + r2*(k + kcube*sqrt(r2))
+  u' = f*u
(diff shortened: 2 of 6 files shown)


larsbrinkhoff:
`.shader` is a very generic file extension.  I'm not a shader language expert; are really all those search result ShaderLab files?

larsbrinkhoff:
I believe these keywords are present in ShaderLab but unlikely to appear in other shader languages: colormask, cginclude, cgprogram, _maintex, subshaders.  You're right, the [vast majority](https://github.com/search?q=extension%3Ashader+colormask+OR+cginclude+OR+cgprogram+OR+_maintex+OR+subshaders&type=Code) of all .shader files is ShaderLab.

However, [some of them are GLSL](https://github.com/search?q=extension%3Ashader+vec3+OR+mat4+OR+gl_position&type=Code).

tgjones:
Ah, thanks for noticing that. Presumably I should add the `.shader` extension back to GLSL - but do I need to do anything, beyond providing samples, to help the classifier?

On a possibly related note, I think there's a bug in [this classifier test](https://github.com/github/linguist/blob/master/test/test_classifier.rb#L49):

``` ruby
Samples.each do |sample|
  language  = Linguist::Language.find_by_name(sample[:language])
  languages = Language.find_by_filename(sample[:path]).map(&:name)
  next unless languages.length > 1

  results = Classifier.classify(Samples.cache, File.read(sample[:path]), languages)
  assert_equal language.name, results.first[0], "#{sample[:path]}\n#{results.inspect}"
end
```

It never gets beyond the `next unless...`, because `Language.find_by_filename` is too specific, and only ever returns a single language. `Language.find_by_extension` seems to work better there, but I don't know the code well enough to know if that's the correct fix.

pchaigno:
> It never gets beyond the `next unless...`, because `Language.find_by_filename` is too specific, and only ever returns a single language. `Language.find_by_extension` seems to work better there, but I don't know the code well enough to know if that's the correct fix.

Nice catch! That is the correct fix :-)
I missed this test when I changed `find_by_filename` in #2006 :/
Could you open a separate pull request to fix it?

larsbrinkhoff:
> do I need to do anything, beyond providing samples, to help the classifier?

If the samples doesn't do the trick, lib/linguist/heuristics.rb is your friend.

tgjones:
> Could you open a separate pull request to fix it?

I've opened #3494 with a fix for the ambiguous classification test.

> If the samples doesn't do the trick, lib/linguist/heuristics.rb is your friend.

It looks like the samples are enough - assuming I fixed the test correctly in #3494, and that I'm testing the right thing :)

tgjones:
Please let me know if you need anything more from me on this PR.

larsbrinkhoff:
Travis says the pull request doesn't pass build/test.

tgjones:
Oops, somehow didn't notice that (or rather I did, but I looked at the build failure for a different Ruby version that had unrelated errors, and thought it wasn't caused by my changes).

Anyway, fixed now.

PR #3493, opened 2017-02-23
Add alias for pandoc

klpn:
I use the modeline `vim: set filetype=pandoc:` for [this documentation Mustache template](https://github.com/klpn/Mortchartgen.jl/blob/master/data/mortchartdoc.mustache) to have it recognized as Pandoc markdown (this filetype is used by e.g. [the vim-pandoc plugin](https://github.com/vim-pandoc/vim-pandoc)). However, the `pandoc` filetype is not recognized by Linguist, and the file is incorrectly classified as `HTML+Django`. So, this request adds an alias for `pandoc` to `Markdown`.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2456,6 +2456,8 @@ Mako:
   language_id: 221
 Markdown:
   type: prose
+  aliases:
+  - pandoc
   ace_mode: markdown
   codemirror_mode: gfm
   codemirror_mime_type: text/x-gfm


Alhadis:
Seems fine by me. 👍 

Should we also include `Commonmark` and `GFM` as aliases? Do you know if Vim modes exist for those Markdown flavours too?

klpn:
Yes, there are many Markdown flavors. I do not know about any vim plugins making use of filetypes other than `markdown` and `pandoc`, however.

Alhadis:
In that case, we should stick to `pandoc` only. Which leaves this PR good-to-go in my book. 👍

klpn:
I do not know about the travis failure: in the error log, it seems to be related to grammar licenses, which do not have anything to do with the change I made.

brandonblack:
@klpn can you update your branch with the latest from master?

klpn:
It is already 1 commit ahead (and no behind).

PR #3504, opened 2017-03-04
Fix grammar hashes

pchaigno:
Due to a recent update of Licensee (with modifications to the normalization process), hashes for grammar licenses changed. This pull request updates those hashes and ensures it doesn't happen again (restrict automatic updates of the licensee gem to minor releases).

Licensee also doesn't recognize a few licenses anymore (additions from `PHP-Twig.tmbundle` to `SublimeEthereum`). I'll open an issue in licensee for those that look like regressions. I'll also try a get a few more normalized (i.e., remove extra title) directly into the grammar repositories.

~~Fixes the Travis build.~~ There's still an issue with Ruby 2.1. @brandonblack Should we drop it for Linguist? Or maybe there's a way to define a conditional dependency (Licensee 8.7 if Ruby < 2.2)?

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -26,5 +26,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'yajl-ruby'
   s.add_development_dependency 'color-proximity', '~> 0.2.1'
   s.add_development_dependency 'licensed'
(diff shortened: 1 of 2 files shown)


brandonblack:
@pchaigno I think it probably makes sense to mark 2.1 as an allowed failure, but lets leave it in CI for now.

PR #3510, opened 2017-03-11
Update grammar whitelist

pchaigno:
This pull request is a follow up to #3504 and removes several grammars from the whitelist.

#3504 added several grammars to the whitelist because they weren't recognized by Licensee anymore. It turns out that some of them (BSD licenses) were not recognized because of a bug in Licensee, fixed in benbalter/licensee#174 (Licensee v8.8.3).

I also updated the submodules since, with the new versions, two grammars (davidhq/SublimeEthereum#13 and Drako/SublimeBrainfuck#1) now have proper `LICENSE` files and are thereby recognized by Licensee.

MagicPython now supports `text.python.console` and `text.python.traceback` so I had to remove the `language-python` grammar.

Fixes the Travis build (again).

EDIT: @brandonblack The Travis build fails for Ruby 2.4.0. I'm not sure what the error means... :s

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,9 +67,6 @@
 [submodule "vendor/grammars/language-javascript"]
 	path = vendor/grammars/language-javascript
 	url = https://github.com/atom/language-javascript
-[submodule "vendor/grammars/language-python"]
-	path = vendor/grammars/language-python
-	url = https://github.com/atom/language-python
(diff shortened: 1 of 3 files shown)


brandonblack:
@pchaigno for now I've added ruby v2.4 as an allowed failure.

PR #3517, opened 2017-03-14
Treat vstemplate and vsixmanifest as XML

fragmer:
`*.vstemplate` files describe VisualStudio templates. They always use XML syntax. [Examples on GitHub (12,000+)](https://github.com/search?utf8=%E2%9C%93&q=extension%3Avstemplate+NOT+nothack&type=Code&ref=searchresults)

`*.vsixmanifest` files describe VisualStudio and VSCode extensions. They also always use XML syntax. [Examples on GitHub (8,000+)](https://github.com/search?utf8=%E2%9C%93&q=extension%3Avsixmanifest+NOT+nothack&type=Code&ref=searchresults)


Alhadis:
Yet another XML-based file format... :(

@pchaigno, @lildude, @brandonblack Is it possible to match languages based on an arbitrary substring? If so, we should consider matching XML files based on their [XML declarations](https://www.w3.org/TR/xml/#sec-prolog-dtd):

~~~xml
<?xml version="1.0" encoding="UTF-8"?>
~~~

This is something that every XML file must start with, and unambiguously identifies the content as being XML. If we supported this, it would obviate listing each and every single XML extension which starts seeing common use...

fragmer:
Looking for `<?xml version=...` would be a great heuristic to add for identifying previously-unknown XML formats, but it cannot be the only rule. Prior to XML 1.1, this declaration was optional. Both file extensions proposed in this PR use XML 1.0, so the declarations may be missing (e.g. as seen in the [official vstemplate docs](https://msdn.microsoft.com/en-us/library/xsxc3ete(v=vs.100).aspx)).

Additional examples:
- [MSBuild props files](https://github.com/github/linguist/blob/master/samples/XML/Default.props)
- [Fuse UX files](https://github.com/github/linguist/blob/master/samples/XML/MyApp.ux)
- [Wix project files](https://github.com/github/linguist/blob/master/samples/XML/wixdemo.wixproj)

fragmer:
Is there anything else I can do to help this PR get merged? It's been a few weeks.

pchaigno:
@fragmer Is there any other filename than `source.extension.vsixmanifest` for the `.vsixmanifest` extension? If not, we should probably add it an a filename, not an extension.

fragmer:
@pchaigno It can be either `source.extension.vsixmanifest` or `extension.vsixmanifest`, and the casing may vary (e.g. I've seen `Extension.vsixmanifest`).

pchaigno:
Ok, thanks for the info. Let's keep `.vsixmanifest` then, it's already quite specific.

PR #3521, opened 2017-03-17
Add C header for more accurate classification from https://github.com/Wetitpig/osmctools

Wetitpig:
The repository with misclassified files: https://github.com/Wetitpig/osmctools

All the files in the repository are created by breaking a large c file into several small c files. However, after breaking, linguist show c++ as the language for three of those files, which doesn't make sense at all.

Can anyone suggest how to improve linguist by detecting c and c++ more accurately?


Wetitpig:
But did it occupy so much space that you cannot afford? Or else why can't you put it into header samples?

By the way, how long does it take to refresh the language statistics bar?

pchaigno:
> But did it occupy so much space that you cannot afford? Or else why can't you put it into header samples?

Yes, space is a concern (see #2117 for instance). I'm also worried that, in part because we don't have much samples, adding a sample file with mostly data will screw the Bayesian classifier.

> By the way, how long does it take to refresh the language statistics bar?

It should happen in the next hour after each push to the repository. If not, you can contact GitHub support and they'll look into it.

PR #3523, opened 2017-03-19
Prefer Markdown over GCC Machine Description

kivikakk:
We currently use a heuristic to determine if a `.md` file should be recognised as "Markdown" or "GCC Machine Description". If neither regex matches the heuristic, the existing guesses are retained. Because "GCC Machine Description" sorts lexicographically before "Markdown", then, it's preferred as the first choice.

This is very unlikely to be what we want in the majority of cases, so let's prefer Markdown unless we're sure.

### DIFF
--- a/lib/linguist.rb
+++ b/lib/linguist.rb
@@ -15,9 +15,9 @@ class << Linguist
   #       see Linguist::LazyBlob and Linguist::FileBlob for examples
   #
   # Returns Language or nil.
-  def detect(blob)
+  def detect(blob, options={})
     # Bail early if the blob is binary or empty.
-    return nil if blob.likely_binary? || blob.binary? || blob.empty?
+    return nil if blob.likely_binary? || blob.binary? || (!options[:allow_empty] && blob.empty?)
 
     Linguist.instrument("linguist.detection", :blob => blob) do
       # Call each strategy until one candidate is returned.
--- a/lib/linguist.rb
+++ b/lib/linguist.rb
@@ -15,9 +15,9 @@ class << Linguist
   #       see Linguist::LazyBlob and Linguist::FileBlob for examples
   #
   # Returns Language or nil.
-  def detect(blob, options={})
+  def detect(blob, allow_empty: false)
     # Bail early if the blob is binary or empty.
-    return nil if blob.likely_binary? || blob.binary? || (!options[:allow_empty] && blob.empty?)
+    return nil if blob.likely_binary? || blob.binary? || (!allow_empty && blob.empty?)
 
     Linguist.instrument("linguist.detection", :blob => blob) do
       # Call each strategy until one candidate is returned.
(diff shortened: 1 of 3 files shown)


lib/linguist.rb
```diff
@@ -15,9 +15,9 @@ class << Linguist
   #       see Linguist::LazyBlob and Linguist::FileBlob for examples
   #
   # Returns Language or nil.
-  def detect(blob)
+  def detect(blob, options={})
```
charliesome: Can we use a keyword argument here? `allow_empty: false` or something

kivikakk: I was about to say "we don't do that anywhere else and maybe there's a compat reason", but I just found a single instance, so yes!

pchaigno:
I'm not sure this is a good idea. This makes Markdown the language for all GCC Machine Description files that don't match the `^(;;|\(define_)` heuristic.

We have some cases where the heuristic rules are almost perfect. For instance, there might be a module or keywords that's needed in every files of a particular language for them to compile properly. In this case, we're almost sure all files of that language contain the keyword. I don't think GCC Machine Description is one of those cases.

> If neither regex matches the heuristic, the existing guesses are retained. Because "GCC Machine Description" sorts lexicographically before "Markdown", then, it's preferred as the first choice.

As far as I know, that's not what happens. If the heuristic rules don't match, [we fall back to the Bayesian classifier](https://github.com/github/linguist/blob/69ba4c5586fc4cab3166a6b8ee2d933e83bb2152/lib/linguist.rb#L60-L67). The classifier will give scores to each language and the language with the highest score will be selected. The classifier might fail and return the wrong language, which is probably what is happening in your case.

I think we should try to add more sample files for the classifier or try to reinforce the `md` heuristic rules.

/cc @larsbrinkhoff

kivikakk:
> The classifier might fail and return the wrong language, which is probably what is happening in your case.

Right, I see. In this case, I'm still heavily biased toward retaining the strong default introduced by this PR simply because we depend strongly on `.md` coming through as Markdown in almost all cases. Getting the Bayesian classifier to recognise these cases more often would be ideal, but even a small false-positive rate for GCC Machine Description will cause real confusion.

Reading the [GCC machine description docs](https://gcc.gnu.org/onlinedocs/gccint/Overview.html#Overview), it looks as if there will always be at least one `(define_insn …)` or `(define_expand …)` in any useful machine description. I suppose it's possible people might have whitespace in unexpected places causing the rule not to hit.

larsbrinkhoff:
I agree that there should be a strong preference for Markdown.

If a file doesn't match `^(;;|\(define_)` it's probably ok not consider it as GCC machine description.

pchaigno:
Okay, I stand corrected.
Thanks for the information @kivikakk & @larsbrinkhoff!

kivikakk:
Thank you for bringing up the concern and for having this discussion! I really appreciate it :heart:

PR #3529, opened 2017-03-21
Try Fixing Ruby 2.4.0 Errors

Wetitpig:
Should be ok.


kivikakk:
Hi there! Could you explain this PR at all? I see you've changed some of the MD5 tests, but I can't quite work out how they relate to Ruby 2.4.

PR #3530, opened 2017-03-21
Update the `Instrumenter` doc

javierhonduco:
Updates the instrumenter doc, initializing it.

### DIFF
--- a/lib/linguist.rb
+++ b/lib/linguist.rb
@@ -74,7 +74,7 @@ class << Linguist
   #       end
   #     end
   #
-  #     Linguist.instrumenter = CustomInstrumenter
+  #     Linguist.instrumenter = CustomInstrumenter.new
   #
   # The instrumenter must conform to the `ActiveSupport::Notifications`
   # interface, which defines `#instrument` and accepts:


Alhadis:
qué

javierhonduco:
The existing docs suggest setting `Linguist.instrument` to the Instrumenter class: `Linguist.instrumenter = CustomInstrumenter`. 

That's not going to work as the `instrument` method calls [`instrumenter.instrument(*args, &bk)`](https://github.com/linguist/linguist/blob/716498e63561e8844f68b68b1fd9c716f2748272/lib/linguist.rb#L94), which is an instance method.

I think that doc embedded in the comments may be confusing as it doesn't work as stated. That's why I created this small PR.

Alhadis:
Ah, veo!

/cc @pchaigno @kivikakk

javierhonduco:
Practising languages on a code collaboration tool, we are living in the future ✨ ✨  

PS: Sounds more natural if you say "entiendo" instead of "veo" 😃

Alhadis:
I pretend to contribute to this repo, but I'm really using the name `Linguist` as a rabid excuse to shit-talk in as many world languages as possible.

PR #3541, opened 2017-03-27
.pod disambiguation heuristic fix

kivikakk:
Look for any line starting with "=\w+", not full lines, otherwise we miss e.g. "=head1 HEADING".

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -346,7 +346,7 @@ module Linguist
     end
 
     disambiguate ".pod" do |data|
-      if /^=\w+$/.match(data)
+      if /^=\w+\b/.match(data)
         Language["Pod"]
       else
         Language["Perl"]


kivikakk:
@jberger Your assumption is correct! The disambiguation rule only runs on files ending in `.pod`, so this won't affect the recognition of `.pl`/`.pm`/etc. files with embedded pod — it just ensures any `.pod` file is actually recognised as POD as long as it has at least one line that starts with `=`, then any number of alphanumeric characters, then either whitespace or the end of the line.

jberger:
Heh, sorry, I'm so used to using the new approval system for work that I forgot that my vote won't count. 👍 anyway! 😄

kivikakk:
Great! I'll do a release and get the Linguist used on github.com bumped this week. 👍

kivikakk:
(In the mean time, including literally any `=xyz` tag without text following it on the same line will cause our renderer to pick it up correctly 😅)

jberger:
Thank you very much for the attention and quick fix!

jberger:
Has this been released? I'm still seeing this behavior on my repo (for which I'm going to adopt the recommended work-around since I'm going to release the project to CPAN soon).

jberger:
Ah I see that there haven't been any releases since this was merged. In that case nevermind, sorry for the noise.

lildude:
@jberger Correct. I plan to do a release in the next few days (before 9 May).

PR #3546, opened 2017-03-29
remove unused disambiguation heuristc for .t

smola:
There were two disambiguations for .t files. The first one
did only disambiguate between Perl and Perl6. The second one
disambiguated Turing and Perl6, but it was being ignored since
it was repeated.

Turing files were still being correctly classified since they fell
back to the classifier strategy though. This change removes the
unused heuristic.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -435,14 +435,6 @@ module Linguist
       end
     end
     
-    disambiguate ".t" do |data|
-      if /^\s*%|^\s*var\s+\w+\s*:\s*\w+/.match(data)
-        Language["Turing"]
-      elsif /^\s*use\s+v6\s*;/.match(data)
-        Language["Perl6"]
-      end
-    end
-    
     disambiguate ".toc" do |data|
       if /^## |@no-lib-strip@/.match(data)
         Language["World of Warcraft Addon Data"]


smola:
Note that "fixing" the heuristic by merging both leads to incorrect classification of `samples/Perl/perl-test.t` as Turing, so it's probably better to use the classifier and remove this.

lildude:
I'd rather not remove this disambiguation.  As I see it, and I'm open to being corrected (🙇 @Alhadis and @pchaigno, veterans of this project), but the classifier is really a "last resort, I'm gonna take a guess at this" effort and shouldn't be relied upon if there are better and more reliable methods.

At the moment the `.t` extension is associated with four languages, of which there is only disambiguation for two of them.  I fear dropping this to zero is going to lead to more misclassification than is currently happening.

I'm going to let this PR sit pending comment for the rest of the week and will probably close it if there isn't a good argument to go ahead and merge it.

Alhadis:
[This was added](https://github.com/github/linguist/pull/3049/files) to fix [misclassification](https://github.com/github/linguist/issues/2940) of Perl6 and Turing, and I fail to see any reason to remove it.

👎  from me, sorry.

pchaigno:
> The second one disambiguated Turing and Perl6, but it was being ignored since it was repeated.

@smola The Heuristic strategy should [go through all sets of heuristic rules until it find one that matches or exist](https://github.com/github/linguist/blob/c0e242358af6a3ad0ace66e29b480d3f507917cb/lib/linguist/heuristics.rb#L19-L24). So, having several sets for the same file extension should still work. I however agree that we could improve that situation and merge the two sets of heuristic rules for `.t`.

> Note that "fixing" the heuristic by merging both leads to incorrect classification of `samples/Perl/perl-test.t` as Turing, so it's probably better to use the classifier and remove this.

The Turing heuristic rule matches `samples/Perl/perl-test.t` because of [this line](https://github.com/github/linguist/blob/c0e242358af6a3ad0ace66e29b480d3f507917cb/samples/Perl/perl-test.t#L9), which starts with a `%`. @Alhadis Do you see any way to restrict that first part of the heuristic rule? Would requiring a white space after the `%` be sufficient (i.e., `^%\s|...`)?

Alhadis:
> Would requiring a white space after the % be sufficient (i.e., ^%\s|...)?

Ah crap, I totally forgot about hash sigils. Erm, kind of. There's still the possibility of Turing comments that have no space between `%` and the first character, though. You're right, this needs to be revised.

Although that test looks wrong to me... I'm not familiar with Perl 6 whatsoever, but Perl 5's strict mode would mandate `%hash` be preceded by `my` or `our` (analogous to `local` / `global`, pretty much).

pchaigno:
> Although that test looks wrong to me... I'm not familiar with Perl 6 whatsoever, but Perl 5's strict mode would mandate `%hash` be preceded by `my` or `our` (analogous to `local` / `global`, pretty much).

Sure, but that's still common usage, right?

Alhadis:
I suppose, but not all Perl code runs with strict mode enabled (to the dismay of any self-respecting Perl programmer).

My gut currently has two feelings: hunger, and the feeling this isn't a good choice for a heuristic. I should probably back over some Turing code.

BTW, it'd be awesome if grammars could be leveraged as a fallback for disambiguating between two languages... which result has more refined highlighting, for example...

pchaigno:
> BTW, it'd be awesome if grammars could be leveraged as a fallback for disambiguating between two languages... which result has more refined highlighting, for example...

That was Pygments' approach to classify files if I remember correctly. I doubt this can scale to 10M+ users though...

Alhadis:
> That was Pygments' approach to classify files if I remember correctly. I doubt this can scale to 10M+ users though.

Well, it could only be used as a fallback when the Bayesian classifier fails to recognise a file's language. But I can pursue that idea some other time.

I *think* we can get away with our current heuristic by including a secondary check for a trailing semicolon, too. Perl mandates their use, whereas Turing doesn't use them at all. Not sure how well that'll scale to Perl lines like this one, though:

~~~perl
my %hash =~ s/
	(multi)line(expression)
/m;
~~~

Alhadis:
You see, one advantage to using TextMate grammars for this sort of analysis is being able to confidently distinguish between `my %hash = ();` and a quoted string containing the same pattern. =) We could leverage scopes to filter out or restrict matches based on matches.

That would require grammars to be well-written, though, and I'm sorry to say that there aren't many grammar authors who go to the pains I do...

Alhadis:
@pchaigno Wait, when I run `linguist` on `samples/Perl/perl-test.t`, it identifies as Perl. What issue are we trying to address, again...?

It sounds like we're potentially **adding** issues by changing what doesn't need fixing... or am I missing something here?

pchaigno:
> @pchaigno Wait, when I run linguist on samples/Perl/perl-test.t, it identifies as Perl. What issue are we trying to address, again...?

We'd only be fixing classification of Perl files that don't match the Perl heuristic rule but contain a hash declaration (`%`).

Merging the two sets of heuristic rules would also mean we run the Perl heuristic rule only once (instead of twice currently) for many cases.

Alhadis:
OH! Sorry, I'm with you now. Somehow in my mind, I must've blended Perl 5 and 6 together as the one language. Gotcha.

I'll take a stab at it. =) Should it be a separate PR, or...?

smola:
I'm closing this, I got the logic wrong, so this is indeed invalid.

Alhadis:
On the plus side, you *did* indeed highlight something that needed improvement, so thanks! :D

PR #3549, opened 2017-04-02
Fix the CodeMirror mode for RPM

tberman:


tberman:
This is incorrect, closing.

PR #3278, opened 2016-10-11
The grand language renaming bonanza

arfon:
Thanks to the stellar work of @chrismwendt and @TwP it's now possible to rename languages on GitHub 🎉 

This is a work-in-progress PR to rename languages that we know are incorrect.

/ cc @pchaigno @larsbrinkhoff @Alhadis as they all seem to have better memories than I :grin:

PR #3550, opened 2017-04-04
fix heuristic for Unix Assembly with .ms extension

smola:
It was left unrenamed from GAS, which lead to heuristic not working for this case.
See here: abfb698550a40dc51212de075fde5fe670f529b4

Without this patch, if I enable only the extension+heuristic strategies, I get:

```
$ bundle exec linguist samples/Unix\ Assembly/hello.ms
hello.ms: 92 lines (63 sloc)
  type:      Text
  mime type: application/x-troff-ms
  language:  MAXScript
```

with this patch:

```
$ bundle exec linguist samples/Unix\ Assembly/hello.ms
hello.ms: 92 lines (63 sloc)
  type:      Text
  mime type: application/x-troff-ms
  language:  Unix Assembly
$ bundle exec linguist samples/MAXScript/volume-calc.ms 
volume-calc.ms: 23 lines (22 sloc)
  type:      Text
  mime type: application/x-troff-ms
  language:  MAXScript
```

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -298,7 +298,7 @@ module Linguist
       if /^[.'][a-z][a-z](\s|$)/i.match(data)
         Language["Groff"]
       elsif /(?<!\S)\.(include|globa?l)\s/.match(data) || /(?<!\/\*)(\A|\n)\s*\.[A-Za-z]/.match(data.gsub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'|\\\s*(?:--.*)?\n/, ""))
-        Language["GAS"]
+        Language["Unix Assembly"]
       else
         Language["MAXScript"]
       end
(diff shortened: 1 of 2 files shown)


smola:
There are probably other languages affected, see: https://github.com/github/linguist/pull/3278

pchaigno:
@smola Thanks for the fix! Could you also add a test in [`test/test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb) to prevent this from happening again?

smola:
@pchaigno Done.

Alhadis:
Looks good to me!

~~~asm
shp $5, %this
~~~

Alhadis:
Can I be the one to merge this?

Heck, it's a fix for the very first PR I submitted to Linguist (December 2015). Nostalgia!

Alhadis:
I'm so happy I could vomit.

PR #3556, opened 2017-04-07
Fix heuristics after rename

smola:
This is similar to https://github.com/github/linguist/pull/3550 for the remaining languages where the name used in heuristics.rb does not match the one in samples directory.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -165,7 +165,7 @@ module Linguist
       elsif data.include?("flowop")
         Language["Filebench WML"]
       elsif fortran_rx.match(data)
-        Language["FORTRAN"]
+        Language["Fortran"]
       end
     end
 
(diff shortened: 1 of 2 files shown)


pchaigno:
Should we add a test in `all_fixtures` to make sure `#{samples_path}/#{language_name}/` exists? It would prevent this from happening again (for cases where we have a test).

Alhadis:
Sounds like a good idea. Although I can't envision this becoming a recurring issue...

lildude:
I see no harm in adding a test as @pchaigno suggested. It'll mean we don't have to manually remember to check this.

@smola could you add a test as suggested by @pchaigno

lildude:
Gonna merge this without the test as I want to build a new release with enough time to ship it before 9 May. I don't think we'll be renaming things again any time soon so I'm happy to leave off the test for now. We can always add one later.

PR #3571, opened 2017-04-15
Correct Arduino's Color, Swap MAXScript's

MarcianoPreciado:
Correct Arduino's color to it's signature blue; Swap Arduino's old purple color to MAXScript whose color is too close. Arduino is a much more commonly used programming language than MAXScript so the color will be correct more of the time.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -258,7 +258,7 @@ Arc:
   language_id: 20
 Arduino:
   type: programming
-  color: "#31a09c"
+  color: "#00979D"
   extensions:
   - ".ino"
   tm_scope: source.c++
@@ -2368,7 +2368,7 @@ M4Sugar:
   language_id: 216
 MAXScript:
   type: programming
-  color: "#bd79d1"
+  color: "#056d6d"
   extensions:
   - ".ms"
   - ".mcr"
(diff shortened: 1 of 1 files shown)


wopian:
MAXScript should probably be changed to a lighter/darker teal variant which its heavily associated with, instead of randomly becoming purple.



 

---

Edit: Ardiuno's colour is `#00979D`. `#056d6d` could work for MS if its still too similar

Alhadis:
👎  from me. MAXScript should stay teal, because that's the colour that's always been associated with 3DS Max. Having it purple would look random and weird.

MarcianoPreciado:
Although MAXScript has always been teal, and for longer than Arduino (which has also always been a teal color), it would satisfy more users to have Arduino's color be correct. 
A simple google search indicates the vast difference in popular use. 3DS Max returns 21,300,000 results, MAXScript returns 462,000 results, Arduino returns 62,000,000 results. Having Arduino be purple "would look random and weird" to more people than MAXScript.

Alhadis:
> Having Arduino be purple "would look random and weird" to more people than MAXScript.

You see, there's a problem with that logic: you're assuming there aren't Arduino users who are used to recognising the language by that specific shade of purple.

Sorry, but nothing about this PR feels right to me.

lildude:
So the general consensus is to _not_ change the colour of Arduino and MAXscript so I'm closing this PR.

PR #3575, opened 2017-04-20
Switch the PHP grammar to the upstream repo

### LINKED ISSUE #3565: Syntax highlighting for PHP 7 does not work well
the grammar repo used for PHP (https://github.com/brandonblack/php.tmbundle) is not the source repo of the grammar, but a fork of https://github.com/textmate/php.tmbundle

The fork is behind the source repo by 8 commits. And some of these commits are about adding support for PHP 7 features.
Could you switch to the source repo for the PHP grammar instead ?

stof:
Closes #3565 (at least for the part to be fixed in Linguist itself).

Note that I have only update the URL. I have not updated the target commit to the latest commit in this PR for now, as the readme says that the grammar is updated when building the gem anyway. But I may contribute the reference update too if you think it is necessary.


stof:
@lildude done

lildude:
Sadly, I'm going to have to revert this as as you pointed out in https://github.com/github/linguist/issues/3601, it breaks PHP highlighting.  I'll work on our test too to catch this.

stof:
@lildude it would be great to work with the textmate maintainer to understand why it breaks it though, to be able to benefit from other upstream improvements.
I have no experience at all using TextMate (and I'm on Ubuntu so I won't have one), so I have no idea how to contribute to the bundle to improve it.

lildude:
> @lildude it would be great to work with the textmate maintainer to understand why it breaks it though, to be able to benefit from other upstream improvements.

Definitely. I plan to look into it to see if I can work out the precise change that introduces the problem. For the moment I've had to revert to stem the firehose.

PR #3578, opened 2017-04-21
Clarify description of implicit alias.

binki:
I was trying to look up the alias to use for DNS Zone. From the docs
the alias I should use would be dns zone, but in reality it is dns-zone.
This change updates the comments to describe how to derive the
implicit name of a given alias.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2,7 +2,7 @@
 #
 # type              - Either data, programming, markup, prose, or nil
 # aliases           - An Array of additional aliases (implicitly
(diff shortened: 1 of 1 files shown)


pchaigno:
Thanks for the pull request! It indeed needs fixing :-)

I'm not sure the line of code will be very clear to readers though. Maybe it would be best to explain it with a complete sentence?

binki:
The original docs required the user to be able to infer that it was Ruby and execute it in their heads, so I just assumed that was acceptable :-p. But I can attempt to turn it into English.

PR #3579, opened 2017-04-21
Revert "Added Jest snapshot test files as generated src (#3572)"

cpojer:
I saw https://github.com/github/linguist/pull/3572 fly by in my Twitter feed. This diff and its explanation is entirely wrong and snapshot files *must not* be hidden from diffs and are important for code review.

It is true that snapshot files are generated by Jest, however users have to accept the new values and approve them and code reviewers should inspect them for unintended changes. While Jest creates them, it is merely a mechanism to reduce manual work to write down the assertion content and it is essential to the test rather than an unimportant side-effect.


pchaigno:
Thank you for raising your concerns. We often don't know the languages (or file format) added to Linguist, which makes it difficult to know if a file should be considered generated, documentation, or even vendored, so other points of view are greatly appreciated :yellow_heart: 

> It is true that snapshot files are generated by Jest, however users have to accept the new values and approve them and code reviewers should inspect them for unintended changes. 

I see your point, but don't all committed files need to be reviewed in any case? IMO, generated files are hidden from diffs to make those more readable (I was under the impression that these files can be huge). You can still review them.

/cc @hnryjms

hnryjms:
For me, I see generated files from Jest as similar generated lockfiles from Yarn..... the auto-generated form is set up to be human readable and diff is logical in most situations; and there are perfectly valid cases to review the diff of these files—GitHub's UI still allows by just selecting "Load Diff."

But I've seen, at least in the few React/RN projects I've been contributing in their early stages, these Snapshot files become very very large, esp. with indentation changes as views become wrapped/unwrapped in parent components.

cpojer:
Hey @pchaigno,

I work on Jest and it is expected that engineers will review snapshots. Here is an example: https://github.com/facebook/jest/blob/master/examples/snapshot/__tests__/__snapshots__/Link.react-test.js.snap#L58-L67 corresponds to https://github.com/facebook/jest/blob/master/examples/snapshot/__tests__/Link.react-test.js#L10-L15. Hiding them from code review will make snapshots lose their value during the review process significantly.

If snapshots are too big to review, then this is a problem with the user and it is recommended to slice the code that is to be snapshotted to only contain the minimal amount of useful information, for example by mocking unimportant components with Jest. This is very similar to a large file of code that is hard to review where breaking it up or optimizing the code will help.

We will also extend Jest's documentation to highlight this part specifically.

hnryjms:
@cpojer I would disagree that it's incorrect to have very large snapshot files. Using React Native, my team has built several components that make sense to be snapshotted directly, but inherit a very large DOM-like structure. For example, we built a styled number pad component that includes 10+ buttons and text components, and the styling to this file changes dramatically when we add a wrap around the buttons to achieve new design ideas.

Other situations we frequently see, especially with React Native is where just simple components for a back arrow button in a `Navigator` will be several layers of child components for rendering icons, text and flex information. Just the arrow button snapshot is ~100 lines, and it's practical to know that a line in this file is changing (i.e. the "Load Diff" button being visible), but it's not as necessary for my team to look at the change unless the diff on the Component being tested is not obvious.

cpojer:
If this is what you believe, I would recommend you and your team to use a custom extension for your browser to hide snapshots but I think it is entirely unreasonable to make this decision for the entire community against the intended use defined by Jest's core team, who built and maintains this feature.

In your argument, you are saying that even in large snapshots it is valuable to know what changed, which doesn't really add up if you are also saying that it will be optional for reviewers to look at them. I'm convinced nobody will end up looking at the snapshots and that many wrong snapshots will make it into your code repository.

hnryjms:
Aside to clarify:
> ...it's practical to know that a line in this file is changing (i.e. the "Load Diff" button being visible), but it's not as necessary for my team to look at the change unless the diff on the Component being tested is not obvious.

I'm glad this discussion is coming up, since I was unsure of both the processes of GitHub and Jest for PR's pertaining to community discussion (#3572 was my first commit to an OSS project affecting such a large developer user base), and I didn't know/still do not know of any channels to start the dialog about this code (other than Twitter 😛).

My process when reviewing code would be 1) Look at component changes, 2) Discuss any feedback, 3) Look at CI system pass/failure status 4) Expand the diff on the snapshot if I need a visual representation of the changes

For large engineering teams that have a separate QA Engineer and Dev Engineering team in projects, I see how it makes sense for these files to be expanded, but in a small startup with two or three people, where every contributor is building the component and the test all together, it's beneficial to focus the review efforts on the component and the unit tests themselves, rather than the lockfile defining the desired render state of a component (which for us is typically 2x+ in size when scrolling through GitHub's diff viewer).

Preferably GitHub's user preferences could control rules for expanded/contracted diff viewing based on comfort of the person/team; but I don't have insight into GitHub's longer-term design and functionality goals.

Daniel15:
Yeah I'm with @cpojer on this. We had the same thing happen internally (someone updated `.snap` files to be collapsed by default in Phabricator) and I reverted it.

Consider a basic function:
```js
function add(a, b) {
  return a + b;
}
```

And a corresponding unit test:
```js
test('adds 1 and 1', () => {
  expect(add(1, 1)).toEqual(2);
});
```

When reviewing this code, a code reviewer should not only review the implementation, but **also** review any assertions made during the unit test to ensure that they're valid. If the test said `expect(add(1, 1)).toEqual(3)`, a reviewer would rightly call it out ("1+1 is 2, not 3! This test is incorrect!")

However, let's change the test to be a snapshot test:
```js
expect(add(1, 1)).toMatchSnapshot();
```
It's a contrived example due to the small amount of output, but it still shows the issue: If you don't verify the assertions made during the unit test, you're not actually verifying that the test is even working as expected. What if the snapshot contained "3"? If it's collapsed by default, it's likely the reviewer won't even see the mistake. This is bad.

If your snapshot is very large, then that's potentially an issue with your test asserting too many different things. A unit test is supposed to just be testing one unit of code, and serializing a very deep React component heirarchy goes against that. If your snapshot is testing a deep React component tree, you probably want to shallow render the tree (for example, mock out the inner components, or use Enzyme's shallow renderer and use `enzyme-to-json` to serialize it). You may want to serialize the entire component tree for an integration test, but you should have many more unit tests compared to integration tests (and many more integration tests compared to end-to-end tests).

cpojer:
This is indeed nuanced. Regardless of what the right choice is, I propose this PR to be merged and the initial change to be reverted. This change is affecting everyone based on a singular opinion and little discussion. It feels to me that we are being charged and have to defend our innocence, rather than the reverse. The right way to go about this is to revert this PR, have a proper discussion on the Jest issue tracker about this with all the right parties involved and then come back with a result.

I think one thing we should do we can probably agree on is to syntax highlight snapshot files so that they provide *more* value during code review, not less like when we hide them.

hnryjms:
So I guess for me, I don't see contracted files on GitHub as having as much significance as @cpojer sees.

If you reference an NPM project using a Git URL/branch in a `package.json` file, and update to the latest commit on that branch (i.e. a new git hash), the `yarn.lock` file will be the only file with a diff, and Linguist will contract that diff even though the only change is a git has, AND that the change is very important to see. I think of `.js.snap` similarly, where the code inside it changing may be significant to review under some conditions, but depending on the sibling files around it also may not be as significant.

Because GitHub's code reviewer screen is a single page with all the files scrolling vertically down the page, the UX during a review involves an alphabetical review of the files not contracted and their diff, as opposed to a service like Atlassian BitBucket, where the viewer only shows one file at a time, and the developer can manually skip or view specific file diffs depending on what other changes are in the PR, and in the order that developer prefers (rather than generally alphabetically).

gensc004:
@Daniel15 That is a good example of what @cpojer has been expressing. I am a coworker of @hnryjms, and I think I have some insight as to why we have felt like reviewing the snapshots are unnecessary. In your example the developer is the one who wrote 3 as the answer to 1 + 1, but, in the case of snapshots, Jest is the thing the writes the answer. Since Jest is not a tool we wrote, it is not our job to ensure that it generates snapshots correctly. For us it has been sufficient to know the the snap was updated, and not necessarily know the contents of the snap file.  We also use a CI tool that asserts the tests that were changed passed, and we have an expectation of branches to pass all their tests before being opened in to a PR to begin with. Those two facts change the value snapshots have for us, and I am interested to see how the rest of the community values them in their projects.

I understand that we shouldn't push our preferences on the rest of the community, and I think @cpojer 's idea to merge this and start a conversation on the Jest Issue tracker, or reopen @hnryjms PR and reference it on the Jest Issue tracker, is a good idea. @hnryjms and I were not expecting the PR to be merged into the repo so quickly, and originally opened it so a discussion could be had.

Daniel15:
> In your example the developer is the one who wrote 3 as the answer to 1 + 1, but, in the case of snapshots, Jest is the thing the writes the answer. Since Jest is not a tool we wrote, it is not our job to ensure that it generates snapshots correctly

Your test code is generating the snapshot. Jest is simply writing it to a file. It's still your job to ensure the output of the test is correct, regardless of whether it's in a snapshot file or not. Fundamentally it's still a part of the test. 😃

rogeliog:
I agree with @cpojer... I think the content of snapshot files is super relevant, they contain the expectation of your test, a change in a snapshot file tells you that some behavior in your application has potentially changed.

I understand that there are cases that snapshot files are really large, but I think that in those cases it should be treated as a per case scenario and it should be up to the reviewer to decide if they just skim through it or not

hnryjms:
> it should be up to the reviewer to decide if they just skim through it or not
> @rogeliog 

But...since GitHub doesn't have any manual contract-a-filediff functionality, wouldn't the decision to review this file or not be based on whether the file gets expanded or not?

gensc004:
@Daniel15 , I think our feelings about this come from having a CI that tells us the tests passed. When reviewing a PR with the knowledge that the tests are passing, we can assume the snapshot tests correctly represent the code changes, and that reviewing the changes in the components and test files is enough without seeing the changes of the snapshot test. I can understand why reviewing the snapshots becomes necessary when you do not make the assumptions that the tests are passing. As @hnryjms said, lets get this merged and continue conversing if people want to 👍

Daniel15:
>  I think our feelings about this come from having a CI that tells us the tests passed. When reviewing a PR with the knowledge that the tests are passing, we can assume the snapshot tests correctly represent the code changes,

The test passing doesn't mean that the test is exhibiting correct behaviour though. It's possible to have a test where the snapshot is totally wrong, much like it's possible to have 100% code coverage without performing any assertions 😃

gensc004:
So if the test file is written correctly, the snapshot file shows as being updated, I have reviewed the changes to the components made in the PR, and the tests related to the changes are passing, what could be wrong with the snapshot that is worthy of reviewing? I guess under these circumstances (which I understand are specific to me :P ) I don't see what I am missing by not reviewing the snapshot.

nickmccurdy:
@gensc004 Snapshots are test code, even though they're generated. As a result, they need to be reviewed in PRs for false positives like other tests. Even if you review a PR that changes the code's implementation so that it would output a different snap shot and you see the snapshot has changed, you still need to read the snapshot diff to ensure that the snapshot has changed appropriately and isn't introducing false positives. If you updated snapshots every time you ran your Jest tests, your snapshots would always pass, but that doesn't mean the output they're expecting is accurate.

blogcommoda:
OK

lildude:
Thanks for bringing this up and clarifying things and bringing this discussion to the table @cpojer.  Would you mind accepting the GitHub CLA and I can go ahead and merge this.

cpojer:
done!

PR #3580, opened 2017-04-22
Add .admx and .adml as extensions for XML

MichaelGrafnetter:
ADMX and ADML files are XML-based and are [used by Windows](https://technet.microsoft.com/en-us/library/cc772507(v=ws.10).aspx) to create a user interface for Group Policy settings. There are approximately [1.5K such files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aadmx+NOT+nothack&type=Code&ref=searchresults) at GitHub.

The samples are sourced from [Microsoft Download Center](https://www.microsoft.com/en-us/download/details.aspx?id=48257).

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4691,8 +4691,8 @@ XML:
   - wsdl
   extensions:
   - ".xml"
-  - ".admx"
   - ".adml"
+  - ".admx"
   - ".ant"
   - ".axml"
   - ".builds"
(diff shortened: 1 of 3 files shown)


lib/linguist/languages.yml
```diff
@@ -4691,6 +4691,8 @@ XML:
   - wsdl
   extensions:
   - ".xml"
+  - ".admx"
```
pchaigno: `.admx` should be after `.adml` in the list.

PR #3584, opened 2017-04-25
[FOR DISCUSSION]: Hide Jest Snapshot Text Files

hnryjms:
- [ ] **[WIP]: Confirm community discussion before merging**

Originally merged #3572, reverted #3579 for future discussion.

Facebook employees @cpojer and @Daniel15 had voiced valid concerns when the original PR for this code change had been merged to instigate a larger community discussion around Jest Snapshot files being contracted in Linguist (and therefore contracted by default in the GitHub code review tool), so I am reopening this PR to host the discussion for the larger community.

Some discussion already has landed on #3579 but I am hoping it can continue via this PR, and links can be posted around the Jest community to voice their own thoughts beyond a small subset of power-users of the project.


Daniel15:
For reference, here's my original comments on the revert diff:

---

Yeah I'm with @cpojer on this. We had the same thing happen internally (someone updated `.snap` files to be collapsed by default in Phabricator) and I reverted it.

Consider a basic function:
```js
function add(a, b) {
  return a + b;
}
```

And a corresponding unit test:
```js
test('adds 1 and 1', () => {
  expect(add(1, 1)).toEqual(2);
});
```

When reviewing this code, a code reviewer should not only review the implementation, but **also** review any assertions made during the unit test to ensure that they're valid. If the test said `expect(add(1, 1)).toEqual(3)`, a reviewer would rightly call it out ("1+1 is 2, not 3! This test is incorrect!")

However, let's change the test to be a snapshot test:
```js
expect(add(1, 1)).toMatchSnapshot();
```
It's a contrived example due to the small amount of output, but it still shows the issue: If you don't verify the assertions made during the unit test, you're not actually verifying that the test is even working as expected. What if the snapshot contained "3"? If it's collapsed by default, it's likely the reviewer won't even see the mistake. This is bad.

If your snapshot is very large, then that's potentially an issue with your test asserting too many different things. A unit test is supposed to just be testing one unit of code, and serializing a very deep React component heirarchy goes against that. If your snapshot is testing a deep React component tree, you probably want to shallow render the tree (for example, mock out the inner components, or use Enzyme's shallow renderer and use `enzyme-to-json` to serialize it). You may want to serialize the entire component tree for an integration test, but you should have many more unit tests compared to integration tests (and many more integration tests compared to end-to-end tests).

---

> In your example the developer is the one who wrote 3 as the answer to 1 + 1, but, in the case of snapshots, Jest is the thing the writes the answer. Since Jest is not a tool we wrote, it is not our job to ensure that it generates snapshots correctly

Your test code is generating the snapshot. Jest is simply writing it to a file. It's still your job to ensure the output of the test is correct, regardless of whether it's in a snapshot file or not. Fundamentally it's still a part of the test. 😃 

---

>  I think our feelings about this come from having a CI that tells us the tests passed. When reviewing a PR with the knowledge that the tests are passing, we can assume the snapshot tests correctly represent the code changes,

The test passing doesn't mean that the test is exhibiting correct behaviour though. It's possible to have a test where the snapshot is totally wrong, much like it's possible to have 100% code coverage without performing any assertions 😃  "Test passing" and "code is covered by tests" aren't meaningful without also ensuring that the test is behaving correctly.

---

cpojer:
Again, I'm voicing my strong opinion that I think this is not the right way to go about this. @Daniel15 pointed out the reasons quite well. I have not heard anything convincing to sway me the other way. The reasonable conclusion I would take away from the discussion so far is that large snapshots aren't great and we are exploring the answers in this discussion thread: https://github.com/facebook/jest/issues/2197

hnryjms:
Here's my original reasoning for opening this PR:

---

Contracted diff's on GitHub don't mean or represent the code shouldn't or doesn't need to be reviewed, but rather it makes the first pass of scrolling through GitHub's viewer easier, as the content is minimized and can then be expanded later. This reduces simple wrapping-DOM-element PR's in the amount of scroll volume during a review immensely, but the diff can still be expanded with the "Load Diff" button during review.

Per @pchaigno in response to initial revert discussion:

> I see your point, but don't all committed files need to be reviewed in any case? IMO, generated files are hidden from diffs to make those more readable (I was under the impression that these files can be huge). You can still review them.

pedrottimark:
@hnryjms Hello, there is a sample project to show how in the future you might be able to reduce the size of snapshots and move them from a separate file into the main test file.

Start reading here: https://github.com/pedrottimark/whimper#interlude-about-snapshot-testing
and then look at these examples of “subset” tests with `toMatchObject` assertion:
* https://github.com/pedrottimark/whimper#create
* https://github.com/pedrottimark/whimper#update-or-undo

Remaining work is outside the control of Jest project, but we are following through with others.

Even if you cannot use this type of testing when it becomes possible, imagine that subset testing existed before snapshot testing. Does that make it clearer why some commenters have made such a strong case why the `.snap.js` file is an essential part of the criterion for snapshot tests

Whenever a snapshot test fails, you have to choose:

* Prevent unexpected **regression**. If change is incorrect, then fix code.
* Confirm expected **progress**. If change is correct, then update snapshot.

hnryjms:
^ My team stopped using Snapshot testing, so I'm going to close this PR, and if someone else stumbles across this...there are many approaches to testing and everyones arguments here are valid—there's no one-size-fits-all for test file code review.

Issue #3568, opened 2017-04-13
C repo partially misclassified as C++


rafael-santiago:
Hi there,

One of my repositories has only [C code](https://github.com/rafael-santiago/kryptos/) but the Linguist is pointing also C++ sources.

The repo has some macros that could remember a "CPP from stone age" but I am not tracking any ``.C|.CC|.cc|.[cC][pP][pP]`` file just ``.c`` files. Considering this, maybe the heuristics could be improved.

lildude:
> Considering this, maybe the heuristics could be improved.

I think you're right on this point. We'll happily accept a PR for these improvements.

rafael-santiago:
Okay. I proposed some improvements [here](https://github.com/github/linguist/pull/3585).

gdonald:
Similar problem here:

https://github.com/gdonald/cpama/search?l=c%2B%2B

No C++ files or .hh file extensions in my repo but quite a few .h files are being mis-labled as C++.  Github Support asked me to post my info here.

PR #3585, opened 2017-04-28
Fine-tuning on heuristics related with the issue#3568 (C/C++ header files)

rafael-santiago:
Hi there,

I have done some improvements on disambiguation for ".h" files trying to solve this [issue](https://github.com/github/linguist/issues/3568). 

I also moved one file from "samples/C++/Field.h" to "samples/C/Field.h", if you analyse this file isolated, it is more like an C source file or at least it is a inconclusive sample.

I think that cases like the old "sample/C++/Field.h" should be solved with a mechanism like a white list but I do not know the ``Linguist`` source code enough to know if it would be really possible.


lib/linguist/heuristics.rb
```diff
@@ -68,6 +68,20 @@ def call(data)
     # Common heuristics
     ObjectiveCRegex = /^\s*(@(interface|class|protocol|property|end|synchronised|selector|implementation)\b|#import\s+.+\.h[">])/
 
+
+    CPPRegexType0 = /^\s*#.*include.*<(stdint|cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream).*>/
```
Alhadis: 👋  Hey @rafael-santiago, thanks for the contribution.

Using a multiline/expanded regex is preferable to having multiple, smaller expressions. Could you combine them, please?

Additionally, this should go inside the relevant heuristic, as they're only used once. We have the `ObjectiveCRegex` assigned to a variable because it's shared by two different tests (`.h, .m`).

rafael-santiago: Hello, @Alhadis 

So just confirming:

Did you mean split the regexes over multiple lines:

```
%r {
(...)
}x

```
or suggested capturing groups?

I think that use capturing groups could obfuscate the code and lead to misunderstandings in the future besides the performance cost.

Alhadis: The former. If readability is a concern, be liberal with whitespace and/or comments. :)

Alhadis: Erm, to clarify: I meant "whitespace and/or comments" within the actual regex.

[Here's a good example](https://github.com/github/linguist/blob/c8094d3/lib/linguist/strategy/modeline.rb#L4-L93) of what I'm talking about.

rafael-santiago: Hi @Alhadis, I have applied the  suggested improvements excepting the regex splitting.

The splitting was influencing on tests (breaking them). The first regex related with common CPP includes is a Frankenstein. I tried %r{(...)}x, ~= /.../ and nothing. Maybe it could be something related with my ruby copy but I am not sure.

Alhadis: @rafael-santiago Try this:

~~~rb
CPPRegex = /
  ^ \s*
  (?: template \s* <
    | \#.*include.*<
      (stdint|cstdint|string|vector|map|list
      |array|bitset|queue|stack|forward_list
      |unordered_(map|set)|(i|o|io)stream).*>
    )
  |
  ^ [\x20\t]*
  (?: try \b
    | catch \s *\(
    | ( class
      | namespace
      | using \h+ namespace
      )
      \s+ \w+
    | (private|public|protected) :$
    )
  | std::\w+
  | .*\#ifdef.*__cplusplus.*
  | ^struct\s+
/xi
~~~

For regex with lots of nested capturing groups, it's much easier to maintain and review when they're indented.

When writing expressions in "expanded form" this way, be sure to escape all whitespace and `#` characters you're matching! Tabs and spaces are ignored, so they need to be written like `\x20` (space) and `\t` (tab). `#` characters are used for comments *inside* the regular expression: that's what I meant earlier before.

I don't think comments are necessary in this case: in fact, they're actually making things *harder* to follow. Stuff like URLs and sources are better off included in pull requests. ;)

pchaigno:
@rafael-santiago Thanks for your work on this!

I'm a bit worried about the impact of these changes on performance. Could we evaluate that on a corpus of `.h` C files?

rafael-santiago:
@pchaigno I have just added 5 smaller regexes. I am afraid of doing them all (even combined with non-capturing groups) because it could stress the ruby regex system. I have no idea how internally it is implemented in ruby, however in other Languages the suggestion is execute smaller regexes to do the whole search. Considering the short circuit used on if-branches (and operator) it could avoid this overhead because it would process only the essential for that specific data. Well, who knows...

Alhadis:
@rafael-santiago It's generally faster to allocate a single, fixed-size unit of memory than several. That being said, it's easier for a single, large expression to be optimised in one compilation unit.

Ruby's [regex engine](https://github.com/kkos/oniguruma) is already tightly-wired for performance. You shouldn't have to worry about what goes on under the hood - this isn't C or assembly. Dynamic programming languages exist to take care of heavy lifting.

rafael-santiago:
@pchaigno, @Alhadis 

Okay, I used your suggested code snippet. Running it twice (``time rake``):

1st:
real 1m25.195s
user 1m20.960s
sys 0m1.836s

2nd:
real 1m21.810s
user 1m17.852s
sys 0m1.772s

The original PR code (C, Assembly old approach):

1st:
real 1m21.718s
user 1m17.632s
sys 0m1.796s

2nd:
real 1m20.875s
user 1m16.572s
sys 0m2.108s

Of course that we should run it several times to get a good average but I really do not have time for this.

@Alhadis  I have pushed your snippet (now this is breaking with "pqiv.h"), if you want to solve and use it feel free.

pchaigno:
@Alhadis Does your script to retrieve sample files on github.com still works with the new page layout? I'd like to check the performance impact on a "representative" corpus of files.

Alhadis:
If it did, I'd be damn surprised. This was exactly what I expected to see when I tried:

Figure 1

I'll need to rewrite it, I'm afraid. This time it might be better to run it within a [headless browser](http://phantomjs.org/), which will make running the script *much* easier.

GitHub's still limiting our filetype-searches to logged-in users only, so it's gonna be interesting dealing with 2FA...

pchaigno:
Hm, thanks. I'll see if I can fix @larsbrinkhoff's.

Alhadis:
@pchaigno I [fixed the script](https://github.com/Alhadis/Harvester/commit/a5dcaab28da807ebed7ab436fa699b1d744d3800). I have this aggravating tendency to assume problems are 10 times more complicated than they really are, because I'm too distracted by something else to actually bother investigating. I drive myself nuts like that.

Seriously though, I'd really love for this to be rewritten and reimplemented in a form that doesn't... well, suck as bad as it does now.

pchaigno:
> I also moved one file from "samples/C++/Field.h" to "samples/C/Field.h", if you analyse this file isolated, it is more like an C source file or at least it is an inconclusive sample.

Could you revert this? If it is inconclusive, then I'd rather we don't change its classification for now.

lib/linguist/heuristics.rb
```diff
@@ -204,11 +204,36 @@ def call(data)
...
+
       if ObjectiveCRegex.match(data)
         Language["Objective-C"]
-      elsif (/^\s*#\s*include <(cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream)>/.match(data) ||
-        /^\s*template\s*</.match(data) || /^[ \t]*try/.match(data) || /^[ \t]*catch\s*\(/.match(data) || /^[ \t]*(class|(using[ \t]+)?namespace)\s+\w+/.match(data) || /^[ \t]*(private|public|protected):$/.match(data) || /std::\w+/.match(data))
-        Language["C++"]
+      elsif CPPRegex.match(data)
+           Language["C++"]
+      else
+        Language["C"]
```
pchaigno: That's going to classify way too many files as C. I'd prefer that we let the classifier decide, even if it is a bit arbitrary for now.

pchaigno:
I've evaluated the performance impact on a corpus of 11,523 `.h` files downloaded from the following search queries using [Harvester](https://github.com/Alhadis/Harvester):
- [`extension:h`](https://github.com/search?q=extension%3Ah+NOT+dfhdfdhfdghjd)
- [`extension:h+-language:C`](https://github.com/search?q=extension%3Ah+-language%3AC+NOT+dfhdfdhfdghjd)
- [`extension:h+-language:Objective-C`](https://github.com/search?q=extension%3Ah+-language%3AC+NOT+dfhdfdhfdghjd)
- [`extension:h+-language:C++`](https://github.com/search?q=extension%3Ah+-language%3AC+NOT+dfhdfdhfdghjd)

I ran Linguist 10 times on a repository consisting of all these files (including duplicates), both with and without the present changes, as currently implemented.
In average, without this changes, Linguist takes 24s to run on the repository, whereas it takes 30s to run with them. The standard deviation over the 10 runs is low (< 1s). Removing the fallback case on C is likely to impact performance, so I'll have to run the test again.

So that's 25% longer. But. This is a huge number of `.h` files for a single repository (Linux has only 4 times as much). The overhead is likely to be much smaller for a normal size repository (especially since we update statistics incrementally) as the time to run heuristic rules would be a smaller portion of the total time and only that time is affected.

I'm flagging this as stale. @rafael-santiago Please tell us if you still intent to work on this.

PR #3611, opened 2017-05-06
Don't show npm lockfiles by default

SimenB:
Like this: https://github.com/npm/node-fetch-npm/commit/efbc7d5bfac8c80f499b15636a05f0a2a39cdb91

/cc @zkat

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -57,7 +57,7 @@ module Linguist
       composer_lock? ||
       node_modules? ||
       go_vendor? ||
-      npm_shrinkwrap? ||
+      npm_shrinkwrap_or_package_lock? ||
       godeps? ||
       generated_by_zephir? ||
       minified_files? ||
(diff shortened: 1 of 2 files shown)


lib/linguist/generated.rb
```diff
@@ -330,7 +330,8 @@ def go_vendor?
     #
     # Returns true or false.
     def npm_shrinkwrap?
-      !!name.match(/npm-shrinkwrap\.json/)
+      return false unless name.match(/npm-shrinkwrap\.json/)
+      return false unless name.match(/package-lock\.json/)
```
pchaigno: I think the second line should be `!!name.match(/package-lock\.json/)`.

SimenB: Ok, changed it

lib/linguist/generated.rb
```diff
@@ -330,7 +330,8 @@ def go_vendor?
     #
     # Returns true or false.
     def npm_shrinkwrap?
-      !!name.match(/npm-shrinkwrap\.json/)
+      return false unless name.match(/npm-shrinkwrap\.json/)
+      !!name.match(/package-lock\.json/)
```
pchaigno: Oh, and the first line should be `return true if name.match(/npm-shrinkwrap\.json/)`, otherwise it's equivalent to a logic AND. You could also write `name.match(/package-lock\.json/) || name.match(/npm-shrinkwrap\.json/)` by the way.

SimenB: Updated to the one-liner 😄

lib/linguist/generated.rb
```diff
@@ -330,7 +330,7 @@ def go_vendor?
     #
     # Returns true or false.
     def npm_shrinkwrap?
```
pchaigno: The name and comment should probably change too.

SimenB: Done 😄

PR #3638, opened 2017-05-24
adding kdb to languages.yml and samples/kdb

simongarland:
for a collection of kdb repositories see:

http://kxsystems.github.io

http://code.kx.com/q/github/

sample search for in the wild repositories:

user:AquaQAnalytics extension:q

user:psaris extension:q

the colour choice is used on kx.com and code.kx.com, we've used it or a similar dark blue since the early '90's.


lildude:
Thanks for the new PR, though you could have carried on in the original 😃 

Your test failure is due to your colour choice:

```
  1) Failure:
TestColorProximity#test_color_proximity 
[/home/travis/build/github/linguist/test/test_color_proximity.rb:21]:
The following 1 languages have failing color thresholds. Please modify the hex color.
- kdb (#0070CD) is too close to ["4A76B8", "3572A5", "0070CD"]
```

You've been pipped on the colour choice by other languages.  You're going to need to tweak the colour choice to make it more distinct.

My only other concern with this PR is the prevalence of the `.q` and `.k` extensions (there are tons of 'em) which are _not_ currently classified are most definitely _not_ kdb and this could result in those being incorrectly identified.

simongarland:
that’s pretty difficult - there are 214 unique colours in languages.yml
already
couple that with this test complaining about ”color proximity” and I could
be here for a while
trying to find one of the last remaining dark blues that’s not also ”near”
one in use.

(existing duplicates: color:"#28431f” and color:"#e4cc98”)

Do you have a list of unused colours that are also not near any existing?
If so I could pick one of the last
remaining dark blues from there

the .k and .q extension frequency - I can see that could be a problem. How
have you dealt with this in the
past? Quite a lot of them may well be ours - but for sure not all. There’s
no other easy test I can think
of - we don’t, for example start every file with a hash-bang, or have some
common construct like #include

lildude:
Sorry about the delay in getting back to you @simongarland.

I understand your frustration surrounding picking a colour... how about using the colour proposed in 
https://github.com/github/linguist/pull/3659 which has since been opened for "q" which appears to be the same as what you're including here. (I've suggested that PR be closed in favour of this one).

> the .k and .q extension frequency - I can see that could be a problem. How
> have you dealt with this in the
> past?

I'm not sure, but have raised the same concern, and a possible solution (use a heuristic with a fallback to "Text") in https://github.com/github/linguist/pull/3689#pullrequestreview-47059649. Will see what others suggest.

If we do go this route, we'll need to come up with a good reliable way of identifying kdb files.

simongarland:
that colour choice in #3659 would be fine

recognising kdb files isn’t easy as there’s nothing like a hash-bang at the
beginning, but there is something
easy to do that would pick up 99.x% of the files. Any .k or .q file is
going to contain function definitions and
that means almost all of them will match the string ”:{[” - that’s assign a
value ”:”, followed by open
function definition ”{”, followed by open argument definition ”[”

an example being:

plus:{[x;y] x+y}

I’m not sure how that’d fit in with your framework, but it should be
easier/faster than a regex to match

abalkin:
Sorry for fueling a trivial discussion, but as an author of [PyQ (Python for kdb+)](https://pyq.enlnt.com/), I have a project that will be affected by a conflict between q and Python color codes.

While 0040cd passes test/test_color_proximity.rb, it is still fairly close to Python's `#3572A5` color:

screen shot 2017-08-23 at 4 39 06 pm

I would propose using something like `#050cb5` instead:

screen shot 2017-08-23 at 4 39 40 pm

I checked that `#050cb5` passes the proximity test by running

```bash
$ bundle exec rake test TEST=test/test_color_proximity.rb
Run options: --seed 16421
# Running:
.
Finished in 0.792886s, 1.2612 runs/s, 1.2612 assertions/s.
1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

With respect to language detection issue, I don't see why resolving it should be a blocker for accepting this PR. Can we start by registering `.kdb` extension for the k language and `.kdb+` for q?  Individual projects can always add `*.k` and `*.q` to custom overrides.  See  #3792.

lildude:
@simongarland...

> easy to do that would pick up 99.x% of the files. Any .k or .q file is
> going to contain function definitions and
> that means almost all of them will match the string ”:{[” - that’s assign a
> value ”:”, followed by open
> function definition ”{”, followed by open argument definition ”[”
> 
> an example being:
> 
> plus:{[x;y] x+y}
> 
> I’m not sure how that’d fit in with your framework, but it should be
> easier/faster than a regex to match

Ooo, you might be onto something here. Looking at the examples in this PR,  https://github.com/github/linguist/pull/3659 and https://github.com/github/linguist/pull/3820, I can see that basic format in all of them, except your `sudoku.k` sample. This is definitely something that can be used for a heuristic.

A quick bit of testing with `/^\w+:{[^\[]+\[/` matches all the `.k` and `.q` files, except your `sudoku.k` sample, included in the three PRs with only one false positive being `samples/JavaScript/jquery-1.4.2.min.js` which wouldn't ever get to the heuristic for `.k` or `.q` files.

This is definitely "close enough" and you can easily override your `sudoku.k` once this lands.

Wanna have a go at implementing a heuristic using this?

@abalkin...

> Sorry for fueling a trivial discussion, but as an author of PyQ (Python for kdb+), I have a project that will be affected by a conflict between q and Python color codes.

Hmmm, I'm not sure this is going to have that much of an impact; the colours are quite distinct and there are already several other languages with colours closer than this project. As I mentioned in https://github.com/github/linguist/pull/3815, blue is incredibly popular so conflicts _will_ happen and whilst we may be able to accommodate you now (it's up to @simongarland and @dstrachan) we almost certainly won't be able to in future as the availability of shades of blue dwindles.

lildude:
Not sure where my mind was yesterday but don't worry about the heuristics I suggested... this isn't going to help as they only come into play when there are two or more identified languages with the same extension.

simongarland:
closing, superseded by psaris' request #3820

PR #3643, opened 2017-05-25
Update MD5 digest for testing under Ruby 2.4

lildude:
Ruby 2.4 deprecated Fixnum & Bignum by unifying them into Integer (see https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/). This means the MD5 digests for the integers in our tests have a class of Integer under 2.4.0 and later instead of Fixnum from earlier releases which means we need to update the digests specifically for Ruby >= 2.4.0.  This PR does just that.

I didn't go the route of modifying Linguist::MD5.hexdigest as any mention of `Fixnum` or `Bignum` results in a deprecation warning which gets annoying fast 💨.

/cc @github/data-engineering

### DIFF
--- a/test/test_md5.rb
+++ b/test/test_md5.rb
@@ -17,7 +17,7 @@ class TestMD5 < Minitest::Test
 
   def test_hexdigest_integer
     # Ruby 2.4.0 merged Bignum and Fixnum into Integer which means we get different digests
-    if RUBY_VERSION >= "2.4.0"
+    if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.4.0")
       assert_equal "eb70b9dc51f70acc91b4d984ef81570e", MD5.hexdigest(1)
       assert_equal "41f415a79361937c28fff32c2c6d056d", MD5.hexdigest(2)
     else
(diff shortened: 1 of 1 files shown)


lildude:
Behold... it's a thing of beauty:

screen shot 2017-05-25 at 15 51 00

We've not had this one green for a very very long time.

test/test_md5.rb
```diff
@@ -38,16 +44,30 @@ def test_hexdigest_nil
 
   def test_hexdigest_array
     assert_equal "4410ec34d9e6c1a68100ca0ce033fb17", MD5.hexdigest([])
-    assert_equal "759f28c1d0c20c22e79c91d339855d95", MD5.hexdigest([1])
-    assert_equal "9efe07b352dd94cd1cdc9d1a8d054f8f", MD5.hexdigest([1, 2])
-    assert_equal "60b1ef7201404f20c3d12f47499c3a1f", MD5.hexdigest([1, 2, 3])
-    assert_equal "2f82cdc166616208077ff1dd0a8faeff", MD5.hexdigest([1, 2, [3]])
+    # Ruby 2.4.0 merged Bignum and Fixnum into Integer which means we get different digests
+    if RUBY_VERSION >= "2.4.0"
```
Caged: This string compare will fail in some common cases:

```
>> "2.3.2" > "2.3.10"
=> true
```
You can use `Gem::Version` to work around that:

```
>> Gem::Version.new("2.3.2") > Gem::Version.new("2.3.10")
=> false
```

lildude: TIL. I nearly did it this way but saw we were using it on github/github so thought it would do. Will change to be safe.

PR #3659, opened 2017-06-01
Added q language

dstrachan:
q language entry added, along with a sample .q file.

[Sample search](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aq+kx&type=Code)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3518,17 +3518,6 @@ QMake:
   - qmake
   ace_mode: text
   language_id: 306
-q:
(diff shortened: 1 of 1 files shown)


lildude:
Hmmm, this appears to be the same thing as being proposed in https://github.com/github/linguist/pull/3638, but there the language is referred to as "kdb".

I think it may be best to continue this work in https://github.com/github/linguist/pull/3638 - an alias for "q" can be added that PR.

lildude:
Closing as q support has landed with #4058

PR #3660, opened 2017-06-02
Added syntax definition for Fantom language

rkoeninger:
[Fantom](http://fantom.org) was a pre-existing language in linguist.

Added [rkoeninger/sublime-fantom](https://github.com/rkoeninger/sublime-fantom) as a submodule
Added license for `sublime-fantom`
Specified `tm_scope` for Fantom
Provided better color for Fantom derived from official logo

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -854,3 +854,6 @@
 [submodule "vendor/grammars/language-webassembly"]
 	path = vendor/grammars/language-webassembly
 	url = https://github.com/Alhadis/language-webassembly
+[submodule "vendor/grammars/sublime-fantom"]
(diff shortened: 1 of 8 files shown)


pchaigno:
@rkoeninger You'll also need to update the submodule to point to your fork's last commit.

rkoeninger:
Submodule updated.

pchaigno:
Do you have any sample Fantom file from a real application under permissive license we could use (`samples/Fantom/`)? We currently don't have any sample for Fantom :/

rkoeninger:
Added a couple of samples under very permissive license.

PR #3662, opened 2017-06-04
Added Ring Language Support 

MahmoudFayed:
Language (GitHub) : https://github.com/ring-lang/ring

Language Page (Wikipedia) : https://en.wikipedia.org/wiki/Ring_(programming_language)

Language in TIOBE Index (TOP 100, In group 51 to 100) : https://www.tiobe.com/tiobe-index/

Language Website : http://ring-lang.net 

Search : https://github.com/search?p=100&q=extension%3Aring+see&ref=searchresults&type=Code&utf8=%E2%9C%93 

Please Add It.


Alhadis:
Hi @MahmoudFayed,

You should be dealing with submodules locally, not through GitHub. If you're having trouble sorting through all these conflicts, it may be easier for you to simply start again with a freshly-cloned fork.

MahmoudFayed:
Hi @Alhadis 

Thank you very much, I have already solved all these conflicts.

.gitmodules
```diff
@@ -658,7 +658,7 @@
 	url = https://github.com/gregory-m/ejs-tmbundle
 [submodule "vendor/grammars/nix"]
 	path = vendor/grammars/nix
-	url = https://github.com/wmertens/sublime-nix
+	url = git@github.com:wmertens/sublime-nix.git
```
pchaigno: We prefer HTTPS links. We've had issues with SSH before.

samples/Ring/weblib.ring
```diff
@@ -0,0 +1,3039 @@
+Import System.Web
```
pchaigno: This file is huge. @lildude Will that be an issue?

samples/Ring/hello.ring
```diff
@@ -0,0 +1 @@
+See "Hello, World"
```
pchaigno: The sample files are used for tests, but also to train the Bayesian classifier. We therefore prefer sample files from "real" applications (under permissive license).

lib/linguist/languages.yml
```diff
@@ -3736,6 +3736,15 @@ RenderScript:
   tm_scope: none
   ace_mode: text
   language_id: 323
+Ring:
+  type: programming
+  color: "#0e60e3"
+  extensions:
+  - .ring
+  - .rh
```
pchaigno: We'll also need search results demonstrating usage on github.com for this file extension.

.gitmodules
```diff
@@ -854,6 +854,9 @@
 [submodule "vendor/grammars/language-webassembly"]
 	path = vendor/grammars/language-webassembly
 	url = https://github.com/Alhadis/language-webassembly
+[submodule "vendor/grammars/language-ring"]
+	path = vendor/grammars/language-ring
+	url = https://github.com/MahmoudFayed/atom-language-ring
```
pchaigno: Your grammar has a few compatibility issues with the PCRE regex engine GitHub uses when interpreting grammars. In particular, it does not recognize `\h` and `{,n}` (replace with `[0-9A-Fa-f]` and `{0,n}` respectively).

Alhadis: Well, it actually *does* recognise `\h` sequences. ;) They just have an entirely unrelated meaning: horizontal whitespace characters. Spaces and tabs, but not newlines or carriage returns, et cetera.

Bane of my regex-obsessed existence.

.gitmodules
MahmoudFayed: @pchaigno Thanks, I have already updated the file.

samples/Ring/hello.ring
MahmoudFayed: @pchaigno Thanks, The file is updated to include a sample from "real" applications under permissive license.

.gitmodules
MahmoudFayed: @pchaigno  @Alhadis Thanks, The grammar file is updated.

lib/linguist/languages.yml
MahmoudFayed: @pchaigno Thanks, The search results
(1) https://github.com/search?p=100&q=extension%3Aring+see&ref=searchresults&type=Code&utf8=%E2%9C%93
(2) https://github.com/search?utf8=%E2%9C%93&q=*.ring&type=
(3) https://github.com/search?utf8=%E2%9C%93&q=*.rh&type=

pchaigno: [I can't find many `.rh` Ring files](https://github.com/search?utf8=%E2%9C%93&q=extension%3Arh+func&type=Code) (those are from a Python dialect it appears).

MahmoudFayed: @pchaigno Yes, the *.rh files are rarely used to define some constants instead of directly adding them to *.ring files.

samples/Ring/weblib.ring
MahmoudFayed: @pchaigno I have already reduces the file size to keep the sample (Small)

lib/linguist/languages.yml
MahmoudFayed: @pchaigno I have already removed *.rh (rarely used), To keep only *.ring

.gitmodules
pchaigno: You'll also need to update the reference commit in this pull request.

MahmoudFayed: @pchaigno Thanks, I have already updated the reference commit in this pull request.

samples/Ring/weblib.ring
lildude: Yes, this file is still too large and contains a lot of data that, whilst relevant to a real-world use, is not relevant to the structure of the language. For example the `htmlcssattributes` is waaaaaaay bigger than is really needed: 65 lines for a single variable 😱!!

There's also a fair amount of "duplication" in the huge number of similar functions.

My naïve "I know nothing about ring" point of view believes this can be cut down quite considerably, or replaced with a smaller, more representative, example.

MahmoudFayed: @lildude @pchaigno  Thanks, I have already updated the file.

samples/Ring/weighthistory
```diff
@@ -0,0 +1,174 @@
+Load "guilib.ring"
```
pchaigno: This file is missing an extension.

MahmoudFayed: @pchaigno  Thanks I have already added the extension to the file name.

Issue #3663, opened 2017-06-04
Cannot run linguist binary from repository


JonasT:
I cannot run the `linguist` tool from the repository as instructed:

```
[jonas@falcon linguist]$ bundle install
Resolving dependencies...
Using rake 12.0.0
Using public_suffix 2.0.5
Using byebug 9.0.6
Using charlock_holmes 0.7.3
Using coderay 1.1.1
Using color_difference 0.0.1
Using escape_utils 1.1.1
Using multipart-post 2.0.0
Using mime-types-data 3.2016.0521
Using rugged 0.25.1.1
Using thor 0.19.4
Using metaclass 0.0.4
Using method_source 0.8.2
Using minitest 5.10.2
Using plist 3.3.0
Using slop 3.6.0
Using yajl-ruby 1.3.0
Using bundler 1.12.5
Using addressable 2.5.1
Using color-proximity 0.2.1
Using faraday 0.12.1
Using mime-types 3.1
Using licensee 8.8.5
Using mocha 1.2.1
Using pry 0.10.4
Using sawyer 0.8.1
Using github-linguist 5.0.11 from source at `.`
Using octokit 4.7.0
Using licensed 0.6.0
Bundle complete! 11 Gemfile dependencies, 29 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
[jonas@falcon linguist]$ bundle exec linguist --breakdown
bundler: failed to load command: linguist (/home/jonas/.gem/ruby/2.3.0/bin/linguist)
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/jonas/Develop/linguist/lib/linguist/samples.json
  /home/jonas/Develop/linguist/lib/linguist/samples.rb:24:in `read'
  /home/jonas/Develop/linguist/lib/linguist/samples.rb:24:in `cache'
  /home/jonas/Develop/linguist/lib/linguist/language.rb:505:in `<module:Linguist>'
  /home/jonas/Develop/linguist/lib/linguist/language.rb:18:in `<top (required)>'
  /home/jonas/Develop/linguist/lib/linguist.rb:5:in `require'
  /home/jonas/Develop/linguist/lib/linguist.rb:5:in `<top (required)>'
  /home/jonas/Develop/linguist/bin/linguist:5:in `require'
  /home/jonas/Develop/linguist/bin/linguist:5:in `<top (required)>'
  /home/jonas/.gem/ruby/2.3.0/bin/linguist:23:in `load'
[...]

PR #3669, opened 2017-06-07
Improve running from cloned repo docs

lildude:
As pointed out in https://github.com/github/linguist/issues/3663, it's not clear that you need to generate the `samples.json` file before you can run Linguist from the cloned repo. This is probably missed by most who run straight into running the tests as the Rake task updates the samples for you.

This PR adds the missing step and advises that this needs to be rerun every time there is a change to any of the samples if the tests aren't being run.

Fixes https://github.com/github/linguist/issues/3663

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,9 +67,11 @@ For development you are going to want to checkout out the source. To get it, clo
     cd linguist/
     script/bootstrap
 
-If you are going to run Linguist from the cloned repository directory, you will need to generate the code samples first:
+To run Linguist from the cloned repository directory, you will need to generate the code samples first:
 
     bundle exec rake samples
+
+If you are not running the tests below, re-run this command every time you modify any of the [samples][samples].
 
 To run Linguist from the cloned repository directory:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,6 +67,14 @@ For development you are going to want to checkout out the source. To get it, clo
     cd linguist/
     script/bootstrap
 
+If you are going to run Linguist from the cloned repository directory, you will need to generate the code samples first:
+
+    bundle exec rake samples
+
+To run Linguist from the cloned repository directory:
+
+    bundle exec bin/linguist --breakdown
+
 To run the tests:
 
     bundle exec rake test
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,13 +67,13 @@ For development you are going to want to checkout out the source. To get it, clo
     cd linguist/
     script/bootstrap
 
-To run Linguist from the cloned repository directory, you will need to generate the code samples first:
+To run Linguist from the cloned repository, you will need to generate the code samples first:
 
     bundle exec rake samples
 
-If you are not running the tests below, re-run this command every time you modify any of the [samples][samples].
+Run this command each time a [samples][samples] has been modified.
 
-To run Linguist from the cloned repository directory:
+To run Linguist from the cloned repository:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -67,6 +67,16 @@ For development you are going to want to checkout out the source. To get it, clo
     cd linguist/
     script/bootstrap
 
+To run Linguist from the cloned repository directory, you will need to generate the code samples first:
```
Alhadis: **Nitpick:** *"cloned repository directory"* is a bit redundant. We can shorten that to simply "cloned repository" instead.

CONTRIBUTING.md
```diff
@@ -67,6 +67,16 @@ For development you are going to want to checkout out the source. To get it, clo
...
+
+    bundle exec rake samples
+
+If you are not running the tests below, re-run this command every time you modify any of the [samples][samples].
```
Alhadis: > *"If you are not running the tests ~~below~~"*

Unless there's a need to be specific, and unless there's another set of tests in Linguist which aren't relevant to the current paragraph, we can 🔥  that.

> "re-run this command every time you modify any of the [samples][samples]."

This sentence feels a bit... clunky to me. I'd word that like this:

> "run this command each time a [sample] has been modified:"

Alhadis: Since `re-run` implies the user has already run the command, and the sentence makes the assumption the user modified a sample file, as opposed to a process or what-have-you.

Alhadis:
Is there no way for the samples to be automagically updated beforehand?

lildude:
Thanks for the feedback @Alhadis 

> Is there no way for the samples to be automagically updated beforehand?

Good point. I think we could add it to `script/bootstrap` though I'm not entirely sure why it's not already done there. I can't help but feel there is a reason for this.

Another option would be to add `bundle exec rake check_samples` to `bin/linguist` though I'd like that task to be enhanced further to automatically check for changes to `samples` before doing so.

Thoughts?

lildude:
> Another option would be to add `bundle exec rake check_samples` to `bin/linguist` though I'd like that task to be enhanced further to automatically check for changes to samples before doing so.

Thinking about it further... this probably isn't the best idea as it'll run each time this is run when shipped as part of the gem. Yes, it'll be a noop, but it would still take time to check the samples dir for changes.  It could also potentially lead to peeps like me forgetting to update the samples when shipping a new gem 😊.

Alhadis:
I'd add it to the bootstrap step, personally. Though if Linguist is savvy enough to tell you when the samples database is out-of-date, it's probably savvy enough to do its own damn chores. 😁

Alhadis:
**Random rambling:** I have `bundle exec rake` [aliased to `brake`](https://github.com/Alhadis/.files/blob/4ba0559dde24a851533b37d5bd50f493a616fad3/aliases.sh#L4).

Spelling aside, it's hilarious when you run `brake test` and end up annihilating every test because you forgot to `brake` some samples beforehand.

I highly recommend everybody adopt my sense of humour, the world is much more fun this way.

Alhadis:
> Good point. I think we could add it to `script/bootstrap` though I'm not entirely sure why it's not already done there. I can't help but feel there is a reason for this.

@pchaigno You're the oldest guy here (as measured in "Linguist-years"): if there was a reason for not including sample regeneration in the bootstrapping process, I'm sure you're closer to remembering than I am. 😞 

Of course, it's also possible that previous maintainers shared our feelings of uncertainty, which is ultimately the only reason it's not been added. 😀

lildude:
Gonna merge this now. We can investigate doing this automatically in a separate PR.

PR #3672, opened 2017-06-08
Fix spelling of Perl 6

### LINKED ISSUE #3671: Rename Perl6 to Perl 6
The official name of the language is "Perl 6".

Alhadis:
Resolves #3671.

I'm surprised it's taken that long for anybody to notice, because nowhere is Perl 6 written as `Perl6` in the official docs. Heck, the Camel Book calls it Perl 6 too:

> The third edition was published in the middle of 2000, just as Perl v5.6 was coming out. As we write this, it’s 12 years later and Perl v5.16 is coming out soon. A lot has happened in those years, including several new releases of Perl 5, and a little thing we call **Perl 6**. That 6 is deceptive though; **Perl 6** is really a “kid sister” language to Perl 5, and not just a major update to Perl 5 that version numbers have trained you to expect. This book isn’t about that other language. It’s still…


Alhadis:
I'd really love to just
~~~perl
while(<languages_yml>){
	s/Perl6/Perl 6/g;
	print;
}
~~~

PR #3676, opened 2017-06-13
Updated CONTRIBUTING

anteriovieira:


pchaigno:
I'd prefer to keep it in its original form. I don't expect potential contributors to read the whole document. Many might only read one of the sections.

anteriovieira:
Hi @pchaigno , I considered the redundancy and the fact that a note draws attention to reading.
But I also agree with your point of view. Maybe just add it as note then?

lildude:
I'm :-1: on this PR. It is best to keep the message in the appropriate section. As it's the first sentence in the section, I don't think it needs more in the way of highlighting either.

Issue #3144, opened 2016-08-04
GitHub Code highlighting for React file with js extension


ktalebian:
If you create a React file but save it with `.js` extension, then GitHub code highlighting breaks when it reaches a single quote in text. Saving the same file in `.jsx` works fine.

See example on [this repo](https://github.com/ktalebian/react-highlighting-problem). Both file `correct-highlighting.jsx` and `incorrect-highlighting.js` have the same content. As you can, the `.js` has wrong syntax highlighting.

A solution is to add `// vim: syntax=JSX` to the top of the file, but this will get annoying really fast. Ideally, Linguist is intelligent to determine that the file is a React file and applies correct syntax.

muescha:
same here:
https://github.com/keybase/client/blob/master/shared/nav.android.js#L113

pchaigno:
We'd first have to add `.js` as a [JSX extension](https://github.com/github/linguist/blob/b958779e3d7364b89a988c0b41dfc012ad7fd450/lib/linguist/languages.yml#L1985-L1994) for Linguist to be able to distinguish the two. Please see the [guidelines to add support for a new extension to Linguist](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language).

PR #3677, opened 2017-06-14
Linguist should understand JSX in .js files

buzinas:
[Search for `.js`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ajs+NOT+nothack&type=Code) / [Search for `.jsx`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ajsx+NOT+nothack&type=Code)

[Search for `.js` with "return \<div>"](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ajs+return+%3Cdiv%3E&type=Code) (the `<` characters are ignored, but you can see that many people uses JSX in `.js` files by opening a few pages).

Please, let me know what I can do to help getting this PR merged at some point.

Refs #3144

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1975,6 +1975,7 @@ JSX:
   type: programming
   group: JavaScript
   extensions:
+  - ".js"
   - ".jsx"
   tm_scope: source.js.jsx
   ace_mode: javascript
--- a/samples/JSX/sample.js
+++ b/samples/JSX/sample.js
@@ -1,0 +1,23 @@
+'use strict';
+
+const React = require('react')
+
+module.exports = React.createClass({
+  render: function() {
+    let {feeds, log} = this.props;
+
+    log.info(feeds);
+    return <div className="feed-list">
+      <h3>News Feed's</h3>
+      <ul>
+        {feeds.map(function(feed) {
+          return <li key={feed.name} className={feed.fetched ? 'loaded' : 'loading'}>
+            {feed.data && feed.data.length > 0 ?
+              <span>{feed.name} <span className='light'>({feed.data.length})</span></span>
+              : 'feed.name' }
+          </li>
+        })}
+      </ul>
+    </div>;
+  }
+});


pchaigno:
Do you want the JSX portions of `.js` files to be included in statistics, or are you trying to fix the syntax highlighting?

If you're trying to recognize the JSX portions, that is not currently supported by Linguist and would require major changes. Currently, Linguist assigns a single language per file.

If you want to fix the syntax highlighting, then you should submit a pull request to [atom/language-javascript](https://github.com/atom/language-javascript), which github.com uses to highlight JavaScript code.

Alhadis:
Uh, there's nothing wrong with the syntax highlighting. At all.

JSX is *not* JavaScript.

pchaigno:
Hm, I read this a bit too quickly :/

@buzinas So you're point is that many JSX files have a `.js` extension and we should therefore try to recognize this extension for JSX?

lib/linguist/languages.yml
```diff
@@ -1975,6 +1975,7 @@ JSX:
   type: programming
   group: JavaScript
   extensions:
+  - ".js"
```
lildude: Is `.js` really the primary extension for JSX? If not, please move this below `.jsx` as the first extension is considered the primary extension.

buzinas:
> JSX is not JavaScript.

@Alhadis JSX is not JavaScript, agreed - but **millions** of people write JSX using the `.js` extension

> So you're point is that many JSX files have a .js extension and we should therefore try to recognize this extension for JSX?

@pchaigno yeah, exactly that!

lib/linguist/languages.yml
buzinas: Currently, it's kind of impossible to do that research on Github because it ignores some important characters - but from my knowledge much more people use `.js` extension for writing JSX code than `.jsx`.

If you go to [React's official repo examples](https://github.com/facebook/react/wiki/Examples), you'll see that all of them uses JSX in `.js` files (and the entire ecosystem does - Redux etc). Based on that, I assume that if the creators of JSX use the `.js` extension, it's the primary extension for the language. But I can't ensure that.

lildude: :+1: Thanks for the confirmation.

pchaigno:
Should we add a heuristic to distinguish JS and JSX files? I'm afraid the classifier alone might not do a very good job.

buzinas:
@pchaigno yeah, that sounds reasonable.

Can you help me by giving the directions on how to do that?

pchaigno:
You can follow the examples in [`heuristics.rb`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb). You'll also need to add a test for that new heuristic in [`test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb).

The heuristic itself doesn't need to be perfect. It should be correct when it identifies a file as being JSX (or JS), but it doesn't need to identify *all* JSX (or JS) files.

lib/linguist/languages.yml
buzinas: Btw, another useful info: https://github.com/facebookincubator/create-react-app/issues/87

lildude:
Just a friendly nudge.

buzinas:
Can anyone jump in and help me with adding the heuristic? I'm having a hard time figuring out how to identify if a file is JSX or JS.

lildude:
I don't know anything about JSX but I'm pretty good at pattern matching 😉. From a cursory glance at https://reactjs.org/docs/jsx-in-depth.html and files already on GitHub, it appears JSX files will almost always include `React` somewhere in them. Either in the form of something like `React.createElement` or `import React from 'react';` or `React = require('react')`, though this last form does bring up a lot of files with just that in them, which is perfectly valid javascript and not really JSX from my understanding, so is not really a good indicator of JSX content.

I've also noticed that some form of `render .... return <[htmltaghere]>` or similar statement occurs in the JSX files I've found so that may be a better multiline string to match for in a regex.

exalted:
My two cents: in a JavaScript file, you could consider JSX anywhere you find a `<tag …/>` or `<tag …>…</tag>` that otherwise could be invalid JavaScript syntax (i.e. `SyntaxError: Unexpected token '<'`).

For example in JavaScript `const hello = <h1>Hello, world!</h1>;` would be invalid, but it’s perfectly valid JSX.

I think it’s virtually impossible to determine JSX based on it’s location (whether inside a component's `render()` function, in a stateless function, in a variable assignment, or something like `ReactDOM.render(<h1>Hello, world!</h1>, document.getElementById('root'));`, etc.); instead JSX, as being a syntax **extension** to JavaScript, it could be detected/treated as such.

buzinas:
@exalted that's what I was trying, but the problem is: there can be a `<tag>` inside a string, and it would be a false positive, since it's a perfectly valid JavaScript.

exalted:
Maybe it's worth looking at how JSX is detected to be then converted into regular JavaScript in tools like babel?

exalted:
@buzinas I don’t think you can look for `<tag>` in a JS file and call it a day.

buzinas:
That's another thing that I tried, but since Babel parses the JavaScript code as an AST, that would be too complex for a heuristic.

exalted:
I am not sure what do you mean by heuristic in this context, however I don’t think a RegEx approach suffice, sadly.

I haven‘t worked with linguist, so I shouldn’t judge but, if an AST is not a viable option here, I think what we have left with is to take a best guess approach. I am sure that doesn't hold for long though and you’d have to add *many* edge cases to _guess_ better and better…

lildude:
Nudge.

zacanger:
Hey, just a friendly bump. I currently have PRs into private projects with `js linguist-language=JSX` in gitattributes in the hope that will fix highlighting in diffs, but having this in would be much nicer/less hacky.

Just a suggestion, but maybe detecting either an `import` or a `require` of React and _also_ any xml-like tag would be enough? React has to be imported or required in any file that has JSX, so that seems reasonably safe.

reyronald:
Jumping in here.

I believe Linguist can already syntax JSX files correctly with no problems, as you know, everything inside a JSX files is valid JavaScript, **_except_** for the HTML-like tags, so... would it be too crazy of an idea to use the already working JSX grammar to parse JS files by default instead and call it a day? I think it would have no actual implications, correct me on this?

In fact, this is what a lot of devs are doing manually now like @zacanger just stated, not only for Linguist, but for Atom too (see https://discuss.atom.io/t/how-do-i-make-atom-recognize-a-file-with-extension-x-as-language-y/26539?u=wliu).

By the way, I'm coming from here: https://github.com/atom/language-javascript/issues/220#issuecomment-360553820, in which I made a similar argument, just in case anyone wants to check that out.

lildude:
> would it be too crazy of an idea to use the already working JSX grammar to parse JS files by default instead and call it a day?

I don't think so, there is however one teeeeny weeeny heeeawg fat problem... https://github.com/github/linguist/issues/3044

In short, the version of language-babel we currently use is pinned to a really old version with a known issue because the maintainer changed some of the regexes in the grammar to those only supported by the Oniguruma engine used in Atom, but not GitHub.com, and isn't prepared to entertain GitHub.com compatibility because "[It works in atom...](https://github.com/github/linguist/issues/3044#issuecomment-224547519)"

If you know of a method of automatically converting the Oniguruma-only regexes to be PCRE-friendly or know another PCRE-friendly grammar that is actively maintained (we looked into switching to https://github.com/babel/babel-sublime but found it lacking - see https://github.com/github/linguist/issues/3775) we'll definitely entertain the solution to our current problem so we can consider your suggestion.

JonAbrams:
I think we can all agree that it's at least common, if not _standard_, for projects to have jsx code in `.js` files. Every code editor that supports JSX, supports this just fine. What's the downside of adding JSX support for `.js` files?

One reason why projects haven't adopted `.jsx` is due to the fact that codebases often have a mix of source code files, some with JSX, some not. It's a pain to have to change the file's extension when JSX happens to be added to an existing source file. Plus, it complicates scripts to have glob/regexes that match both `.js` and `.jsx` (I do realize the irony here).

Can we plz merge this for the benefit of people that love both JSX and Github? Arguing over whether `.jsx` or `.js` is more popular, and how to measure it, is pure bike-shedding.

Alhadis:
> @Alhadis mentioned in #4030 (Comment) that he's working on such a JS/JSX grammar, but that's obviously a huge task and I don't know if he found time to make progress.

Not just a huge task, but slow as well. Much of this has to do with the fact that Atom doesn't live reload grammars as they're being worked on, reducing me to reloading the entire workspace to see changes. This cumbersome workflow, in addition to the predicted complexity of the grammar code, has motivated me to start work on a compiler for an intermediate grammar format optimised for reading/writing regexp-based grammars.

The long-term goal is to have a maintainable format which can be kept updated by the Babel maintainers, the TC39 committee, or whoever else is guaranteed to keep it updated. The compiler won't be limited to TextMate-style grammars alone, meaning it'll generate output for CodeMirror and Pygments as well.

If I were to continue the way I started, with frequent repetition of CSON blocks and hacky idioms, I 100% guarantee the grammar would become unmaintainable in future.

Qix-:
For what it's worth, Facebook (more specifically, Dan Abramov) decrees that the `.jsx` extension is no longer relevant since the pre-Babel days. His reasoning is pretty solid.

https://github.com/facebook/create-react-app/issues/87#issuecomment-234627904

This should happen as `.js` files with JSX are currently monotone, of course. Having the JSX statistic would also be nice, but honestly since Javascript is an amalgamation of bizarre syntax features and proposals and whatnot, the statistics become a bit less important in my opinion.

Alhadis:
I'm not sure why we're even bothering to differentiate between JS and JSX anymore, actually. Since it falls under the usage statistics of the former, we may as well roll it into the `JavaScript` entry and spare ourselves the hassle over semantics.

Since JSX is a superset of JavaScript, it should be possible for one grammar to accommodate both JS and sugary extensions like embedded XML and type annotations. Moreover, Linguist has no way of knowing if something like this is a JS or JSX file:

~~~js
export const foo = "Valid ES6";
~~~

So in a large codebase, many files are likely to be flagged as ordinary JS because they don't use tag syntax, meaning users end up with a jarring classification like "20% JavaScript, 80% JSX" or what-have-you. And Linguist has no way of identifying files based on the presence of another language in the repository (which would be helpful, as it would vastly improve its interpretation of C++/C).

Alhadis:
Self-assigning this to shut Stalebot up. 👍 See #4358

Qix-:
Thank you @alhadis :)

lildude:
This is now resolved by moving to using the treesitter JavaScript grammar GitHub uses in https://github.com/github/linguist/pull/5133. GitHub.com will reflect this after the next release due some time this month.

Closing.

PR #3691, opened 2017-06-23
Improve Pan language support

jrha:
An atom package providing grammar for Pan is now available and can be used to once again provide highlighting for Pan (which stopped being supported after the move away from Pygments).

While we're working on this, we can add a more representative sample of valid Pan source files to the samples directory.


samples/Pan/ca.pan
```diff
@@ -0,0 +1,235 @@
+#
+# Generated by RepositoryTask on 11/10/14 12:21 PM
```
lildude: This is a pretty big file with almost identical data repeated within the single `contents` dict: over 100 almost identical uncommented lines 😱!!  

Whilst this may be relevant to a real-world use, is not relevant to the structure of the language. Could you simplify this file or find a smaller real-world example?

jrha: Sure! I've removed all but one of this type as they are normally produced by automated tasks.

lildude:
Thanks for the smaller example. I've taken a better look at the other samples, and whilst a lot of samples are great, many of these are duplicating quite a lot of the same structures within the language... several of the samples are also almost identical, eg `cluster-A.pan` and `cluster-B.pan`.

Whilst we require real-world examples of files, we don't need quite so many files.  Are you able to cut down the number of files so we've got a reduced amount of duplication?

jrha:
No worries. I've cut down the number of examples, what is left should cover the majority of commonly used language features in a mixture of older and more modern coding styles.

PR #3696, opened 2017-06-27
Classify some project files as XML

am11:
Also added disambiguation rule for `.proj` and `.user`.

##### CSCFG

https://github.com/search?utf8=%E2%9C%93&q=extension%3Acscfg+NOT+nothack&type=Code
(16.7K hits)

##### CSDEF

https://github.com/search?utf8=%E2%9C%93&q=extension%3Acsdef+NOT+nothack&type=Code
(12.7K hits)

##### CCPROJ

https://github.com/search?utf8=%E2%9C%93&q=extension%3Accproj+NOT+nothack&type=Code
(5K hits)

##### DEPPROJ

https://github.com/search?utf8=%E2%9C%93&q=extension%3Adepproj+NOT+nothack&type=Code
(505 hits)

##### NDPROJ

https://github.com/search?utf8=%E2%9C%93&q=extension%3Andproj+NOT+nothack&type=Code
(194 hits)

##### PROJ

https://github.com/search?utf8=%E2%9C%93&q=extension%3Aproj+%28project+OR+Property+OR+Import+OR+xml+OR+xmlns%29&type=Code
(35K hits)

##### SHPROJ

https://github.com/search?utf8=%E2%9C%93&q=extension%3Ashproj+NOT+nothack&type=Code
(13K hits)

<del>
##### USER

https://github.com/search?utf8=%E2%9C%93&q=extension%3Auser+%28project+OR+Property+OR+Import+OR+xml+OR+xmlns%29&type=Code
(693K hits)
</del>


am11:
@pchaigno, can you please review this? 😊

am11:
@Alhadis, is it ok to combine similar types in one PR? Alternatively, I can split them into separate ones. There are so many `*proj` types that are classify as XML.

Another option is that we disambiguate based on `.*proj` (suffix) pattern and make one rule matching `/^(\s*)(<Project|<Import|<Property|<?xml|xmlns)/i` to classify the XML. After that we can remove all those `*proj` listed under XML node in languages.yml to reduce the volume.

Alhadis:
> @Alhadis, is it ok to combine similar types in one PR?

Yes, it's okay. :) Don't worry, the way you've arranged this PR is perfectly acceptable: it's consistently XML-related.

Now, I'm not sure it'd be wise to register an extension as generic as `.user` as being an XML type... as you might expect from the [sheer volume of results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Auser+%28project+OR+Property+OR+Import+OR+xml+OR+xmlns%29&type=Code), there are probably numerous unrelated files with names ending in `.user`. Adding this extension would probably lead to a lot of false positives, so I'd recommend dropping it, or refining it to be more specific (`.xproj.user`, `.csproj.user`, etc...)

@lildude I'm still of the belief that XML deserves a separate classification mechanism based on doctype. The over-fragmentation of different file extensions is getting way out of hand...

lib/linguist/heuristics.rb
```diff
@@ -379,6 +379,14 @@ def call(data)
       end
     end
 
+    disambiguate ".proj" do |data|
+      if /^(\s*)(<Project|<Import|<Property|<?xml|xmlns)/i.match(data)
+        Language["XML"]
+      else
+        Language["Text"]
```
Alhadis: This won't work because the `.proj` extension hasn't been registered as a filetype for the `Text` language. I believe heuristics are only for differentiating *shared* file extensions (used by more than one language type).

In other words, there's no need for this heuristic whatsoever.

lib/linguist/heuristics.rb
```diff
@@ -491,5 +499,14 @@ def call(data)
         Language["XML"]
       end
     end
+
+    disambiguate ".user" do |data|
```
Alhadis: See my comment above... as well as the one below.

am11: I agree *.user is quite broad. Does it make sense to register it under Text as well in languages.yml and keep the XML disambiguation? In future, folks will add other (else) branches to this block for varied languages. If it incurs performance penalties in platform or there are other reasons, I am all for removing it. :)

lib/linguist/heuristics.rb
am11: Removed.

lib/linguist/heuristics.rb
Alhadis: No, actually, I'd leave `.user` out as a registered file extension altogether. Unless you can link to some *canonical* documentation where that extension is promulgated as official.

am11: Removed.
I found this document https://docs.microsoft.com/en-us/cpp/ide/working-with-project-properties suggesting:

> Tip
The .user file is a legacy feature and we recommend that you delete it in order to keep properties correctly grouped according to configuration/platform. 

Then it explains in `Important` quote under [this section](https://docs.microsoft.com/en-us/cpp/ide/working-with-project-properties#bkmkPropertySheets) why it's a bad idea to use such global settings per-user.

I think it will make sense to skip it as the vendor is not happy with it either :smile:

Alhadis: Yep, superb idea. Let's pretend it was never brought up. 😝 

LGTM. /cc @lildude for final review

lildude:
> @lildude I'm still of the belief that XML deserves a separate classification mechanism based on doctype. The over-fragmentation of different file extensions is getting way out of hand...

I agree, that list is becoming heeeawg and I imagine it'll keep getting bigger.

Gonna approve this PR in the meantime.

PR #3698, opened 2017-06-27
Add C sample

DavidAylaian:
Sample was incorrectly being identified as C++

### DIFF
--- a/samples/C/asm.h
+++ b/samples/C/asm.h
@@ -1,6 +1,6 @@
 /* CarbonOS System/Kernel
- * Copyright 2015-2017 David Aylaian
- * Licensed under GitHub CLA: https://cla.github.com/agreement
+ * Copyright 2015-2017 David Aylaian, No Rights Reserved
+ * Licensed under the Unlicense: http://unlicense.org
  */
 
 #ifndef ASM_H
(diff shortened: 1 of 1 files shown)


samples/C/asm.h
```diff
@@ -0,0 +1,27 @@
+/* CarbonOS System/Kernel
+ * Copyright 2015-2017 David Aylaian
+ * Licensed under GitHub CLA: https://cla.github.com/agreement
```
pchaigno: I don't think you can license a file under a CLA...

DavidAylaian: I'll put it under the unlicense then.

Alhadis: Well, that works, but a [permissive license](https://github.com/github/linguist/blob/9289704/vendor/licenses/config.yml#L2-L11) (e.g., MIT/Expat-compatible) is really all we need. =) It is of course your personal choice, however.

Alhadis: I also noticed that the project the code was originally authored for is [licensed under the Apache 2.0 License](https://github.com/DavidAylaian/CarbonOS/blob/6dbd6ec/LICENSE.md). Which, [as you'll see](https://github.com/github/linguist/blob/master/vendor/licenses/config.yml#L2), is an acceptably permissive license. =)

In other words, you didn't have to relicense anything. ;-)

DavidAylaian: @Alhadis Ok, sounds great. I wasn't sure if Apache 2.0 was an acceptably permissive license.

PR #3706, opened 2017-07-03
Add support for XPM

smola:
* .xpm and .pm extensions associated with XPM.

* .pm is disambiguated by searching the /* XPM */ string.
  This is how `file` performs detection and should work with
  every XPM3 file (most XPM generated by software later than 1991).

Added XPM samples:

* stick-unfocus.xpm: extracted from Fluxbox (MIT License)
  https://github.com/fluxbox/fluxbox/blob/0c13ddc0c85ed4445ff8e289387c142e0ac98c77/data/styles/Emerge/pixmaps/stick-unfocus.xpm

* cc-public_domain_mark_white.pm: public domain image from
  https://commons.wikimedia.org/wiki/File:Cc-public_domain_mark_white.svg
  converted to XPM with ImageMagick (convert input.svg output.xpm).


smola:
I couldn't find a specific grammar for this. Note that it uses C syntax, so for syntax highlighting purposes it should be treated just like C.

lib/linguist/languages.yml
```diff
@@ -4883,6 +4883,14 @@ XML:
   - Web.config
   - packages.config
   language_id: 399
+XPM:
+  type: data
+  extensions:
+  - ".xpm"
+  - ".pm"
+  ace_mode: c_cpp
+  tm_scope: none
```
pchaigno: To treat this as C during syntax highlighting, you should write `source.c` here. You can use [Lightshow](https://github-lightshow.herokuapp.com/) to check that it works as expected.

smola:
@pchaigno Done.

Issue #3589, opened 2017-04-29
Classifier gives false positive if no heuristic gives a positive (and giving negative impossible)


veox:
This has been extracted (and somewhat expanded) from [a comment](https://github.com/github/linguist/pull/3560#issuecomment-293001855) in PR #3560. See that for context.

-----

### Set in question

As commit d852415204baaa94d53ac9fd98d7dbe1c873e230 notes, and as is discussed in PR #2634, there are [many files with the `.sol` extension](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+NOT+nothack&type=Code).

Known categories:

1. Solidity.
1.1. Modern uses can be recognised by use of `pragma solidity` ([search](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+%22pragma+solidity%22&type=Code)).
1.2. Older files (written before the introduction of `pragma`) can be rec-d by use of `contract` keyword ([search](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+contract&type=Code)), but this is incomplete. Different from what was written in PR #2634, `contract` keyword is not mandatory, and may not be present for libraries - they will have `library` instead.
2. Eagle (electronics CAD) solder mask (or similar), rec-d by ending in `M02*` ([search](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+M02*&type=Code)).
3. Solution files to [UVa Online Judge](https://uva.onlinejudge.org/) problems, e.g. [this one](https://github.com/thnkndblv/UVa-Online-Judge-Solutions/blob/88f534303d2aaccf22950f8927ca2fda8fb4a64d/10892%20-%20LCM%20Cardinality/p10892.sol). The site seems to be a library of computational problems with means for automated solution submission. The files in question mostly contain space-separated numbers. This extension/format is not official: their [submission specification page](https://uva.onlinejudge.org/index.php?option=com_content&task=view&id=15&Itemid=30) states that no file read/writes are to be performed, so the `.sol` is most likely a user convention to present results more easily to each other.
4. Other? Drawing CAD solids? Solar flare measurements? SoLame scripts?..

### Classifier gives false positive if no heuristic gives a positive

Using the current approach with extension-based classifiers and content-based heuristics, and implementing them naïvely as in PR #3560, gives the following results.

(1.1) and (2) are properly detected as (respectively) `Solidity` and `Eagle` by the simple heuristic included in this PR.

[...]

PR #3707, opened 2017-07-03
improve .ms disambiguation (Unix Assembly / MAXScript)

smola:
A few MAXScript files were misclassified as Unix Assembly.
Some of them can be found at github.com/davestewart/maxscript

* This commit changes the heuristic which looked for labels
  such as ".LG7E0:" to actually match the full label including
  the colon. This reduced the number of MAXScript files
  misclassified as Unix Assembly, without any new Unix Assembly
  misclassified so far.

* add MAXScript sample rolloutCreator.ms, extrated from MIT repo:
  https://github.com/davestewart/maxscript/blob/master/3D/3ds2ae/02_resources/max%20scripts/3ds%20ax%20scripts/rolloutCreator.ms

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -297,7 +297,7 @@ module Linguist
     disambiguate ".ms" do |data|
       if /^[.'][a-z][a-z](\s|$)/i.match(data)
         Language["Roff"]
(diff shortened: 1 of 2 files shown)


smola:
There are still some files misclassified as MAXScript for languages that are not supported yet (MethodScript and mason-lang). I guess those could be fixed once something like https://github.com/github/linguist/issues/3589 is in place

Alhadis:
Ugh, I should clean up MAXScript's syntax highlighting... :( Was the first language-grammar I ever wrote, pretty rough compared to my later work... :\ Fun learning curve, though. ^^

PR #3708, opened 2017-07-03
shell: add more extensions and interpreters

smola:
* ash: only interpreter, extension is more commonly used for
  Kingdom of Loathing scripting, e.g. github.com/twistedmage/assorted-kol-scripts

* dash: only interpreter, extension is more commonly used for
  dashboarding-related stuff

* ksh: extension was already present

* mksh

* pdksh

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4122,7 +4122,12 @@ Shell:
   - PKGBUILD
   - gradlew
   interpreters:
+  - ash
   - bash
+  - dash
+  - ksh
+  - mksh
+  - pdksh
   - rc
   - sh
   - zsh
--- a/test/fixtures/Shell/ash-env
+++ b/test/fixtures/Shell/ash-env
@@ -1,0 +1,2 @@
+#!/usr/bin/env ash
+echo "ash"
--- a/test/fixtures/Shell/dash-env
+++ b/test/fixtures/Shell/dash-env
@@ -1,0 +1,2 @@
+#!/usr/bin/env dash
+echo "dash"
--- a/test/fixtures/Shell/ksh-env
+++ b/test/fixtures/Shell/ksh-env
@@ -1,0 +1,2 @@
+#!/usr/bin/env ksh
+echo "ksh"
--- a/test/fixtures/Shell/mksh
+++ b/test/fixtures/Shell/mksh
@@ -1,0 +1,2 @@
+#!/bin/mksh
+echo "mksh"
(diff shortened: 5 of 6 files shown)


samples/Shell/mksh-env
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env mksh
```
Alhadis: According to @mirabilos, this line should be used:

~~~sh
#!/bin/mksh
~~~

> “This search” is broken, the standard shebang is `#!/bin/mksh`,
> only some (predominantly older) scripts use `#!/usr/bin/env mksh`.
>
> Please make sure both are recognised for scripts.

smola: :+1: Updated.

samples/Shell/ash-env
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env ash
+echo "ash"
```
pchaigno: We prefer to have samples from "real" applications (under permissive license). Alternatively, you can move those to `tests/fixtures/`.

lib/linguist/languages.yml
```diff
@@ -4110,6 +4110,8 @@ Shell:
   - ".command"
   - ".fcgi"
   - ".ksh"
+  - ".mksh"
+  - ".pdksh"
```
pchaigno: I'm not sure there are enough [`.mksh`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Amksh+NOT+kdjskfjdjf&type=) or [`.pdksh`](https://github.com/search?utf8=%E2%9C%93&q=extension%3Apdksh+NOT+dkfjdkfkd&type=) files on github.com to warrant their addition as file extensions. Adding the shebang might be sufficient anyway...

mirabilos: Neither pdksh used to nor mksh uses the shell name as script extension, we use just `.sh` or no extension in the general case, `.ksh` is seen rarely and more for AT&T ksh scripts.

smola: * 13 mksh https://github.com/search?utf8=%E2%9C%93&q=extension%3Amksh+mksh&type=
* 14 pdksh https://github.com/search?utf8=%E2%9C%93&q=extension%3Apdksh+pdksh&type=

Indeed, it's pretty rare. Removed.

samples/Shell/ash-env
smola: I can move both these and the previous ones (`bash` and `dash`) which I used as template to `tests/fixtures`. Although I'm not sure which test would be checking it?

pchaigno: https://github.com/github/linguist/blob/37979b26b04e10868017469e5cc56263b0a39c84/test/test_file_blob.rb#L640-L664 tests fixture files.

smola:
@pchaigno Done.

PR #3709, opened 2017-07-03
fix classifier tests

smola:
test_classify_ambiguous_languages was not running any test, since
it was looking only for languages that are ambiguous on
filename for known filenames (rather than ambiguous for filename
or extension).

Note that test time and assertions.
Before:
  Finished in 0.149294s, 40.1892 runs/s, 46.8874 assertions/s.
After:
  Finished in 3.043109s, 1.9717 runs/s, 224.7702 assertions/s.

### DIFF
--- a/test/test_classifier.rb
+++ b/test/test_classifier.rb
@@ -47,7 +47,10 @@ class TestClassifier < Minitest::Test
     Samples.each do |sample|
       language  = Linguist::Language.find_by_name(sample[:language])
       languages = Language.find_by_filename(sample[:path]).map(&:name)
-      next unless languages.length > 1
+      next if languages.length == 1
+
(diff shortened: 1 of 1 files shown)


test/test_classifier.rb
```diff
@@ -47,6 +47,9 @@ def test_classify_ambiguous_languages
     Samples.each do |sample|
       language  = Linguist::Language.find_by_name(sample[:language])
       languages = Language.find_by_filename(sample[:path]).map(&:name)
+      next unless languages.length != 1
```
lildude: Double negatives are confusing. Can you switch this to `next if languages.length == 1`

smola:
@lildude Done.

PR #3710, opened 2017-07-03
fix Coq sample JsNumber.v

smola:
It was fetched as HTML from GitHub instead of raw.


smola:
:+1: I'm using this branch with cross-validation for the classifier to catch this kind of problem (or more generally, languages that might need more samples): https://github.com/smola/linguist/tree/x-validation

PR #3715, opened 2017-07-08
Added required file extension for Modula-2 and sample files.

trijezdci:
Modula-2 libraries consist of a definition module (file extension .def) and an implementation module (file extension .mod). This is comparable to .h and .c files in C and C++. Linguist not recognising .def, but .mod alone for Modula-2 would be like not recognising .c for C and C++. This patch adds (1) an entry to recognise ".def" as Modula-2 to languages.yml and (2) a def/mod sample file pair in the samples/modula-2 directory.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2665,6 +2665,7 @@ Modelica:
 Modula-2:
   type: programming
   extensions:
+  - ".def"
   - ".mod"
   tm_scope: source.modula2
   ace_mode: text
--- a/samples/Modula-2/Newline.def
+++ b/samples/Modula-2/Newline.def
@@ -1,0 +1,18 @@
+(*!m2iso*) (* Copyright (c) 2017 Modula-2 Software Foundation *)
+
+DEFINITION MODULE Newline;
+
+(* Newline mode management *)
+
+TYPE Mode = ( LF, CR, CRLF );
+
+
+PROCEDURE SetMode ( mode : Mode );
+(* Sets the default newline mode. *)
+
+
+PROCEDURE mode ( ) : Mode;
+(* Returns the default newline mode. *)
+
+
+END Newline.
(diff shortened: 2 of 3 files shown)


trijezdci:
FYI, the license on the two samples files is LGPL-2.1. This can be verified by looking them up in the github repo https://github.com/m2sf/m2pp which shows the licensing.

https://github.com/m2sf/m2pp/blob/master/src/Newline.def
https://github.com/m2sf/m2pp/blob/master/src/imp/Newline.mod

Alhadis:
👋 Hey @trijezdci,

> FYI, the license on the two samples files is LGPL-2.1. This can be verified by looking them up in the github repo https://github.com/m2sf/m2pp which shows the licensing.

Can you find another sample? Preferably one released under an Expat/MIT-compatible license? We generally prefer samples which are covered by a [whitelisted license](https://github.com/github/linguist/blob/24e196d/vendor/licenses/config.yml#L2-L12). Since LGPL isn't mentioned on that list, I'm skeptical as to whether even a "lesser" form of the GPL can be accepted. 

@mlinksva, could you clarify? The [Wikipedia page for LGPL](https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License) makes it sound acceptably permissive:

> The license allows developers and companies to use and integrate software released under the LGPL into their own (even proprietary) software without being required by the terms of a strong copyleft license to release the source code of their own components. The license only requires software under the LGPL be modifiable by end users via source code availability.

trijezdci:
The LGPL is not "lesser". It is more permissive than the GPL. It isn't viral if the code itself is not going to be modified. For the purpose of using it as a sample to test syntax highlighting, deliberately chosen to be extremely simple, this practically matches the MIT license because nobody is going to ever change the code in the sample. There isn't anything in there. Nor is the code going to be compiled to do anything or fix bugs. In fact the files are so short that under copyright law they are by themselves not capable of receiving any protection, You have to have a certain size and complexity for something to be copyrightable. Otherwise you could copyright the sentence "That will be 2 dollars" and ask every shop to pay license fees to you when they charge a customer 2 dollars for a snack or whatever. Don't be so nitpickety on something completely and utterly irrelevant.

PR #3716, opened 2017-07-08
Added required file extension for Modula-2

trijezdci:
Modula-2 libraries consist of a definition module (file extension .def) and an implementation module (file extension .mod). This is comparable to .h and .c files in C and C++. Linguist not recognising .def, but .mod alone for Modula-2 would be like not recognising .h files for C and C++.

This patch adds:

(1) an entry for the .def file extension under Modula-2 in languages.yml
(2) a def/mod sample file pair in the samples/Modula-2 subdirectory

The sample files are part of a github project by the Modula-2 Software Foundation which is LGPL 2.1 licensed . The files have been relicensed to the Linguist project under MIT license terms for exclusive use with Linguist. A notice to this effect has been added at the bottom of each of the two files.


lildude:
Friendly nudge.

lildude:
This PR has been pending changes for quite some time. Marking it as stale and will close in a week's time if no further updates.

trijezdci:
I have absolutely no idea what it is you expect me to do. The grammars do the disambiguation (you call it heuristics) that you ask for, but then you seem to be complaining that I provided grammars.

The fact remains that your handling of Modula-2 is utterly broken. There isn't any Modula-2 source code without .def files, just as there is no C source code without any .h files. And the .mod files don't display correctly either.

I am beginning to regret that we moved from Bitbucket to Github. Bitbucket uses a syntax highlighter called Pygments, I wrote a multi-dialect parser module for that and the Pygments folks were rather more welcoming of people contributing. By contrast, you appear to be annoyed by folks who want to help you fix problems in your code.

So the question is, do you or don't you want to support as many languages for which there are projects on github? If the answer is yes, perhaps you should be a tad more supportive.

For the time being, your Modula-2 highlighting is broken.

lildude:
> I have absolutely no idea what it is you expect me to do.

I'm sorry if I wasn't clean in my initial note. Essentially, we need this PR moulded into the shape we detail in the [Adding a language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language) section of our contributing guide before we can accept it.  The biggest issue with it is you have included the syntax highlighting grammar files directly in the PR. We need these files removed in favour of using a submodule. ([Step 2](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language)).

The heuristic is used to better identify files with identical extensions as extensions have a higher priority when it comes to identifying a language over the classifier. The actual syntax highlighting is used for nothing more than making the code look good. You can ignore this point for the moment.

Now speaking of the the syntax highlighting and your comment:

> For the time being, your Modula-2 highlighting is broken.

Linguist doesn't do the syntax highlighting. It only does the language identification. The syntax highlighting is done by the grammars Linguist pulls in, and these are often the exact same grammars BitBucket uses.

In this case, we use the grammar from https://github.com/harogaston/Sublime-Modula-2. If there is a problem with the syntax highlighting, the [Fixing syntax highlighting](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#fixing-syntax-highlighting) section of our contributing docs detail what needs to be done. In short:

1. Log an issue in the repo of the used grammar, or
2. Submit a new PR to Linguist pointing to a more accurate and actively maintained grammar repo that doesn't show the problem you are experiencing.

So with all of that said, you have three options:

1. Close this PR and open an issue at https://github.com/harogaston/Sublime-Modula-2 detailing the problems you are seeing with the syntax highlighting. Once the issue has been fixed by the grammar maintainer, it'll automatically be pulled into a later release of Linguist and subsequently on to GitHub.com
2. Update this PR to...
  - remove the grammar files you added directly, 
  - add the new grammar as a submodule as per [Changing the source of a syntax highlighting grammar](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#changing-the-source-of-a-syntax-highlighting-grammar)
  - remove the addition of the `.def` extension as it is not needed for syntax highlighting and will cause problems for other languages that use this extension. You can use an [override](https://github.com/github/linguist/blob/master/README.md#using-gitattributes) to force the `.def` files to be identified as Modula-2 in your repos.
3. Leave things as-is and hope someone else will take the syntax highlighting issue up with the grammar maintainer. This isn't the nicest option for the wider Modula-2 community, but it's the most commonly chosen route.

I hope I've not added more confusion with this long explanation. Please feel free to ask any further questions, no matter how simple. I'm more than happy to help.

trijezdci:
Yes, I am very well aware that syntax highlighting/colouring is a multi-stage operation and the rendering of the source code is only the last stage. Nevertheless, the syntax analysis done prior to that has no other purpose than to allow the rendering later, its only rationale is to highlight or colour the source. If it wasn't for that, nobody would bother. It is thus perfectly correct to speak of syntax colouring when referring to the whole process, because that is the ultimate purpose.

It may well be that some of the grammars you are using are the same as used by Pygments parsers (the package used by Bitbucket) but this is certainly not the case for Modula-2.

I know this because I am the author of the Pygments Modula-2 parser which is one of the most sophisticated parsers in the entire Pygments package. It recognises and distinguishes between all major Modula-2 dialects and language extensions of some 15 or so compilers most commonly in use. It further recognises identifiers of various libraries depending on dialect and compiler.

To disambiguate between dialects and compilers, it recognises special comment tags that indicate the dialect (and optionally the compiler) for which the source is written. The tags follow a community standard that had been agreed between the maintainers of compilers still under active development or maintenance. Sources that aren't tagged (that is to say lacking a special comment that embeds the dialect tag) are analysed using a lowest common denominator grammar that produces acceptable results for all dialects but ignores many tokens.

The parser also marks various syntax errors (not merely lexical glitches) so they can be rendered in red (or any other way depending on style sheets).

It took several months of hard work to write and test this. No matter what dialect and compiler Modula-2 code is written for, Pygments (and thus Bitbucket) will render it properly.

By contrast, the grammar used in Linguist is not correct for any Modula-2 dialect, past or present.

Also, your suggestion that it should be acceptable for every repo owner to have to specifically add a custom rule to treat .def files as Modula-2 sources is flabbergasting.

I had already filed bug reports for the existing Modula-2 treatment by Linguist and those bug reports were eventually closed down for lack of Modula-2 expertise within the Linguist team. The responses were along the lines of "We neither have the expertise to verify if Linguist does Modula-2 correctly, nor do we have the expertise to fix it if it doesn't. We could use a Modula-2 expert". I am paraphrasing this but that was the tenor.

It is for this reason that I volunteered to provide proper Modula-2 recognition. I found Linguist to be extremely confusing and badly documented. Somewhere, I read it would simply port that which Pygments does in Python to Ruby but it seems to me that isn't actually the case, even though it might be an ultimate goal not yet realised. Instead, I found Linguist to be a mishmash of different approaches, which is extremely difficult for any outsider to work with because which of the different approaches should one use then? If it was indeed a rewrite of Pygments in Ruby, one could simply learn a bit of Ruby and then convert the Pygments Modula-2 parser into Ruby. It would then have all the sophistication of the Pygments parser as a result.

Unfortunately I had to aim much lower. So I provided three grammars for the main three Modula-2 dialects and a rule in each to recognise the generic dialect tag embedded in a comment in the very first line (the Pygments parser doesn't need to find it in the very first line). Also, there is no lowest common denominator fallback if the source code is not tagged with any dialect tag as is the case with Pygments. Such a fallback could be added later on for legacy code (which is naturally untagged).

In any event, the grammar for Modula-2 that Linguist currently uses is not correct for any dialect.

If I am not permitted to provide any replacement grammars, I can't see how to fix this.

The reason why a Modula-2 source code rendering needs to distinguish between different dialects is simple. If a developer hosts their code on github and finds that it doesn't render according to the dialect they are using, they might be inclined to submit a patch, but that patch would then break rendering code written for a different dialect. Eventually somebody else will submit a patch breaking the former again. The only way to avoid this is to support all the major dialects and the active compiler maintainers have agreed on special comment tags for this very purpose.

So, please do not suggest to me to again file a bug report for the existing grammar. We have been down that road and it didn't work out. Let's move on.

Rather, help me to do what I set out to do:

(1) to provide replacement grammars with dialect tag recognition for the main three Modula-2 dialects within Linguist as an interim solution;
(2) to later on add a fallback method for untagged Modula-2 source files;
(3) as a long term goal, to replicate the functionality of the Pygments Modula-2 parser/renderer in Linguist.

regards
benjamin

trijezdci:
For your information:

There is no grammar maintainer for Linguist's Modula-2 grammar.

The existing grammar is **abandonware** and even then I think whoever did this in the first place simply copied a Pascal grammar from somewhere and replaced a handful of reserved words from memory, but never bothered to research this properly, let alone verify that it matches any spec.

lildude:
Wow! Thanks for the thorough explanation @trijezdci. It definitely sounds like your grammar is going to be a huge improvement.

> Whatever we do, it cannot be based on a single grammar. So, please explain how to go about
> (1) completely remove the existing broken Modula-2 support in Linguist.
> (2) add Modula-2 support to Linguist based on three separate grammars.

I think we can address all of these in one go, assuming the multiple grammars you are referring to are the files you've attached to this PR. We consider the grammars as you would a TextMate "bundle". If you can get all the functionality for all the variants in a single "bundle", then Linguist will pull them all in one go. Things get tricky when it comes to multiple distinct grammar "bundles" as Linguist only supports a one-to-one language-to-grammar mapping.

So if we take just the files you've included in this PR and assume they're all to be treated as part of the single "modula2-tmbundle" they're shown to be part of in this PR, the steps you'd need to take are as follows:

1. Create a new GitHub, BitBucket or GitLab repository in an account you own or have full control over. You are and will be the owner and maintainer of your grammar, so the location is entirely up to you. These are the three we pull in from.
2. Add the TextMate bundle files you've included in this repo to your newly created repo.
3. Add a license to your repo. As Linguist is MIT licensed, we can only accept grammars with one of [these licenses](https://github.com/github/linguist/blob/257425141d4e2a5232786bf0b13c901ada075f93/vendor/licenses/config.yml#L2-L11).
4. Remove the grammar files from this PR.
5. Use `script/add-grammar --replace Modula-2 <url to the repo created in step 1>` to replace the current sub-module reference with a reference to your grammar.
6. Push your updated changes to your fork of the Linguist repo. This PR will automatically reflect your changes.
7. Check and confirm Linguist will be using your grammar correctly and as you expect using https://github-lightshow.herokuapp.com/ - please provide links in this PR to Lightshow showing the old and new grammars in this PR.

Once you've done that, you'll end up with a PR that looks very similar to https://github.com/github/linguist/pull/3871 and everything should be in place for Linguist to stop using the current grammar in favour of your grammar.

pchaigno:
> Things get tricky when it comes to multiple distinct grammar "bundles" as Linguist only supports a one-to-one language-to-grammar mapping.

@lildude I think we might be able to apply a distinct grammars for each dialect if we create one entry per dialect in `languages.yml`, with each their dedicated grammar scope, and regroup them under the Modula-2 language. We can then distinguish them with heuristic rules on the comment tag. What do you think? I think it's going to be easier than distinguishing them in the grammar itself.

@trijezdci Looking at the sample file you provided, I'm guessing the comment tags are `(*!m2iso*)`, `(*!m2pim*)`, and `(*!m2r10*)`. Is it really as simple as that?

trijezdci:
@lildude Thanks for your suggestions, I am going to take a closer look at that.

@pchaigno As I had mentioned, I had to aim lower for Linguist than for Pygments. Thus, for the intended functionality for Linguist at least initially, the comment tags are indeed just those three.

The Pygments parser also recognises tags that show compiler extensions, for example code written for ISO Modula-2 using language extensions of the GNU Modula-2 compiler could be tagged `(*!m2iso+gm2*)` where gm2 is a mnemonic for GNU Modula-2. For each supported compiler such a mnemonic has been agreed between the compiler maintainers. Pygments also doesn't need the tag to be in the very first line.

Also note that no spaces and no control codes/line breaks are permitted within those tags, which makes the regex simpler. This particular restriction is part of the agreed spec and thus also applies to Pygments.

The idea is to get support for the generic dialects working in Linguist first without creating too many test cases.  With the proposed initial scheme there will only be three test cases. Eventually, functionality can be extended to also recognise tags that indicate the targeted compiler but let's not get too fancy at first.

trijezdci:
One more thing ...

We have also implemented this for the `vim` editor and if I remember correctly, the maintainer of the GNU Modula-2 compiler has done the same for emacs. The vim implementation searches the first 100 lines of source of any file ending in .def, .DEF, .mod and .MOD for either a dialect tag or the reserved words MODULE, DEFINITION MODULE, IMPLEMENTATION MODULE and BLUEPRINT in which case it assumes the file is a Modula-2 source file even without a dialect tag, then using a fallback scheme.

For the avoidance of confusion it is useful to know that some versions of vim still use the older recognition scheme which used file endings such as .d, .i, .m2d, .m2i (IIRC) in addition to def and mod but those are non-standard. There is an old unmaintained compiler by a German university that used those file endings but this compiler has been patched a few years ago to no longer use them and use the standard .def and .mod endings instead. As far as I know, there are only two projects published online that use these non-standard file extensions: the compiler/library sources of the afore-mentioned compiler (since patched) and the sources of another compiler that was used in the former Soviet Union and has since become abandonware. It is thus safe to only recognise the standard file endings.

lildude:
> @lildude I think we might be able to apply a distinct grammars for each dialect if we create one entry per dialect in languages.yml, with each their dedicated grammar scope, and regroup them under the Modula-2 language. We can then distinguish them with heuristic rules on the comment tag. What do you think? I think it's going to be easier than distinguishing them in the grammar itself.

Oooo, good idea. I completely forgot about the grouping. Def something to look at once we've got the initial implementation sorted out.

pchaigno:
Thanks for the details @trijezdci! We can probably do the same as vim with [our heuristic rules](https://github.com/github/linguist/blob/512f077da8afb4390db05dce44355febe6d66d42/lib/linguist/heuristics.rb).
Could you follow [the steps suggested by @lildude](https://github.com/github/linguist/pull/3716#issuecomment-357442864), and we'll look into the disambiguation in a second PR?

lildude:
Closing as stale. Please feel free to submit a new PR when you have the chance to implement all the suggestions.

Issue #3094, opened 2016-06-30
Include support for JSP tag files


rn4n:
Hi,

I would like to suggest support for [files with the extension ".tag"](http://docs.oracle.com/javaee/1.4/tutorial/doc/JSPTags5.html). They have the same syntax used by JSP language, but it is not displayed with the same syntax highlighting:
- [JSP file example](https://github.com/caelum/mamute/blob/master/src/main/webapp/403.jsp).
- [Tag file example](https://github.com/caelum/mamute/blob/master/src/main/webapp/WEB-INF/tags/answerWith.tag).

:pray:

pchaigno:
If we add the `.tag` file extension to JSP, we also need to add it to every other language that use it (otherwise all `.tag` files will be indentified as JSP). Thus, we first need to identify languages that use the `.jsp` file extension.
@reebr Are [those files](https://github.com/search?p=1&q=extension%3Atag+NOT+xml+NOT+tags+NOT+jsp+NOT+attribute+if&ref=searchresults&type=Code&utf8=%E2%9C%93) JSP tag files?

So far, I found:
- [10k XML files](https://github.com/search?p=1&q=extension%3Atag+xml+version&ref=searchresults&type=Code&utf8=%E2%9C%93)
- [at least 35k JSP tag files](https://github.com/search?p=1&q=extension%3Atag+if+tags+div+OR+section&ref=searchresults&type=Code&utf8=%E2%9C%93)

reebr:
Woah, there're several languages that use this file format, I didn't know that.
Thanks for the search @pchaigno.

I'm not sure on [these files][1], but it seems they use the [Play framework template tags][2].
But [those files][3], they are definitely not related to Java / JSP.

Maybe it's a bad idea implement this for a specific language, there are a lot of ".tag" types. :disappointed: 

[1]:https://github.com/search?p=1&q=extension%3Atag+NOT+xml+NOT+tags+NOT+jsp+NOT+attribute+if&ref=searchresults&type=Code&utf8=%E2%9C%93
[2]:https://www.playframework.com/documentation/1.2/tags
[3]:https://github.com/search?p=1&q=extension%3Atag+xml+version&ref=searchresults&type=Code&utf8=%E2%9C%93

pchaigno:
[...]

PR #3720, opened 2017-07-16
Added Riot Language

crisward:
https://github.com/search?utf8=%E2%9C%93&q=language%3AJavaScript+extension%3Atag+riot&type=Repositories&ref=advsearch&l=&l=


lildude:
You've got a couple of problems here that need to be resolved.

The colour failure should be obvious. The license issue is because you've manually added the license file rather than using the script as detailed in https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language

lildude:
This PR has been pending changes for quite some time. Marking it as stale and will close in a week's time if no further updates.

crisward:
Sorry about that.. will aim to fix this soon. Thanks for the nudge.

crisward:
The color issue is a bit of a problem. The language authors want red, because their branding is red. They specifically want #fc0e49, any ideas on how to work out a color that is close enough to that without overlapping another?

crisward:
Thanks, pretty sure that's what I did before. The trial and error ended up with something quite far away. Oh well... 😞

GianlucaGuarini:
@lildude may I help you guys on anything here? I will be really glad to see finally the riot tags landing into linguist

GianlucaGuarini:
@crisward please use ```#AD0E49``` as color and update the license using the [official one (MIT)](https://github.com/riot/riot/blob/master/LICENSE.txt) and I think this PR should be fine. Thank you

pchaigno:
Thank for your contribution @crisward!

Do you know what [these files](https://github.com/search?p=96&q=extension%3Atag+NOT+riot&type=Code&utf8=%E2%9C%93) are? They don't look the same as your Riot sample files.

GianlucaGuarini:
> Do you know what these files are? They don't look the same as your Riot sample files.

@pchaigno they don't look like riot files but they are arbitrarily named `.tag` since they come always for the same account and they are html files anyway so we shouldn't have any conflict in this case

crisward:
😢  Is there no way around this, perhaps using `.gitattributes`

GianlucaGuarini:
@pchaigno the other `.tag` files matched by the search actually should be `.jsp` files coming from [java taglib directive](http://www.tutorialspoint.com/jsp/taglib_directive.htm) I don't think riot will change the file extension anytime soon (even because `.riot` is also used by [some users already](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ariot+NOT+riot&type=Code) ).

pchaigno:
@crisward @GianlucaGuarini We did some previous work on the `.tag` file extension in #3094. If we're exhaustive enough in our search of languages using `.tag`, we might be able to add the extension.

For now, we have:
- [4k Riot files](https://github.com/search?p=99&q=extension%3Atag+riot&type=Code&utf8=%E2%9C%93)
- [10k XML files](https://github.com/search?p=1&q=extension%3Atag+xml+version&ref=searchresults&type=Code&utf8=%E2%9C%93)
- [at least 35k JSP tag files](https://github.com/search?p=1&q=extension%3Atag+if+tags+div+OR+section&ref=searchresults&type=Code&utf8=%E2%9C%93)
- [at least 750 Play framework template files (366 repositories)](https://github.com/search?p=1&q=extension%3Atag+NOT+xml+NOT+tags+NOT+jsp+NOT+attribute+NOT+my+doBody&ref=searchresults&type=Code&utf8=%E2%9C%93)

@crisward Do you have a better search query for Riot files? Specific keywords for example?

crisward:
@pchaigno I don't, do you @GianlucaGuarini ?

GianlucaGuarini:
@pchaigno i guess [this query](https://github.com/search?utf8=%E2%9C%93&q=extension%3Atag+this.on%28&type=Code) is a bit better since many riot tags use [riot-observable](https://www.npmjs.com/package/riot-observable) for their lifecycle events. Of course my query excludes all the users aliasing `this` with `self`.. and all the others using riot tags with the `.html` extension because of their better highlighting on github

pchaigno:
@GianlucaGuarini What about [this](https://github.com/search?p=1&q=extension%3Atag+%22this.on%22+OR+%22self.on%22&type=Code&utf8=%E2%9C%93)?

GianlucaGuarini:
@pchaigno the query will never be precise since in js you can do:
```js
const me = this
const self = this
const that = this
const _this = this
```
😄

Alhadis:
> @pchaigno the query will never be precise since in js you can do:

In the days before ES6, my approach to this was to simply write `this` as `THIS`. Hurray for case-sensitivity! 😉 Not such good news in this case though.

Thank God we now have arrow functions and block scoping (ES5 feels like a different language... 😨)

crisward:
@Alhadis I'll try and push a merged update today, up to my eyes in server updates at the moment... it's not only the cpus suffering with meltdowns...

GianlucaGuarini:
@pchaigno @Alhadis why is the travis CI failing? Can we do anything for you guys to merge this PR?

Alhadis:
It's due to [this](https://travis-ci.org/github/linguist/jobs/330560199#L1271-L1273):

> **1271** Error updating vendor/grammars/Sublime-HTTP
> **1272** error: pathspec 'vendor/grammars/Sublime-HTTP' did not match any file(s) known to git.
> **1273** Did you forget to 'git add'?

Which is entirely unrelated to the PR. Not entirely sure what's going on there. :confused:

pchaigno:
> @pchaigno @Alhadis why is the travis CI failing? Can we do anything for you guys to merge this PR?

I think you made a mistake while merging; you deleted the sublime-http submodule.

GianlucaGuarini:
@crisward could you please rebase fixing the issue?

crisward:
Colour seems to be failing now.

crisward:
Could go back to `#AD0E49` instead of `#A71E49`?

crisward:
No idea about the last fail...??

lib/linguist/languages.yml
```diff
@@ -3894,6 +3894,15 @@ Ring:
   tm_scope: source.ring
   ace_mode: text
   language_id: 431
+Riot:
+  type: programming
+  color: "#A71E49"
+  group: HTML
```
Alhadis: @crisward This is the reason for the [most recent build failure](https://travis-ci.org/github/linguist/jobs/332305128#L1388-L1390):

~~~
 1) Failure:
TestLanguage#test_no_unused_colours [/home/travis/build/github/linguist/test/test_language.rb:467]:
Unused colour assigned to Riot
~~~

Languages which are assigned a `group` will inherit their parent language's colour and usage stats, For example, Less [inherits CSS's name and colour too](https://github.com/jonikorpi/Less-Framework).

In other words, you'll want to either ditch the `group` or the colour field, one or the other. =)

crisward: Thanks! I'll dump the group them.

lib/linguist/languages.yml
```diff
@@ -3894,6 +3894,14 @@ Ring:
   tm_scope: source.ring
   ace_mode: text
   language_id: 431
+Riot:
+  type: programming
```
Alhadis: This should be `markup`, not `programming`.

crisward: Updated

lildude:
If @pchaigno is quick with the final :+1:, this PR might just sneak into the next Linguist release 😉

pchaigno:
I don't think my previous review is solved. If we merge this as-is, all `.tag` files will be recognized as Riot files. A large majority of them don't appear to be Riot files.

lildude:
Oh bollox!! I missed that. Yeah, we can't merge this until we know more about those files and have a heuristic to ensure they're not incorrectly identified.

GianlucaGuarini:
@pchaigno can't linguist filter files by regex? A simple filter for [`/<%@|<\?xml|#{doBody/`](https://regex101.com/r/piUB5V/1) should be enough to detect only the riot tags

lildude:
> @pchaigno can't linguist filter files by regex? A simple filter for /<%@|<\?xml|#{doBody/ should be enough to detect only the riot tags

Yup, however the way Linguist works is like a funnel... it takes in all languages at the top and whittles the list down based on the various strategies until it gets a single language. Prior to getting to the heuristics, it filters the list by extension.

As things currently stand with this PR, _all_ `.tag` files will be identified as Riot and Linguist will exit at the extension stage (cos languages == 1) and wouldn't get down to the heuristics which would perform the regex match.

In order to get down to the heuristics, we need `.tag` associated with all languages that commonly use it so we have more than one language going into the heuristics strategy. We mention this as a possible requirement in the CONTRIBUTING.md file.

The heuristics would then need to do something like: (pseudo code to explain the point):

```
if /<%@|<\?xml|#{doBody/`
  return "Riot"
elif /some-other-regex-for-another-language/
  return ?
[ rinse and repeat ]
else
  return ?
end    
```

... to whittle down those `.tag` languages to a single language. If it can't get down to a single language at this point, the remaining languages are passed down to the naïve Bayesian classifier which will use the samples to try and guess. This is far from perfect so we prefer to leave this as an absolute last resort.

GianlucaGuarini:
As side note:

For the Playframework the `.tag` format extension **is a fallback** extension https://www.playframework.com/documentation/1.2.3/templates#Formatspecifictags

For the `.xml` files having the `.tag` extension I have not meaningful explanation... but  of course **the `.xml` extension is the most correct one in these cases**

While for the `JSP` tags a regex is required..

Riot.js on the other end as for `JSP Tags` has the `.tag` **extension as default file format**

lildude:
Nudge.

Alhadis:
I've extended your nudge in a [linked issue](https://github.com/riot/riot/issues/2529#issuecomment-379290366) at the Riot repository.

This merge conflict is an endangered species now that we have the module-sorting script in-place, ensuring that no more of its kind can ever breed in this codebase again. 😉

Alhadis:
@pchaigno Hey, I just realised something.

That button labelled <kbd><b>Resolve conflicts</b><kbd> actually does something. 😀

I did that something in question, and now this LGTM. 👍 Thoughts?

PR #3731, opened 2017-07-23
Add Adobe Font Metrics to recognised filetypes

Alhadis:
Adobe Font Metrics (`*.afm` files) are a data format for storing character measurements, encoding data, and other PostScript/Type 1-specific metadata. Wikipedia's page on PostScript fonts [touches on it briefly](https://en.wikipedia.org/wiki/PostScript_fonts#Adobe_Font_Metrics.2C_Adobe_Composite_Font_Metrics.2C_Adobe_Multiple_Font_Metrics):

> (AFM files)… contain general font information and font metrics information for the font program. These files are generally used directly only in Unix environments. An AFM file provides both global metrics for a font program and the metrics of each individual character.
>
> The metrics of a multiple master font are described by one AMFM file, which specifies the control data and global font information, plus one AFM file for each of the master designs in the font.
>
> An ACFM file provides information about the structure of a composite font. Specifically, the global metrics of the composite font program and the global metrics of each of its immediately descendent font programs. ACFM file does not associate with a base font, but act as the top-level structure of a composite font. The character metrics of individual characters in the composite font are described completely by one of more associated AFM files.
>
> The formats are sufficiently similar that a compliant parser can parse AFM, ACFM, and AMFM files.

Fundamentally, all three formats are the same for our purposes: they share the same file-extension and perceivable syntax, so I haven't bothered adding `*.acfm` and `*.amfm` files as associated extensions. Not that there [even](https://github.com/search?q=extension%3Aamfm+NOT+nothack&type=Code) [are](https://github.com/search?q=extension%3Aacfm+NOT+nothack&type=Code), anyway. I have included them as language aliases, however.

### In-the-wild usage

Scrape and logs stashed in the silo, as per usual: [`Alhadis/Linguist-Silo/tree/afm`](https://github.com/Alhadis/Linguist-Silo/tree/afm). Nearly [269,000](https://github.com/search?q=extension%3Aafm+NOT+nothack&type=Code) results in-the-wild, most of them duplicates. Which doesn't surprise me, considering a good deal of those results point to GhostScript forks/mirrors. GhostScript (the most popular free implementation of Adobe PostScript) distributes free versions of the base 35 PDF fonts in [several formats](http://git.ghostscript.com/?p=urw-core35-fonts.git;a=commit;h=1f28a6fcd2176256a995db907d9ffe6e1b9b83e9), which includes AFM files.

### Origin of samples

Both of the samples I added were converted from TTF files sourced from [`google/fonts`](https://github.com/google/fonts), both released under Apache-friendly licenses:

* `SpecialElite.afm` -> https://github.com/google/fonts/tree/8839397/apache/specialelite
* `OpenSansCondensed-Bold.afm` -> https://github.com/google/fonts/tree/44438ae/apache/opensanscondensed

To be honest, I really don't think it matters, though. AFM files don't contain any glyph information: just the width and heights, spacing adjustments, et al. I erred on the side of caution anyway. GhostScript's fonts are released under the [GNU Affero License](https://www.gnu.org/licenses/agpl.html), so... suffice to say that was a no-go.


pchaigno:
The 2 sample files are huge and will likely increase `samples.json`'s size in memory. Couldn't we reduce them by removing similar chunks of code?

Alhadis:
HAH! I looked at their filesizes before pushing and thought 416 KBs and 161 KBs were reasonable filesizes. Totally lost my sense of scale watching 111 MBs of search results being pulled in...

Will fix, hang on.

Alhadis:
Three times now I've accidentally written `GFM` (GitHub-flavoured Markdown) instead of `AFM`. Ugh.

All done!

PR #3734, opened 2017-07-24
Add Printer Font ASCII to recognised PostScript extensions

Alhadis:
[PFA files][1] are simple PostScript wrappers for Type 1 and Type 42 font instructions. This commit adds `.pfa` as a recognised PostScript extension, and updates the [logic][2] for recognising generated PostScript files. With very [rare][3] [exceptions][4], PFA files are almost always generated by a program converting a font to a PostScript-friendly format.

There are [~15,853][5] PFA files on GitHub; results of a search result scrape [are here][6]. Only a handful of non-PostScript files were encountered, which appear to be [genome-related sciency crapola](https://github.com/aron0093/KB8019/blob/0a418dd1bc60a457c7aac83b27f78ff19c281c3c/data/Proteomes/08.fa.txt.pfa). They're limited to two repositories, both by Swedish users. Origin of samples unknown, but presumably case-specific.

[1]: https://en.wikipedia.org/wiki/PostScript_fonts#Printer_Font_ASCII
[2]: https://github.com/github/linguist/blob/81957f90e9416661aee24e326b1b958222c696e5/lib/linguist/generated.rb#L249-L266
[3]: https://github.com/gitGNU/gnu_enscript/blob/af89c2ad63ca63befc630adfa0819214ddd99e66/afm/matrix.pfa
[4]: https://github.com/gitGNU/gnu_a2ps/blob/a572e93acf79191fbc65bc0dd08d19cd2a646977/fonts/pcfont.pfa
[5]: https://github.com/search?q=extension%3Apfa+NOT+nothack&type=Code
[6]: https://github.com/Alhadis/Linguist-Silo/tree/pfa


lib/linguist/generated.rb
```diff
@@ -244,7 +244,11 @@ def generated_parser?
     #
     # Returns true or false.
     def generated_postscript?
-      return false unless ['.ps', '.eps'].include? extname
+      return false unless ['.ps', '.eps', '.pfa'].include? extname
+
+      # Type 1 and Type 42 fonts converted to PostScript are stored as hex-encoded byte streams; these
+      # streams are always preceded the `eexec` operator (if Type 1), or the `/sfnts` key (if Type 42).
+      return true if data =~ /(\n|\r\n|\r)\s*(?:currentfile eexec\s+|\/sfnts\s+\[\1<)\h{8,}\1/
```
pchaigno: Couldn't `\n|\r\n|\r` be written as `^` here?

Isn't `\/sfnts` sufficient to match Type 42 fonts?

Alhadis: Not really. It's a multiline match for Type 42; and no, `/sfnts` alone would match literally any PostScript program which used that name for an identifier. It's part of a "Dictionary" in PostScript terminology. It'd be like identifying generated JavaScript based solely on an object literal containing `{ sfnts: …`

pchaigno: Oki. Thanks for the explanation :-)

Alhadis: Since you like explanations, I'll also add that at least 8 hexadecimal characters are matched, because the Adobe Type 1 specification mandates that encoded `eexec` sections are prefixed with four random, non-ASCII bytes for encryption purposes. So `4 * 2 = 8`. I imagine this was initially done to keep the format secret (recalling this was once proprietary software back before they open-sourced it).

Issue #3733, opened 2017-07-24
TypeScript file is being JavaScript.


gluons:
Some `.ts` file is treated as JavaScript file.

<img width="779" alt="default" src="https://user-images.githubusercontent.com/4688092/28508902-551b352a-7068-11e7-8441-5e8a17f2de36.PNG">

pchaigno:
Oo, you can interpret TypeScript with the Node interpreter??

gluons:
@pchaigno Did you mean that this file can run directly by Node? If you mean that, I think it can't.

This file contains some variable type declaration that isn't valid JavaScript.

pchaigno:
Hm. So this file is detected as JavaScript because of the [shebang on the first line](https://github.com/gluons/changelog/blob/6f8c6fc314dae149882e8ef5624aa090abcd3010/src/changelog.ts#L1).

gluons:
Haha. **shebang** make this file to be detected as JavaScript. 🤣

Alhadis:
> Oo, you can interpret TypeScript with the Node interpreter??

@pchaigno [Relevant issue](https://github.com/file-icons/atom/issues/606#issuecomment-304558298) at `file-icons` which has an explanation:

>  No, it cannot [execute], but the shebang string gets propagated from the TS source to the JS compiled with `tsc`. If I omit it in the source, it won't appear in the output JS.

This isn't something Linguist currently accommodates, AFAIK. 😕

pchaigno:
Thanks for the explanation @Alhadis!

@gluons You can override the language detected using [Linguist overrides](https://github.com/github/linguist#overrides).

gluons:
@pchaigno OK. I'll try. Thank you.

PR #3735, opened 2017-07-25
Allow POD with Perl shebangs

kivikakk:
Right now a POD-extension file with a Perl shebang will only get recognised as Perl. This lets the Shebang classifier not narrow too much.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -18,8 +18,7 @@ module Linguist
 
       @heuristics.each do |heuristic|
         if heuristic.matches?(blob.name)
-          languages = Array(heuristic.call(data))
-          return languages if languages.any? || languages.all? { |l| candidates.include?(l) }
+          return Array(heuristic.call(data))
         end
       end
 
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -27,8 +27,8 @@ module Linguist
 
     # Internal: Define a new heuristic.
     #
-    # extensions - String names of file extensions and languages to
-    #              disambiguate.
+    # exts_and_langs - String names of file extensions and languages to
+    #                  disambiguate.
     # heuristic - Block which takes data as an argument and returns a Language or nil.
     #
     # Examples
@@ -41,16 +41,16 @@ module Linguist
     #       end
     #     end
     #
-    def self.disambiguate(*extensions, &heuristic)
-      @heuristics << new(extensions, &heuristic)
+    def self.disambiguate(*exts_and_langs, &heuristic)
+      @heuristics << new(exts_and_langs, &heuristic)
     end
 
     # Internal: Array of defined heuristics
     @heuristics = []
 
     # Internal
-    def initialize(extensions, &heuristic)
-      @extensions, @candidates = extensions.partition {|e| e =~ /\A\./}
+    def initialize(exts_and_langs, &heuristic)
+      @exts_and_langs, @candidates = exts_and_langs.partition {|e| e =~ /\A\./}
       @heuristic = heuristic
     end
 
@@ -59,7 +59,7 @@ module Linguist
     def matches?(filename, candidates)
       filename = filename.downcase
       candidates = candidates.compact.map(&:name)
-      @extensions.any? { |ext| filename.end_with?(ext) } ||
+      @exts_and_langs.any? { |ext| filename.end_with?(ext) } ||
         (candidates.any? &&
          (@candidates - candidates == [] &&
           candidates - @candidates == []))
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -17,7 +17,7 @@ module Linguist
       data = blob.data
 
       @heuristics.each do |heuristic|
-        if heuristic.matches?(blob.name)
+        if heuristic.matches?(blob.name, candidates)
           return Array(heuristic.call(data))
         end
       end
@@ -27,7 +27,8 @@ module Linguist
 
     # Internal: Define a new heuristic.
     #
-    # languages - String names of languages to disambiguate.
+    # extensions - String names of file extensions and languages to
+    #              disambiguate.
     # heuristic - Block which takes data as an argument and returns a Language or nil.
     #
     # Examples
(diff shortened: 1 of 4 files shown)


Alhadis:
Well, I'll be. Linguist allows more than one language to use the same interpreter.

Guess my [last comment](https://github.com/github/linguist/issues/3733#issuecomment-317358951) was kinda nonsensical...

pchaigno:
We currently don't have any sample Pod file with a shebang. Maybe we should add one so that this actually gets tested?

pchaigno:
On second through, looking at the code, I don't think we support having several languages with the same interpreter. Interpreters are [indexed in an associative array](https://github.com/github/linguist/blob/b94e018c3af1ca5b3858f2249253d076b26df6a0/lib/linguist/language.rb#L185-L187), and the next strategy, [`Linguist::Strategy::Extension`](https://github.com/github/linguist/blob/b94e018c3af1ca5b3858f2249253d076b26df6a0/lib/linguist.rb#L64), [doesn't take candidate languages from the previous strategy as input](https://github.com/github/linguist/blob/b94e018c3af1ca5b3858f2249253d076b26df6a0/lib/linguist/strategy/extension.rb) anyway.

@Alhadis I guess you were right :stuck_out_tongue_winking_eye:

Alhadis:
> We currently don't have any sample Pod file with a shebang. Maybe we should add one so that this gets tested?

Uhm... yeah, that's the other thing. About that...

daphuq

pchaigno:
@Alhadis I'm not sure I understand... There was initially a sample in this pull request?

Alhadis:
If it was deleted, it would've appeared in the changes... :\

@kivikakk This looks to be a bug with GitHub's diff-displaying.

kivikakk:
You might've missed this:

screen shot 2017-07-25 at 5 31 52 pm

I removed it as it wasn't actually appropriate, as it's a test for the classifier (only), not the whole detection process.

> the next strategy, `Linguist::Strategy::Extension`, doesn't take candidate languages from the previous strategy as input anyway.

When a strategy doesn't return _any_ results, results from the previous strategy are retained. This  fix is for cases where a script has a `#!/usr/bin/perl` shebang but contains POD documentation.  The original filename often will have no extension when a shebang is used, in which case `Linguist::Strategy::Extension` returns nothing, and the two results from `Shebang` are passed onto `Heuristics` (no result for no extension) and `Classifier`.

(If there is an extension, then `.pl` or `.pod` will be detected and selected as appropriate.)

iow, I think this does what we want, but I'll add some test cases to demonstrate.

Alhadis:
> You might've missed this:

Nope, it missed me. I can't see that listed anywhere, and trying to open [`07760c1`](https://github.com/github/linguist/pull/3735/commits/07760c1) gives me this:

NIGGA, THEY'RE NOT GONNA BE IN OUTER SPACE

Anyway, doesn't matter. =)

kivikakk:
(This in fact won't do what I want it to, after reviewing what I wrote above, because the heuristic for `.pod` won't fire without a filename. Grrr.)

kivikakk:
Okay — new sample added, and some changes to heuristics so we can accept a list of candidates as input. In other words, if filename extension disambiguation fails, we can optionally disambiguate the list of candidates previous strategies have given us. (So far this just adds the one case for disambiguating Perl/Pod.)

Anyone got :eyes: for this?

Alhadis:
LGTM. 👍 

Also, please tell me [`Kommunikatsioonimoodul`](https://github.com/adamsont/Kommunikatsioonimoodul/) isn't a real Estonian word. 😨  Holy hell.

lib/linguist/heuristics.rb
```diff
@@ -17,9 +17,8 @@ def self.call(blob, candidates)
       data = blob.data
 
       @heuristics.each do |heuristic|
-        if heuristic.matches?(blob.name)
-          languages = Array(heuristic.call(data))
-          return languages if languages.any? || languages.all? { |l| candidates.include?(l) }
```
pchaigno: Any reason why you removed the second check (`languages.all? { |l| candidates.include?(l) }`)? This is in place to make sure heuristic rule don't return new languages we've already cross off the list.

lib/linguist/heuristics.rb
```diff
@@ -50,14 +50,19 @@ def self.disambiguate(*extensions, &heuristic)
 
     # Internal
     def initialize(extensions, &heuristic)
```
pchaigno: We should probably rename `extensions` to something more appropriate...?

lib/linguist/heuristics.rb
kivikakk: `languages.any? || languages.all? { … }` is always true; if `languages.any?` is false, then `languages.all? { … }` will always return true (because `.any?` on an empty list is always true). So the `if` was always happening.

lib/linguist/heuristics.rb
kivikakk: Good point!

lib/linguist/heuristics.rb
pchaigno: Oo :scream: You're right, that should've been an AND. And I'm the one who switched it to an OR :sweat: 

This was for cases where we'd use the same set of heuristic rules for two extensions with different candidates. For example if we use heuristic rules on C, C++, and Objective-C for both `.h` (the three languages are candidates) and `.inc` (only C and C++ are candidates). In that case, the extra check would ensure Objective-C is never returned for the `.inc` extension.

We currently don't have any such case, so we're probably fine.

kivikakk: Ohh, I see now. Feel free to reintroduce it in a PR if you like!

dgholz:
This change recognizes files with no extension and a perl shebang and POD contents as POD files, but I'd expect them to be shown as syntax-highlighted perl

e.g. https://github.com/sitaramc/gitolite/blob/5c2fe87019dadbecb1883bedd075cd07b07b2078/src/gitolite previously displayed as a highlighted perl script (good), but now displays as the embedded POD (requiring me to use the Raw view to be able to see the code).

Please fix it to correctly detect these kinds of files as perl (not POD) or revert it, thanks!

PR #3560, opened 2017-04-10
Add Solidity language (take 5)

veox:
### Summary

* Conflicts with PR #3559.
* Adds [Solidity](https://en.wikipedia.org/wiki/Solidity) as a language ([search result](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asol+%22pragma+solidity%22&type=Code)).
* Changes current vendor of `source.solidity` grammar from [davidhq/SublimeEthereum](https://github.com/davidhq/SublimeEthereum) to [caktux/language-ethereum](https://github.com/caktux/language-ethereum). The latter includes work on the former, and provides a few more grammar files. Language detection for those can be more easily integrated later on, if needed.
* Includes a simple heuristic to differentiate between Eagle soldermask `.sol` files and Solidity, and a test for that. The new Eagle samples are from [storborg/regerberate](https://github.com/storborg/regerberate), which is [MIT-licensed](https://github.com/storborg/regerberate/tree/ed85bb545109950946af90118b058abf9b0bcd3b/samples/eagle).
* Provides an example file written by your truly for [pygments](https://bitbucket.org/veox/pygments-main/src/95ac90c00cc6448404d751bf878cfb116b70cd1c/tests/examplefiles/example.sol?fileviewer=file-view-default), which has a BSD license ([Lightshow highlight](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fgithub.com%2Fcaktux%2Flanguage-ethereum%2Fblob%2Fmaster%2Fgrammars%2Fsolidity.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fveox%2Flinguist%2Fblob%2F894bae2d7bfe80165aa69f6da2004ff8f94ed396%2Fsamples%2FSolidity%2Fpygments-example.sol&code=)).
* In general, makes sure no GPLv3 code makes it into github/linguist, including git history.
* Attempts to follow [contributing guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language).

### Previous attempts

[Sorted chronologically](https://github.com/github/linguist/pulls?q=is%3Apr+solidity+sort%3Acreated-asc):

* PR #2634 - closed as stale, due to possible mis-classification of many non-Solidity files with `.sol` extension;
[...]

PR #3736, opened 2017-07-25
Added a new extension for solidity language

quasisamurai:
Usually, solidity files has a '.solidity' extension. But me and over the 70% blockchain programmers using Solidity as main smart contracts langugage use '.sol' extension as well. So this string allow toh ighlite a .sol files. Cheers :)

### DIFF
--- a/grammars.yml
+++ b/grammars.yml
@@ -119,6 +119,7 @@ vendor/grammars/SublimeClarion:
 - source.clarion
 vendor/grammars/SublimeEthereum:
 - source.solidity
+- source.sol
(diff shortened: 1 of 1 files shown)


lildude:
Unfortunately, I'm going to need to close this in favour of prior art, which is blocked by the fact Linguist doesn't currently have a concept of "not a language" or a default fallback which means the many many many other `.sol` files which aren't Solidity will be incorrectly classified.

See https://github.com/github/linguist/pull/3560 for more details.

Duplicate of #3560

PR #3740, opened 2017-07-26
Change Type Language type to markup

goodmind:
I looked at Protobuf and it has `type: markup` but initially I thought that this type should be used for HTML-like languages only and not RPC languages like Protobuf.

(markup type is very misleading at first)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4514,7 +4514,7 @@ Twig:
   codemirror_mime_type: text/x-twig
   language_id: 377
 Type Language:
-  type: data
+  type: markup
   aliases:
   - tl
   extensions:


Alhadis:
> but initially I thought that this type should be used for HTML-like languages only and not RPC languages like Protobuf

No, actually, you're 100% right. I'm not sure Protobuf should be classified as `markup` either. I haven't had much expose to it, but what I've seen is very much data-typed in nature.

@goodmind, thanks for bringing this up. I'll look at getting Protobuf corrected (or at least reviewed) shortly. In the meantime, Type language's current, uh, "type" of `data` is correct.

(That last sentence was hard to write and even harder to read... 😢 )

guidenbond:
`[https://en.wikipedia.org/wiki/Wikipedia:Community_portal(url)`

PR #3741, opened 2017-07-27
Add flow-typed to vendor.yml

iWader:
flow-typed provide third-party library definitions for flow. They're generated js files and should be always committed to the repo.

https://github.com/flowtype/flow-typed


lildude:
Can you merge master into your branch so the recent CI changes can be picked up thus allowing this PR to pass CI.

PR #3748, opened 2017-07-31
tokenizer: capture non-ASCII identifiers

smola:
`\w` captures only ASCII letters and numbers. Changed to [[:alnum:]]
to capture any Unicode letter and digit. This makes the tokenizer
work properly on non-English based langiages (e.g. 1C Enterprise).

### DIFF
--- a/lib/linguist/tokenizer.rb
+++ b/lib/linguist/tokenizer.rb
@@ -107,7 +107,7 @@ module Linguist
           tokens << token
 
         # Regular token
-        elsif token = s.scan(/[\w\.@#\/\*]+/)
+        elsif token = s.scan(/[[[:alnum:]]_\.@#\/\*]+/)
           tokens << token
 
         # Common operators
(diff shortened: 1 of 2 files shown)


Alhadis:
> `\w` captures only ASCII letters and numbers.

That's dependent on the regex engine in question. Ruby uses Oniguruma, which is [Unicode-aware](https://github.com/kkos/oniguruma/blob/1838f19/doc/RE#L45-L46) by default, IIRC.

smola:
@Alhadis If I run the test I added without the fix, this is its output:

```
  1) Failure:
TestTokenizer#test_utf8_tokens [test/test_tokenizer.rb:118]:
Expected: ["Функция", "الكون"]
  Actual: []
```

pchaigno:
@Alhadis `[[:alnum:]]` is safer in any case, no?

Alhadis:
Possibly; it seems like Oniguruma only defaults to Unicode in Atom, but not Ruby... never mind then. :)

smola:
Closing. I guess #3846 makes this obsolete.

Issue #353, opened 2013-01-13
Invalid byte sequence error when reading files with malformed utf-8 sequences


mrorii:
For example, running linguist on [this file](https://raw.github.com/leoniedu/CongressoAberto/a4785785cb37e8095893dc411f0a030a57fd30f8/CongressoAbertoWP/wp-includes/js/swfupload/swfupload.js) throws an invalid byte sequence error:

<pre>
$ wget https://raw.github.com/leoniedu/CongressoAberto/a4785785cb37e8095893dc411f0a030a57fd30f8/CongressoAbertoWP/wp-includes/js/swfupload/swfupload.js
$ linguist swfupload.js
/Users/orii/.rvm/gems/ruby-1.9.3-p286/gems/github-linguist-2.4.0/lib/linguist/blob_helper.rb:209:in `split': invalid byte sequence in UTF-8 (ArgumentError)
        from /Users/orii/.rvm/gems/ruby-1.9.3-p286/gems/github-linguist-2.4.0/lib/linguist/blob_helper.rb:209:in `lines'
        from /Users/orii/.rvm/gems/ruby-1.9.3-p286/gems/github-linguist-2.4.0/lib/linguist/blob_helper.rb:240:in `loc'
        from /Users/orii/.rvm/gems/ruby-1.9.3-p286/gems/github-linguist-2.4.0/bin/linguist:24:in `<top (required)>'
        from /Users/orii/.rvm/gems/ruby-1.9.3-p286/bin/linguist:23:in `load'
        from /Users/orii/.rvm/gems/ruby-1.9.3-p286/bin/linguist:23:in `<main>'
</pre>

arfon:
I'm going to close this. Ruby 2.0 is nearly two years old now and I just don't see us investigating this any time soon sorry.

If anyone else wants to take a stab at this then please be my guest :smile:

pchaigno:
@arfon I get this error on [the original file of this post](https://raw.github.com/leoniedu/CongressoAberto/a4785785cb37e8095893dc411f0a030a57fd30f8/CongressoAbertoWP/wp-includes/js/swfupload/swfupload.js) with Ruby 2.2:
```
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux]
```

arfon:
Well then.

echatman:
I have also encountered this many times using ruby 2.2. Here is a recent stack trace:

    $ linguist swfupload.js 
[...]

PR #3749, opened 2017-07-31
replace non-UTF-8 characters before tokenizing

smola:
This prevents later encoding errors in the tokenizer
whenever a file with non-UTF-8 characters reaches the
classifier.

Added GLSL sample with non-UTF-8 characters. Extracted
from wangdingqiao/noteForOpenGL, licensed unther the MIT
license. See:
https://github.com/wangdingqiao/noteForOpenGL/blob/40c014ca23ebdb9a33da86278ba232594c1b9f1e/geometryShader/geometryShader1/scene.gs

### DIFF
--- a/lib/linguist/tokenizer.rb
+++ b/lib/linguist/tokenizer.rb
@@ -13,6 +13,7 @@ module Linguist
     #
     # Returns Array of token Strings.
     def self.tokenize(data)
+      data = data.encode('UTF-8', 'UTF-8', :invalid => :replace)
       new.extract_tokens(data)
     end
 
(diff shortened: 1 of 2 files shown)


pchaigno:
Thanks for the fix!

Have you checked if it fixes #353?

smola:
@pchaigno It was supposed to, but I've realized it doesn't (see Travis results). I'm working on it.

pchaigno:
Many have tried... (none of them came back the same)
If you succeed you'll be our everlasting hero! :bow:

lildude:
@smola are you still working on this?  The classifier has changed so may not be needed or your changes may need a slight tweak.

PR #3743, opened 2017-07-28
Refine definition of KiCad

Alhadis:
What the site [currently classifies](https://github.com/github/linguist/blob/b94e018/lib/linguist/languages.yml#L2117-L2125) as **KiCad** is, according to [their documentation](http://kicad-pcb.org/help/file-formats/), three entirely different formats:

* **Schematic files:** Two different files, same format (and extension). Still in use.
	* `.sch`: [5,501 samples](https://github.com/Alhadis/Linguist-Silo/tree/sch) collected out of [~224,968 results](https://github.com/search?q=extension%3Asch+NOT+nothack&type=Code)

* **Board files:** Original PCB layout format, now superseded by new S-expression format.

	* `.brd`: [2,907 samples](https://github.com/Alhadis/Linguist-Silo/tree/brd) collected out of [~107,658 results](https://github.com/search?q=extension%3Abrd+NOT+nothack&type=Code)

* **Pcbnew layout:** Primary PCB format used by KiCad suite, with a syntax based on S-expressions:
	* `.kicad_pcb`: [16,808 results](https://github.com/search?q=extension%3Akicad_pcb+NOT+nothack&type=Code)
	* `.kicad_mod`: [284,942 results](https://github.com/search?q=extension%3Akicad_mod+NOT+nothack&type=Code)
	* `.kicad_wks`: [306 results](https://github.com/search?q=extension%3Akicad_wks+NOT+nothack&type=Code)
	* `fp-lib-table`: [4,482 results](https://github.com/search?q=filename%3Afp-lib-table+NOT+nothack&type=Code)

Technical specifications of each format are covered in a [PDF file](http://bazaar.launchpad.net/~stambaughw/kicad/doc-read-only/download/head:/1115%4016bec504-3128-0410-b3e8-8e38c2123bca:trunk%252Fkicad-doc%252Fdoc%252Fhelp%252Ffile_formats%252Ffile_formats.pdf/file_formats.pdf), downloadable from KiCad's website.

I collected a fresh harvest of both `sch` and `brd` files, with the breakdown graphs included in each silo:
* [`.sch` breakdown](https://github.com/Alhadis/Linguist-Silo/blob/sch/breakdown.log)
* [`.brd` breakdown](https://github.com/Alhadis/Linguist-Silo/blob/brd/breakdown.log)

## Notes
[...]

PR #3751, opened 2017-08-02
Fix classification of bogus "markup" languages

Alhadis:
## UPDATED
Okay, further digging around the corners of `languages.yml` revealed some other egregious cases of bogus "markup" languages, which my comment below describes in depth. I've updated the OP with a summary to recap what's changed:

* **ApacheConf** and **Nginx** configuration files recategorised from `markup` to `data`. The [latest "trending" Nginx repos](https://github.com/trending/nginx) reveals no physics-defying typesetting logic that can explain how a list of server variables controls the flow of typography (I've seriously been spending too much time in `troff`, honestly...)

* **Web Ontology Language** was tagged as `markup`, presumably by a submitter who was mislead by the connection to XML (possibly before XML was deemed to be data-typed, rather than real markup).

* **Eagle** CAD files specify PCB drawing data for the Eagle CAD suite. See https://github.com/github/linguist/pull/3743#issuecomment-319643063 for long-winded rationale and too many tangential anecdotes.

* **XPages** was a dead giveaway with its only two file-extensions being `.xsp-config` and `.xsp.metadata`. Didn't take much research to figure out they're just [XML-based configuration files](https://github.com/jesse-gallagher/Raidomatic/blob/6a3b28ebcb591f6dfb8d9eb8dd3f7b1c18035401/forums/CustomControls/DateTime.xsp-config) for a [web application framework](http://xpages.info/XPagesHome.nsf/Home.xsp) named "XPages". On the plus side, I also added XML highlighting for it, so at least its data [will look nicer now](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=text.xml&grammar_url=&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fjesse-gallagher%2FRaidomatic%2Fblob%2F6a3b28ebcb591f6dfb8d9eb8dd3f7b1c18035401%2Fforums%2FCustomControls%2FDateTime.xsp-config&code=). =)

* **Pure Data** was only investigated for laughs because of the name, which looked ironic juxtaposed to a `type` of `programming`. After looking it up though, it turns out that the... frikkin' language really is ["pure data"](https://github.com/halfbyte/PureDataLibrary/blob/master/zerhacker.pd). I'll quote [PD's FAQ here](https://puredata.info/docs/faq/whatispd):

	> **Q:** _"What Is Pd?"_
	> **A:** `Pd` stands for `Pure Data.` Pd is a real-time software system for live musical and multimedia performances. It is in active development by Miller Puckette, and perhaps others. The system is unfinished, but quite usable for sophisticated projects.

	And the gloss on [PD's homepage](https://puredata.info/) makes it pretty obvious it isn't even programmer-related:

	> Pure Data is an open source visual programming environment [ … ] Pd enables musicians, visual artists, performers, researchers, and developers to create software graphically without writing lines of code. Pd can be used to process and generate sound, video, 2D/3D graphics, and interface sensors, input devices, and MIDI

	One look at the [sort of stuff](https://github.com/breedx2/pure-data/blob/master/abstractions/ctable.pd) that `.pd` files contain, and you know it's not been hand-written... Though to be fair, this misclassification was added [in the same commit](https://github.com/github/linguist/commit/8f46cd07487b6c5ce034ef7a9a3e19f999914513#diff-3552b1a64ad2071983c4d91349075c75R621) that Linguist's first two language-types were defined: which were limited to `Programming` and `Markup`. Guess it escaped notice and remained unnoticed for ~6 years.

--------------

### Original post that started this search:

@goodmind pointed out a confusing oddity in #3740:

> I looked at Protobuf and it has `type: markup` but initially I thought that this type should be used for HTML-like languages only and not RPC languages like Protobuf. (markup type is very misleading at first)

He's right. My understanding of Protobuf is that it's used for defining APIs, not marking up presentational content.

The current `type` classification for Protocol Buffer was added by @qqshfox in February 2013 (#392). I'd say Linguist's handling of language-types has become much more refined in the years since, but I feel I should ping @qqshfox to ask if he feels `markup` would still be a more suitable descriptor for Protocol Buffer?

My exposure with Protobuf is rather minimal, but I was actually reading about it recently after submitting [`google/fonts#1094`](https://github.com/google/fonts/pull/1094), and then wondering about the missing syntax highlighting for its "text-based component".


Alhadis:
Found a few more egregious cases:

* [`ApacheConf`](https://github.com/github/linguist/blob/7451424/lib/linguist/languages.yml#L211-L212): Good LORD, that's [been there](https://github.com/github/linguist/commit/04a2845e91e407f2846269aba19d40cf41fe1574) for ~5 years. 😨  Seems Linguist's definition of language-types certainly was... rough-around-the-edges. Does anybody else agree that this could only possibly be classified as `data`...?

* [`API Blueprint`](https://github.com/github/linguist/blob/7451424/lib/linguist/languages.yml#L86-L93): Not at all familiar with what this is, but the fields which reference Markdown/GFM looked a little bit... off. After taking a cursory glance, I'd say this is [closer to prose than markup](https://github.com/github/linguist/commit/828fc7a19ba7dbf36da48aa3b02486c6c9a6a4cd#diff-1fc482854b262123c604495c7f0e1a1d), as its content looks very reminiscent of rST or Markdown. I'm dye that alone, but bringing it up in case somebody with more knowledge of API Blueprints wants to chime in.

* [`nginx`](https://github.com/github/linguist/blob/7451424/lib/linguist/languages.yml#L2786-L2790): Added at the same time as [`ApacheConf`](https://github.com/github/linguist/commit/ae753e6e88d743f47b1f43f1488e39af6baf08ca). Definitely configuration data. Also, I now recall one of my repositories being misclassified as ApacheConf in the past, due to one or two server-config files, which happened to be very heavy with commented-out lines. I'd say this would fix a lot of repositories out there that've been erroneously categorised as "ApacheConf" codebases...

Will update this again soon-ish.

Alhadis:
Okay, there's no room for subjectivity here. `XPages` is [full-blown XML](https://github.com/jesse-gallagher/Raidomatic/blob/6a3b28ebcb591f6dfb8d9eb8dd3f7b1c18035401/forums/CustomControls/DateTime.xsp-config), so I've reclassified it as `data`, with `Eagle` given the same treatment.

If anybody here has managed to typeset a published document using [an `nginx` config](https://github.com/h5bp/server-configs-nginx/blob/3bda5b93edba147d51760e900c2079828a7dc274/nginx.conf), please let me know how you managed that, because that's some Houdini-level sh*t.

goodmind:
Well, so `data` type shouldn't be displayed in repository languages list?

Alhadis:
Would you prefer to be drowning in JSON and XML repositories everywhere you turned your head?

Because that's precisely what would happen. The volume of data-type languages often far outweighs the volume of actual programming or markup languages.

sbassoon:
I'm writing about the decision to count "Pure Data" as a type: data language, as opposed to programming language, as it had previously been counted until this update. For reference, quoted below is the part of the [original logic](https://github.com/github/linguist/pull/3751#issue-133680421) behind changing the designation:

> Pure Data was only investigated for laughs because of the name, which looked ironic juxtaposed to a type of programming. After looking it up though, it turns out that the... frikkin' language really is "pure data". I'll quote PD's FAQ here:
> 
> > Q: "What Is Pd?"
> > A: Pd stands for Pure Data. Pd is a real-time software system for live musical and multimedia performances. It is in active development by Miller Puckette, and perhaps others. The system is unfinished, but quite usable for sophisticated projects.
>
> And the gloss on PD's homepage makes it pretty obvious it isn't even programmer-related:
> 
> > Pure Data is an open source visual programming environment [ … ] Pd enables musicians, visual artists, performers, researchers, and developers to create software graphically without writing lines of code. Pd can be used to process and generate sound, video, 2D/3D graphics, and interface sensors, input devices, and MIDI
>
> One look at the sort of stuff that .pd files contain, and you know it's not been hand-written... Though to be fair, this misclassification was added in the same commit that Linguist's first two language-types were defined: which were limited to Programming and Markup. Guess it escaped notice and remained unnoticed for ~6 years.

Pure Data is a visual programming language, as well as a data flow programming language, which is primarily edited through the Pd environment itself. However, the programs produced are able to be run independent of the program and have been used in mobile implementations, internet streaming implementations, and as audio synthesis support in larger programs.

The classification of "type: data" is incorrect, because though the name of the language is called "Pure Data," the actual files are programs that perform functions primarily within audiovisual playback, synthesis, and processing, as well as other algorithmic procedures operating on the data needed to work with audio and video.

I understand the purpose of these revisions, especially when it comes to XML-like languages that are just data configurations files, or database files that only house data. But as it stands, the rationale listed in #3743 that elucidates what counts as type:data:

> KiCad formats are data formats, not programming languages. I'm not sure why @pchaigno chose the latter in #2309 / #2187, but these formats are essentially just lists of coordinates, property lists, and object descriptions. XML has more in common with programming languages than these do.

does not actually include what a Pure Data file is. Though there are indeed lists of coordinates, but the coordinates are the location of objects and functions on the canvas of the program as it would be rendered in a visual environment. There are no "property lists" or "object descriptions." Instead, these objects call functions and denote the connections between objects in order to run the program itself. Just as Java code files need a Java compiler to be run, Pure Data needs the Pure Data compiler to be run -- but not universally, because as mentioned earlier, Pd files are used in mobile, web, and polyglot implementations where the compiling happens behind the scenes and the sound works without the canvas being displayed, or the Pure Data environment itself being utilised.

Finally, I'd like to close with an objection to some of the language that's used in the original post:

> And the gloss on PD's homepage makes it pretty obvious it isn't even programmer-related:

and

> One look at the sort of stuff that .pd files contain, and you know it's not been hand-written...

Obviously this is going to immediately descend into a philosophical argument about what a programming language can be, but

1. Pure Data is used by programmers in primarily audiovisual implementations, but having a (comparatively to Python or C++) small implementation market doesn't disqualify something from being a programming language, and thus, programmer-related;

2. Pure Data is "hand-written" in the sense that the program files were created by actual people making actual decisions about what functions to call and how to control the flow of data within the program. The files themselves contain coordinates for objects on a canvas, as is the standard operating procedure for many visual programming languages. The logic that governs how these programs work is entirely up to the programmer, but the coordinates that appear in the program file are necessary for the compiler to understand how objects connect to one another.

I think it's troubling to assert the above two points, especially on languages that one is unfamiliar with, because it's a bit gate-keepery regarding who can be a programmer. Pure Data has a wide share of the audio programming market, particularly for creative implementations, along with Max/MSP, which is also a visual data flow programming language. Pd projects are implemented and developed on Github (the Pure Data repo itself [is distributed here](https://github.com/pure-data/pure-data)), and while the program files they create don't look like a standard written language such as Python or C++, that's only because of their implementation within a visual data flow paradigm. My primary concern with this point is that if Linguist's idea of "programming language" only includes specific (ie written) kinds of programming paradigms, then it excludes certain fields where other paradigms like visual or data flow languages are much more common.

Hopefully I've done some work to convince you that these are not type:data files, but actual programs that do more than just store information or function as a XML-like markup file. (At the very least, you should also list Max/MSP as type:data instead of also counting it as a programming language, because it's the exact same program architecture.)

PR #3752, opened 2017-08-02
Make Language methods more resilient to non-String input

jdennes:
The erroring test in 3e594a58ed70c278b0a5aa299dea45c18f40e531 shows the exception that we're seeing when using Linguist on GitHub.com and GitHub Enterprise when calls like `Language.find_by_alias(true)` end up being made:

```
  1) Error:
TestLanguage#test_does_not_blow_up_with_non_string_lookup:
NoMethodError: undefined method `downcase' for true:TrueClass
    /home/travis/build/github/linguist/lib/linguist/language.rb:128:in `find_by_alias'
    /home/travis/build/github/linguist/test/test_language.rb:274:in `test_does_not_blow_up_with_non_string_lookup'
```

Instead of raising an error, this handles non-String input so that `nil` is returned instead.

The problem can be recreated by creating a `.gitattributes` file that sets `linguist-language` to `true`, rather than a `String`.

So like this:

```
*.rb linguist-language
```

Rather than using the documented instructions like this:

```
*.rb linguist-language=Ruby
```

From the docs for the _Set_ state at https://git-scm.com/docs/gitattributes:

> The path has the attribute with special value "true"; this is specified by listing only the name of the attribute in the attribute list.

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -109,7 +109,7 @@ module Linguist
     #
     # Returns the Language or nil if none was found.
     def self.find_by_name(name)
-      return nil if name.to_s.empty?
+      return nil if !name.is_a?(String) || name.to_s.empty?
       name && (@name_index[name.downcase] || @name_index[name.split(',').first.downcase])
     end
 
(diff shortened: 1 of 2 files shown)


jdennes:
> How soon do you think we should make a new release for this @jdennes?

@lildude: Soon would be good. There was a closed issue about this, which I reopened because of noise being generated again by repositories that have `.gitattributes` files that cause this problem.

PR #3559, opened 2017-04-09
Add the Solidity language without an associated extension

frangio:
As discussed in #2634 we can't associate the `.sol` extension with Solidity until the `not-a-language` feature is implemented.

Meanwhile, though, the language should be included in Linguist so that repositories can manually set the language via `.gitattributes`.

PR #3755, opened 2017-08-04
Add the Solidity language without an associated extension (updated)

frangio:
After the discussion in #3560 and following @pchaigno's suggestion in #3559, this PR adds the Solidity language, together with two sample files, but without associating an extension to it.

To sum things up, the associated `.sol` extension is shared by files on GitHub that don't correspond to a language, and it's impossible for a linguist heuristic to return that as a result. This is referred to as the `not-a-language` feature, which is apparently non-trivial to implement. In the meantine, we've agreed on adding the language without an extension, so that repositories can at least use `.gitattributes` overrides to get syntax highlighting.

Here's [a Github search](https://github.com/search?utf8=%E2%9C%93&q=pragma+solidity&type=Code) for the language with more than 25000 results.

One of the sample files is taken [from OpenZeppelin](https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/crowdsale/RefundVault.sol), which is MIT licensed. The other one was written by @veox specifically to demonstrate the language syntax, and is BSD licensed.


samples/Solidity/RefundVault.sol
```diff
@@ -0,0 +1,56 @@
+pragma solidity ^0.4.11;
```
pchaigno: The Travis CI build fails because of these samples. Files in `samples/` are used by the Bayesian classifer and are expected to be registered in `languages.yml` (through their file extension or filename). Since it's not the case here, could you move these example files to `tests/fixtures/`?

lib/linguist/languages.yml
```diff
@@ -4227,6 +4227,11 @@ Smarty:
   codemirror_mime_type: text/x-smarty
   tm_scope: text.html.smarty
   language_id: 353
+Solidity:
+  type: programming
+  ace_mode: text
+  tm_scope: source.solidity
+  language_id: 237469032
```
pchaigno: Could you add a color for Solidity? We usually try to use colors from official logos if any.

samples/Solidity/RefundVault.sol
frangio: Sure! Maybe there should be a note about that in the contributing guide.

pchaigno: It's a pretty rare case. It's only the second language we'd be adding that doesn't have an extension or a filename defined.

lib/linguist/languages.yml
frangio: The [logo](https://solidity.readthedocs.io/en/latest/_images/logo.svg) only has dark shades of gray. I'm not sure it'll look very nice but I've set it to one of those.

frangio:
I'm not sure I understand what I'm supposed to do with the sample files. The Travis CI build is still failing.

frangio:
Regarding the color, it seems to be too close to an existing one. I've tried all the colors of the logo and they're too close too. Would it be ok to submit without color, or should I make one up?

veox:
For the [color failure](https://travis-ci.org/github/linguist/jobs/261126837#L1283), try `#373737`. That seems to work for me locally.

veox:
From https://github.com/github/linguist/pull/3755#discussion_r131441202:

> The Travis CI build fails because of these samples. Files in samples/ are used by the Bayesian classifer and are expected to be registered in languages.yml (through their file extension or filename). Since it's not the case here, could you move these example files to tests/fixtures/?

From https://github.com/github/linguist/pull/3755#discussion_r131464062:

> It's a pretty rare case. It's only the second language we'd be adding that doesn't have an extension or a filename defined.

@pchaigno Would the first one be `Data`?

If so, does that mean that the ["No language" test cases](https://travis-ci.org/github/linguist/jobs/261126837#L1275) still failing would require special handling, like `Data`?

For ref, see failing tests:

* https://github.com/frangio/linguist/blob/dee142e2731dbdcf9bf2f595344dd31e263351de/test/test_blob.rb#L270
* https://github.com/frangio/linguist/blob/dee142e2731dbdcf9bf2f595344dd31e263351de/test/test_file_blob.rb#L655

frangio:
@pchaigno Any thoughts?

onbjerg:
@veox @frangio The other language that doesn't have an extension or a filename defined is "Python Console" I think.

There's no fixtures or sample files for that language, so I suppose the only way to make this pass is to remove the fixtures (which feels scary).

The files in `test/fixtures` are also tested like the ones in samples: https://github.com/github/linguist/blob/b94e018c3af1ca5b3858f2249253d076b26df6a0/test/test_blob.rb#L256-L279

Vishesh-Gupta:
Is there any update on this issue? I think linguist does not check for solidity currently.

oktapodia:
Thanks for this pull request, any update?

lildude:
Closing in favour of https://github.com/github/linguist/pull/3973 which will add Solidity support with the `.solidity` extension and will then allow peeps to use an override for the `.sol` or any other extension.

Issue #3775, opened 2017-08-15
Javascript/JSX single-quote/apostrophe breaks syntax highlighting


helios1138:
Basically, this is the issue https://i.gyazo.com/45d9921e631ea8274d97c4af9c062628.png

helios1138:
I know it says to report issues to specific language repo, but the JSX repo seems to be version-locked/deprecated: https://github.com/github-linguist/language-babel

lildude:
Yeah, this is a bit of a known issue and stems from changes made to the upstream grammar that aren't compatible with the regex engine (Ruby's PCRE compiled without the `PCRE_UCP` flag - see https://github.com/github/linguist/issues/3291#issuecomment-255750842 why) forced us to lock to a particular version. More details in https://github.com/github/linguist/issues/3044

I've done a quick crude test and maybe switching to https://github.com/babel/babel-sublime may be the best solution going forward. It's not as actively developed as https://github.com/gandm/language-babel but it is recommended for Sublime by Facebook [here](https://github.com/facebook/react/wiki/Complementary-Tools#editor-integration) and doesn't experience the problem...

[...]

PR #3779, opened 2017-08-16
Switch JSX grammar to bable-sublime

lildude:
In an attempt to stop having to continue using an outdated and old release of the language-babel grammar, which we're not even attempting to keep in sync, this PR proposes switching out our outdated github-linguist/language-babel in favour of Facebook's [recommended grammar for Sublime](https://github.com/facebook/react/wiki/Complementary-Tools#editor-integration) babel/babel-sublime.

Limited testing suggest this solves the recently identified problems and the reason for the old grammar (see https://github.com/github/linguist/issues/3775#issuecomment-322459597) but the only thing I'm not sure about is testing picks up a duplicate for the scope:

```
WARN: Duplicated scope source.js
  Current package: vendor/grammars/language-javascript
  Previous package: vendor/grammars/babel-sublime
```

Whilst both do indeed use the `source.js` scope, `convert-grammars` only pulls in this scope for the babel-sublime grammar now.

Is this going to cause us a problem @Alhadis or @pchaigno?

This PR also does a little tidying up thanks to the `script/add-grammar` pulling other scripts which found two missing references in the `vendor/README.md` file and also an old license.

This should fix https://github.com/github/linguist/issues/3775


Alhadis:
Something I've been meaning to ask: how is it that a scope can be referenced by `languages.yml`, yet have no entry in `grammars.yml`?

lildude:
> Something I've been meaning to ask: how is it that a scope can be referenced by languages.yml, yet have no entry in grammars.yml?

No idea 😄. I was wondering the same thing. I wonder if it's something to do with the duplication that isn't being handled well by the `convert-grammar` script, assuming `grammars.yml` _should_ have the mentioned scope.

Alhadis:
I've a very bad feeling about all this. First of all, the grammar we're switching to should be using a `scopeName` of `source.jsx`, not `source.js`.

Secondly, I think this *will* cause conflicts with the JavaScript grammar, although that's going by assumption. :\ This just isn't sitting right with me...

I honestly *this* goddamn close to going through `language-babel` by hand and translating every incompatible RegExp feature myself. :p

lildude:
> I've a very bad feeling about all this. First of all, the grammar we're switching to should be using a `scopeName` of `source.jsx`, not `source.js`.

Yeah, I'm not 100% comfortable either now I've dug into this. Interestingly, this grammar _used_ to scope to `source.jsx` but switched to `source.js` in https://github.com/babel/babel-sublime/commit/b915873046084fe785cc04e74e15116ca39479e3 - I can't see why though.

> I honestly this goddamn close to going through language-babel by hand and translating every incompatible RegExp feature myself. :p

Well, you do maintain a few grammars we use already, so what's one more 😉 . Seriously, it would be great if we could get this into the upstream repo, but I don't think we'd succeed.

Given the uncertainty from both @Alhadis and me, I'm canning this idea. Back to the drawing board.

Alhadis:
Haha, I wouldn't attempt to submit it upstream. =) The author has made his opinion on the subject quite clear. 

I'd be looking at updating GitHub's fork of it, I mean.

pchaigno:
Can't we fork his grammar and change the scope? It's under MIT...

Alhadis:
Well, that's not gonna help distract me from depression for very long, now, is it? ;p

Better this is done sooner or later, I suppose.

PR #3780, opened 2017-08-16
Updated color for Ceylon language

rkoeninger:
Golden-brown color taken from [Ceylon elephant logo](https://ceylon-lang.org/).

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -628,6 +628,7 @@ Ceylon:
   color: "#dfa535"
   extensions:
   - ".ceylon"
+  tm_scope: source.ceylon
   ace_mode: text
   language_id: 54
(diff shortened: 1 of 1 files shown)


lib/linguist/languages.yml
```diff
@@ -625,6 +625,7 @@ CartoCSS:
   language_id: 53
 Ceylon:
   type: programming
+  color: "#563a08"
```
Alhadis: Try `#dfa535` instead. Much closer to the colour of Ceylon's logo than `#563a08`.

rkoeninger: I tried a handful of similar colors, but they kept getting rejected. I thought only really dark colors would be left.

Thanks for the hint.

Issue #3712, opened 2017-07-05
Ruby language source repository


adaedra:
Hello,

Currently ruby language definitions (and a few others) are fetched from `aroben/ruby.tmbundle#pl`. However, this repository is not updated anymore. It contains a few enhancements, but also lacks a lot of some that have been added upstream.

I discussed the issue with the owner of the current repository here: aroben/ruby.tmbundle#2.

IMO, the ideal situation on this one would be to use `textmate/ruby.tmbundle` directly as source, but that may require the improvements or aroben's repos to be pushed upstream.

Another solution hinted in the issue above would be to keep a separate live fork with most recent improvements, but it requires someone to maintain it.

So what is your opinion on this issue? What are the next steps to resolve this, and have GitHub be more up-to-date on Ruby syntax highlighting?

lildude:
I watched the conversation unravel in https://github.com/aroben/ruby.tmbundle/pull/2 and I think the best thing for the long-term is to get those three changes you mentioned in that PR merged upstream so we can switch to using the upstream syntax.

Another solution would be to switch to the Atom grammar if it already has those changes.

adaedra:
Since #3782 is merged, this can be closed.

PR #3782, opened 2017-08-17
Change Ruby grammar source

adaedra:
As discuted in #3712, this changes the source repository for Ruby grammars from the aroben/ruby.tmbundle repository to the official Atom one.

I hope to have done things correctly.


Alhadis:
Good grief, *seriously?*

~~~
Warnings:
vendor/licenses/grammar/language-ruby.txt:
  - license needs reviewed: other.
288 dependencies checked, 1 warnings found.
The command "script/licensed verify" exited with 1.
cache.2
store build cache
~~~

/cc @pchaigno License detection is really becoming problematic... 😞

pchaigno:
I haven't dug into this bug. Is it an error message from Licensee or from us?

Alhadis:
The former, I believe...

Maybe it's time we added a pre-license check to cater for all the `language-*` grammars that have "strange" licenses with TextMate references? 😁

Alhadis:
@adaedra Travis seems stuck. You might need to force-push an amended commit, just to trigger the build again. Sometimes it does this...

adaedra:
Relaunched, thanks.

pchaigno:
I don't think we're suppose to change the `license.txt` files manually. There's something else that's causing this issue.

adaedra:
So what's the next step for this?

Alhadis:
Well, it all looks fine to me. I'm not 100% sure editing the license by hand is the right approach: normally, we whitelist the grammar's checksum instead.

I'll leave approval to @lildude though. ;)

pchaigno:
> normally, we whitelist the grammar's checksum instead.

Exactly! We should be whitelisting a hash value, not a repository name. Licensee is able to compute the hash for this repository (and even [recognizes the license](https://github.com/atom/language-ruby/blob/master/LICENSE.md)) so we shouldn't need to do this. The switch doesn't look urgent, so I'd prefer if we fixed our tests first...

lildude:
Ooops, I've caused a conflict by merging another PR. @adaedra can you please merge master into your branch and resolve the conflict.

> Exactly! We should be whitelisting a hash value, not a repository name. Licensee is able to compute the hash for this repository (and even recognizes the license) so we shouldn't need to do this. The switch doesn't look urgent, so I'd prefer if we fixed our tests first...

Yes, I agree. I'd prefer if we got to the bottom of why this is showing up as "other" when it is recognised correctly on GitHub.com. It's worth pointing out that GitHub.com is using newer versions of Licensed and Licensee. I'll give the last person to touch Licensed a nudge to get them to build a gem for the latest release so we can start using it and the later Licensee too.  This might help.

I'm going to hold off approving this PR until we've got :point_up: sorted.

adaedra:
I rebased the PR.

However, does this mean that I should remove the commit changing the license tag?

adaedra:
Hello,

Is there a way where to track the other issue, to know how it is progressing?

Alhadis:
Hey @adaedra,

Let's wait for @lildude to get back to us with an update before moving forward. =) Don't revert the commit just yet until we're given a green light.

lildude:
Sorry, been AFK for a while so I dropped the ball on this. I'll hopefully have time to look into the licensing issue early next week.

lildude:
I've looked into this and the problem here is down to the license file included in the repo being too different from what Licensee expects and the way it determines the actual license.

I've not managed to work out how GitHub.com gets the right license as it was set so long ago.

The best thing to do here is reverse the change to the license file in https://github.com/github/linguist/pull/3782/commits/9824c565d3fee8078f09c689d1df8d3c87608475 and instead add the hash to https://github.com/github/linguist/blob/fc1404985abb95d5bc33a0eba518724f1c3c252e/test/test_grammars.rb#L12

adaedra:
I currently only have rebased the commit.

I didn't find which hash I should add to this file, either the shasum of the license file, the cached license, or some other things. Also, `test/test_grammars.rb` does not seem unhappy about the added file, but `script/licensed verify` is, and changes I did on the first file did not seem to change anything for the second (It did, however, make the `test_grammars` not pass due to extra hash).

I have probably missed something somewhere, but I can't find what 🙃

lildude:
> I currently only have rebased the commit.

:+1:

As for the hash... I thought the hash was displayed in the event of a failure or similar issue but I think I may be mixing up my licensee/licensed/linguist work.

The hash we need is that reported by running `licensee` from within the `vendor/grammars/language-ruby` directory:

```
$ licensee
License: Other
Matched files: ["LICENSE.md", "package.json"]
LICENSE.md:
  Content hash: c87e7e574fca543941650e5b0a144b44c02c55d8  <--- THIS ONE
  Attribution: Copyright (c) 2014 GitHub Inc.
  License: Other
package.json:
  Confidence: 90.00%
  Matcher: Licensee::Matchers::NpmBower
  License: MIT License
$
```

You need to take the hash I've pointed out above and add it to the array I referenced earlier, ie at the bottom of this array https://github.com/github/linguist/blob/fc1404985abb95d5bc33a0eba518724f1c3c252e/test/test_grammars.rb#L12

pchaigno:
@lildude It looks like it's *licensed* complaining here. Do we really need this closed source gem for license checking? We seem to have better tests with the hash mechanism in `test_grammar.rb`.

lildude:
> It looks like it's licensed complaining here.

Ooops, you're right. I completely missed that - I really should stop juggling so many balls at once 😉.

@adaedra ignore what I said above, it won't help. 

> Do we really need this closed source gem for license checking?

We do as it does all the heavy lifting around extracting, caching and verifying the licenses. Whilst it's a PITA here, it does protect us from any other unapproved licenses sneaking in.

I have however since just discovered a PR for Licensed that might just help us here. I'll need to do some more testing and come back with my findings.

pchaigno:
> We do as it does all the heavy lifting around extracting, caching and verifying the licenses.

Okay for the extracting and caching part. I'm only proposing that we don't use it for the "verifying" :-)

> Whilst it's a PITA here, it does protect us from any other unapproved licenses sneaking in.

What does it do that we don't already have a mechanism for?

lildude:
> What does it do that we don't already have a mechanism for?

Validate the results of its caching behaviour 😁 . If it didn't verify its own caching we may have missed it picking "other" instead of MIT for the license when caching it.

> I have however since just discovered a PR for Licensed that might just help us here. I'll need to do some more testing and come back with my findings.

I've tested this PR and it does exactly what we want it to do and will result in the correct license being set in the cached license file.

So with that in mind, I'm sorry to say this @adaedra but your manual change of "other" -> "mit" in https://github.com/github/linguist/commit/9824c565d3fee8078f09c689d1df8d3c87608475 is the correct fix for this issue as this is essentially what the updated Licensed will do.

adaedra:
Updated and green, hope it's all good now! 😸

lildude:
LGTM. Thanks @adaedra

adaedra:
Thank you!

Do you have an idea of when it would be applied to GitHub?

lildude:
When I next have time to make a release which will probably be early October. 😄

50Wliu:
Hmm, looks like I have even more incentive to start working on language-ruby now 😀.

PR #3786, opened 2017-08-19
Update escape_utils requirement to >= 1.1, < 1.3

greysteil:
Updates requirements on [escape_utils](https://github.com/brianmario/escape_utils) to permit the latest version.

Generated on my fork using [Dependabot](https://dependabot.com), if you want to automate this stuff for the future.

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.executables = ['linguist', 'git-linguist']
 
   s.add_dependency 'charlock_holmes', '~> 0.7.5'
-  s.add_dependency 'escape_utils',    '>= 1.1', '< 1.3'
+  s.add_dependency 'escape_utils',    '>= 1.1', '< 2.0'
   s.add_dependency 'mime-types',      '>= 1.19'
   s.add_dependency 'rugged',          '>= 0.25.1'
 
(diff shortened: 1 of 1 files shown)


lildude:
Thanks for the PR. Is there any special requirement or need for this newer, yet still restricted, update? 

If so, why the restriction on <1.3?

greysteil:
Figured I knew less about the gem than you do - if you don't trust SemVer on it then I figured I'd keep your gemspec that way! I can update the upper bound to be `2.0` if you want?

lildude:
Thanks, but I'm still curious about if there's a particular reason for this other than it's out of date. 

For context, Linguist is tightly tied to the gem versions used on GitHub.com so updating dependency versions needs to be properly considered.

greysteil:
Was hoping to pull [this commit](https://github.com/brianmario/escape_utils/commit/10098ded5d61c80d4cb2285ab35a05c721df7f2e) into a project I'm working on, and Linguist is blocking me. Can obviously work around it if there are complexities to updating the gemspec here that I hadn't realised.

lildude:
Hmmm that commit was included in escape_utils 1.1.1 which is what Linguist is pulling in already.

greysteil:
Woah, hadn't realised they'd backported that. In that case I guess fine to close this.

Issue #3784, opened 2017-08-18
Kicad 


5N44P:
I've recently had some issue getting one of my repositories recognised as kicad by GitHub.
Then i started looking arround into the linguist and found some recent updates in the kicad handling process. See #3765 and #3743.

So, as a KiCad user: what's the point of doing that? It doesn't make much sense.
Kicad files usually travel together. While it's possible to have a repository made out of only layouts or schematics, that usually doesn't make much sense.

Moreover there are some important kicad files that are at the moment unclassified or misclassified:
* `.pretty` folders: **unclassified**. These are footprint libraries. But i don't know if it will be possible to classifiy a folder. 
* `.kicad_mod` files: **misclassified**. These are the single footprints. Usually these file are the content of a .pretty folder and they are the components available from the library. Now they are classified in "KiCad Layout", which isn't really the best option.
* `.pro` files: **misclassified**. These are project files, currently they are recognized as INI files. They contain all the proprieties of a given project. Usually when you open a kicad project, you open these files and from there the rest.
* `-cache.lib` file: **unclassified**. This is a project-specific file: it contains all the schematic symbols from user added libraries.
* `.net` file: **unclassified**. This is the netlist file, although it can be generated from the .sch or the .kicad_brd files, it usually travels along with the project. 
* `fp-lib-table` file: **misclassified**. This file contains the list of non-standard kicad libaries used in that particular project. Currently it is classified in "KiCad Layout", which makes no sense.
* `zones_dump.txt` file: **misclassified**. There's not much documentation on this, but exists in many kicad repos. At the moment it's classified as a TXT file, but it's a KiCad file in this case.

But, as i said, all those files make sense together, and they all form **a single kicad project**. 

So, in concusion, i think that [all the kicad files](http://kicad-pcb.org/help/file-formats/) should be classified in just a plain _KiCad_ language. Adding or moving the extension in the current system doesn't make any sense: a kicad project is made of **all these files, together**. Would you classify differently the `.c`, `.cpp` and  `.h` files from a c/c++ library?

Alhadis:
Hey @5N44P,

A repository's classification is drawn from the languages of the files it's comprised of. It's not (currently?) possible to manually declare the language of a repository, although several users have requested such a feature in the past, I believe.

[...]

PR #3744, opened 2017-07-28
Add support for Gerber files

Alhadis:
This is it. This is the moment that nobody's been waiting for. The anxiously awaited addition of **GERBER** support for GitHub!

What's **Gerber**, you ask? Well, if you haven't figured out from the name that it's [the de facto standard](https://en.wikipedia.org/wiki/Gerber_format) for PCB imaging formats, then I don't know what else to tell you. In all seriousness, the format has been standardised since at least 1998, and with more file extension schizophrenia than Markdown AND Roff together.

I wouldn't have expected such a format to have a widespread presence on GitHub, and... well, I was wrong:
* `.gbr`: [~37,727 results](https://github.com/search?q=extension%3Agbr+NOT+nothack&type=Code) / [2,953 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gbr)
* `.gbl`: [~14,608](https://github.com/search?q=extension%3Agbl+NOT+nothack&type=Code) / [2,862 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gbl)
* `.gbo`: [~12,124](https://github.com/search?q=extension%3Agbo+NOT+nothack&type=Code) / [2,847 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gbo)
* `.gbp`: [~4,773](https://github.com/search?q=extension%3Agbp+NOT+nothack&type=Code) / [2,599 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gbp)
* `.gbs`: [~21,498](https://github.com/search?q=extension%3Agbs+NOT+nothack&type=Code) / [2,889 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gbs)
* `.gko`: [~2,656](https://github.com/search?q=extension%3Agko+NOT+nothack&type=Code) / [2,237 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gko)
* `.gtl`: [~14,481](https://github.com/search?q=extension%3Agtl+NOT+nothack&type=Code) / [2,856 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gtl)
* `.gto`: [~13,105](https://github.com/search?q=extension%3Agto+NOT+nothack&type=Code) / [2,860 sampled results](https://github.com/Alhadis/Linguist-Silo/tree/gto)
[...]

PR #3799, opened 2017-08-29
Change KiCad Board language to KiCad Legacy Layout

seppestas:
KiCad .brd files and .kicad_pcb files have the same purpose, they are both source files for PCB layouts. Having one of the file types named "KiCad Board" and the other one "KiCad Layout" can cause confusion since it implies they are not the same thing.

The [.brd files use the old, legacy layout format](http://kicad-pcb.org/help/file-formats/#_native_file_formats) that is [not actively used anymore](https://github.com/search?utf8=%E2%9C%93&q=language%3A%22KiCad+Board%22&type=Repositories&ref=advsearch&l=KiCad+Board&l=). Having it come before the KiCad Layout language in the Language Selection list and not having it flagged as legacy can cause people to select it when searching for KiCad layout files:

screen shot 2017-08-29 at 10 27 56

See #3784 for the ongoing discussion about the KiCad file formats. This PR builds upon the changes introduced by #3743.


seppestas:
cc @Alhadis

Alhadis:
Alright. I'm the one who started this mess, so I should logically be the one to help with cleaning up.

First of all, thank you sincerely for stepping in and lending your knowledge here. =) I think it's pretty clear from #3743 that my knowledge of PCB-related technologies is at utmost zero, and I found working with PCB programs to be a frustrating and awkward affair. Had I known there were that many things unaccounted for, I surely would've bugged the owners of the projects I sourced samples from for their assistance.

Secondly, it was mentioned several times in #3784 that there were still some project files being unaccounted for. Which of a KiCad project's files have yet to be recognised by Linguist?

Alhadis:
@seppestas As an aside, it would also be nice to have educated feedback on the [grammar-package](https://github.com/Alhadis/language-pcb) I wrote to provide syntax highlighting for PCB formats... ;)

seppestas:
This PR just fixes an issue introduced in #3743 that makes it harder to find KiCad projects, namely confusion between "KiCad Boards" and "KiCad Layouts" and the fact that the barely used and deprecated "KiCad board" layout is listed before the "KiCad Layout" format in the advanced search list.

I think the further separation of KiCad library / project files requires some more discussion, and I don't plan to add these kind of changes in this PR.

lib/linguist/languages.yml
```diff
@@ -2170,6 +2163,13 @@ KiCad Layout:
...
   codemirror_mime_type: text/x-common-lisp
   language_id: 187
+KiCad Legacy Layout:
+  type: data
+  extensions:
+  - ".brd"
+  tm_scope: source.pcb.board
+  ace_mode: text
+  language_id: 140848857
 KiCad Schematic:
```
Alhadis: Are schematics still in use by KiCad projects? From what I saw, they featured a very similar syntax to the legacy `.brd` files.

5N44P: Are you referring to `.sch` files? 
Yes, those are not deprecated

Alhadis:
There was another file format I noticed that appeared related to KiCad projects... it had an extension of `.drl` and looked quite similar to the `.brd` file format, IIRC. [Here's one example](https://github.com/r2aiv/motor_driver/blob/eaa5e3c2f745a467b6b55ff642caf41fba41be7c/Gerber/pwm_motor_ctrl_2_order/pwm_motor_ctrl_2.drl).

Should this be included as a KiCad-related file too?

5N44P:
`.drl` is an extension used for Gerber drill files. Gerbers are the main fabbrication output for pcbs, and they are kind of standard. (some EDAs export `txt` files instead of `drl`). 

Drill files should contain, for each drill hole in the pcb, the coordinates and the tool number. Sometimes there's also a table correlating tool number with the hole diameter, so the manufacturer knows which drill bit sizes load into the machine. 

As an example: when you're done with the layout of your board, you generate your gerber and drill files, pack them in a `zip` folder and send it to the manufacturer. 

So `.drl` it's a generic Gerber file, not specific to KiCad. If it has to be classified, l'd put it in a "Gerber" language (which I don't know if it exists yet) 

If you have any pcb/KiCad/EDA questions, please ask, i'd be glad to help :smile:

Alhadis:
Glad to know I have somebody to turn to, finally! =) Thank you. 😄 

> If it has to be classified, l'd put it in a "Gerber" language (which I don't know if it exists yet)
Already taken care of; see #3744. ;) And I now recall it was when I was researching Gerber images that I encountered the `.drl` extension and wondered what it was for... I left it out due to uncertainty. Though I should probably think about adding it back, haha.

lib/linguist/languages.yml
seppestas: The plan is to switch to an s-expression based syntax similar to the one used in the new KiCad layout in KiCad 5 though (but I think the plan is to change the file extension as well when doing this).

seppestas:
Oops, looks like I accidently pushed my commit to my fork's master branch -.- Forgot about Github automatically creating patch branches.

seppestas:
All issues should be fixed now. I think squashing the commits is probably a good idea :).

Alhadis:
> I think squashing the commits is probably a good idea :)

Don't worry, we always squash-merge pull-requests. ;) It makes for a much tidier revision history.

Thank you!

/cc @lildude

seppestas:
🎉

PR #3800, opened 2017-08-29
Update Go (golang) handling

### LINKED ISSUE #3719: New rule for generated Go files
See https://github.com/golang/go/issues/13560#issuecomment-288457920.

Related to #2724.

AlekSi:
* Use https://golang.org/s/generatedcode in addition to existing rule for generated files.
* Do not treat vendored Go code as generated.

Closes #3719.
Refs #2724.

/cc @lildude @cespare


AlekSi:
I'm conflicted about the second point. Vendored files are not generated, they are vendored, that is clear. But why all vendored files are not suppressed in diffs?

lib/linguist/generated.rb
```diff
@@ -325,29 +327,13 @@ def node_modules?
...
       name.match(/npm-shrinkwrap\.json/) || name.match(/package-lock\.json/)
     end
 
-    # Internal: Is the blob part of Godeps/,
-    # which are not meant for humans in pull requests.
-    #
-    # Returns true or false.
-    def godeps?
-      !!name.match(/Godeps\//)
-    end
```
lildude: 'scuse my ignorance, but why is this being removed? Is it because all `Godeps` will _always_ match `Code generated .* DO NOT EDIT` implemented above?

test/test_blob.rb
```diff
@@ -221,11 +221,6 @@ def test_generated
     # Generated by Zephir
     assert !sample_blob_memory("Zephir/Router.zep").generated?
 
-    # Go vendored dependencies
-    refute sample_blob("vendor/vendor.json").generated?
-    assert sample_blob("vendor/github.com/kr/s3/sign.go").generated?
-    refute fixture_blob("go/food_vendor/candy.go").generated?
```
lildude: Wouldn't these test still "pass" with the changes above?

test/test_generated.rb
```diff
@@ -42,11 +42,6 @@ def test_check_generated
     generated_sample_without_loading_data("Dummy/foo.xcworkspacedata")
     generated_sample_without_loading_data("Dummy/foo.xcuserstate")
 
-    # Go-specific vendored paths
-    generated_sample_without_loading_data("go/vendor/github.com/foo.go")
-    generated_sample_without_loading_data("go/vendor/golang.org/src/foo.c")
-    generated_sample_without_loading_data("go/vendor/gopkg.in/some/nested/path/foo.go")
-
```
lildude: Same here, wouldn't these test still "pass" with the changes above?

test/test_generated.rb
```diff
@@ -60,10 +55,6 @@ def test_check_generated
     generated_sample_without_loading_data("Dummy/npm-shrinkwrap.json")
     generated_sample_without_loading_data("Dummy/package-lock.json")
 
-    # Godep saved dependencies
-    generated_sample_without_loading_data("Godeps/Godeps.json")
-    generated_sample_without_loading_data("Godeps/_workspace/src/github.com/kr/s3/sign.go")
-
```
lildude: And here too?

lib/linguist/generated.rb
AlekSi: Because vendored files are not generated files. Vendoring is already covered by https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml#L51

lildude:
Nudge

lildude:
No response in three months. Marking as stale.

AlekSi:
> Vendored files are not generated, they are vendored, that is clear. But why all vendored files are not suppressed in diffs?

That is the important question, and answering may answer your questions. It looks like different parts of linguist treat "vendored" and "generated" words differently.

lildude:
> But why all vendored files are not suppressed in diffs?

Hmmm, vendored files generally aren't suppressed in diffs. Only generated should be.  That said, very large diffs are minimised in diffs, regardless of whether they're vendored or not. Do you have examples of where you're seeing discrepancies that I can see and try explain?

AlekSi:
> Hmmm, vendored files generally aren't suppressed in diffs. Only generated should be.

PTAL: https://github.com/percona/pmm-managed/commit/6e3e7d2ef368cc0c7a5f1d93bf1bef583b54bc1f Here vendored files are not rendered, but the message says "Some _generated_ files are not rendered by default." But those files are vendored, not generated.

pchaigno:
There's some background to this in #2843 and #2620.

As far as I understand, we mark these files as generated---even though they're not really---precisely because we want to hide them by default in diffs.

lildude:
Nudging this for one last time.

dmitshur:
@AlekSi Are you planning to continue to work on this PR?

If not, I can send a new PR.

AlekSi:
@dmitshur Feel free to work on it. And thank you.

AlekSi:
> As far as I understand, we mark these files as generated---even though they're not really---precisely because we want to hide them by default in diffs.

That feels right. Right now [README.md](https://github.com/github/linguist#overrides) provides a good description of what is considered [vendored](https://github.com/github/linguist#vendored-code) and what is [generated](https://github.com/github/linguist#generated-code). So what we call "vendored" in Go is "generated" in terms of Linguist.

@dmitshur While you work on that PR, may you please add a few comments in code to mention this vendored/generated terms clash? They could save time for the next contributor. 😄

PR #3806, opened 2017-08-31
Add some linguist-detectable attributes

seppestas:
This allows to test a use-case where markdown should be reported
in the language stats (e.g because the main purpose of the repo is to
hold prose in markdown format) while HTML should be excluded from the
language stats (e.g because it's example output).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,5 @@ Rakefile linguist-generated
 test/fixtures/* linguist-vendored=false
 README.md linguist-documentation=false
 samples/Arduino/* linguist-documentation
+samples/Markdown/*.md linguist-detectable=true
+samples/HTML/*.html linguist-detectable=false


seppestas:
This is required by the tests added in #3807.

lildude:
My comment at https://github.com/github/linguist/pull/3807#issuecomment-327733602 applies here too.

lildude:
As this PR makes no sense on its own and is only needed for the functionality you're proposing in https://github.com/github/linguist/pull/3807, can you please incorporate the changes into https://github.com/github/linguist/pull/3807 and close this PR.

pchaigno:
@lildude The 2 pull requests are based on different branches. We have a special `test/attributes` branch for some of the tests.

lildude:
Oh yes. I completely missed that. 😊 Ignore my comment @seppestas

seppestas:
Any chance this could get changed by the way?

It took me a bit to figure out how testing Linguist gitattributes work and it looks like even some regular contributors are not aware of how it works.

IMO it would make sense to have a seperate test repository on which to run tests instead of using Linguist's own repo. This allows to test gitattributes without the need for a test/gitattributes branch, and it would make a lot of other test relying on git a lot faster, since Linguist's history is pretty big.

The test repository could be added to the main repo as a submodule (possibly even multiple times), allowing to update the submodule in PR's to the main repo when e.g changes to the gitattributes are needed for new tests.

lildude:
> Any chance this could get changed by the way?

Not at the moment. I agree it's a bit _odd_ and I didn't know about it until now (I've not touched that side of the code) and it makes sense, however I don't think this and the repo size is enough substantiation to pull the test data away from the code actually being tested at this point primarily as it would add more complexity to contributing to Linguist: we have way more contributions that _don't_ affect git attributes than we do that do.  Maybe in future, until then I think improving `CONTRIBUTING.md` to highlight this is better in the interim.

Before approving this, why have you chosen to test with markdown and HTML rather than KiCad, the language motivating all these changes?

Just wondering if it makes more sense to test the override for KiCad to keep the change in context with the changes in https://github.com/github/linguist/pull/3807.

seppestas:
> Before approving this, why have you chosen to test with markdown and HTML rather than KiCad, the language motivating all these changes?

Because my implementation was built on top of the changes that make KiCad detectable by default. While I plan to remove that feature from #3807, I still think files like KiCad design files should be included in the language stats by default.

I mainly added the linguist-detectable field to address concerns like
> it could open up a common vector for disputes (i.e. some folks want XML files to be detectable by default and some don't)
-- @rafer in #3807 

Markdown and HTML are very commonly known, and I think the use-case I described makes sense for e.g repo's containing some sort of website like a blog. XML files in Android projects would be another great use-case, but I'm not that familiar with that.

As a sidenote: one of the reasons KiCAD motivated these changes is because it used to be included in language stats, but since #3743 this stopped working. For some reason Github still includes Eagle files in the language stats, even though Linguist stopped reporting them since #3751 (according to @Alhadis this is probably caused by some sort of caching). Eagle is still a lot more popular than KiCAD, so if it wasn't for this glitch I guess Eagle would have been the motivation 😅.

Alhadis:
Just out of modest curiosity, what exactly distinguishes Eagle from any other XML-derived language? Every example I've seen is essentially just XML data, and probably belongs in the ever-growing morass of [vendor-specific XML files](https://github.com/github/linguist/blob/4cd558c37482e8d2c535d8107f2d11b49afbc5b5/lib/linguist/languages.yml#L4879-L4979).

seppestas:
As for changing the way Linguist is tested: I think having separate test repositories would reduce complexity and increase reproducibility.

Also, AFAIK most contributions to Linguist don't add that much test code. Typically contributions just add a language and some test files for the language.

Because of the slow Ruby VM and the big git repo history, running all the test of Linguist takes a lot of time, making a task that should be done often pretty tedious.This is coming from someone used to building huge C++ projects.

seppestas:
> Just out of modest curiosity, what exactly distinguishes Eagle from any other XML-derived language? Every example I've seen is essentially just XML data, and probably belongs in the ever-growing morass of vendor-specific XML files.

I believe Eagle is indeed just vendor-specific XML (but I'm not that familiar with the Eagle format). That being said, I would very much like to keep having it included in the language stats. For the purpose of syntax highlighting I guess calling it XML would work fine.

Alhadis:
> I would very much like to keep having it included in the language stats.

Authors of [TextMate themes](https://github.com/github/linguist/blame/4cd558c37482e8d2c535d8107f2d11b49afbc5b5/lib/linguist/languages.yml#L4949), [WiX websites](https://github.com/github/linguist/blame/4cd558c37482e8d2c535d8107f2d11b49afbc5b5/lib/linguist/languages.yml#L4961), and [X3D models](https://github.com/github/linguist/blame/4cd558c37482e8d2c535d8107f2d11b49afbc5b5/lib/linguist/languages.yml#L4967) would probably prefer that degree of recognition too.

Not to worry, I don't recall a language ever being *removed* in Linguist's history... although I could be wrong.

Issue #3228, opened 2016-09-17
Language Stats skewed by not showing reStructuredText, SVG, etc. (but for instance TeX).


inoas:
Could it be that github detects sphinx docs (reStructuredText) as JavaScript?

Ref: https://github.com/isaacs/github/issues/768#issuecomment-247761390

If that's the case then it probably accounts all restructured text (and thus sphinx docs) wrongly towards JavaScript and - in case you gather stats of javascript by this code - impacts the global stats wrongly.

pchaigno:
Except for the `jquery.js` file, [these files](https://github.com/cakephp/docs/search?l=javascript) make up for the 44.6% of JavaScript (we count the number of lines, not the number of files). You can mark those files as vendored using [Linguist overrides](https://github.com/github/linguist#overrides).

inoas:
So re structured text is not accounted for. E.g. the whole source docs files are being ignored?

Alhadis:
Not ignored. reStructuredText is excluded from the statistics bar because it's a prose language. I'll copy you [an explanation](https://github.com/github/linguist/issues/3159) I wrote last month:

> There are [four distinct classifications](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3) of language on GitHub:
>
> * **Prose:** Material intended to be consumed by a human reader, with little to no processing
> * **Markup:** Material intended to be presented to a reader after being processed or parsed by a program
> * **Programming:** Material with instructions designed to be interpreted or compiled by software
> * **Data:** Virtually anything that doesn't fit into the above
> 
> Language statistics are only affected by markup and programming languages. Data and prose are ignored, and they'd unfairly bloat a repository's language statistics.
> 
> Markdown is classed as "prose", as it's designed to be read in its raw form as easily as its processed form. HTML is classed as "markup", because it's not designed to be read in its raw form.

inoas:
That is a really bad decision. 
Especially the difference between reStructured text and simple html documentation is marginal at best

Alhadis:
> It really de-evaluates community driven documentation and its used tools/languages.

That's a really silly thing to say, especially since GitHub has gone out of their way to support the displaying of rendered reStructuredText in repositories.

inoas:
There is nothing silly about it. Either the stats are relevant or they are not. Disregarding reStructuredText is a slap in the face towards the docs teams, stuff where open source can shine, but doesn't always (due to lack of documentation).

Alhadis:
[...]

Issue #3159, opened 2016-08-21
There should be a way to configure linguist from github repo using config file


jcubic:
SVG is not marked as language (#1020) same happen to markdown (#3158) it would be nice if you could add language in some config file that will be read by linguist on init.

Alhadis:
First, please read the [troubleshooting notes](https://github.com/github/linguist#troubleshooting) on adjusting your project's language classification.

Secondly, Markdown isn't going to be listed in language statistics. This is why:

There are [four distinct classifications](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3) of language on GitHub:

* **Prose:** Material intended to be consumed by a human reader, with little to no processing
* **Markup:** Material intended to be presented to a reader after being processed or parsed by a program
* **Programming:** Material with instructions designed to be interpreted or compiled by software
* **Data:** Virtually anything that doesn't fit into the above

Language statistics are only affected by markup and programming languages. Data and prose are ignored, and they'd unfairly bloat a repository's language statistics.

Markdown is classed as "prose", as it's designed to be read in its raw form as easily as its processed form. HTML is classed as "markup", because it's not designed to be read in its raw form.

sunjay:
What about repositories where "prose" is most of the content? For example blogs hosted on GitHub pages are mostly markdown but this isn't shown in the language statistics.

arfon:
@sunjay - please see this related discussion for some additional background: https://github.com/github/linguist/issues/3228

PR #3807, opened 2017-08-31
Override languages being included by language statistics

### LINKED ISSUE #3805: Improve support for non-code languages
Currently there are [4 different language types: data, programming, markup, prose or nill][language types]. When a language is not considered a programming, markup or prose language it is typically classified as data. Recently there have been a lot of changes to languages that have been incorrectly classified as programming or markup languages. E.g KiCad files used to be classified as programming language (PR #3743), Eagle files used to be classified as markup (PR #3751), ...

However, when classifying a language as something else than a programming or markup language, [it is not included in the repository's language statistics][detectable types].

While the new classification is probably correct (depending on your interpretation of what a programming and markup language is), these languages are considered to be "primary" languages used in a lot of repositories that contain them. This causes a lot of confusion and criticism on these files not being recognized, especially for languages that used to be reported because they were misclassified before. See e.g #3784, #3795, #3484.

It would be nice if Linguist and Github could improve support for languages that can be considered primary languages that are not programming or markup languages. E.g PCB design / EDA files, files for specialised image manipulation software, files used to encode sheet music, ... The open source world extends beyond just code!

This could be done by e.g:
- Adding new types for languages like EDA tools and other languages that can be considered "primary"
- Making Linguists' definition of programming languages and/or markup languages very loose so these languages can be classified as one of these types
- Adding a more explicit "include in language stats"/`detectable` field to all languages and use it instead of/together with the type to decide whether or not to include the language in the language stats. This would have the added bonus of making it easily configurable on a per repo basis (e.g to allow repos with primarily XML files).

[language types]: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3
[detectable types]: https://github.com/github/linguist/blob/8dc4a1308f7d309ebdf5566251eca3f5562d8901/lib/linguist/blob_helper.rb#L336

seppestas:
Adds a way to override ~~languages and~~ blobs being included in the language statistics.

This improves support for non-code languages (#3805) by making it possible ~~to flag languages as being `detectable` in the languages.yml file. It is also possible~~ to override the detectable flag for specific blobs in the repo using the .gitattributes file.

Note: the tests require #3806 to be merged in order to work properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Linguist supports a number of different custom override strategies for language 
 
 ### Using gitattributes
 
-Add a `.gitattributes` file to your project and use standard git-style path matchers for the files you want to override using the `linguist-documentation`, `linguist-language`, `linguist-vendored`, and `linguist-generated` attributes. `.gitattributes` will be used to determine language statistics and will be used to syntax highlight files. You can also manually set syntax highlighting using [Vim or Emacs modelines](#using-emacs-or-vim-modelines).
+Add a `.gitattributes` file to your project and use standard git-style path matchers for the files you want to override using the `linguist-documentation`, `linguist-language`, `linguist-vendored`, `linguist-generated`  and `linguist-detectable` attributes. `.gitattributes` will be used to determine language statistics and will be used to syntax highlight files. You can also manually set syntax highlighting using [Vim or Emacs modelines](#using-emacs-or-vim-modelines).
 
 ```
 $ cat .gitattributes
--- a/README.md
+++ b/README.md
@@ -164,6 +164,21 @@ $ cat .gitattributes
 Api.elm linguist-generated=true
 ```
 
+#### Detectable
+
+Currently there are 4 different language types: data, programming, markup, prose or nill. When a language is not considered a programming, markup or prose language it is typically classified as data. Recently there have been a lot of changes to languages that have been incorrectly classified as programming or markup languages. E.g KiCad files used to be classified as programming language (PR #3743), Eagle files used to be classified as markup (PR #3751), ...
+
+By default only programming languages are "detected", i.e counted towards a repository's language statistics. In some repositories it might be preferable to include some other files in the language statistics. In other cases source files that are not vendored, documentation or generated code might want to be be excluded from the language statistics.
+
+Use the `linguist-detectable` attribute to mark or unmark paths as detectable.
+
+```
+$ cat .gitattributes
+*.kicad_pcb linguist-detectable=true
+*.sch linguist-detectable=true
+tools/export_bom.py linguist-detectable=false
+```
+
 ### Using Emacs or Vim modelines
 
 If you do not want to use `.gitattributes` to override the syntax highlighting used on GitHub.com, you can use Vim or Emacs style modelines to set the language for a single file. Modelines can be placed anywhere within a file and are respected when determining how to syntax-highlight a file on GitHub.com
--- a/README.md
+++ b/README.md
@@ -166,9 +166,7 @@ Api.elm linguist-generated=true
 
 #### Detectable
 
-Currently there are 4 different language types: data, programming, markup, prose or nill. When a language is not considered a programming, markup or prose language it is typically classified as data. Recently there have been a lot of changes to languages that have been incorrectly classified as programming or markup languages. E.g KiCad files used to be classified as programming language (PR #3743), Eagle files used to be classified as markup (PR #3751), ...
-
-By default only programming languages are "detected", i.e counted towards a repository's language statistics. In some repositories it might be preferable to include some other files in the language statistics. In other cases source files that are not vendored, documentation or generated code might want to be be excluded from the language statistics.
+Only programming languages are included in the language statistics. Languages of a different type (as defined in [`languages.yml`](/lib/linguist/languages.yml)) are not "detectable" causing them not to be included in the language statistics.
 
 Use the `linguist-detectable` attribute to mark or unmark paths as detectable.
 
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -380,7 +380,7 @@ module Linguist
 
     # Internal: Should this blob be included in repository language statistics?
     def include_in_language_stats?
-      defined?(detectable) && !detectable?.nil? ? language && detectable? :
+      defined?(detectable?) && !detectable?.nil? ? language && detectable? :
         !vendored? &&
         !documentation? &&
         !generated? &&
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -380,11 +380,13 @@ module Linguist
 
     # Internal: Should this blob be included in repository language statistics?
     def include_in_language_stats?
-      defined?(detectable?) && !detectable?.nil? ? language && detectable? :
-        !vendored? &&
-        !documentation? &&
-        !generated? &&
-        language && DETECTABLE_TYPES.include?(language.type)
+      !vendored? &&
+      !documentation? &&
+      !generated? &&
+      language && ( defined?(detectable?) && !detectable?.nil? ?
+        detectable? :
+        DETECTABLE_TYPES.include?(language.type)
+      )
     end
   end
 end
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -376,17 +376,15 @@ module Linguist
       language && language.tm_scope
     end
 
+    DETECTABLE_TYPES = [:programming, :markup].freeze
+
     # Internal: Should this blob be included in repository language statistics?
     def include_in_language_stats?
-      !detectable?.nil? ?
-        language && detectable? :
-        language && (
-          detectable?.nil? &&
-          !vendored? &&
-          !documentation? &&
-          !generated? &&
-          language.detectable?
-        )
+      defined?(detectable) && !detectable?.nil? ? language && detectable? :
+        !vendored? &&
+        !documentation? &&
+        !generated? &&
+        language && DETECTABLE_TYPES.include?(language.type)
     end
   end
 end
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -376,14 +376,12 @@ module Linguist
       language && language.tm_scope
     end
 
-    DETECTABLE_TYPES = [:programming, :markup].freeze
-
     # Internal: Should this blob be included in repository language statistics?
     def include_in_language_stats?
       !vendored? &&
       !documentation? &&
       !generated? &&
-      language && DETECTABLE_TYPES.include?(language.type)
+      language && language.detectable?
     end
   end
 end
--- a/lib/linguist/blob_helper.rb
+++ b/lib/linguist/blob_helper.rb
@@ -378,10 +378,15 @@ module Linguist
 
     # Internal: Should this blob be included in repository language statistics?
     def include_in_language_stats?
-      !vendored? &&
-      !documentation? &&
-      !generated? &&
-      language && language.detectable?
+      !detectable?.nil? ?
+        language && detectable? :
+        language && (
+          detectable?.nil? &&
+          !vendored? &&
+          !documentation? &&
+          !generated? &&
+          language.detectable?
+        )
     end
   end
 end
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -33,6 +33,7 @@ module Linguist
 
     # Valid Languages types
     TYPES = [:data, :markup, :programming, :prose]
+    DETECTABLE_TYPES = [:programming, :markup].freeze
 
     # Detect languages by a specific type
     #
@@ -295,9 +296,10 @@ module Linguist
       @interpreters = attributes[:interpreters]   || []
       @filenames  = attributes[:filenames]  || []
 
-      # Set popular, and searchable flags
+      # Set popular, searchable and detectable flags
       @popular    = attributes.key?(:popular)    ? attributes[:popular]    : false
       @searchable = attributes.key?(:searchable) ? attributes[:searchable] : true
+      @detectable = attributes.key?(:detectable) ? attributes[:detectable] : false
 
       # If group name is set, save the name so we can lazy load it later
       if attributes[:group_name]
@@ -477,6 +479,15 @@ module Linguist
       @searchable
     end
 
+    # Public: Is it detectable?
+    #
+    # Detectable languages will by include in the language stats.
+    #
+    # Returns true or false
+    def detectable?
+      @detectable || DETECTABLE_TYPES.include?(self.type)
+    end
+
     # Public: Return name as String representation
     def to_s
       name
@@ -559,6 +570,7 @@ module Linguist
       :wrap              => options['wrap'],
       :group_name        => options['group'],
       :searchable        => options.fetch('searchable', true),
+      :detectable        => options.fetch('detectable', false),
       :language_id       => options['language_id'],
       :extensions        => Array(options['extensions']),
       :interpreters      => options['interpreters'].sort,
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -33,7 +33,6 @@ module Linguist
 
     # Valid Languages types
     TYPES = [:data, :markup, :programming, :prose]
-    DETECTABLE_TYPES = [:programming, :markup].freeze
 
     # Detect languages by a specific type
     #
@@ -296,10 +295,9 @@ module Linguist
       @interpreters = attributes[:interpreters]   || []
       @filenames  = attributes[:filenames]  || []
 
-      # Set popular, searchable and detectable flags
+      # Set popular, and searchable flags
       @popular    = attributes.key?(:popular)    ? attributes[:popular]    : false
       @searchable = attributes.key?(:searchable) ? attributes[:searchable] : true
-      @detectable = attributes.key?(:detectable) ? attributes[:detectable] : false
 
       # If group name is set, save the name so we can lazy load it later
       if attributes[:group_name]
@@ -479,15 +477,6 @@ module Linguist
       @searchable
     end
 
-    # Public: Is it detectable?
-    #
-    # Detectable languages will by include in the language stats.
-    #
-    # Returns true or false
-    def detectable?
-      @detectable || DETECTABLE_TYPES.include?(self.type)
-    end
-
     # Public: Return name as String representation
     def to_s
       name
@@ -570,7 +559,6 @@ module Linguist
       :wrap              => options['wrap'],
       :group_name        => options['group'],
       :searchable        => options.fetch('searchable', true),
-      :detectable        => options.fetch('detectable', false),
       :language_id       => options['language_id'],
       :extensions        => Array(options['extensions']),
       :interpreters      => options['interpreters'].sort,
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -14,8 +14,6 @@
 #                     listed alphabetically)
 # interpreters      - An Array of associated interpreters
 # searchable        - Boolean flag to enable searching (defaults to true)
-# detectable        - Boolean flag to include language in language stats (defaults to true for programming
-#                     and markup types, false for other types)
 # language_id       - Integer used as a language-name-independent indexed field so that we can rename
 #                     languages in Linguist without reindexing all the code on GitHub. Must not be
 #                     changed for existing languages without the explicit permission of GitHub staff.
@@ -1122,7 +1120,6 @@ EQ:
   language_id: 96
 Eagle:
   type: data
-  detectable: true
   extensions:
   - ".sch"
   - ".brd"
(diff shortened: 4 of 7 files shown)


seppestas:
Note: Travis Build 8308 will probably fail because of the weird way this library tests .gitattributes.

5N44P:
This would also fix #3228 and probably many more past issues. For what i could read, this feature has been asked for a long time. 
Let's hope that Github's staff and the contributors will agree with these changes!

seppestas:
So... any thoughts?

While I think this is the most flexible and forward looking approach, I understand it's also has the most impact on the rest on the Linguist system compared to the other approaches I described in #3805.

As stated by @5N44P this approach allows users that want XML, SVG, doc, ... files to represented in the language statistics of a project, like the cases stated in #3159 and #3228, to do so.

Note that the override in the .gitattributes works on a blob level. While flexible and simple, this might not be the most user-friendly approach. Having the option to override what languages are "detectable" in the .gitattributes file might be a more consistent solution, however this is harder to combine with the "sensible defaults" in the languages.yml file.

lildude:
I'm not 100% sure this is the right approach to solve this problem as this paves the way for a million more PRs asking for similar functionality from other non-programming languages which don't show up in the colourful language bar. I'm also not sure of the impact allowing such an override will have on the performance of Linguist on GitHub.com - it's already a pretty hungry beast and I imagine opening the door to overriding things like XML & SQL will lead to heeeeawg problems.

Accordingly, I've deferred the judgment on how best to approach this to the team within GitHub that "owns" the Linguist product (I'm not part of that team) as they have a much clearer idea of the direction they'd like to take such things, in conjunction with future GitHub.com development plans, in the long run.

seppestas:
> I'm not 100% sure this is the right approach to solve this problem as this paves the way for a million more PRs asking for similar functionality from other non-programming languages which don't show up in the colourful language bar.

Why is this not a good thing? Having people indicate what they expect their projects to be recognized as by Github is an awesome way to handle this IMO. Anyhow, I tried to find some other languages where it would make sense to have them "detectable" by default, but it wasn't so easy to find those since most languages in GH are either already programming languages or just plain data. The only languages I could think of (like music sheet files and music mastering settings) are not currently not part of Github. I think the amount of requests for other languages to be detectable by default will be not that high. The only problem might be people wanting to recognise generic files (like XML or SVG), for this reason I think having the option to override the detectable flag makes sense.

>  I'm also not sure of the impact allowing such an override will have on the performance of Linguist on GitHub.com

I don't think the detectable field for languages I added will have a big impact on performance, but the way the override works might since it uses file globbing. Having an override on a language level (or no override at all) would probably make it more performant. However most overrides in the .gitattributes file currently work on a file / blob level, so I felt the current approach was the most consistent.

Should I separate the addition of the "detectable" language flag (0f7c0df5c4683a134c8e0794d672f1aef4479aee) and the override feature into their own PR?

lildude:
> Should I separate the addition of the "detectable" language flag (0f7c0df) and the override feature into their own PR?

Nah, you can leave things as-is and we'll let my colleagues decide how best to approach this and advise further.

Alhadis:
Personally, I believe giving users the option of setting their project's language classification a.l.a BitBucket is the ideal solution. It needn't be mandatory; by default, the classification would be based on the codebase's language usage (as it is now), until manually set by the owner. This needn't affect the stats in the language breakdown, of course... but it would offer an intuitive way to affect how their projects are being advertised to other developers.

seppestas:
I would really like to have languages like KiCad and Eagle design files and other languages that can be considered primary to a repository detectable by default.

Most users of these systems are not that intimately familiar with the workings of git and Github, let alone the role of Linguist in language detection. These users expect their repositories to be recognised correctly, and when this is not the case this causes confusion, and in a lot of cases frustration.

Best case this will result in a bunch of people creating issues on Linguist asking why their repo is being mis-identified, with a whole lot of "you should add a `linguist-detectable` line in your .gitattributes file" answers. Worst case people never find out about Linguist and/or leave things as-is, negatively impacting discoverability of their projects on Github.

I'll update this PR to only add the .gitattributes feature, but I would like to keep #3805 open until a solution is added that allows non-code languages to be included in the language statistics by default. Could #3806 already be merged so the tests pass when I update this PR?

seppestas:
> Personally, I believe giving users the option of setting their project's language classification a.l.a BitBucket is the ideal solution.

While I agree this would be more intuitive for people that are not familiar with the .gitattributes file, I'm not a big fan of this solution because:

- In the case of hardware projects repositories often contain a combination of hardware design files and firmware source files.
- It's not automatic, meaning people can overlook / ignore this.
- The language stats bar just looks cool and adds an easy way to navigate to different types of source files in the repo.

Alhadis:
> It's not automatic, meaning people can overlook / ignore this.

... erh, it literally says in the second sentence that the proposed default *would* be automatic:

> It needn't be mandatory; by default, **the classification would be based on the codebase's language usage (as it is now)**, until manually set by the owner.

As for this:

> negatively impacting discoverability of their projects on Github.

There *is* a topics feature, as I said. Furthermore, I've found that to be much more reliable than searching repositories by language. Searching for repositories that've been identified as PostScript has, more often than not, given me a bunch of misclassified projects. Same for Turing and various lesser-known assembler languages. Searching for something that's been *explicitly tagged* by humans has proven more useful than the results of a Bayesian classifier.

Alhadis:
> The language stats bar just looks cool

I agree, but don't get too attached to what you see in GitHub's UI. Design elements are subject to change without notice, and believe me, it often happens when you least expect it. 😜 If GitHub decide to do away with the language bar someday, for reasons only known to internally, well...

seppestas:
> by default, the classification would be based on the codebase's language usage (as it is now), until manually set by the owner

This would still require proper support for non-code languages in order to be recognised by Linguist though.

Having the option to manually change the language a project is classified as in a way that is more user-friendly than .gitattribute files would already be a big improvement. It might be interesting to tie this in to Github's topic feature. Anyhow, these are things that GH themselves need to implement, since it can't be done by Linguist...

lildude:
> Having the option to manually change the language a project is classified as in a way that is more user-friendly than .gitattribute files would already be a big improvement. 

Please don't confuse the language bar showing the breakdown of languages found within a repository with a specific repository language. GitHub has no concept of a repository language and never has as I pointed out in https://github.com/github/linguist/issues/3805#issuecomment-328325527. I'm repeating this as I keep getting the feeling that you're thinking the language stats bar is an indicator of a repository's language rather than a breakdown of the languages within it.

The `.gitattributes` file is _not_ for setting a repository language and nor should it considered a method of doing so either. The Linguist-specific options are meant as a tool to help Linguist identify the languages within a repo when it doesn't quite get it right; it's not perfect after all. If adjusting the content just so happens to result in a shift in the proportion of languages such that your favourite language is now the predominant language, then great, but there is still no concept of a repository language and using the `.gitattributes` file to try and set one is really missing the point.

If you really want the ability to set a repository language and you feel topics are't the right thing, you'll need to contact GitHub support and log a feature request.

seppestas:
> I'm repeating this as I keep getting the feeling that you're thinking the language stats bar is an indicator of a repository's language rather than a breakdown of the languages within it.

I'm aware of this. The problem I'm trying to address is it not showing the languages I'm / (Ki)CAD users / maintainers of none code repos in general are interested in, and only the remaining, typically miss-identified files being exposed in the language stats causing confusion.

Anyhow, this PR is an implementation of one of my proposals on how this can be fixed. I think we should keep the higher level discussion on how Linguist could / should handle non-code languages in #5805.

rafer:
Hi All, just wanted to provide some further context for the decisions that @lildude outlined. Basically, we're trying to get some of the functionality you want (the ability for your projects to have KiCad files show up in the language bar), without making significant changes to linguist (adding detectable to the `languages.yml`), because we'd need to further investigate the impact of this on internal github systems, and also it could open up a common vector for disputes (i.e. some folks want XML files to be detectable by default and some don't). Long story short we're trying to be cautious and make sure that we've fully thought out the repercussions before acting. That's why I think the project-level overrides solution is a good interim step.

* It alleviates at least some of the pain y'all are currently experiencing.
* We can :thumbsup: this change without a lengthy investigation on the GitHub side.
* We can continue to collect data. If we do indeed get a ton of "you should add a linguist-detectable line in your .gitattributes file" type issues as @seppestas suggests, that will be good to know and can help inform future decisions.

Anyway, looking forward to merging this in once @lildude's [review comments](https://github.com/github/linguist/pull/3807#pullrequestreview-61652849) are addressed.

seppestas:
Ok, I have a commit ready where I reverted the changes of 0f7c0df5c4683a134c8e0794d672f1aef4479aee and made some required adjustments to make this work with the .gitattributes related code, as well as fixing a problem with other blob types.

Having #3806 merged before pushing this commit would make the tests pass (if everything goes well). It's a bit of a chicken and egg story though. Should I go ahead and push my commit to update this PR and have you rerun the test once #3806 is merged, or do you want to merge #3806 first so the tests pass as soon as I update this PR?

seppestas:
Huh, looks like the required commit on the test branch (7af53039a2d18d90d8d49e14b2091d4e437166bb) is still not found... Am I missing something?

seppestas:
Ah, the commit hash got changed when the PR was changed -.-

seppestas:
Should be fixed now. Maybe it's a good idea to tag the commits in test/attributes and using the tag in the tests. This way the person merging PRs could tag the correct commit when a PR is merged. IMO this would also make the test appear a bit clearer, especially if the tags get namespaced to the branch, e.g `test/attributes/override_detectable`.

Alhadis:
I believe that's what "lightweight" Git tags are for (as opposed to the annotated ones we use daily). ;-)

See [**git-tag**(1)](https://git-scm.com/docs/git-tag).

seppestas:
Yeey, all green now ^^

I don't think the type of tag matters (though using a lightweight tag does make sense for this). It might even make more sense to use branches, this would make it possible to update the commit required for the tests without having to update the tests.

Either way, someone with merge access would have to create a tag or (update a) branch after a PR is merged into test/attributes for this to work. Also, I'm not sure if it's possible to do a git fetch using a wildcard in the branch name. Apparently [this requires configuration of the refspec](https://stackoverflow.com/questions/28115668/how-do-i-fetch-pull-multiple-branches-matching-a-wildcard-string).

seppestas:
Hmm, I should probably add a test to see if `detectable` actually causes `include_in_language_stats` to be changed...

Alhadis:
Erh no, that's precisely what lightweight tags are for... keeping branches in-sync isn't just a pain, it also takes an ungodly amount of time to switch between for a project of Linguist's size.

Then, of course, there's the racket caused by submodules:

~~~console
λ GitHub-Linguist (master): time git checkout origin/puppet-fix
warning: unable to rmdir vendor/grammars/Sublime-Pep8: Directory not empty
warning: unable to rmdir vendor/grammars/dartlang: Directory not empty
warning: unable to rmdir vendor/grammars/data-weave-tmLanguage: Directory not empty
warning: unable to rmdir vendor/grammars/language-closure-templates: Directory not empty
warning: unable to rmdir vendor/grammars/language-pan: Directory not empty
warning: unable to rmdir vendor/grammars/language-pcb: Directory not empty
warning: unable to rmdir vendor/grammars/language-webassembly: Directory not empty
…
M	vendor/CodeMirror
M	vendor/grammars/Elm
M	vendor/grammars/Handlebars
M	vendor/grammars/LiveScript.tmbundle
M	vendor/grammars/MagicPython
M	vendor/grammars/Modelica
M	vendor/grammars/NimLime
M	vendor/grammars/Sublime-SQF-Language
M	vendor/grammars/SublimeGDB
M	vendor/grammars/SublimePuppet
M	vendor/grammars/TLA
M	vendor/grammars/Terraform.tmLanguage
M	vendor/grammars/antlr.tmbundle
M	vendor/grammars/api-blueprint-sublime-plugin
M	vendor/grammars/atom-language-clean
M	vendor/grammars/atom-language-perl6
M	vendor/grammars/atom-language-rust
M	vendor/grammars/c.tmbundle
M	vendor/grammars/capnproto.tmbundle
M	vendor/grammars/chapel-tmbundle
M	vendor/grammars/cucumber-tmbundle
M	vendor/grammars/cython
M	vendor/grammars/d.tmbundle
Note: checking out 'origin/puppet-fix'.

real	0m2.866s
user	0m0.451s
sys	0m1.758s
~~~

Yeah, no thanks. A lightweight tag is designed for exactly this scenario.

seppestas:
Hmm, I agree keeping all those branches in sync is a pain, using tags is probably easier. Why specifically lightweight tags though? Just because there's not much point in using annotated tags or is there a reason lightweight tags must be used?

Alhadis:
Lightweight tags are essentially a means of assigning a human-friendly name to an object (usually a commit). There's nothing stopping you from using an annotated tag, but it's complete overkill for something as mundane as a label - it includes date, author name and e-mail, and a tagging message. All of that data is useless and just pollutes the repository's object store over time.

Using annotated tags will also make it harder to distinguish release tags from temporary/internal ones.

seppestas:
Is it me or is it impossible to detect the language of a `LazyBlob` using `Linguist.detect`?

I'm trying the following:

```ruby
def test_linguist_override_detectable?
  attr_commit = "8f86998866f6f2c8aa14e0dd430e61fd25cff720"
  linguist_repo(attr_commit).read_index

  # markdown is overridden by .gitattributes to be detectable, html to not be detectable
  markdown = Linguist::LazyBlob.new(rugged_repository, attr_commit, "samples/Markdown/tender.md")
  html = Linguist::LazyBlob.new(rugged_repository, attr_commit, "samples/HTML/pages.html")

  assert_predicate markdown, :detectable?
  refute_predicate html, :detectable?

  assert_predicate markdown, :include_in_language_stats?
  refute_predicate html, :include_in_language_stats?

  # detectable flag for text and c is not set in .gitattributes
  text = Linguist::LazyBlob.new(rugged_repository, attr_commit, "samples/Text/foo.txt")
  c = Linguist::LazyBlob.new(rugged_repository, attr_commit, "samples/C/main.c")

  refute_predicate text, :detectable?
  assert_predicate text, :documentation?
  refute_predicate c, :detectable?
  assert c.language.type == :programming

  refute_predicate text, :include_in_language_stats?
  assert_predicate c, :include_in_language_stats?
end
```

But it seems like `language` isn't set for `text` and `c`...

seppestas:
Nevermind, figured out mocking `detectable` in `test_include_in_language_stats` is probably a better idea (I'm not used to mocking being so simple :o).

I'm pretty sure this feature works as intended now.

seppestas:
Hey @lildude, could you update your review? I would love to start using this feature ^^.

lildude:
I don't have the bandwidth right now but plan to catch up with linguist PRs and make a new release early next month.

lib/linguist/blob_helper.rb
```diff
@@ -337,10 +337,11 @@ def tm_scope
 
     # Internal: Should this blob be included in repository language statistics?
     def include_in_language_stats?
-      !vendored? &&
-      !documentation? &&
-      !generated? &&
-      language && DETECTABLE_TYPES.include?(language.type)
+      defined?(detectable?) && !detectable?.nil? ? language && detectable? :
```
lildude: This is going to need refining as it's too greedy and overrides any of the subsequent attributes in this method.

For example, creating a repo containing:

```
index.html
docs/index.html
vendor/index.html
vendor/docs/index.html
```

... and analysing using the default attributes results in:

```
$ ../seppestas-linguist/bin/linguist --breakdown .
100.00% HTML

HTML:
index.html

$
```

If we then implement a `.gitattributes` like this:

```
vendor/* linguist-vendored=false
vendor/docs/* linguist-documentation
```

... and run again, we get:

```
$ ../seppestas-linguist/bin/linguist --breakdown .
100.00% HTML

HTML:
index.html
vendor/index.html

$
```

Now change the first line to `vendor/* linguist-detectable` and test again we get:

```
$ ../seppestas-linguist/bin/linguist --breakdown .
100.00% HTML

HTML:
index.html
vendor/docs/index.html
vendor/index.html

$
```

As you can see, the `linguist-documentation` attribute is being completely ignored.

seppestas: Mmh, true, though it kind of does that by design...

Would it be a good idea to fix this in `LazyBlob .detectable?`? I.e do something like
```
def detectable?
  if documentation? || generated? || vendored?
    return false
  elseif attr = git_attributes['linguist-detectable']
    return boolean_attribute(attr)
  else
    nil
  end
end
```

lildude: Possibly, though I feel we're repeating ourselves unnecessarily.

5N44P:
Hey guys, how are you doing? 
Is there any news with this? 
@seppestas @Alhadis @lildude

seppestas:
I was still looking into the change request by @lildude, but I'm not 100% sure what the expected result should be.

I got distracted by more urgent projects, but I'll have another look at this.

seppestas:
Hey @lildude, I addressed your remark by making `include_in_language_stats` ignore the detectable override (the same way it ignores the language being of a detectable type) if the blob is vendored, documentation or generated.

I added unit tests for this in test_blob.rb. Do you think it's usefull/necessary to also add some integration tests?

lildude:
Thanks @seppestas. This is looking great now.

> I added unit tests for this in test_blob.rb. Do you think it's usefull/necessary to also add some integration tests?

Yes, I think this is probably a good idea as it was the way I confirmed the change in behaviour I noted before.

lildude:
@seppestas any news on those tests? I'd like to get this PR in the next Linguist release.

seppestas:
Sadly adding these tests is not as easy as I first expected. It will again require some changes on the test/attributes branch.

I'l have another look this weekend.

lildude:
>>  I added unit tests for this in test_blob.rb. Do you think it's usefull/necessary to also add some integration tests?
> 
> Yes, I think this is probably a good idea as it was the way I confirmed the change in behaviour I noted before.

Thinking about this again, I think we can do without them for the moment. We've not had integration tests for the current functionality to-date and I'd rather not hold this PR back much longer. We can always come back and add them later.

Please update your fork/branch to pull in the recent changes to the README and update the README to include a section on this new option.

seppestas:
> Please update your fork/branch to pull in the recent changes to the README and update the README to include a section on this new option.

Ok, will do.

I looked into the integration test, but I kind of got stuck figuring out whether I wanted to create 2 separate commits on the test branch (considering your "step-by-step" example) or coming up with an example that works with a single commit. I also tried referencing a light tag so I don't have to wait with updating this PR until the changes on the test branch are changed.

I might have also gotten distracted by the holidays and my GF at some point...

seppestas:
Finally gotten around to rebasing my branch onto master. Took me a longer than expected since I got confused because the tests were failing. Seems like running `bundle exec rake samples` fixed everything.

lildude:
:+1: Can you please add a little section for the new attribute like we've got for all the others, for example like the [documentation attribute](https://github.com/seppestas/linguist/blob/794af605627be3c3d23eee66e0abe09ac65e52db/README.md#documentation), to explain the usage and purpose.

README.md
```diff
@@ -164,6 +164,21 @@ $ cat .gitattributes
 Api.elm linguist-generated=true
 ```
 
+#### Detectable
+
+Currently there are 4 different language types: data, programming, markup, prose or nill. When a language is not considered a programming, markup or prose language it is typically classified as data. Recently there have been a lot of changes to languages that have been incorrectly classified as programming or markup languages. E.g KiCad files used to be classified as programming language (PR #3743), Eagle files used to be classified as markup (PR #3751), ...
```
lildude: I don't think we need this paragraph; it adds too much context for the introduction of the change rather than concentrating on the purpose of the attribute. The next paragraph, with a few tweaks, is probably sufficient.

README.md
```diff
@@ -164,6 +164,21 @@ $ cat .gitattributes
...
+
+By default only programming languages are "detected", i.e counted towards a repository's language statistics. In some repositories it might be preferable to include some other files in the language statistics. In other cases source files that are not vendored, documentation or generated code might want to be be excluded from the language statistics.
```
lildude: The content of this paragraph is good though it feels a little "off" to me, though I can't put my finger on why just yet. I'll need to mull it a little bit.

@Alhadis you're the better wordsmith, you may have some ideas on this too.

README.md
seppestas: Oops, I added this paragraph by accident... I just copy-pasted it to see what I wrote in my original issue, forgot to remove it and committed it (this is what you get when you start working on something one evening and only finish up quickly the next day) 😳

seppestas:
Great, thanks! It's been quite a journey, but it was also educative ^^.

seppestas:
🎉  Thanks!

lildude:
Linguist v6.0.0 is now live on GitHub.com. 🎉

5N44P:
Thank you guys, you did an amazing job ❤

jcubic:
Is there a way to set SVG as language? If not, then you just exclude lot of cases. What about If user have a book written in TeX and it's the source of the book. There are lot of cases when user wants something to be the source of the repo. It don't need to be code but written in Language.

PR #3809, opened 2017-09-03
Added ".monkey2" extension to Monkey Programming Language

DoctorWhoof:
The latest Monkey Programming Language extension is ".monkey2", so I added it to this list. The language description is available at "http://monkeycoder.co.nz". Thanks!

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2740,6 +2740,7 @@ Monkey:
   type: programming
   extensions:
   - ".monkey"
+  - ".monkey2"
   ace_mode: text
   tm_scope: source.monkey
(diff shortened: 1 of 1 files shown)


DoctorWhoof:
I have added an example to the samples folder. It is similar to the Monkey1 example already there in order to showcase some differences between the two versions.

DoctorWhoof:
Added a few more examples. Closed the previous pull request, opened a new one with all the added examples. How do I set those files as Public Domain?

Alhadis:
You declaring them as public domain *should* be enough, I believe. But if you're uncertain, you can include a comment in the header which states the author and license type.

PR #3728, opened 2017-07-22
Change Haskell's colour to use official branding

Alhadis:
Resolves github/linguist#3727.

Quoting @adius:
> The Haskell community has obviously settled with purple (check out https://haskell.org, https://haskell-lang.org, https://reddit.com/r/haskell, ...). `rgb(94,80,134)` to be precise.

He's right, and now that I think about it, I'm not entirely sure where the colour green came from. Everything in the language's branding uses either purple or dark blue, so that's obviously much more representative of the language than, well, `rgb(41,181,68)`.

/cc @lildude Reckon we could squeeze this under the door for Linguist [`v5.1.0`](https://github.com/github/linguist/pull/3725)?

**EDIT:** Additional rationale to justify this particular shade of purple: it's the same colour used by the [central lambda in Haskell's logo](https://www.haskell.org/static/img/haskell-logo.svg), arguably the image's most prominent element.

PR #3815, opened 2017-09-08
Update Racket color based on new logo

pmaddams:
Some time ago, the Racket folks updated their logo as part of the rebranding from PLT Scheme. You can see the new logo here: https://github.com/racket/icons/blob/master/racket-logo.svg.

Racket is a great language and deserves a cool GitHub color to match their logo. Also, Haskell's color was recently changed from green to purple: https://github.com/github/linguist/pull/3728. We don't want people to think Haskell is the only functional programming game in town.

### DIFF
--- a/test/test_color_proximity.rb
+++ b/test/test_color_proximity.rb
@@ -9,7 +9,7 @@ class TestColorProximity < Minitest::Test
 
   def test_color_proximity
     langs_with_colors = Language.all.reject { |language| language.color.nil? }
-    cp = ColorProximity.new(0.02, langs_with_colors.map { |lang| cut_hash(lang.color) })
+    cp = ColorProximity.new(0.01, langs_with_colors.map { |lang| cut_hash(lang.color) })
     failing_threshold = langs_with_colors.map do |lang|
       state = cp.past_threshold?(cut_hash(lang.color))
       if !state.first && lang.color != "##{state.last.first}"


lildude:
Unfortunately this PR can't be merged yet as it's not passing CI because the colour you've chosen is too similar to that used by another language. You're going to need to experiment with the colour until you find one that didn't conflict. I understand this may be quite a challenge.

pmaddams:
Apologies for the failed commits. Because I don't have a development machine right now, I am making changes on my phone. Several points:

1. The chosen color is the same one used in Racket's logo. If you look at the raw SVG file you will see:
`<path id="blue-piece" fill="#3E5BA9"` ...
2. I understand the color is somewhat similar to Go and PHP. However, they are clearly not identical, and blue is a very popular color. There is little danger that someone will confuse Racket with Go or Racket with PHP.
3. The threshold should be decreased anyway because with an increasing number of languages on GitHub, conflicts are bound to occur. It is like filling up a hash table; you must expect collisions. Having done a search of the possible values, 0.01 is the only one that works without going to a higher number of decimal places.

lildude:
> Apologies for the failed commits. Because I don't have a development machine right now, I am making changes on my phone. Several points:

No problems, though you'll find it much easier to test this stuff on an actual computer 😉

> The chosen color is the same one used in Racket's logo. If you look at the raw SVG file you will see:
> <path id="blue-piece" fill="#3E5BA9" ...

I saw that.

> I understand the color is somewhat similar to Go and PHP. However, they are clearly not identical, and blue is a very popular color. There is little danger that someone will confuse Racket with Go or Racket with PHP.

Indeed. Blue is incredibly popular and it is for this reason we've had to implement this threshold test and it's why, if you're going to want a shade of blue, you're going to have to pick another.

`#3e5ba9` (your chosen colour) and `#4F5D95` are so remarkably close that it's almost impossible to tell the two apart when they're side-by-side as they _could_ be in the language statistics bar (I know, highly unlikely, but remotely possible):

option1

`#3e5ba9` (your chosen colour) and `#375eab` are a little more distinguishable, but not easily:

option2

(I cunningly changed the proportions between the two screenshots to make guessing harder 😉)

Why not keep the current colour `#22228f`? It's not that far off the colour you're trying to set it to 
 and many other languages stick with their original choice so as not to confuse their users. 

Or how about a variant of the red (`#9F1D20`)? This currently conflicts with only one other language so a slight tweak to the hue or something should get past this. A quick little testing shows `#9F1D26` is close enough to the red in the logo to be almost indistinguishable _and_ without any conflicts.  You could try a similar tweak with your blue if you absolutely want a blue, though this will much more of a laborious task due to the high number of blues already in use.

> The threshold should be decreased anyway because with an increasing number of languages on GitHub, conflicts are bound to occur. It is like filling up a hash table; you must expect collisions. Having done a search of the possible values, 0.01 is the only one that works without going to a higher number of decimal places.

Indeed there are an increasing number of languages and conflicts do occur, however we (the Linguist community) can't unilaterally decide on lowering the threshold without a full and comprehensive analysis of the languages on GitHub and how such a change would impact the overall appearance of the language statistics bar on GitHub.com, especially when there are a lot of unused, and significantly more unique, colours still available. 

Not every language has a colour directly associated with their logo or website colours as a deliberate choice to make it more unique on GitHub, that's also an option.

pmaddams:
Thanks Colin, I appreciate your effort to resolve this issue in a reasonable manner. It seems leaving things alone is probably the best choice. It is just a color and not a bug. Closing.

PR #3820, opened 2017-09-13
add new language grammar for Q sourced from KX.tmbundle

psaris:
many examples of *.q files can be found here:

https://github.com/KxSystems/kdb
https://github.com/psaris/qtips


lildude:
Thanks for the PR. Unfortunately it's a duplicate of https://github.com/github/linguist/pull/3638 which is currently blocked due to limitations in Linguist.

psaris:
#3638 agreed to your new color proposal and if you accept the grammar without the extension then we can use custom overrides.  if we modify the color of the PR and remove the .q/.k extensions but keep the name of the programming language as q and/or k, can you approve it?

lildude:
If/when https://github.com/github/linguist/pull/3638 lands, the only thing we'd need from this PR is the grammar as that PR doesn't currently reference a grammar.

psaris:
#3638 has been closed in deference to mine. i have changed the color to an approved and agreed on hue of blue. please approve.

psaris:
@lildude, please review this PR and comment if anything else is needed

lildude:
I won't have the bandwidth for a full review until next week at the earliest. @Alhadis and/or @pchaigno may have the bandwidth for a first round review in the mean time.

.gitmodules
```diff
@@ -871,9 +871,13 @@
...
+[submodule "vendor/grammars/KX.tmbundle"]
+	path = vendor/grammars/KX.tmbundle
+	url = https://github.com/psaris/KX.tmbundle
 [submodule "vendor/grammars/language-ballerina"]
 	path = vendor/grammars/language-ballerina
 	url = https://github.com/ballerinalang/plugin-vscode
 [submodule "vendor/grammars/language-ruby"]
 	path = vendor/grammars/language-ruby
 	url = https://github.com/atom/language-ruby
+
```
lildude: Nit: please remove this blank line.

lib/linguist/languages.yml
```diff
@@ -3601,6 +3601,18 @@ Python traceback:
   tm_scope: text.python.traceback
   ace_mode: text
   language_id: 304
+Q:
+  type: programming
+  color: "#0040CD"
+  aliases:
+  - Kdb+
+  extensions:
+  - ".q"
```
lildude: https://github.com/github/linguist/pull/3638 attempted to include support for the `.k` extension too, which would have posed a problem due to clashing with non-Kdb, yet undefined, languages. This PR doesn't mention that extension at all. Whilst this isn't a problem, is `.q` the preferred extension for Q/Kdb+?

I notice that https://github.com/KxSystems/kdb which you reference includes `.k` files which won't be recognised as Q/Kdb+ once this PR has been merged, hence the question.

psaris: i have removed the extra new line, and added two more samples.  the `.k` samples from #3638 are not needed because we are not looking for the `.k` extension to be formally recognized in linguist. `.q` is the proper extension for Q/Kdb+.  in fact, users can use Q syntax highlighting to markup their .k files as well, so this PR will allow that - just not automatically.

lildude:
Thanks @psaris.  You've addressed my questions and nits. I'm happy with this PR but would like @pchaigno or @Alhadis to cast their 👀 over it too before giving my final 👍 .

Alhadis:
LGTM. 👍

lildude:
> Am I missing something? From what I see there are many SQL-like files (related to Spark) that would be misclassified by this pull request as-is...?

And that's why I wanted a second set of 👀. I completely missed this 😊. 🙇 @pchaigno

Alhadis:
I think my set of 👀 might be screwed, because I misread the above paragraph and found myself judging on [`.kdb` files](https://github.com/search?q=extension%3Akdb+NOT+nothack&type=Code) instead of the `.q` extension. 😢 Sorry.

psaris:
@pchaigno how can i provide a classifier to distinguish the language?  also, can i change the PR to support language syntax without the automatic extension handling - so that .gitattributes and vim/emacs modelines enable the highlighting?

Alhadis:
> also, can i change the PR to support language syntax without the automatic extension handling - so that `.gitattributes` and vim/emacs modelines enable the highlighting?

You won't need to change a thing. :) Once this is merged (and its effects are live on GitHub with the next release of Linguist), users will be able to do this themselves and correct the highlighting used for misclassified Q/Kdb+ files.

psaris:
apologies, is there anything else required from me before this PR can be merged?

psaris:
@alhadis this PR is still open. is there anything i need to do to get it merged?

Alhadis:
If there was, I would've told you already.

/cc @pchaigno

lib/linguist/languages.yml
abalkin: > in fact, users can use Q syntax highlighting to markup their .k files as well

Since k is accessible from q in `k)` prefixed lines, a good q syntax highlighter should know how to markup k code, but the k language is sufficiently different from q that using a q highlighter on k files is unlikely to produce good results.  The differences include over 100 q keywords that are not special in k and the treatment of `_` character in identifiers.

psaris: Q.tmbundle inherits from K4.tmbundle. so my suggestion is that people can update they .gitattributes file to link the .k extension to the K4 syntax. i agree that it doesn’t handle the _ difference between k and q.

abalkin: People will not be able to link to K4 syntax unless K4 language is known to Linguist.  It is frustrating that the dispute over default extensions prevents the Linguist from learning about the Q and K languages.  Personally, I would not care if the default extensions were `.kxq` and `.kxk`.  I would still be able to configure my repositories to use `.q` and `.k`.  However this will only be possible once `languages.yml` has two new entries.

Yes, having just Q will be a progress, but let's not give an impression that Q and K are the same language.

pchaigno:
Sorry for the late answer.

> Once this is merged (and its effects are live on GitHub with the next release of Linguist), users will be able to do this themselves and correct the highlighting used for misclassified Q/Kdb+ files.

@Alhadis I'd prefer if we don't break everyone's highlighting and language statistics... Could we add `.q` as an SQL extension with, eventually, heuristic rules to distinguish between Q/Kdb+ and SQL?

lildude:
Closing as q support has landed with https://github.com/github/linguist/pull/4058

As the language now exists, peeps can implement an override to force the `.k` extensions to be recognised as q.

PR #3823, opened 2017-09-14
added misrepresented c headers to samples

Petelliott:


pchaigno:
Thanks for the pull request!

We already have a good number of C sample files. What exactly is misrepresented in these two sample files that you think it warrants inclusion?

Petelliott:
align.h was considered a c++ file, which is understandable, as it's very simple.
memory.h was considered an objective-C file, despite being a very typical c header.

misson20000:
Even simpler case of a C header file being misrepresented as Objective-C: https://github.com/reswitched/libtransistor/blob/4920e0c52e6b53d38acf31e63cac00463b341d21/include/libtransistor/tls.h
```
#pragma once

void *get_tls();
```

lildude:
Thanks for the contribution, but in this case I don't think adding more samples is going to help improve the accuracy of the classifier when it comes to C-family header files, especially very small files or files that can be used by multiple variants of the C-family of languages.

Closing this in favour of the work taking place in https://github.com/github/linguist/pull/4025

PR #3829, opened 2017-09-18
Update languages.yml for EJSON

peterzhu2118:
Adds the EJSON (encrypted JSON) filetype to the languages list.


pchaigno:
It looks like standard JSON...? Should it be added as a new extension for JSON rather than a new language?

peterzhu2118:
@pchaigno I guess it looks the same as JSON, but it works differently. Placing EJSON as a new extension for JSON will be a little misleading as every value in EJSON file will be encrypted (vs. plain text for a JSON file).

lildude:
Closing as stale.

PR #3422, opened 2017-01-09
Add AngelScript scriping language

ifarbod:
From http://angelcode.com/angelscript:

>The AngelCode Scripting Library, or AngelScript as it is also known, is an extremely flexible cross-platform scripting library designed to allow applications to extend their functionality through external scripts. It has been designed from the beginning to be an easy to use component, both for the application programmer and the script writer.
For the script writer the scripting language follows the widely known syntax of C/C++, but without the need to worry about pointers and memory leaks. Contrary to most scripting languages, AngelScript uses the common C/C++ datatypes for more efficient communication with the host application.

The selected color is also from the same website.

It's basically a C/C++-like language, so I guess there's no need for a custom grammar.
[Here](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aangelscript+NOT+nothack&type=Code&ref=searchresults)'s a search result with some angelscript files.

PR #3844, opened 2017-10-04
Add Angelscript language

codecat:
**Note:** This is a continuation of #3422 by @ifarbod, please see that PR for a bit more details, including [link to search results](https://github.com/search?utf8=✓&q=extension%3Aangelscript+NOT+nothack&type=Code&ref=searchresults), however I believe most Angelscript developers use the .as extension which is ambiguous with Actionscript.

Contributing.md mentions that @bkeepers is able to help with ambiguity testing? (I can provide a sample of > 500 Angelscript files from our game, but there's a lot of Angelscript codebases out there.)

Otherwise, I could also write a disambiguate rule in `heuristics.rb` if necessary.

On top of the original PR, I have added [this Angelscript grammar](https://github.com/wronex/sublime-angelscript).


codecat:
It looks like Travis is failing with some cryptic errors that looks like it's confused about the ambiguity. Should I write a heuristic for it, or..?

The license for the grammar also says "other", which it is warning about. It is however "Unlicense", so I believe that's just fine and it didn't properly detect that.

It's also complaining about missing **ActionScript** samples! There are indeed no samples there, so it's probably good to add some. Should I do that in this PR?

codecat:
I'm going to mark the license for the grammar file as `unlicensed` manually. I hope that's okay.

codecat:
Okay, there is 1 issue left. It says "once approved", so I will wait for a maintainer to approve it I suppose.

codecat:
I have added a heuristic as well, though not entirely sure how to properly test it.

lildude:
Wow!!! Thanks for the great PR, and welcome to the Linguist community. You move so fast we've not had a chance to respond 😄

> Contributing.md mentions that @bkeepers is able to help with ambiguity testing? (I can provide a sample of > 500 Angelscript files from our game, but there's a lot of Angelscript codebases out there.)

This is my responsibility now (update to `CONTRIBUTING.md` coming :soon:). No need for this at the moment as the corpus gathering side of things has been bust for a while and I've not fixed it yet 😁

> The license for the grammar also says "other", which it is warning about. It is however "Unlicense", so I believe that's just fine and it didn't properly detect that.

Yeah, this is a bit of a limitation of the version of Licensed we pull in. A newer version is in the works which will handle situations like this a lot better. I'm pushing my colleagues to open source it too and have buy-in from the one of the devs (FYI @Alhadis & @pchaigno).

> I'm going to mark the license for the grammar file as unlicense manually. I hope that's okay.
> 
> Okay, there is 1 issue left. It says "once approved", so I will wait for a maintainer to approve it I suppose.

We normally like to give a 👍 before you do this and whitelist the grammar, but I'll let it slide this time
 😉 as I'd have given the 👍 in this case anyway.

> I have added a heuristic as well, though not entirely sure how to properly test it.

This is definitely needed due to the clash with ActionScript as you pointed out earlier, and very much appreciated 🙇.  I'll comment on your heuristic inline in a mo.

lib/linguist/heuristics.rb
```diff
@@ -73,6 +73,14 @@ def call(data)
     # Common heuristics
     ObjectiveCRegex = /^\s*(@(interface|class|protocol|property|end|synchronised|selector|implementation)\b|#import\s+.+\.h[">])/
 
+    disambiguate ".as" do |data|
+      if /^\s*(package\s+[a-z0-9_\.]+|import\s+[a-zA-Z0-9_\.]+;|class\s+[A-Za-z0-9_]+\s+extends\s+[A-Za-z0-9_]+)/.match(data)
```
lildude: Caveat pre-emptor: I know nothing about either AngelScript or ActionScript....

How confident are you that this will match most of the ActionScript files and won't match any AngelScript?

Is there a similar AngelScript-specific regex we could use to be even more precise with the "else"?  It's not essential, just trying to be extra safe.

samples/ActionScript/FooBar.as
```diff
@@ -0,0 +1,35 @@
+// A sample for Actionscript.
+
```
lildude: Thanks for the great samples. They're small and show a nice diverse range in how the language could be used.  Just one teeny weeny question: what is the source of these examples and what license covers them?

We don't need an explicit license added or comment, just clarification as we can't include any samples that comes from sources licensed under licenses that are _not_ in this list https://github.com/github/linguist/blob/master/vendor/licenses/config.yml#L1-L12

lib/linguist/heuristics.rb
codecat: I don't know a lot about Actionscript myself. I actually wrote the samples myself after doing a bit of research on the language. During that I've found that you need to have a package (which is the first match here), in which any classes will be contained. (This seems like a requirement, though [this Adobe packages article](http://www.adobe.com/devnet/actionscript/learning/as3-fundamentals/packages.html) doesn't seem to mention it as a requirement.) Either way, this is exclusive to Actionscript.

The import statement in Actionscript is used like:
```
import com.example.ClassName;
```

Angelscript *also* has an import keyword, which is used like:
```
import void MyFunction(int a, int b) from "Another module";
```

Angelscript however only lets you import functions, so that shouldn't clash with the regex used in the heuristic.

Class inheritance in Actionscript uses the `extends` keyword which is not a valid keyword in Angelscript. To inherit in Angelscript, you simple use `:`.

I can come up with a few Angelscript-specific matches, but those are not required to be used at all, such as: `[A-Za-z0-9_::]+@` for reference types, `@[A-Za-z0-9_::]\s*=` for setting references instead of values, or also even using the `auto` keyword, perhaps with a match like `^\s*auto\s+`, although it's still possible that someone names their Actionscript class `auto`, or if that appears in a block comment.

There's also the `is` operator for reference comparing, so a match like `^\s*if\s*\(.*\s+is\s+` could maybe even work. (Only `\s+is\s+` probably won't, since that can appear in comments/strings.)

Would love to know your thoughts on this, I'm unsure what's normally the most accepted way of detecting these.

samples/ActionScript/FooBar.as
codecat: As I wrote above, I made these samples after reading through some Actionscript getting started articles. I was initially going to copy paste the actual samples from the Actionscript documentation, but their documentation is licensed as CC BY-NC-SA 3, which I presume is not an acceptable license.

Could of course search through some Github projects to find some real properly licensed samples.

As for the Angelscript samples, one of which is written by me and available here: https://github.com/codecat/ssbd-payload It's currently not licensed, but if you need it to be licensed I can slap an MIT license on it and call it a day. :)

Not sure where @ifarbod got their sample from.

codecat:
I would've waited putting the license in the whitelist, but I wanted to make the tests pass for that sweet green checkmark. Sorry about that! 😊

lib/linguist/heuristics.rb
lildude: Wow! This is just the substantiation I was looking for. Thanks.

samples/ActionScript/FooBar.as
lildude: Ditto here. 🙇

ifarbod:
Regarding the sample, It's mine, But since I don't care anymore about it, Think of it as Public Domain licensed.

lib/linguist/heuristics.rb
codecat: Good to hear! If you do want me to do any additional changes to the heuristic, let me know.

Alhadis: I used to be highly fluent in ActionScript back in the days when Flash were relevant, and I can honestly tell you I've never seen a line like `import void MyFunction(int a, int b) from "Another module";` appear in any conventional AS3 files. It's pretty much full-on Java-esque with their dot-delimited import style:

~~~actionscript
import com.Java.*
import com.Gosling.Ryan.Cafebabe;
import com.oracle.CorporateTakeover;
~~~

codecat: I think you've misread, that line is an Angelscript line, not Actionscript. The language names are similar I admit.. :D

codecat: So @Alhadis, what do you think, should I change anything in the heuristic, or is this one fine?

Alhadis: Ah sorry, I missed this. No, everything's fine. I was simply lending assurance that this was the correct approach, since you'd both mentioned you didn't know ActionScript.

> Also, does the automated testing test the heuristics against samples?

It does, but I believe they need to be included in [`test/test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb).

codecat: I've added the test for the heuristic and it looks like it works.

samples/AngelScript/payload.as
```diff
@@ -0,0 +1,396 @@
+// Sample script.
```
pchaigno: If, in AngelScript, `.as` and `.acs` file have exactly the same meaning and syntax, would it be possible to rename this file to `payload.acs`?

I'm worried we might have a conflict on `.acs` in the future, in which case we'll have to find a `.acs` sample file for AngelScript. We might as well take care of this now.

test/test_grammars.rb
```diff
@@ -43,6 +43,7 @@ class TestGrammars < Minitest::Test
     "82c356d6ecb143a8a20e1658b0d6a2d77ea8126f", # idl.tmbundle
     "9dafd4e2a79cb13a6793b93877a254bc4d351e74", # sublime-text-ox
     "8e111741d97ba2e27b3d18a309d426b4a37e604f", # sublime-varnish
+    "23d2538e33ce62d58abda2c039364b92f64ea6bc", # sublime-angelscript
```
pchaigno: I'm surprised this is needed since Licensee seems to already correctly recognize the license on the repository: https://github.com/wronex/sublime-angelscript (right side of the top bar).

codecat: Here's the travis test that failed before this was added: https://travis-ci.org/github/linguist/builds/283335180?utm_source=github_status&utm_medium=notification

Indeed then, it also recognized the license correctly. Maybe @lildude knows more?

samples/AngelScript/payload.as
codecat: Quite honestly, I didn't modify the extensions part to honor @ifarbod's initial PR. Personally I've never seen the `.acs` or `.angelscript` file extensions being used. In fact, all samples in the Angelscript SDK are using `.as` for file extensions, and I can't find any reference to "acs" in the Angelscript documentation.

Personally I'd lean more towards removing those 2 file extensions altogether, but I'd like to hear from @ifarbod first! 😊 (If not `.angelscript`, then only `.acs`, since that seems like the strangest file extension to me.)

ifarbod: I've talked to AngelScript's author regarding the extension situation before, asking them for alternatives to `.as` where it's not possible, They said it's fine to use either `.acs` or `.angelscript`. They even initially wanted to use `.acs` as the "official" file extension for it. But if an extension's going to be removed, I'd say remove `.acs`, but keep support for `.angelscript` as the usage for this extension is more common than `.acs`.

Cheers

codecat: You're right, searching github for `.angelscript` brings up a lot of stuff while searching for `.acs` brings up a lot of unrelated stuff (some unrelated Doom scripting language as well as XML-formatted files.) I guess I'll remove the `.acs` extension.

ifarbod: I think It's all about legacy. AngelScript built its legacy on `.as` and `.angelscript` file extensions. 😉

test/test_grammars.rb
pchaigno: Okay, it looks like [Licensee](https://github.com/benbalter/licensee/releases/tag/v8.4.0) does not recognize the license for that repository since v8.4.0. I think we're fine with adding the license hash for now and I'll open an issue on the Licensee repository to investigate.

codecat:
Thanks so much for merging!! Looking forward to seeing Angelscript identified on Github! :tada:

PR #3853, opened 2017-10-10
Fix typo in script/add-grammar

lildude:
Just noticed this typo and thought best to fix it now whilst I remember as I _will_ forget otherwise.


Alhadis:
I must've been wired at God knows what hour... 😥

PR #3855, opened 2017-10-11
Add .hsig extensions for Haskell highlightning

alexbiehl:
Since version 8.2.1 GHC supports declaration of so called signatures which also use Haskell syntax.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1824,6 +1824,7 @@ Haskell:
   extensions:
   - ".hs"
   - ".hsc"
+  - ".hsig"
   interpreters:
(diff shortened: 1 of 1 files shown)


pchaigno:
Please have a look at the [guidelines to add support for a new extension to Linguist](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language). In particular, notice the requirement on in-the-wild usage.

pchaigno:
Closing in favor of #4219. We'll monitor the popularity of the extension there.

If you have an improved search query compared to the one used in #4219, please post it here and I'll update #4219.

PR #3856, opened 2017-10-11
Ping @lildude from now on

lildude:
@bkeepers is no longer actively involved with Linguist so it doesn't makes sense to ping him anymore. Updating the docs to get peeps to ping me.

As an aside, the corpus gathering on the GitHub side doesn't work at the mo so the ping is a bit moot at the moment, but it is on my list to fix it :soon:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ To add support for a new extension:
 In addition, if this extension is already listed in [`languages.yml`][languages] then sometimes a few more steps will need to be taken:
 
 1. Make sure that example `.yourextension` files are present in the [samples directory][samples] for each language that uses `.yourextension`.
-1. Test the performance of the Bayesian classifier with a relatively large number (1000s) of sample `.yourextension` files. (ping **@bkeepers** to help with this) to ensure we're not misclassifying files.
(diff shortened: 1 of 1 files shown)


Alhadis:
What do you mean the corpus gathering doesn't work on GitHub's side? Is this referring to the way public search results can't be browsed using the `NOT nothack` hack unless you're logged into GitHub?

lildude:
I'm referring to the thing we used to do where we created a big fat zip file and bunged it on S3 for peeps to download and test against. That part doesn't work. It used to take advantage of the ability to search public repos anonymously without limiting to a user/repo/org that borked it. The interface has also now been removed so I'll need to implement it some other way.

Alhadis:
Definitely a much better and efficient solution than my [hacky harvester script](https://gist.github.com/Alhadis/cae3452d49d640b48a507cadd02f86ce). :( Which has to be operated semi-manually, and the downloading process needs to happen separately. It's all a colossal mess because of no API from GitHub that allows filetypes/filenames to be searched *without* specifying a user or organisation. 😢

lildude:
@Alhadis or @pchaigno can one of your guys :+1: this so I can merge it. 🙇

PR #3857, opened 2017-10-11
Don't count VCL as Perl for statistics.

dcmoore-gd:
While the Varnish-specific language was apparently inspired by C and Perl, there's no reason to group it as Perl for repo statistics.  Also re-adding the VCL `color` definition.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4689,7 +4689,6 @@ UrWeb:
   ace_mode: text
   language_id: 383
 VCL:
-  group: Perl
   type: programming
   extensions:
   - ".vcl"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4690,6 +4690,7 @@ UrWeb:
   language_id: 383
 VCL:
   type: programming
+  color: "#0298c3"
   extensions:
   - ".vcl"
   tm_scope: source.varnish.vcl


dcmoore-gd:
I could also see setting the group as "Other" or "Vendor-Specific"/"Domain-Specific" if these kinds of languages should be grouped together, but that doesn't seem to be a pattern in `languages.yml` at this time.

pchaigno:
@gjtorikian added the Perl group to VCL in #2298. Could be an oversight or he had some legitimate reason. @gjtorikian Do you remember?

gjtorikian:
Oh, it was almost certainly an oversight. In fact, I’m a bit embarrassed to see that I deleted the color, too. 😨 It might be nice to define that back in here again.

dcmoore-gd:
I can make that color change too as part of this PR.

Alhadis:
> In fact, I’m a bit embarrassed to see that I deleted the color, too

The colour deletion was justified if it was being grouped under Perl, because it would never have had a chance to be used or seen in the language bar. Nothing to feel embarrassed about. =)

dcmoore-gd:
Thanks @Alhadis and @gjtorikian for your work on this project and feedback on this PR!  VCL's `color` has been restored in this PR now too and CI tests, they are a'passin.

gjtorikian:
Thanks!

Issue #3637, opened 2017-05-23
Perl 5 code being mistaken for Perl 6


wchristian:
This is recognized as Perl 6, wrongly:

```

=head1 NAME

class

=cut

package PH;

use 5.010;
```

This is recognized as Perl 5:

```

=head1 NAME



=cut

package PH;

use 5.010;
```

wchristian:
I don't get reliable results when i push this to github with changes, but this looks like it also is recognized as Perl 6, wrongly:

```
use 5.006;
use strict;
use warnings;

=head1

module

=cut
```

And removing `module` changes it to perl 5.

wchristian:
Forgot to mention, in both examples the files were `Word.pm`.

zoffixznet:
Looks like just swapping [these conditionals](https://github.com/github/linguist/blob/58ab593a64088be56e3323a9fb121fe0988bdb3c/lib/linguist/heuristics.rb#L347-L353) to check for Perl 5 regex first will solve the issue.

kentfredric:
https://github.com/github/linguist/blob/58ab593a64088be56e3323a9fb121fe0988bdb3c/lib/linguist/heuristics.rb#L348

^ That heuristic would false positive on anything using MooseX::Declare  style perl-5 syntax anyway.

So yeah, +1 on inversion.

pchaigno:
We'd welcome a pull request with a test ([`test/test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb)) to implement that change :smiley:

zoffixznet:
> Is it really worth it to have Perl 6 separated from Perl

Yes, considering it's an entirely different language. It's like C vs C++, except with poorer name.

Alhadis:
@smola Perl 5 and Perl 6 are, as @zoffixznet put it, two entirely separate languages. To quote *"Programming Perl"*:

> That 6 is deceptive though; Perl 6 is really a “kid sister” language to Perl 5, and not just a major update to Perl 5 that version numbers have trained you to expect.

tessarin:
@pchaigno Do you know if this was in fact fixed? I've got a similar problem in a repository for a Perl 5 module which has its main file being incorrectly interpreted as Perl 6, making the repository also Perl 6 (funny enough, named [**p5**-POD-Auto](https://github.com/tessarin/p5-POD-Auto)). Thanks!

pchaigno:
[...]

PR #3860, opened 2017-10-16
Fix Perl 5 and 6 disambiguation bug

tessarin:
Perl 5 code containing `class` or `module` was incorrectly interpret as Perl 6, as described in #3637.

This fixes the issue and adds a test which would cause a failure without the change to `heuristics.rb`.


tessarin:
Thank you @pchaigno

PR #3863, opened 2017-10-17
Only recognise Pod for .pod files

### LINKED ISSUE #3832: perl files with no extension and POD and a shebang should be rendered as syntax-highlighted perl, not POD
This bug was introduced in #3735. Here's an expanded version of my comment from there.

This change recognizes files with no extension and a perl shebang and POD contents as POD files, but I'd expect them to be shown as syntax-highlighted perl. I would expect POD-only files to (maybe) have a `.pod` extension, and (always) no shebang. Some examples from CPAN:

https://metacpan.org/source/LEONT/Module-Build-0.4224/lib/Module/Build/API.pod has no shebang and a `.pod` extension.

https://metacpan.org/source/LEONT/Module-Build-0.4224/lib/Module/Build/Cookbook.pm has no shebang and a `.pm` extension.

Here's an example of where linguist chooses to render as POD rather than as syntax-highlighted perl: https://github.com/sitaramc/gitolite/blob/5c2fe87019dadbecb1883bedd075cd07b07b2078/src/gitolite
It previously displayed as a highlighted perl script (good), but now displays as the embedded POD (requiring me to use the Raw view to be able to see the code). It has a shebang.

Please fix it to correctly detect these kinds of files as perl (not POD) or revert it, thanks!

kivikakk:
We uncomplicate matters by removing ".pod" from the Perl definition entirely.

Fixes #3832.

/cc @lildude


kivikakk:
@Alhadis thanks! testing this on dotcom locally, then will merge/release/get a live deploy going.

Alhadis:
I'm honestly surprised by how many `.pod` files there are which use hashbangs...

kivikakk:
Now live.

PR #3865, opened 2017-10-17
Add Cocoapods to generated list so it doesn't show in PR diffs

freak4pc:
## What ?

This PR adds the `Pods` folder to `generated.rb` to blacklist it and its contents from PR diffs, etc.

## Why ?

While many developers don't commit their `Pods` folder, there are many developers and companies who do, due to the following reasons:

1. Quicker build times in CI/CD _(this is the top reason, usually)_
2. Ability for new developers to build immediately after cloning
3. Many companies prefer keeping dependencies cloned to protect themselves from the day a project is "gone"

The problem for pushing PRs that contain changes to dependencies is that they create an impossible-to-follow PR, as well as the fact the _actual files_ changed by the developer don't get diffs as there are too many files. 

Here's just a single example but we, as well as many other companies and devs I know, have this issue:
e1d96f3f-c3ac-40a3-a1ee-35603375e735

I would appreciate your consideration of merging this change, for the sanity of iOS developers :neckbeard:

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -100,7 +100,7 @@ module Linguist
     #
     # Returns true or false.
     def cocoapods?
-      !!name.match(/Pods\//)
+      !!name.match(/(^Pods|\/Pods)\//)
     end
 
     # Internal: Is the blob minified files?
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -52,6 +52,7 @@ module Linguist
     # Return true or false
     def generated?
       xcode_file? ||
+      cocoapods? ||
       generated_net_designer_file? ||
       generated_net_specflow_feature_file? ||
       composer_lock? ||
@@ -93,6 +94,13 @@ module Linguist
     # Returns true of false.
     def xcode_file?
       ['.nib', '.xcworkspacedata', '.xcuserstate'].include?(extname)
+    end
+
+    # Internal: Is the blob part of Pods/, which contains dependencies not meant for humans in pull requests.
+    #
+    # Returns true or false.
+    def cocoapods?
+      !!name.match(/Pods\//)
     end
 
     # Internal: Is the blob minified files?
--- a/test/test_file_blob.rb
+++ b/test/test_file_blob.rb
@@ -491,9 +491,6 @@ class TestFileBlob < Minitest::Test
     # Carthage
     assert sample_blob('Carthage/blah').vendored?
 
-    # Cocoapods
-    assert sample_blob('Pods/blah').vendored?
-
     # Html5shiv
     assert sample_blob("Scripts/html5shiv.js").vendored?
     assert sample_blob("Scripts/html5shiv.min.js").vendored?
--- a/test/test_file_blob.rb
+++ b/test/test_file_blob.rb
@@ -188,6 +188,10 @@ class TestFileBlob < Minitest::Test
     assert fixture_blob("Binary/MainMenu.nib").generated?
     assert !sample_blob("XML/project.pbxproj").generated?
 
+    # Cocoapods
+    assert sample_blob('Pods/blah').generated?
+    assert !sample_blob('My-Pods/blah').generated?
+
     # Gemfile.lock is NOT generated
     assert !sample_blob("Gemfile.lock").generated?
 
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -232,9 +232,6 @@
 # Carthage
 - ^Carthage/
 
-# Cocoapods
-- ^Pods/
-
 # Sparkle
 - (^|/)Sparkle/
 
--- a/test/test_generated.rb
+++ b/test/test_generated.rb
@@ -42,6 +42,12 @@ class TestGenerated < Minitest::Test
     generated_sample_without_loading_data("Dummy/foo.xcworkspacedata")
     generated_sample_without_loading_data("Dummy/foo.xcuserstate")
 
+    # Cocoapods
+    generated_sample_without_loading_data("Dummy/Pods/Pods.xcodeproj")
+    generated_sample_without_loading_data("Dummy/Pods/SwiftDependency/foo.swift")
+    generated_sample_without_loading_data("Dummy/Pods/ObjCDependency/foo.h")
+    generated_sample_without_loading_data("Dummy/Pods/ObjCDependency/foo.m")
+
     # Go-specific vendored paths
     generated_sample_without_loading_data("go/vendor/github.com/foo.go")
     generated_sample_without_loading_data("go/vendor/golang.org/src/foo.c")
--- a/test/test_generated.rb
+++ b/test/test_generated.rb
@@ -43,6 +43,10 @@ class TestGenerated < Minitest::Test
     generated_sample_without_loading_data("Dummy/foo.xcuserstate")
 
     # Cocoapods
+    generated_sample_without_loading_data("Pods/Pods.xcodeproj")
+    generated_sample_without_loading_data("Pods/SwiftDependency/foo.swift")
+    generated_sample_without_loading_data("Pods/ObjCDependency/foo.h")
+    generated_sample_without_loading_data("Pods/ObjCDependency/foo.m")
     generated_sample_without_loading_data("Dummy/Pods/Pods.xcodeproj")
     generated_sample_without_loading_data("Dummy/Pods/SwiftDependency/foo.swift")
     generated_sample_without_loading_data("Dummy/Pods/ObjCDependency/foo.h")


freak4pc:
P.S Ruby isn't my top language, I added some relevant tests but couldn't get them to run locally due to `rugged` not installing unfortunately. 

b6038f19-14bd-44ab-93b1-6effd11b01ba

lildude:
Hmmm, I'm pretty sure CocoaPods are _not_ generated files. The usage you've described and my limited knowledge, suggests these are more inline with being "vendored" rather than generated.

freak4pc:
@lildude How does that differ than having [`node_modules`](https://github.com/freak4pc/linguist/blob/2ca563c92d129445390b3c02dbebe979af668ad4/lib/linguist/generated.rb#L59) or [`go_vendor`](https://github.com/freak4pc/linguist/blob/2ca563c92d129445390b3c02dbebe979af668ad4/lib/linguist/generated.rb#L60) as generated? It seems

lildude:
Yeah, I know about those and I'm not entirely sure why they're under generated when they're clearly not. Almost everything else is. Interesting, `node_modules` is also vendored as is CocoaPods https://github.com/github/linguist/blob/15885701cd883b633847979583f128e4ba1beac9/lib/linguist/vendor.yml#L235-L236

I think this all comes back to the blurred line between vendored and generated. Generally peeps don't want either in their language stats or diffs, but GitHub.com doesn't exclude vendored files from diffs, only generated files.

pchaigno:
@lildude Until we've got more accurate classifications between vendored and generated, it's probably best to add `Pods/` to generated files no?

freak4pc:
@lildude I think the problem is vendor.yml doesn't dictate whether certain files are included in the diff, while generated does. Even though the terminology is incorrect at the moment, it seems for the time, as @pchaigno suggested, it might be good to "keep as is" and perhaps have a separate issue to deal with separating those somehow

freak4pc:
@lildude P.S I also think the tricky part is Pods includes vendored contents (the framework sources), but it also includes _a bunch_ of generated stuff (xcodeprojs, target support files, etc.) . So it lies somewhere in the middle of those regardless.

lildude:
> @lildude Until we've got more accurate classifications between vendored and generated, it's probably best to add Pods/ to generated files no?

I'd looked into this more, had a walk and a think, and I agree... lets go for it. CocoaPods seems to pull in a metric 💩 ton of stuff in almost every case I found. 😱 

As we're marking these as generated, @pchaigno think we should remove the entry from vendored too to avoid confusion?

freak4pc:
@lilsude would it cause the language stats to break though ? If not I don’t see a problem with it personally :)

pchaigno:
> As we're marking these as generated, @pchaigno think we should remove the entry from vendored too to avoid confusion?

Don't have a strong opinion, but would rather go for adding `go_vendor` to `vendor.yml`. I don't think it has any impact anyway and the three entries are supposed to be vendored (`Pods`, `node_modules`, and `go_vendor`).

freak4pc:
Maybe in the future migrating to something along the lines of whitelist_languages.ymk vs whitelist_diffs.ymk will give more proper context, since those are sort of the classification of these to begin with I believe

freak4pc:
@lildude @pchaigno Thank you guys for reviewing this and I'm glad to hear your open on moving forward. Is there anything I can do to help bring this home?

freak4pc:
@lildude @pchaigno Sorry for poking again but we're stale for about a week. Are we missing anything here or do we have any worries/concerns about merging this?

orta:
I'm one of the maintainers of CocoaPods - I think this is a good idea BTW 👍

lildude:
@freak4pc I've been AFK for a week so it wouldn't have been merged anyway. I also prefer to leave the initial approval to the community maintainers before putting my final 👍/👎 on PRs. Additionally, this won't appear on GitHub.com until I make another release which won't be for a few weeks due to other commitments, though I may be able to sneak it in if I've got a free moment next week - if my flights across the pond had WiFi I'd do it then, but they don't 😭. 

So for the moment, hold fast and I'm sure @pchaigno or @Alhadis will take a peek when they get a break in their own schedules.

freak4pc:
Sounds good ! No rush at all, I just wanted to know we're all on the same page. Thanks @lildude !

freak4pc:
🎉🎉🎉

freak4pc:
Hey guys, since we have approval can we merge this in ? 
Also, I know there's a release process here after merge - is there a plan for upcoming version here as well? 

Thanks again for the assistance in reviewing this change! @lildude

lildude:
Before approving this, @pchaigno & @Alhadis: do you think it's a good idea to remove Cocoapods from the vendoring classification (and associated tests if applicable) whilst we're at it? I'd prefer not to have redundant classification as it'll lead to confusion down the line, but I'd like your thoughts on it.

Alhadis:
I'm afraid I'm the wrong guy to be asking about CocoaPods, my friend.;) But if these `Pods` file do indeed filll the niche of `node_modules` folders for Cocoa projects, then by all means give 'em the boot (so to speak). 👍

freak4pc:
They fill the same role as the folder can be “rebuilt” using ‘pod install’. Technically there would be no difference no difference regarding the resulted language display for the project here, right? I think it’s safe to remove from vendored @lildude @Alhadis

freak4pc:
I’ll remove those from this PR in a few :)

freak4pc:
@Alhadis @lildude I'm confused though, it seems you also have node_modules on both files... 
https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml#L41

lildude:
I was referring to the CocoaPods section: https://github.com/github/linguist/blob/9c4dc3047ca36e810f4584361cb3f04d76bee4a2/lib/linguist/vendor.yml#L247-L248

This is not needed and never encountered now the files are considered "generated".

Don't worry about the `node_modules` for the moment as it's not relevant for this PR.

freak4pc:
Yeah I understood, just thought its weird to have node_modules but not Pods as they are the same thing pretty much :) Anyways, sure I'll get rid of it, I agree it fits in a different PR.

freak4pc:
@lildude Ready for you :)

freak4pc:
Hey guys, I think we have all feedback addressed as well as several approvals here. Is there anything else preventing us from moving forward? Thanks ! @lildude @Alhadis @pchaigno

ashfurrow:
Seconded – I understand that this won't be on Github immediately once merged, but this PR is going to significantly improve the day-to-day work of _thousands_ of iOS developers, myself included. I'm very keen to see this move forward.

lildude:
The only thing holding this back is my own free time 😀 This is a relatively large change in the grand scheme of things (now you see it, now you don't) so I want to do a bit of testing and tinkering before I merge this.

freak4pc:
@lildude Sure, that makes sense. Is there anything we can do to help testing-wise ? 

I understand being "Careful" around this but this is not different than node_modules in that sense so I think most people would actually expect it. As you can see from this thread, there are no "againsts" for this, and many many people who could really use this addition. 🙏 

I would really want to see if it would be possible to merge and release before mid-December, as I'm worried if we won't make it before Holiday season it would be delayed to end of January 😄 which is fine, but would be nice to get it in before since its such a straightforward change.

Again, thank you so much for all the help and consideration of merging this PR! 💯

lib/linguist/generated.rb
```diff
@@ -95,6 +96,13 @@ def xcode_file?
       ['.nib', '.xcworkspacedata', '.xcuserstate'].include?(extname)
     end
 
+    # Internal: Is the blob part of Pods/, which contains dependencies not meant for humans in pull requests.
+    #
+    # Returns true or false.
+    def cocoapods?
+      !!name.match(/Pods\//)
```
lildude: This needs tightening as it'll match `I-am-not-Pods/` or any other combination of `*Pods/` which we probably don't want.

test/test_file_blob.rb
```diff
@@ -491,9 +491,6 @@ def test_vendored
     # Carthage
     assert sample_blob('Carthage/blah').vendored?
 
-    # Cocoapods
-    assert sample_blob('Pods/blah').vendored?
```
lildude: Can this test be move to `[test_generated()`](https://github.com/freak4pc/linguist/blob/fca6ebbcfd4101561944d387573a3b202025e650/test/test_file_blob.rb#L183`) and flipped to `.generated?`

Can a new test be added to ensure something like `Dummy/not-Pods/` is _not_ treated as generated.

lib/linguist/generated.rb
freak4pc: No problem! 
P.S just so you're aware - this same method applies to node_modules etc, which means this would also ignore a folder called `MyThing-node_modules`

test/test_file_blob.rb
freak4pc: @lildude I already have these tests in generated, see here: https://github.com/github/linguist/pull/3865/files#diff-bd71790729c9298dbbdadc57b93d1038R45

As for the "not" test - I just want to confirm, we don't have these kind of test for any other of the generated folders (go/vendor, node_modules, etc). Also, not entirely sure on the test structure of that file and have some hard problem running these tests locally :) I'll try to count on CI for this.

lib/linguist/generated.rb
lildude: Yup, I saw that. I'd like to only concentrate on cocoapods in this PR to keep things simple and limit the scope of the changes to be relevant to the original purpose of this PR.

test/test_file_blob.rb
freak4pc: @lildude Would you want me to add a method to test_generated.rb along the lines of `not_generated_without_loading_data` to assert the negative of `generated_without_loading_data` ? 

Or would you want me to manually assert that test case? Since there isn't any other test case in that file to assert something _doesn't_ match

lildude: That's not quite the same thing. The tests you've added test the Generated behaviour. The test changes I've requesting are for FileBlob.

As for the "not" tests - good point. I was thinking about trying to start testing the regexs to ensure they continue to remain accurate and limited to the scope initially intended, but as the tests stand, this will involve adding another irrelevant sample.  Ignore the "not" test for the moment.

freak4pc:
Waiting for CI to make sure I didn't mess anything up, but let me know if the changes I made make sense to you @lildude :) Thanks again for taking the time !

freak4pc:
All green 💚

test/test_file_blob.rb
```diff
@@ -188,6 +188,10 @@ def test_generated
     assert fixture_blob("Binary/MainMenu.nib").generated?
     assert !sample_blob("XML/project.pbxproj").generated?
 
+    # Cocoapods
+    assert sample_blob('Pods/blah').generated?
+    assert !sample_blob('My-Pods/blah').generated?
```
freak4pc: @lildude It's a bit more straightforward in this test file to add the "negative" test, so I did. I think this is OK for now to test the regex validity but up to you :)

lildude:
@freak4pc you can have the pleasure of clicking the big green button 😄 . There's no need to change any of the content for the commit message etc.

freak4pc:
@lildude Would be my honor, but I don't seem to have write-access (yet) 😜 

0cf7a01a-7cc2-4538-b501-ac32af164add

lildude:
Doh! I'll do it.

freak4pc:
Thanks ! Just of if curiosity, is there a planned release that will include this ? @lildude

lildude:
There's no planned release date yet, but this will definitely be in the next release.

freak4pc:
Hey @lildude, Really don't want to put pressure or anyways but just wanted to ping and see what's the status on this release?

This feature was merged almost 3 months ago. Many thanks 🙏

lildude:
I'm working on a release right now, however I've hit a little niggle with the handling of symlinks that needs to be fixed first. I'm aiming for the release to be live by the end of the month.

freak4pc:
Sounds great! Thanks a lot :)

PR #3866, opened 2017-10-18
Set color for Idris language

rkoeninger:
Idris is probably [big enough](https://github.com/search?l=Idris&q=idris&type=Repositories&utf8=%E2%9C%93) now to earn is its own color.

Medium red color (`#b30000`) taken from language logo:


Alhadis:
If the language has enough usage to qualify for recognition on GitHub, then it certainly qualifies to have its own colour. ;)

There's no prerequisites for a registered language to have a colour, in other words.

PR #3868, opened 2017-10-18
Correct Arduino's Color

MarcianoPreciado:
Change arduino's purple color representation to it's signature blue.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2499,7 +2499,7 @@ M4Sugar:
   language_id: 216
 MAXScript:
   type: programming
-  color: "#00a6a6"
+  color: "#016b6b"
   extensions:
   - ".ms"
   - ".mcr"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -281,7 +281,7 @@ Arc:
   language_id: 20
 Arduino:
   type: programming
-  color: "#bd79d1"
+  color: "#00979d"
   extensions:
   - ".ino"
   tm_scope: source.c++


MarcianoPreciado:
The color I took for 3D max was taken from a color picker of their logo. I don't see how that isn't representative. The net change of the repo is more accurate. Arduino becomes represented correctly, as opposed to 100% inaccurate representation. 3D Max goes from slightly misrepresented to slighly more misrepresented (with a color that is on it's logo).

Alhadis:
In all fairness, the minimum required difference between colours should really be dropped. It's doing more harm than good (as PRs like this have repeatedly evidenced), and colours that are too similar are unlikely to be featured side-by-side in the same codebases anyway.

/cc @lildude Is there anything we can do about this? Many colours misrepresent their languages this way... 😕

Alhadis:
@MarcianoPreciado Randomly selecting a pixel isn't a good way of selecting a colour, FYI. The particular shade of teal used by 3DS Max is `#00aaaa`, which can be seen on its [Autodesk product page](https://www.autodesk.com.au/products/3ds-max/overview).

When possible, try to consult an official styleguide before resorting to the eyedropper. 😉

pchaigno:
> In all fairness, the minimum required difference between colours should really be dropped. It's doing more harm than good (as PRs like this have repeatedly evidenced), and colours that are too similar are unlikely to be featured side-by-side in the same codebases anyway.

I'm convinced that, if we could see the available colors on a color panel, we'd find that there are plenty of options remaining... (before we start looking into the likeliness of languages to be featured side-by-side)

lildude:
> /cc @lildude Is there anything we can do about this? Many colours misrepresent their languages this way... 😕

I'm reluctant to relax or remove the colour proximity requirement as even with it we still end up with colours incredibly close as witnessed by https://github.com/github/linguist/issues/3910

> I'm convinced that, if we could see the available colors on a color panel, we'd find that there are plenty of options remaining... (before we start looking into the likeliness of languages to be featured side-by-side)

Well, that could be a heeeeeaaaawg page. We can however see the colours already used... https://github.github.com/linguist/ (I discovered this the other day 😄 ).

lildude:
That said, I'm closing this issue as Arduino is _not_ a language and will be removed as a language once https://github.com/github/linguist/pull/3933 has been merged. See https://github.com/github/linguist/issues/3929 for more context.

PR #3874, opened 2017-10-23
Check generated Jest snap file

yuya-takeyama:
This is to mark snapshot files generated by Jest (a testing framework by Facebook) as generated.
I want such files to be collapsed in GitHub diff.

ref: https://daveceddia.com/snapshot-testing-react-with-jest/

btw, does this project have a guideline for adding generated file rules?


lib/linguist/generated.rb
```diff
@@ -82,7 +82,8 @@ def generated?
       generated_roxygen2? ||
       generated_jison? ||
       generated_yarn_lock? ||
-      generated_grpc_cpp?
+      generated_grpc_cpp? ||
+      generated_jest_snap?
```
pchaigno: Could you add it after `generated_by_zephir?` in the list? We prefer to place first the rules that don't require opening the file (to open it only if necessary).

lib/linguist/generated.rb
```diff
@@ -509,5 +510,12 @@ def generated_grpc_cpp?
       return false unless lines.count > 1
       return lines[0].start_with?("// Generated by the gRPC")
     end
+
+    # Internal: Is this a snap file of Jest?
+    #
+    # Return true or false
+    def generated_jest_snap?
+      name.include?("__snapshots__") && name.end_with?(".snap")
```
pchaigno: Could you switch the two predicates? I expect the extension to be a better (and cheaper) discriminator than the path.

yuya-takeyama:
@pchaigno 
Thanks for the quick reaction!
I addressed the 2 points, thanks.

lildude:
@pchaigno nudgy nudge nudge 😉

pchaigno:
> @lildude There are only [195 such files](https://github.com/search?p=8&q=extension%3Asnap+path%3A__snapshots__+NOT+kfjdkfdjf&type=Code&utf8=%E2%9C%93) (but well distributed: 1-2 per repository). Is that enough? The rules run for every single file, so we probably want to impose some adoption requirement here as well...

@lildude nudgy nudge nudge back :stuck_out_tongue_winking_eye:

lildude:
Ooops. Missed my own ping 😊

Looks like we're up to 249 files. Still not really popular enough. Tagging as such.

yuya-takeyama:
@pchaigno @lildude 
Looks like the search query is not correct.
`__snapshots__` directories can not be in the root of the project.

Found 64k files in this query.
https://github.com/search?utf8=%E2%9C%93&q=extension%3Asnap+Jest+Snapshot&type=Code

pchaigno:
Uh, you're right. I updated mine and found ~150K files matching. (Looks like `path` takes an absolute path whereas `in:path` doesn't.) Let's go for this then!

yuya-takeyama:
Thank you! 🎉 😂

lildude:
Oh 💩 I'm going to have to revert this PR. Sorry @yuya-takeyama.

We marked Jest snapshot files as generated before in https://github.com/github/linguist/pull/3572 and had to revert it in https://github.com/github/linguist/pull/3579 following feedback from the community. This PR has effectively re-introduced the first PR and I completely forgot and missed it.

We have https://github.com/github/linguist/pull/3584 open for discussion, though that discussion seems to have stalled.  Please feel free to wake it up by commenting in that PR.

If you want snapshots in your projects to be collapsed in diffs, please use the `linguist-generated` override.

pchaigno:
Arf! I forgot that and did not check :-/

PR #3880, opened 2017-10-28
Test Perl before Turing when running heuristics

Alhadis:
I'm rewriting the Perl 5 grammar from scratch, and I noticed [one of the test-files](https://github.com/Alhadis/language-perl5/blob/2315b8c5727cb4a64b5bf9f562eecd1c6d9f7191/t/6.2.pod.embeds.t) was being misclassified as Turing. The file in question contains numerous blocks of code from other languages in order to stress-test embedded highlighting. So the Turing heuristic probably matched a line of foreign code by mistake.

Perl supports here-docs, embedded documentation, and virtually every text-processing trick known to man. The odds of Turing's heuristic testing positive against embedded content is pretty high, so we simply test Turing last. Turing is pretty vanilla in terms of syntax, so there's nothing lto worry about. =)

__Before:__  

~~~console
λ language-perl5 (master): $LING/bin/linguist t/6.2.pod.embeds.t 
6.2.pod.embeds.t: 128 lines (84 sloc)
  type:      Text
  mime type: application/x-troff
  language:  Turing
~~~

__After:__

~~~console
λ language-perl5 (master): $LING/bin/linguist t/6.2.pod.embeds.t 
6.2.pod.embeds.t: 128 lines (84 sloc)
  type:      Text
  mime type: application/x-troff
  language:  Perl
~~~


tessarin:
I just noticed that a problem could happen here, similar to #3860. Since Perl 5 may contain `module` or `class`, a documented or Moose-style Perl 5 `.t` has a high risk of being considered Perl 6.

Can I suggest the order to be changed to _Perl, Perl 6, Touring_ instead?

After searching `heuristics.rb` for other similar cases, I notice the only other related test is already in the correct order (Perl 6 last):

https://github.com/github/linguist/blob/21babbceb1532793695ee903d7e5e4ec7c98ba39/lib/linguist/heuristics.rb#L350-L358

tessarin:
Thanks @Alhadis, now I think we've got the Perl 5/6 thing fully dealt with. Hopefully 😄

Alhadis:
This might be worth adding a unit test for. @tessarin, you seem to know much more about Moose-related stuff in Perl (I've never ventured that far out). If you could find me a nasty case of Perl 5/6 ambiguity that this PR's (hopefully) solved, I'll add it and include it with our regression tests. =)

tessarin:
> you seem to know much more about Moose-related stuff in Perl (I've never ventured that far out)

I highly doubt that...I'm just a musician

> I forgot that Perl 6 runs in strict mode by default, whereas Perl 5 has to opt in with the `use strict` pragma

Yes, and yes for code using `use strict` is likely never going to be Perl 6. However it is important to remember this is not the only way to enable strict. Version 5.12 made strict automatically enabled when you `use` the version (delta ref. [here](https://perldoc.perl.org/perl5120delta.html#Implicit-strictures)) and some modules such as Moose itself do enable strict by default.

Regarding Moose, it's not very complicated to understand —it's just a module designed to facilitate the OO features. And I'd say Moose alone may not cause problems with the current regex used in disambiguation. The problem begins with its extensions such as [MooseX::Declare](http://search.cpan.org/~ether/MooseX-Declare-0.43/lib/MooseX/Declare.pm). It enables the `class` keyword just like in Perl 6. It's all obviously inspired by it, so the ambiguity is natural. The module is considered deprecated in recent versions of Perl because of the easier internals to support a new keyword. There is the more recent [Moops](http://search.cpan.org/~tobyink/Moops-0.034/lib/Moops.pm) following the same idea. These modules allow code like this:

    use 5.027;
    use Moops;
    class MyClass { }

This is Perl 5 and would match as 6 due to the `/^class/` regex, had the order of tests not been swapped.

Disambiguating this all can be complicated. On top of that there is the whole POD thing. Virtually any file can contain POD with for example the words _class_ or _module_ at the beginning of a line (that's how I found this bug by the way, explaining about _"this↩︎module"_ in my POD).

Also, in my example above the first line is technically not necessary and if left out it would still be Perl 5, although very hard to distinguish unless Linguist starts parsing used modules. My conclusions:

 - If it has `^use\s+v?5` it's Perl 5
 - Similarly, `v6` for Perl 6
 - If it has `use strict` very likely to be Perl 5
 - If it has `^class` depends on the presence of MooseX::Declare, Moops, or others that may exist.

Searching GitHub for `language:"Perl 6" moosex declare` returns currently ~3k mismatched Perl 5 files, for instance.

Now, bringing this to the real world and closer to what generally may happen, the test swap should solve cases like my example above. Dealing with POD and these modules would require a different strategy.

I have a question about Linguist internals: if the heuristics test fails to disambiguate, how is the final decision made? If you check for example [briandfoy/cpan-packagedetails](https://github.com/briandfoy/cpan-packagedetails) all of the files considered Perl 6 are `.t` Perl 5 code. And if I understood the v6 regex properly, it should not match there.

https://github.com/github/linguist/blob/1f5ed3b3fe7c8f0ef377ca2e6b6dee8b321dda8b/lib/linguist/heuristics.rb#L464-L465

I'm sorry that I keep expanding on this subject...

Alhadis:
> I'm sorry that I keep expanding on this subject...

Well, you're helping to elevate my Perl knowledge past the level of "seasoned apprentice". ;-) The more Perl I learn, the more I love it and realise just how much there is to learn in this language.

> On top of that there is the whole POD thing

Right, yeah. Well, we could always rip out the Pod blocks with a regex or three, but as I learned recently, it's more complicated than a `s/^=.+^=cut//gms`:

~~~perl
=pod

Text, text text
=cut

Yet more pod text

=cut
~~~

And, of course, let's not forget embedded code samples:

~~~perl
=begin roff
=cut Ignored because =begin requires a blank line before its content.

.TH ATOM-ROFF 1
.  de XS \\n(.X \" etc
.  UR https://github.com/Alhadis/language-roff
=end roff
\" Still roff

=begin man \n[still-roff]

.TH ATOM-ROFF 5
.B Go install \fI`language-roff`\fP.
.ps -0.5n
.I (You\(cqll be glad you did)
=end man
\" Still roff

=end roff
~~~

:(

tessarin:
Perl is a wonderful language, very complete and with a high level of maturity in its development.

Yes, that’s a fantastic example to describe the problems of parsing such grammar with regex!

By the way, since you’re working on the grammar, here goes something for your leisure:  
https://github.com/Perl/perl5/blob/blead/perly.y

lildude:
@Alhadis do you plan to address https://github.com/github/linguist/issues/3906 as part of this PR? Just checking before rubber stamping this.

Alhadis:
Aye, I do! :-) I'll come back to this soon-ish; just trying to tie up loose ends here.

Alhadis:
@lildude Alrighty, taken care of. Have done the smart thing by using variables. ;)

/cc @zoffixznet

PR #3883, opened 2017-10-31
Improve heuristic for XML/TypeScript

pchaigno:
The heuristic for XML .ts files might match TypeScript generics
starting with TS.

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -477,7 +477,7 @@ module Linguist
     end
 
     disambiguate ".ts" do |data|
-      if data.include?("<TS")
+      if /<TS\b/.match(data)
         Language["XML"]
       else
         Language["TypeScript"]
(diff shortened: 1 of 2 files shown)


lib/linguist/heuristics.rb
```diff
@@ -477,7 +477,7 @@ def call(data)
     end
 
     disambiguate ".ts" do |data|
-      if data.include?("<TS")
+      if /<TS\b/.match(data)
```
Alhadis: Don't XML `.ts` files have a well-formed XML doctype we could check for...?

Do you have an example of an XML `.ts` file?

pchaigno: My understanding is that without the `TS` tag is doesn't work, whereas, without the `xml` tag or the doctype, it's *just* not valid: [examples without doctype](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ats+ts+AND+context+NOT+doctype&type=Code), [examples without `xml` tag](https://github.com/search?p=5&q=extension%3Ats+ts+AND+context+NOT+xml&type=Code&utf8=%E2%9C%93).

pchaigno:
@Alhadis Okay with you?

Alhadis:
I think so. I'm no means an expert on TypoScirpt, so LTGM. 👍

PR #3895, opened 2017-11-06
Add color for LFE language.

prertik:
LFE language had no color to display in the status bar! 
LFE, I guess shouldn't be grouped. `group: Erlang` should be removed. The language is different in its own aspect as Elixir is different to Erlang. LFE contains many language designs as of CL, Scheme, Erlang. So, it wouldn't be justified to group it as LFE.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2248,7 +2248,6 @@ LFE:
   color: "#4C3023"
   extensions:
   - ".lfe"
-  group: Erlang
   tm_scope: source.lisp
   ace_mode: lisp
   codemirror_mode: commonlisp
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2245,7 +2245,7 @@ Kotlin:
   language_id: 189
 LFE:
   type: programming
-  color: "#7F2119"
+  color: "#4C3023"
   extensions:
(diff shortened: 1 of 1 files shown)


prertik:
Oh no! Color was too near Ruby! Need to run a test!

pchaigno:
@oubiwann added support for LFE 4 years ago in #549.
@oubiwann - If you remember, was there any particular reason to add it to the Erlang group?

prertik:
@pchaigno I believe @oubiwann wanted to add it to Erlang group because LFE wasn't much more evolved. I believe it has evolved much more and it can be removed from the group so that it can be displayed in the repository when I write LFE code.
new

oubiwann:
There was no significant reason for keeping LFE in the Erlang group; mostly it reflected Robert Virding's aim of keeping LFE tightly aligned with Erlang.

I'm cool with it being in its own group, though.

prertik:
I can't help but wonder why this hasn't been merged?

lildude:
> I can't help but wonder why this hasn't been merged?

Because I've the final say as Linguist directly affects GitHub.com and I've been tied up in other work 😁. This won't take effect on GitHub until the next release is made and I can't say when they'll be.

prertik:
Thanks, @lildude. It was fast after the comment though.

PR #3896, opened 2017-11-08
Add C sample incorrectly recognized as Objective-C

yegortimoshenko:
See: https://github.com/yegortimoshenko/libsquashfs/search?l=objective-c

### DIFF
--- a/samples/C/squashfuse_dir.h
+++ b/samples/C/squashfuse_dir.h
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2012 Dave Vasilevsky <dave@vasilevsky.ca>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


yegortimoshenko:
It actually does fix a specific issue: code formatted via `llvm-format` is more likely to be classified as Objective-C. See: https://github.com/yegortimoshenko/linguist-3896

yegortimoshenko:
File with `#pragma once` is more likely to be classified as Objective-C. Compare:

https://github.com/yegortimoshenko/linguist-3896/blob/master/dir.h
https://github.com/yegortimoshenko/linguist-3896/blob/master/dir_pragma.h

Only the later is classified as Objective-C. Is there a way to fix this in Linguist? By the way: I was not correct regarding `llvm-format` being the problem, so I've deleted my previous comment.

PR #3899, opened 2017-11-09
Revert "Replace the tokenizer with a flex-based scanner"

kivikakk:
Reverts github/linguist#3846.

It went belly up! Currently looking at results from fuzzing but until it's fixed this isn't stable.

/cc @lildude

PR #3900, opened 2017-11-10
Lexer crash fix

kivikakk:
Even better than #3899, a straight-up fix.

* Ensure we handle `input` returning `0` as EOF so as to not trigger an extra `YY_INPUT` call into nothing.
* Stop using all trailing context; it's equivalent to `REJECT` and results in an unresizeable buffer.
* Remove the superstitious version number restraint.

/cc @philipturnbull who fuzzed this for me to help point the way.


kivikakk:
@lildude Thank you! 🙇‍♀️

ext/linguist/tokenizer.l
```diff
@@ -9,25 +9,25 @@
...
       } else if (c == q) \
         break; \
     } \
-    if (c == EOF) \
-      yyterminate(); \
+    if (c == EOF || !c) \
+      return 0; \
   } while (0)
 
 %}
```
Richardlucas123: - ###
Duplicate of #](url)`

PR #3871, opened 2017-10-20
Update Julia definitions to use Atom instead of TextMate

ararslan:
Currently linguist uses our TextMate plugin for its Julia definitions, which isn't well maintained and has more than its fair share of issues. The Atom plugin for Julia, [atom-language-julia](https://github.com/JuliaEditorSupport/atom-language-julia), is actively maintained and used. Thus we'd like to switch our linguist definitions to use atom-language-julia instead of Julia.tmbundle to ensure that things keeping looking good in the GitHub web UI.

Please let me know whether I've set this up correctly; I wasn't able to find explicit documentation in this repository regarding how to change the grammar that's used for an existing language.

PR #3903, opened 2017-11-13
Release v5.3.3

kivikakk:
* Improve heuristic for XML/TypeScript (#3883)
* Update Julia definitions to use Atom instead of TextMate (#3871)
* Lexer crash fix (#3900)

/cc @lildude


kivikakk:
@lildude :heart: !

PR #3907, opened 2017-11-14
Add BUILD.bazel to Python filenames

jin:
BUILD.bazel and BUILD are used by Bazel, and both are valid filenames. BUILD.bazel is used in favor of BUILD if it exists.

https://stackoverflow.com/a/43205770/521209

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3592,6 +3592,7 @@ Python:
   - ".gclient"
   - BUCK
   - BUILD
+  - BUILD.bazel
   - SConscript
   - SConstruct
   - Snakefile


lib/linguist/languages.yml
```diff
@@ -3592,6 +3592,7 @@ Python:
   - ".gclient"
   - BUCK
   - BUILD
+  - BUILD.bazel
   - SConscript
   - SConstruct
   - Snakefile
```
Alhadis: Should this be including `WORKSPACE.bazel` files too?

jin:
I signed it!

lib/linguist/languages.yml
jin: No need.

jin:
Thank you!

jin:
Syntax highlighting is not showing up - is there anything else that needs to be done?
 e.g. https://github.com/abseil/abseil-cpp/blob/master/absl/synchronization/BUILD.bazel

lildude:
> is there anything else that needs to be done?

Yup, a new release needs to be made and deployed to GitHub.com. PR was merged just after the last release. 

I'm holding off on a release at the moment as there are a few changes coming and the next release is going to be quite significant. We're also about to enter the Christmas period where we limit deploys, so unless there's a major bug that needs fixing, this isn't likely to be until early next year. 

/cc @kivikakk and @vmg as you've both submitted PRs recently and have the powers to produce releases 🙇

jin:
Thank you for the explanation!

kivikakk:
I'll start performing a release now, and if it gets reviewed before our freeze, it'll get deployed then. Will keep you updated here!

jin:
Hi, any idea why this BUILD.bazel file is not highlighted? https://github.com/angular/angular/blob/master/BUILD.bazel

lildude:
@jin yup, see my last comment to that last time you asked. 😉I plan to make a new release and upgrade GitHub.com by the end of the month.

jin:
Ah, I saw that the release issue was closed so I thought it's been pushed :) 

Thanks!

lildude:
@jin Linguist v6.0.0 is now live on GitHub.com. 🎉

jin:
Thanks @lildude!

PR #3921, opened 2017-11-28
Changed Carthage vendor regex to allow folder in any subdirectory

NachoSoto:
In monorepro projects, it's not uncommon for `Carthage` to not be in the root directory.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -240,7 +240,7 @@
 - \.imageset/
 
 # Carthage
-- ^Carthage/
+- (^|/)Carthage/
 
 # Sparkle
 - (^|/)Sparkle/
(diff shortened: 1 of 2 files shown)


lib/linguist/vendor.yml
```diff
@@ -240,7 +240,7 @@
 - \.imageset/
 
 # Carthage
-- ^Carthage/
+- (^Carthage|\/Carthage)\/
```
pchaigno: Could we use the shorter syntax `(^|/)Carthage/`?

NachoSoto: Good point, l'll change this.

NachoSoto: Changed, tests still passing.

lildude:
Thanks for the PR @NachoSoto and welcome to Linguist.

NachoSoto:
Thank you!

PR #3922, opened 2017-11-30
Limit tokens to 64 characters or less

kivikakk:
These don't add much to our tokenizer's accuracy and waste space duplicating large amounts of the samples in the sample JSON.

### DIFF
--- a/ext/linguist/linguist.c
+++ b/ext/linguist/linguist.c
@@ -2,6 +2,9 @@
 #include "linguist.h"
 #include "lex.linguist_yy.h"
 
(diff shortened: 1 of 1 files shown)


kivikakk:
You're right. I'm looking at some of the tokens as a result and there's are very few useful tokens that long. I'll bring it down to 32.

kivikakk:
Dropping the token limit to 32 characters shaved ~150kb from the samples.json (down to 3,598kb).

kivikakk:
Aaaaand I forgot to commit the drop to 32 characters. 🎉

PR #3923, opened 2017-11-30
Spelling

jsoref:
I don't actually expect this to be taken. (And yes, I'm aware this includes spelling fixes to the `Apollo Guidance Computer` -- it shows the authors were humans.) 

I'm going to slowly use these changes as indications of other projects to scan/patch (I'm slowly doing this already).

---
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

---
If anyone else wants to try these scripts out on some of the source repositories, they're welcome to do so :-)


Alhadis:
These samples aren't meant to be read by humans; they exist only for the Bayesian classifier. Therefore, this PR isn't really doing anything more than add diff-noise to Linguist's revision history.

> I'm going to slowly use these changes as indications of other projects to scan/patch (I'm slowly doing this already).

You should consider working off a fork instead.

pchaigno:
I agree with @Alhadis and am going to close this.

PR #3927, opened 2017-12-01
Add Javascript packet manager files

simsieg:


lildude:
Closing this PR as it's not needed. These files are already excluded from the language stats thanks to being marked as generated: 

https://github.com/github/linguist/blob/bee7e5561849055cd2f9c0a88c6b6b1a1b8e792b/lib/linguist/generated.rb#L355-L357

... and ...

https://github.com/github/linguist/blob/bee7e5561849055cd2f9c0a88c6b6b1a1b8e792b/lib/linguist/generated.rb#L507-L514

PR #3930, opened 2017-12-02
Handle repo cleanup race more elegantly

lildude:
Due to recent changes in the way repositories are cleaned up after deletion on GitHub.com, we can find ourselves in a race where the queued linguist job runs moments after the repository cleanup job has removed the repo from disk resulting in Linguist attempting to run on a non-existent directory.

This leads to the following exception:

```
git-linguist failed: fatal: Unable to read current working directory: No such file or directory
/data/github/releases/deploy-b/vendor/gems/2.4.3/ruby/2.4.0/gems/github-linguist-5.3.3/bin/git-linguist:122:in `pwd': No such file or directory - getcwd (Errno::ENOENT)
	from /data/github/releases/deploy-b/vendor/gems/2.4.3/ruby/2.4.0/gems/github-linguist-5.3.3/bin/git-linguist:122:in `git_linguist'
	from /data/github/releases/deploy-b/vendor/gems/2.4.3/ruby/2.4.0/gems/github-linguist-5.3.3/bin/git-linguist:146:in `<top (required)>'
	from /data/github/releases/deploy-b/bin/git-linguist:8:in `load'
	from /data/github/releases/deploy-b/bin/git-linguist:8:in `<main>'
```

Yes, ideally the job should be removed before it's run, but until then, this PR goes towards addressing this by a) not attempting to print the current directory in the event it goes missing, and b) printing the error to STDERR instead of raising another exception.

The lack of the directory in the error message reduces user-friendliness ever-so-slightly, but the removal of the exception improves it vastly.

Previously when run on a non-repo directory:

```
$ ~/github/linguist/bin/git-linguist stats
fatal: Not a git repository (or any of the parent directories): .git
/Users/lildude/github/linguist/bin/git-linguist:123:in `git_linguist': git-linguist must be run in a Git repository (/Users/lildude/Downloads/trash/foo.XX) (RuntimeError)
	from /Users/lildude/github/linguist/bin/git-linguist:150:in `<main>'
$
```

Now...

```
$ ~/github/linguist/bin/git-linguist
fatal: Not a git repository (or any of the parent directories): .git
git-linguist must be run in a Git repository
$
```

I did think of keeping the directory in the message and catching `Errno::ENOENT` but that would still have left a pretty 🤢  user-experience for peeps running on the CLI and would then have led to a RuntimeError we'd need to catch to keep our internal monitoring happy.  I like to think this approach is a happy compromise.

/cc @carlosmn & https://github.com/github/github/issues/81249

### DIFF
--- a/bin/git-linguist
+++ b/bin/git-linguist
@@ -117,36 +117,33 @@ def git_linguist(args)
   end
 
   parser.parse!(args)
+  git_dir = `git rev-parse --git-dir`.strip
+  raise "git-linguist must be run in a Git repository" unless $?.success?
+  wrapper = GitLinguist.new(git_dir, commit, incremental)
 
-  begin
-    git_dir = `git rev-parse --git-dir`.strip
-    raise "git-linguist must be run in a Git repository" unless $?.success?
-    wrapper = GitLinguist.new(git_dir, commit, incremental)
-
-    case args.pop
-    when "stats"
-      wrapper.linguist do |linguist|
-        puts JSON.dump(linguist.languages)
-      end
-    when "breakdown"
-      wrapper.linguist do |linguist|
-        puts JSON.dump(linguist.breakdown_by_file)
-      end
-    when "dump-cache"
-      puts JSON.dump(wrapper.load_language_stats)
-    when "clear"
-      wrapper.clear_language_stats
-    when "disable"
-      wrapper.disable_language_stats
-    else
-      $stderr.print(parser.help)
-      exit 1
+  case args.pop
+  when "stats"
+    wrapper.linguist do |linguist|
+      puts JSON.dump(linguist.languages)
     end
-  rescue Exception => e
-    $stderr.puts e.message
-    $stderr.puts e.backtrace
+  when "breakdown"
+    wrapper.linguist do |linguist|
+      puts JSON.dump(linguist.breakdown_by_file)
+    end
+  when "dump-cache"
+    puts JSON.dump(wrapper.load_language_stats)
+  when "clear"
+    wrapper.clear_language_stats
+  when "disable"
+    wrapper.disable_language_stats
+  else
+    $stderr.print(parser.help)
     exit 1
   end
+rescue Exception => e
+  $stderr.puts e.message
+  $stderr.puts e.backtrace
+  exit 1
 end
 
 git_linguist(ARGV)
--- a/bin/git-linguist
+++ b/bin/git-linguist
@@ -119,7 +119,10 @@ def git_linguist(args)
   parser.parse!(args)
 
   git_dir = `git rev-parse --git-dir`.strip
-  raise "git-linguist must be run in a Git repository" unless $?.success?
+  unless $?.success?
+    $stderr.puts "git-linguist must be run in a Git repository"
+    exit 1
+  end
   wrapper = GitLinguist.new(git_dir, commit, incremental)
 
   case args.pop
--- a/bin/git-linguist
+++ b/bin/git-linguist
@@ -118,30 +118,33 @@ def git_linguist(args)
 
   parser.parse!(args)
 
-  git_dir = `git rev-parse --git-dir`.strip
-  unless $?.success?
-    $stderr.puts "git-linguist must be run in a Git repository"
-    exit 1
-  end
-  wrapper = GitLinguist.new(git_dir, commit, incremental)
+  begin
+    git_dir = `git rev-parse --git-dir`.strip
+    raise "git-linguist must be run in a Git repository" unless $?.success?
+    wrapper = GitLinguist.new(git_dir, commit, incremental)
 
-  case args.pop
-  when "stats"
-    wrapper.linguist do |linguist|
-      puts JSON.dump(linguist.languages)
+    case args.pop
+    when "stats"
+      wrapper.linguist do |linguist|
+        puts JSON.dump(linguist.languages)
+      end
+    when "breakdown"
+      wrapper.linguist do |linguist|
+        puts JSON.dump(linguist.breakdown_by_file)
+      end
+    when "dump-cache"
+      puts JSON.dump(wrapper.load_language_stats)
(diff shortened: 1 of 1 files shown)


carlosmn:
This provides a somewhat nicer message, but not that it won't stop the flurry of exceptions. For that we'll have to change the GitRPC call to ignore the error if the target directory doesn't exist (which we've done for other calls previously).

lildude:
> This provides a somewhat nicer message, but not that it won't stop the flurry of exceptions.

Wouldn't these changes stop the `Errno::ENOENT` exceptions and any other that come directly from Linguist?

carlosmn:
Oh huh, these needles do look like they're coming out of linguist itself, rather than gitrpc raising when it exists with an error. No wonder there's a lot of them, each time this happens we run linguist six times concurrently.

I was thinking of https://haystack.githubapp.com/github/rollups/9cbdc9f4d37ff9c7dd4a31ca2fc06dd1 which is what the gitrpc system then generates (but it knows to simplify into a single exception).

Where is linguist set up to raise needles itself? Is that something we want to do for _any_ uncaught exception? Or should we trust gitrpc to propagate the errors to haystack?

lildude:
> Where is linguist set up to raise needles itself? Is that something we want to do for any uncaught exception? Or should we trust gitrpc to propagate the errors to haystack?

AFAIK Linguist doesn't raise needles itself at all. 

I'm not sure how the exceptions are being propagated up to haystack. Most of the exceptions I've seen have been via gitrpc or from github/github.

carlosmn:
The stack trace shows that this is happening inside linguist. GitRPC would (and does, like the rollup I linked above) throw a `GitRPC::Error` https://github.com/github/github/blob/00b559e2f28b60147c38e9f7703cae21ebf61f77/vendor/gitrpc/lib/gitrpc/backend/language_stats.rb#L32 It's a script we shell out to, so it has no way of telling GitRPC to raise anything else.

The `ENOENT` needles' stack trace looks like

```
endor/gems/2.4.3/ruby/2.4.0/gems/github-linguist-5.3.1/bin/git-linguist:122:in `pwd'
vendor/gems/2.4.3/ruby/2.4.0/gems/github-linguist-5.3.1/bin/git-linguist:122:in `git_linguist'
vendor/gems/2.4.3/ruby/2.4.0/gems/github-linguist-5.3.1/bin/git-linguist:146:in `<top (required)>'
bin/git-linguist:8:in `load'
bin/git-linguist:8:in `<main>'
```

which is not a stack trace that GitRPC would even know how to recreate. The stack traces from GitRPC mention all the code in ernicorn, and usually all the stack trace from the app or job as well.

We have some setup in `github/github` config/initialization which sets up the failbot `exit_hook`. If linguist indirectly loads that, it might be setting it up (but that's just a guess at this point).

Either way, these `ENOENT` needles could only come from inside the `git-linguist` script, as there is no way to indicate to GitRPC that it should generate that needle. Inadvertent loading of github/github magic would be my bet, but I don't really know where they would actually be generated.

carlosmn:
I guess I posted too quickly. I just ran `strace -f git linguist` and I can see the following interesting output:

```
[pid 47564] open("/data/github/releases/deploy-b/lib/github/config/failbot.rb", O_RDONLY|O_NONBLOCK|O_CLOEXEC) = 11
...
[pid 47564] open("/data/github/releases/deploy-b/vendor/gems/2.4.3/ruby/2.4.0/gems/failbot-2.0.1/lib/failbot/exit_hook.rb", O_RDONLY|O_NONBLOCK|O_CLOEXEC) = 11
```

which is how github/github sets itself up to report uncaught exceptions. So yeah, these are definitely coming from the `git-linguist` script.

Which still means that this change is going to shut them up by not bubbling up this uncaught exception. But we might still throw others. We can punt on that, but maybe the `git_linguist` method could `rescue` anything, print it out and `exit 1` so this kind of thing doesn't happen in the future.

lildude:
> But we might still throw others. We can punt on that, but maybe the git_linguist method could rescue anything, print it out and exit 1 so this kind of thing doesn't happen in the future.

I did think about going that route, but wasn't sure this was "best practice".

lildude:
> print it out

Do you mean just the single error or the full stack too?

carlosmn:
> but wasn't sure this was "best practice".

I think i'ts fine because we're catching an exception and erroring out in our main function. Anything calling this script needs to react to an error exit code anyway.

>Do you mean just the single error or the full stack too?

Either works, but I think I would print out stack trace too, so we know more exactly what went wrong. Otherwise we'll just have the stack trace from us calling it over RPC and that just shows uninteresting code.

bin/git-linguist
```diff
@@ -118,27 +118,33 @@ def git_linguist(args)
...
-  when "clear"
-    wrapper.clear_language_stats
-  when "disable"
-    wrapper.disable_language_stats
-  else
-    $stderr.print(parser.help)
+  rescue Exception => e
+    $stderr.puts e.message
+    $stderr.puts e.backtrace
     exit 1
```
lildude: Something like this @carlosmn?

carlosmn: Yeah. I would have just put the `rescue` a the bottom since a method implicitly starts a `begin` but this works just as well.

lildude: Doh! Of course. I like that idea better. Coming up.

bin/git-linguist
```diff
@@ -117,9 +117,8 @@ def git_linguist(args)
   end
 
   parser.parse!(args)
-
   git_dir = `git rev-parse --git-dir`.strip
-  raise "git-linguist must be run in a Git repository (#{Dir.pwd})" unless $?.success?
+  raise "git-linguist must be run in a Git repository" unless $?.success?
```
quiterodroo: Awesome

quiterodroo: New

bin/git-linguist
lildude: Done!

PR #3937, opened 2017-12-05
languages: add *.BUILD support for Bazel (Python)

adamyi:
Many projects compiled with [Bazel](https://bazel.build) utilizes not only BUILD file but also *.BUILD files. They should be parsed as [Python](https://docs.bazel.build/versions/master/build-ref.html#core_build_language).

See examples:
* https://github.com/tensorflow/tensorflow/tree/master/third_party
* https://github.com/google/protobuf
* https://github.com/bazelbuild/rules_closure/tree/master/closure/templates

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3592,6 +3592,7 @@ Python:
   color: "#3572A5"
   extensions:
   - ".py"
+  - ".BUILD"
   - ".bzl"
   - ".cgi"
   - ".fcgi"
@@ -3609,7 +3610,6 @@ Python:
   - ".tac"
   - ".wsgi"
   - ".xpy"
-  - ".BUILD"
   filenames:
   - ".gclient"
   - BUCK
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3609,6 +3609,7 @@ Python:
   - ".tac"
   - ".wsgi"
   - ".xpy"
+  - ".BUILD"
   filenames:
   - ".gclient"
   - BUCK
--- a/samples/Python/sqlite.BUILD
+++ b/samples/Python/sqlite.BUILD
@@ -1,0 +1,16 @@
+# Description:
+#   Sqlite3 library. Provides utilities for interacting
+#   with sqlite3 databases.
+
+licenses(["unencumbered"])  # Public Domain
+
+# exports_files(["LICENSE"])
+
+cc_library(
+    name = "sqlite",
+    srcs = ["sqlite3.c"],
+    hdrs = ["sqlite3.h"],
+    includes = ["."],
+    linkopts = ["-lm"],
+    visibility = ["//visibility:public"],
+)


adamyi:
This was raised before in #2929 and #2930

adamyi:
@pchaigno I just found the previous PRs. Regarding the fact that linguist's extensions are case-insensitive. I think we should be case-insensitive if the defined extension in yml file contains only small-case letters (just in case developers use strange capitalization e.g. .Cpp). However, if the defined extension contains capitalized letters, they should be treated case-sensitively. This is similar to the smart case mode for vim search and replace. I would be happy to work on this if you agree.

pchaigno:
> However, if the defined extension contains capitalized letters, they should be treated case-sensitively. This is similar to the smart case mode for vim search and replace. I would be happy to work on this if you agree.

That might be a sensible way to approach case-sensitive file extensions. @Alhadis @lildude What do you think? Not sure how many such extensions we have though.

lildude:
> However, if the defined extension contains capitalized letters, they should be treated case-sensitively. 

> @lildude What do you think?

This is a nice idea, but it's gonna require changes on the GitHub side of things as search is case-insensitive, so `extension:BUILD NOT hack` will return exactly the same results as `extension:build NOT hack`. This makes it hard for us (the Linguist community) to quantify usage but more importantly, the links in the sidebar for each of the languages on the search results pages won't take into account the capitalised results.

It's also not a simple matter of flipping GitHub.com to be case-sensitive for extensions as there is a lot of search indexing involved as the extensions are indexed in lowercase based off a decision to make all searches case-insensitive.

I was going to suggest treating the `.build` extension case-insensitively, however this is a really really popular extension with a lot of different uses, so will require quite a bit more work and adding support for all the language-specific uses as mentioned by @pchaigno at https://github.com/github/linguist/pull/2930#issuecomment-204657415

As a side note, the current implementation came off feedback from you @pchaigno and others around 3.5 years ago when searches for capitalised extensions returned no results - you opened https://github.com/github/linguist/pull/2087 at the same time 😄.

pchaigno:
> As a side note, the current implementation came off feedback from you @pchaigno and others around 3.5 years ago when searches for capitalised extensions returned no results - you opened https://github.com/github/linguist/pull/2087 at the same time 😄.

Yes, it was a mess at the time. We had no way to tell one extension is case-insensitive while another is case-sensitive. We often had to duplicate extensions to have both a lower-case and an upper-case extension. @adamyi's suggestion looks to me like a sensible way to solve this; we would be able to tell an extension is case-sensitive while most remain case-insensitive.

> the links in the sidebar for each of the languages on the search results pages won't take into account the capitalised results.

How do you mean? The links in the side-bar are equivalent to a `language:the_language_you_clicked` search query, right? They currently display files with both lower-case and upper-case extensions: https://github.com/github/linguist/search?q=language%3Ac&unscoped_q=language%3Ac.

> This makes it hard for us (the Linguist community) to quantify usage but more importantly

I agree it might make it a bit harder for us to quantify usage, although [some cases](https://github.com/search?q=extension%3Abuild+licenses&type=Code) are fairly easy. It's probably more of an issue to check that one case-sensitive extension doesn't include other languages.

lildude:
> How do you mean? The links in the side-bar are equivalent to a language:the_language_you_clicked search query, right?

Sorry, I wasn't very clear.  I was trying to convey that peeps will expect the results to return _only_ case-sensitive results if they used a case-sensitive extension.

[`extension:BUILD NOT hack`](https://github.com/search?q=extension%3ABUILD+NOT+hack) produces sidebar language links like https://github.com/search?l=Python&q=extension%3ABUILD+NOT+hack&type=Code which is case-insensitive, and will always be as long as GitHub implements case-insensitive extension searches.

pchaigno:
> peeps will expect the results to return only case-sensitive results if they used a case-sensitive extension.

But that's already true, no? Someone searching for [`extension:BUILD licenses`](https://github.com/search?q=extension%3ABUILD+licenses&type=Code) could think the search engine is case-sensitive. I don't think support for case-sensitive extensions on Linguist would change that.

lildude:
> Someone searching for extension:BUILD licenses could think the search engine is case-sensitive.

Now that's a good search to prove your point 😁

> I don't think support for case-sensitive extensions on Linguist would change that.

I'm still apprehensive - search is a PITA for us and I fear we may be giving ourselves more explaining to do in future - but lets give it a go. Lets see what happens.

pchaigno:
> I'm still apprehensive - search is a PITA for us and I fear we may be giving ourselves more explaining to do in future - but lets give it a go. Lets see what happens.

I'm also a bit wary because of the risk to create a bit of an unclear situation for users. The plus side being that as long as it only requires changes in Linguist, it'll be easy to revert if necessary.

> I would be happy to work on this if you agree.

@adamyi Do you still have time for that?

PR #3946, opened 2017-12-12
Do not traverse symlinks in heuristics

kivikakk:
This is part of what should be a correct fix for https://github.com/github/markup/pull/1133. Changes are required in `markup` itself and dotcom to pass the symlink info through from the repository.

### DIFF
--- a/lib/linguist/file_blob.rb
+++ b/lib/linguist/file_blob.rb
@@ -26,6 +26,11 @@ module Linguist
       @mode ||= File.stat(@fullpath).mode.to_s(8)
     end
 
+    def symlink?
+      return @symlink if !@symlink.nil?
+      @symlink = (File.symlink?(@fullpath) rescue false)
+    end
+
     # Public: Read file contents.
     #
     # Returns a String.
--- a/lib/linguist/blob.rb
+++ b/lib/linguist/blob.rb
@@ -11,11 +11,13 @@ module Linguist
     #
     # path    - A path String (does not necessarily exists on the file system).
     # content - Content of the file.
+    # symlink - Whether the file is a symlink.
     #
     # Returns a Blob.
-    def initialize(path, content)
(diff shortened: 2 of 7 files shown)


lib/linguist/file_blob.rb
```diff
@@ -26,6 +26,11 @@ def mode
       @mode ||= File.stat(@fullpath).mode.to_s(8)
     end
 
+    def symlink?
+      return @symlink if !@symlink.nil?
+      @symlink = (File.symlink?(@fullpath) rescue false)
+    end
```
lildude: @kivikakk I just noticed this is reporting a bunch of warnings in the tests: https://travis-ci.org/github/linguist/jobs/323823193

```
/home/travis/build/github/linguist/lib/linguist/file_blob.rb:30: warning: instance variable @symlink not initialized
/home/travis/build/github/linguist/lib/linguist/file_blob.rb:30: warning: instance variable @symlink not initialized
/home/travis/build/github/linguist/lib/linguist/file_blob.rb:30: warning: instance variable @symlink not initialized
/home/travis/build/github/linguist/lib/linguist/file_blob.rb:30: warning: instance variable @symlink not initialized
```

They're not failures hence we missed this before, but I think it'd be best to clean these up.

lildude: I've also just realised this PR didn't intro any tests to verify the symlink behaviour. Sorry I missed this before.

kivikakk: No, my bad. I've cleaned up the uninitialised instance var messages in https://github.com/github/linguist/commit/16732843249d434969fbe5d21580a73fd4a73acf. Will add some tests in that PR.

kivikakk: Also a test in https://github.com/github/linguist/pull/3948/commits/3eabf455a2f259e71669a4073896807e4d2ded94.

PR #3992, opened 2018-01-18
script: Add build-grammars-tarball

vmg:
This is a very simple scripts that packages the compiled grammars so we can push a tarball for them with each Linguist release. That should make updating grammars in the website much more straightforward.

cc @github/linguist

### DIFF
--- a/script/build-grammars-tarball
+++ b/script/build-grammars-tarball
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -e
+cd "$(dirname "$0")/.."
+rm -rf ./linguist-grammars
+./script/grammar-compiler compile -o linguist-grammars || true
+tar -zcvf linguist-grammars.tar.gz linguist-grammars
+
+


vmg:
@Alhadis :bow:

@kivikakk: Heyyy, would you have a moment to tag a new Linguist release, so I can try to deploy it using the new Prettylights? Alternatively, do we have docs on the release process?

lildude:
@vmg I'm planning a new release :soon: but it's going to be a big one and will require bumping escape_tools on github/github at the same time.

If you want to test a release yourself, you can build the gem using the details at https://github.com/github/linguist/blob/master/CONTRIBUTING.md#releasing (these are quite rough but I'm sure you'll be able to work it out).

You'll want to give Linguist a version based on the git SHA so we don't hit any caching issues before building the gem.

Once you've got the gem, you need to manually put it in place in a github/github branch and update the Gemlock etc as vendor-gems doesn't play nice with Linguist.

PR #4017, opened 2018-01-30
.h file is now not recognized as C++

KrzysztofSzewczyk:
.h file is NOT c++ header file extension, C++ header files extensions are .hpp, .hh, .hxx and .h++.
.h IS C header file extension. This affected my repository and will help in language recognization
of many repositories

## Description
This PR is removing .h from C++ file extensions.

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):


lildude:
> .h file is NOT c++ header file extension

Not quite... generally speaking `.h` is used for C/C++ compatible or pure C headers. We attribute it to both for precisely this reason. 

What's more, `.h` is certainly way more prevalent than the other extensions from what I've seen, and it appears to be suggested use too:

From http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-file-suffix

> SF.1: Use a .cpp suffix for code files and .h for interface files if your project doesn’t already follow another convention
Reason
> 
> It’s a longstanding convention. But consistency is more important, so if your project uses something else, follow that. Note
> 
> This convention reflects a common use pattern: Headers are more often shared with C to compile as both C++ and C, which typically uses .h, and it’s easier to name all headers .h instead of having different extensions for just those headers that are intended to be shared with C. On the other hand, implementation files are rarely shared with C and so should typically be distinguished from .c files, so it’s normally best to name all C++ implementation files something else (such as .cpp).
> 
> The specific names .h and .cpp are not required (just recommended as a default) and other names are in widespread use. Examples are .hh, .C, and .cxx. Use such names equivalently. In this document, we refer to .h and .cpp as a shorthand for header and implementation files, even though the actual extension may be different.

Sadly, this sharing of an extension with no easy way to distinguish it is a common problem we encounter, but sadly whipping out the `.h` extension from C++ is not the right solution.  It certainly _might_ become the right solution if C++ ever decides to enforce the use of another extension, but I don't see that happening given the age of the language and prevalence of `.h` files already.

You may want to track and help out in https://github.com/github/linguist/pull/3585 in which we're slowly attempting to improve the heuristic to better differentiate the two. It won't be perfect once implemented, but should hopefully be better.

In the mean time, for your own repo, may I recommend you implement an override to force the language of your choice.

That all said, I'm closing this PR as it is not the correct solution to the C/C++ `.h` language identification issue.

PR #4025, opened 2018-02-07
Added C regex to prevent ambiguity in header file

jweinst1:
Adding a C regex for disambiguation of `.h` files

## Description

Basically, my C header file is getting incorrectly classified as Objective-C. While overrides can fix this, it would be good to have a regex detecting C-only standard header includes. 

https://github.com/jweinst1/DataBlock/blob/master/src/DataBlock.h

That is the url to the incorrectly categorized file.

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):

    - Sample source(s): https://github.com/jweinst1/DataBlock/blob/master/src/DataBlock.h
    - Sample license(s):

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -76,8 +76,8 @@ module Linguist
 
     # Common heuristics
     CRegex = Regexp.union(
-        /^(char|double|short|int|long|float|unsigned|void)*\w\s*\w\s*\(\s*void\s*\)/,
-        /^(\w|\s|\*)*restrict/,
+        /^(char|double|short|int|long|float|unsigned|void)\w+\s*\(\s*void\s*\)/,
+        /\([\s\w\*]+restrict\s+/,
         /^(char|short|int|long|float|unsigned)\s+\w+\s*\[\]\s*;/
       )
     CPlusPlusRegex = Regexp.union(
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -75,6 +75,7 @@ module Linguist
     end
 
     # Common heuristics
+    CRegex = /^\s*#\s*include <(stdlib|stdio|stdarg|stdbool|stdint|stddef|setjmp|errno)\.h>/
     CPlusPlusRegex = Regexp.union(
         /^\s*#\s*include <(cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream)>/,
         /^\s*template\s*</,
@@ -231,7 +232,9 @@ module Linguist
     end
 
     disambiguate ".h" do |data|
-      if ObjectiveCRegex.match(data)
+      if CRegex.match(data)
+        Language["C"]
+      elsif ObjectiveCRegex.match(data)
         Language["Objective-C"]
       elsif CPlusPlusRegex.match(data)
         Language["C++"]
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -76,9 +76,9 @@ module Linguist
 
     # Common heuristics
     CRegex = Regexp.union(
-        /^(char|double|short|int|long|float|unsigned|void)\w+\s*\(\s*void\s*\)/,
+        /\b(char|double|short|int|long|float|unsigned|void)\s+\w+\s*\(\s*void\s*\)/,
         /\([\s\w\*]+restrict\s+/,
-        /^(char|short|int|long|float|unsigned)\s+\w+\s*\[\]\s*;/
+        /\b(char|short|int|long|float|unsigned)\s+\w+\s*\[\]\s*;/
       )
     CPlusPlusRegex = Regexp.union(
         /^\s*#\s*include <(cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream)>/,
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -78,7 +78,7 @@ module Linguist
     CRegex = Regexp.union(
         /^(char|double|short|int|long|float|unsigned|void)*\w\s*\w\s*\(\s*void\s*\)/,
         /^(\w|\s|\*)*restrict/,
-        /^(char|short|int|long|float|unsigned)*\s+\w+\s*\[\]\s*;/
+        /^(char|short|int|long|float|unsigned)\s+\w+\s*\[\]\s*;/
       )
     CPlusPlusRegex = Regexp.union(
         /^\s*#\s*include <(cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream)>/,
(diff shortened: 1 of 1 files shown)


jweinst1:
Yea, I was confused since I thought that you had to do `#include <cstdio>` in C++ but I was mistaken. Ill change this with some truly C only features

jweinst1:
So, following @Elronnd, suggestion on #1626, I changed the C regex to detect `(void)` signified functions, 

Additionally since the [restrict](https://en.wikipedia.org/wiki/Restrict) is only valid in C, I added that to the regex, in addition to [flexible array parameters](https://en.wikipedia.org/wiki/Flexible_array_member)

lib/linguist/heuristics.rb
```diff
@@ -75,6 +75,11 @@ def call(data)
     end
 
     # Common heuristics
+    CRegex = Regexp.union(
+        /^(char|double|short|int|long|float|unsigned|void)*\w\s*\w\s*\(\s*void\s*\)/,
+        /^(\w|\s|\*)*restrict/,
+        /^(char|short|int|long|float|unsigned)*\s+\w+\s*\[\]\s*;/
```
pchaigno: Why the asterisk after `(char|short|int|long|float|unsigned)`?

jweinst1: good catch

jweinst1:
@pchaigno let me know if this needs any more changes :)

lib/linguist/heuristics.rb
pchaigno: I think there's still the same error with the two above regular expressions, or am I mistaken?

jweinst1: No, because u need a return type for any void function, like ` foo(void)` with no return type is not valid.

for the second one i think it's optional to have a word character or a `*` pointer type, since I want to match just the restrict keyword.

pchaigno: > No, because u need a return type for any void function, like ` foo(void)` with no return type is not valid.

Sure, but currently, you can have several return types because of the asterisk. For example, `chard d(void)` matches your regular expression. Also, I think you should require a space after the type and probably meant to write `\w+` instead of just `\w`. Not sure why you have `\w\s*` twice after the type.

> for the second one i think it's optional to have a word character or a `*` pointer type, since I want to match just the restrict keyword.

Could you give an example of C construction you're trying to match with the second regular expression?

jweinst1: > Sure, but currently, you can have several return types because of the asterisk. For example, doublechard d(void) matches your regular expression. Also, I think you should require a space after the type and probably meant to write \w+ instead of just \w. Not sure why you have \w\s* twice after the type.

Alright, will fix this.

> Could you give an example of C construction you're trying to match with the second regular expression?

Would be:

```c
void copy(void* restrict dst, void* restrict src){...}
```

The restrict keyword is always used with pointers, but most always used in function parameters.

pchaigno: Maybe `\([\s\w\*]+restrict\s+` for the second expression then?

jweinst1:
alright, made your most recent revisions.

jweinst1:
@pchaigno or @lildude , let me know if this PR needs more revisions :)

jweinst1:
@pchaigno feel free to let me know if i need to make any more changes on this.

lib/linguist/heuristics.rb
```diff
@@ -75,6 +75,11 @@ def call(data)
     end
 
     # Common heuristics
+    CRegex = Regexp.union(
+        /^(char|double|short|int|long|float|unsigned|void)\w+\s*\(\s*void\s*\)/,
+        /\([\s\w\*]+restrict\s+/,
+        /^(char|short|int|long|float|unsigned)\s+\w+\s*\[\]\s*;/
```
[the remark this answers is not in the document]
jweinst1: @pchaigno , yes, for the first and the third, the word boundary is more helpful as it will also match common c-types like `unsigned char`, `unsigned int`, `long long` or `unsigned long` made the changes

lildude:
As this is the most recently active of attempts to resolve this, I'm going to reference https://github.com/github/linguist/pull/3585 as work that has also started to try and address this. Maybe we can use ideas from both that PR and this to get a good solution in place.

jweinst1:
@pchaigno is this ready to merge? it says only those with write access can merge :(

pchaigno:
> @pchaigno is this ready to merge?

No. We still need to wait for the :+1: from someone from GitHub's staff.

lib/linguist/heuristics.rb
```diff
@@ -75,6 +75,11 @@ def call(data)
     end
 
     # Common heuristics
+    CRegex = Regexp.union(
+        /\b(char|double|short|int|long|float|unsigned|void)\s+\w+\s*\(\s*void\s*\)/,
+        /\([\s\w\*]+restrict\s+/,
+        /\b(char|short|int|long|float|unsigned)\s+\w+\s*\[\]\s*;/
```
pchaigno: What makes you think this last heuristic rule is only valid in C?

jweinst1: @pchaigno Flexible struct parameters are not valid C++, see https://stackoverflow.com/a/4413035/5129504

pchaigno: `.h` is also [an Objective-C extension](https://github.com/github/linguist/blob/5363e045bb2728eec05ff2225a87e742e0bff586/lib/linguist/languages.yml#L3067).

lildude: … and precisely why the tests are currently failing, suggesting further refinement is needed.

jweinst1:
Alright so i fixed the test failure there, i think the issue was there is a seldom used keyword in Objective-C called `__restrict`, so i removed the underscores from the regex.

jweinst1:
@pchaigno or @lildude , if a keyword is removed from a newer version of a language, does this count as a linguist standard as not being part of the language? 

The [register keyword is removed in C++17](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4340) and has been deprecated for a while before that, yet still exists in C. Can linguist consider register as a C only regex?

lildude:
> @pchaigno or @lildude , if a keyword is removed from a newer version of a language, does this count as a linguist standard as not being part of the language?
>
> The register keyword is removed in C++17 and has been deprecated for a while before that, yet still exists in C. Can linguist consider register as a C only regex?

I've just test the spec change you've mentioned and I don't think we can:

> The purpose in removing the usage of register is to preserve the ability to use this valuable keyword for future proposals, such as modules. It is therefore retained as a reserved keyword, with a note on intent, similar to the handling of export in C++11. All other references to the register keyword are removed.

As I understand this, it's no longer used but is still reserved for future use so we'd probably end up having to reverse the change in future if this is resurrected. Probably best to leave this keyword alone for the moment.

jweinst1:
@lildude understood, any other changes you suggest making to this PR? since the tests are passing now

jweinst1:
@pchaigno , @lildude , any other changes need to be made here?

pchaigno:
@jweinst1 See my above review :wink: 

> You'll also need to update the tests in `tests/test_heuristics.rb`.

lildude:
Nudge.

pchaigno:
Flagging as stale.

PR #4026, opened 2018-02-08
Treat "python3" as an alias for "python"

njsmith:
Pygments has separate highlighters for "python" (meaning python 2) and "python3" (meaning python 3). As a result, there are lots of files out there (especially ReSTructured text) that contain code blocks whose language is explicitly given as "python3". Currently these are unrecognized by linguist. Instead, we should use our python highlighter for them (which works for both python 2 and python 3).

References:
  http://pygments.org/docs/lexers/#pygments.lexers.python.Python3Lexer
  https://github.com/github/markup/issues/1019
  https://github.com/python-trio/async_generator/pull/12

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3632,6 +3632,7 @@ Python:
   - python3
   aliases:
   - rusthon
+  - python3
   language_id: 303
 Python console:
   type: programming


njsmith:
Pygments doesn't currently recognize "python2" as a lexer name, so no, I've never seen it :-).

If we want to be super-thorough, I guess the following aliases could also be added (reading from that pygments.org link up above):

* `numpy`, `py3`, `py`, `sage` → "Python language"
* `py3tb` → "Python traceback"

I've personally never seen any of these in use, though.

pchaigno:
Let's keep the `python3` alias only for now. We can always add new aliases if/when we see them being used.

njsmith:
:+1:

njsmith:
Thanks!

PR #4029, opened 2018-02-12
add CoNLL-U format

odanoburu:
I'd like to add support for the [CoNLL-U data format](universaldependencies.org/format.html).

## Description
the CoNLL-U format is used to annotate natural language, being widely used among the NLP community.

## Checklist:
      > as CoNLL-U is a data format, Github doesn't seem to search for its extension, however:
      -  we can search for `conllu` and `conll` and see code that references conllu files or gitignore files that include them: https://github.com/search?q=conllu&type=Code&utf8=✓ https://github.com/search?utf8=✓&q=conll&type=Text (of course there is a big intersection between these two results)
      - we can search for `universal dependencies` or `UD`, which is the organization that standardized the format: https://github.com/search?utf8=✓&q=universal+dependencies+UD&type=Code
      - in fact, as CoNLL-U is a data format, most people will not include them on their repos, as they can be generated by parsing raw text. **however**, there is an important exception in all this: the CoNLL-U datasets that are used for training parsers are version controlled, because they are manually revised by humans. most of these are under the [UD organization page](https://github.com/UniversalDependencies/) -- and there seem to be more than 100 of them only under the UD page. 
    - Sample source(s):
      - en-ud-test.conllu: https://github.com/UniversalDependencies/UD_English/blob/master/en-ud-test.conllu
      - ug-ud-test.conllu: https://github.com/UniversalDependencies/UD_Uyghur/blob/master/ug-ud-test.conllu
      - CF1.conllu: https://github.com/own-pt/bosque-UD
    - Sample license(s): CC BY-SA 4.0
  - [x] I have included a syntax highlighting grammar.
  - [N/A] I have included a change to the heuristics to distinguish my language from others using the same extension.


odanoburu:
the `bundle exec rake test` doesn't seem to check for the absence of the license file in `vendor/licenses/`, which is what triggered the failed builds.

samples/CoNLL-U/ug-ud-test.conllu
```diff
@@ -0,0 +1,13030 @@
+# sent_id = s1
```
pchaigno: @lildude I'm guessing the size of these sample files will be an issue?

.gitmodules
```diff
@@ -898,3 +898,6 @@
 [submodule "vendor/grammars/atom-language-nextflow"]
 	path = vendor/grammars/atom-language-nextflow
 	url = https://github.com/nextflow-io/atom-language-nextflow
+[submodule "vendor/grammars/conllu-linguist-grammar"]
+	path = vendor/grammars/conllu-linguist-grammar
+	url = https://github.com/odanoburu/conllu-linguist-grammar
```
pchaigno: I'm actually impressed this works with Linguist, but nevertheless, you might want to change the extension of your grammar file to `.json` ;)

lildude: Yup, it's definitely best to change this to `.json` just to be safe.

samples/CoNLL-U/ug-ud-test.conllu
lildude: Oh my word!! Yes, definitely way too big. Chrome is also struggling with this diff expanded.

Please replace the samples with smaller samples that give a good representation of the structure of the language whilst still being real-world samples  (as opposed to written specifically for Linguist or a "hello world" equivalent).

odanoburu: @pchaigno @lildude can I simple cut the file at some length? what's the ideal size?

.gitmodules
odanoburu: will do!

samples/CoNLL-U/ug-ud-test.conllu
lildude: > @pchaigno @lildude can I simple cut the file at some length? 

Looking at the content, it does appear there is a fair amount of repetition so you could indeed cut the file down to only be a single highly representative "stanza".

> what's the ideal size?

Less than about 20KB so it doesn't bump into GitHub.com's [diff limits](https://help.github.com/articles/limits-for-viewing-content-and-diffs-in-a-repository/) is normally a good place to aim for, but we can accept slightly larger if it's absolutely necessary. 

The Farsi/Persian sample (Google thinks that's the language 😉 )  will be larger due to the high number of multi-byte characters so switching to English or other predominantly single-byte charactered language will also bring down the size.

odanoburu:
I've abridged the samples (although I added a new one for more linguistic diversity), readded the secondary extensions `.conll` and `.conllx`, and changed the grammar file extension to `.json`.

thanks for the feedback!

lib/linguist/languages.yml
```diff
@@ -730,6 +730,18 @@ Closure Templates:
   - ".soy"
   tm_scope: text.html.soy
   language_id: 357046146
+CoNLL-U:
+  type: data
+  extensions:
+  - ".conllu"
+  - ".conll"
+  - ".conllx"
```
pchaigno: I'm seeing this only in my second review, but it looks like all `.conllx` files are [from a single user](https://github.com/search?p=1&q=extension%3Aconllx+NOT+djfhdjfgnd&type=Code&utf8=%E2%9C%93). I'm afraid you'll have to remove it as it doesn't classify as widespread (in use in hundreds of repository).

odanoburu: fixed in 2aaee98! 

indeed, I hadn't noticed that! the CoNLL-X format is actually a very similar format standard that has been superseded by the CoNLL-U format, so it seems that no one is using it anymore !

odanoburu:
hello @lildude , are the changes satisfactory?

lildude:
> hello @lildude , are the changes satisfactory?

Ooops. Yes. Been a tad busy and completely forgot. One mo.

odanoburu:
@lildude no problem! thanks :smile:

PR #4031, opened 2018-02-12
Add Dart files generated by agilord/owl and dart-lang/appengine

chances:
Add Dart files generated by [agilord/owl](https://github.com/agilord/owl) and [dart-lang/appengine](https://github.com/dart-lang/appengine) so they aren't included in PR diffs.

## Checklist:


lib/linguist/generated.rb
```diff
@@ -525,5 +526,18 @@ def generated_grpc_cpp?
...
+    #
+    # An own generated file contains:
+    # // GENERATED CODE - DO NOT MODIFY
+    # on the first line.
+    #
+    # Return true or false
+    def generated_dart?
+      return false unless extname == '.dart'
+      return false unless lines.count > 1
+      return lines[0].include?("GENERATED CODE - DO NOT MODIFY")
```
pchaigno: Looks like `//  Generated code. Do not modify.` [is common too](https://github.com/search?p=1&q=extension%3Adart+%22GENERATED+CODE%22+%22DO+NOT+MODIFY%22&type=Code&utf8=%E2%9C%93). Maybe add a second case for it?

lib/linguist/generated.rb
```diff
@@ -537,7 +537,9 @@ def generated_grpc_cpp?
     def generated_dart?
       return false unless extname == '.dart'
       return false unless lines.count > 1
-      return lines[0].include?("GENERATED CODE - DO NOT MODIFY")
+      genStatement = lines[0].include?("Generated code. Do not modify.")
+      getStatementCaps = lines[1].include?("GENERATED CODE - DO NOT MODIFY")
```
pchaigno: What about this?
```ruby
return true if lines[0].include?("Generated code. Do not modify.")
return lines[1].include?("GENERATED CODE - DO NOT MODIFY")
```
If we're going to run this on all `.dart` files of more than 1 line, we might as well make it as fast as we can.

lildude:
Nudge.

lib/linguist/generated.rb
```diff
@@ -525,5 +526,19 @@ def generated_grpc_cpp?
       return false unless lines.count > 1
       return lines[0].start_with?("// Generated by the gRPC")
     end
+    
+    # Internal: Is this an owl-generated Dart file?
+    #
+    # An own generated file contains:
```
lildude: `s/own/owl/`?

lib/linguist/generated.rb
```diff
@@ -525,5 +526,19 @@ def generated_grpc_cpp?
...
+    # // GENERATED CODE - DO NOT MODIFY
+    # on the first line.
+    #
+    # Return true or false
+    def generated_dart?
+      return false unless extname == '.dart'
+      return false unless lines.count > 1
+      return true if lines[0].include?("Generated code. Do not modify.")
+      return lines[1].include?("GENERATED CODE - DO NOT MODIFY")
```
lildude: I think these two lines can probably be merged into one case-insensitively checking for “generated code”.

lildude: Back to my previous comment... the last two lines can be done in a single check using:

```
lines.first(2).any? { |l| l.downcase =~ /generated code.+do not modify/  }
```

... however it does have a few assumptions and does raise a few questions...

1. Given the comment you've added for this section, do we even need the check for `"Generated code. Do not modify."`?
2. Your comment says the `// GENERATED CODE...` line will be on the _first_ line, yet you're testing for it on the second line. Which is it or is it likely to be in either the first or second line?
3. Will the "generated code" line _always_ be in the first two lines or could it be a little further down too, for example on line 5?

chances:
@lildude Addressing your questions:

1. Checking for `"Generated code"` check for Dart files generated by [dart-lang/appengine](https://github.com/dart-lang/appengine). I'll clarify that in the doc comment.
2. `lines[1]` was a mistake. Fixed.
3. I have not seen other generated Dart files where a this-is-generated comment does not occur on the first line. However, I can't say for sure that's _always_ true.

lildude:
> lines[1] was a mistake. Fixed.

Great. As this is checking a single line, we can definitely combine the two checks into a single case-sensitive check.

lildude:
Thanks @chances. This looks good to me now, but I won’t give my final review until @pchaigno has reviewed.

lildude:
I took the liberty of resolving the merge conflicts, cos I could. I'll approve if all goes green.

PR #4032, opened 2018-02-12
Whitelist injectionSelector in grammars

pchaigno:
`injectionSelector` is an (undocumented?) key for Atom. See cdibbs/language-jison#17 for details.

/cc @vmg


pchaigno:
@Alhadis Do you know if it's documented anywhere?

Alhadis:
Nowhere officially, no. But there are wayyy too many parts of the TextMate grammar system that suffer from sketchy documentation. :confused

For example, there's also a little known property called `matchEndFirst` that (too my knowledge) has only ever been used by the original Perl grammar. It's never been documented anywhere.

PR #4036, opened 2018-02-16
CONTRIBUTING: note license.

MikeMcQuaid:
This is to be more explicit about the contribution process and license.


lildude:
We can ignore the test failures as they're completely unrelated - gonna fix those now.

PR #4042, opened 2018-02-19
Detect Maven wrapper "mvnw"

cyrille-leclerc:
## Description

Add detection of [Maven wrapper files](https://github.com/takari/maven-wrapper) `mvnw` and `mvnw.cmd` so that Java Maven projects using the Maven wrapper don't get recognized as "Shell" project but as "Java" projects.

The Maven wrapper is broadly used in the Java ecosystem. For example, the Spring Framework project kickstart start.spring.io generrates projects with the `mvnw` wrapper.

## Checklist:

  - [x] The Maven wrapper (mvnw / mvnw.cmd) is used in 410K repositories on GitHub.com (2018/02/19: 350K, 2018/04/6: 410k).
    - Search results for "mvnw": 
        - https://github.com/search?utf8=%E2%9C%93&q=+filename%3Amvnw+&type=Code
    - Sample source(s):
      - Sample "mvnw" added with the PR: https://github.com/cyrille-leclerc/linguist/blob/77c5cb1d44e0d6bebb0fcdf4e4111e296738746e/samples/Shell/filenames/mvnw
      - Unit tests added by the PR: https://github.com/cyrille-leclerc/linguist/commit/77c5cb1d44e0d6bebb0fcdf4e4111e296738746e#diff-4ddef002fed17b34709d41546a40729c

### Current behavior: undesired detection of the Maven project as a Shell project

:x: project detect as Shell, desired is Java

```
$ git clone https://github.com/cyrille-leclerc/my-jar.git
$ cd my-jar
$ linguist --breakdown
54.06%  Shell
43.03%  Batchfile
2.92%   Java

Shell:
mvnw

Batchfile:
mvnw.cmd

Java:
src/main/java/com/example/myjar/MyJar.java
src/test/java/com/example/myjar/MyJarTest.java
```

### Desired behavior with the PR: detection of the Maven project as a Java project

:white_check_mark: project detected as Java

```
$ git clone https://github.com/cyrille-leclerc/my-jar.git
$ cd my-jar
$ linguist --breakdown
100.00% Java

Java:
src/main/java/com/example/myjar/MyJar.java
src/test/java/com/example/myjar/MyJarTest.java
```


pchaigno:
As far as I can see, these files (e.g., [this one](https://github.com/eneasanulhak/vinhos/blob/e684a1154e70ef550995bcdc23ebf1a1835ca701/mvnw)) are Bash scripts, Bash scripts for Maven, but still Bash scripts. It would be incorrect to recognize them as Java and it would break their syntax highlighting.

lildude:
Hmmm, these changes mark the files as vendored rather than Java/Maven hence the second "working" demo appears to work and classify them as Java. Marking them as vendored excludes them from the stats, which raises the question: is it generally considered that these files should be excluded from repo language stats across the Java/Maven community?

lildude:
Oh yes, and the test failure is legit - you'll need to address that.

Alhadis:
> As far as I can see, these files (e.g., this one) are Bash scripts, Bash scripts for Maven, but still Bash scripts.

Actually, that's a very solidly portable POSIX shell-script, not merely for Bash. 😉

Even if the constructs used don't give it away, any script whose comments cite compatibility with Solaris 10 is almost certainly guaranteed to be written for portability/POSIX-compliance. =)

So, most definitely *not* Java, or even anything remotely resembling it.

cyrille-leclerc:
@pchaigno @lildude @Alhadis I mimicked was was done for the Gradle `gradlew` file and I have added `mvnw` to the `vendor.yml` ("Vendored files and directories are excluded from language statistics.").

My understanding of `vendor.yml` is that I didn't say that `mvnw` and `mvnw.cmd` are Java files, I just said that they are "helper" files that come along Maven projects and that they can be ignored from statistics.
The build tool file that matter is `pom.xml` that is used by Maven projects and I don't mention it in this pull request so its detection will not change.

The programming language files that matter are the `*.java ` files and other files built by Maven (html, css...) and I don't change their detection either.

> Oh yes, and the test failure is legit - you'll need to address that.

I'm looking

lib/linguist/languages.yml
```diff
@@ -4325,6 +4325,7 @@ Shell:
   - bashrc
   - cshrc
   - gradlew
+  - mvnw
```
cyrille-leclerc: filenames should be alphabetically sorted, should be added after "man" and before "profile"

Alhadis: How many of these `mvnw` lack shebangs? Any portable shell-script is guaranteed to start with `#!/bin/sh` as per long-standing convention.

If there's no risk of a `mvnw` file lacking the hashbang, then you should omit this entry from `languages.yml` altogether. It's unrelated to your classification, and I believe the cause of the problem is to do with vendored-file detection, not languages.

cyrille-leclerc: > How many of these `mvnw` lack shebangs?

All the `mvnw` files should start with `#!/bin/sh` as it has been there since the Maven Wrapper project has been started 5 years ago (see git blame [here](https://github.com/takari/maven-wrapper/blame/master/mvnw#L1)).

> If there's no risk of a mvnw file lacking the hashbang, then you should omit this entry from `languages.yml` altogether

ack, I'll push a commit removing `mvnw` from `languages.yml` as there is no risk of having `mvnw` files not starting by a shebang `#!/bin/sh`.

Thanks for your help @Alhadis !

cyrille-leclerc:
@Alhadis I removed the `mvnw` file from the languages/shell/filenames section as these `mvnw` files are expected to always start from a shebang, the build is green.

Please tell me if something more is needed.

Alhadis:
Great, thanks!

I'll leave the full review to @pchaigno or @lildude to handle, as they're more familiar with the mechanics of vendored-file filtering than I am. 😉

lib/linguist/languages.yml
```diff
@@ -5380,3 +5380,4 @@ xBase:
   tm_scope: source.harbour
   ace_mode: text
   language_id: 421
+
```
pchaigno: Looks like you committed this by mistake.

pchaigno:
> My understanding of `vendor.yml` is that I didn't say that `mvnw` and `mvnw.cmd` are Java files, I just said that they are "helper" files that come along Maven projects and that they can be ignored from statistics.

My bad, I read the initial post a bit too quickly.

LGTM otherwise!

cyrille-leclerc:
@lildude thanks, I am removed the extra white spaces accidentally added to languages.yml. Please feel free to squash the commits to cleanup the history of all my iterations to get this PR right.

PR #4051, opened 2018-03-01
enable syntax highlighting for .jinja2 ext

stringfellow:
## Description
This is a commonly used jinja (2!) extension

## Checklist:

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Ajinja2+endblock
    - Sample source(s):
      - https://raw.githubusercontent.com/osks/gearman-dashboard/18198eda768e1d460a641bbc95fdf3067f507290/gearmandashboard/templates/_workers.jinja2
    - Sample license(s): 
      - MIT


stringfellow:
@pchaigno cool - I have picked out one from the search and added to the description

lildude:
> @pchaigno cool - I have picked out one from the search and added to the description

You need to add it to the PR 😉 It also needs to be a real world use example. The one you've referenced looks like a "hello world" example which is never production-ready code 😉 😉

lildude:
Oh yes, we also need the template updated to mention the license of the sample.

stringfellow:
@lildude Aha, ok. So here's a load of samples from the jinja2 docs. It's "real world" in that it is a neat example of likely usage (since its direct from the docs) and I based that decision on the example of HTML+Django being quite from-the-docsy ( https://github.com/github/linguist/blob/master/samples/HTML%2BDjango/nunjucks.njk )

stringfellow:
urgh, I guess I did that wrong.

samples/Jinja2/jinja2.jinja2
```diff
@@ -0,0 +1,83 @@
+{% extends layout_template if layout_template is defined else 'master.jinja2' %}
```
pchaigno: You need to move this file to `samples/HTML+Django/`

pchaigno:
One last thing: could you update the template with the link to the original sample file (i.e., where it was taken from)?

stringfellow:
@pchaigno as I constructed it from the documentation for jinja2 (see comment above) - should I link it to that?

lildude:
> @pchaigno as I constructed it from the documentation for jinja2 (see comment above) - should I link it to that?

Yup, pointing to the docs should suffice.

stringfellow:
Is there anything else y'all need me to do to get this progressed?

pchaigno:
@stringfellow Could you replace the sample file with [this one](https://github.com/osks/gearman-dashboard/blob/18198eda768e1d460a641bbc95fdf3067f507290/gearmandashboard/templates/_workers.jinja2)? I think it'd be more representative and it looks like it is [under MIT license](https://github.com/osks/gearman-dashboard/blob/rewrite-to-pyramid/LICENSE.txt). You don't need to add a link in the sample file, in the original post is enough.

stringfellow:
@pchaigno Have done so, though I'm not sure I agree that it's any more representative than one constructed straight from the docs with a lot of the syntax covered, but if it makes more sense I'm OK with that!

stringfellow:
🎉

PR #4052, opened 2018-03-01
Add notes about gitattributes usage (see #4050)

Pawamoy:
`.gitattributes` must be committed again after each modification.

Also a `licenses/*` pattern won't match the `licenses` sub-sub-directory's files, we need to use `**/licenses/*`.

But it's confusing me a lot because [Git documentation itself](https://www.git-scm.com/docs/gitattributes) says the patterns in `.gitignore` and `.gitattributes` will match the files the same way:

> The rules how the pattern matches paths are the same as in .gitignore files; see gitignore.

Did I miss something here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,9 +131,7 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
-Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that you might need to use wildcards to match files deeper than top-level ones.
-
-```
+```console
 $ cat .gitattributes
 build/* linguist-vendored     # will match build/* but not dist/build/*
 **/build/* linguist-vendored  # will match build/* and dist/build/*
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
-Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that Linguist requires that you match files and not directories in `.gitattributes`. You might need to use wildcards to match files deeper than top-level ones.
+Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that you might need to use wildcards to match files deeper than top-level ones.
 
 ```
 $ cat .gitattributes
--- a/README.md
+++ b/README.md
@@ -131,10 +131,12 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
-Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that `.gitattributes` patterns are not equivalent to `.gitignore` ones. You may need to use wildcards to match files in subdirectories:
+Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that Linguist requires that you match files and not directories in `.gitattributes`. You might need to use wildcards to match files deeper than top-level ones.
 
-```$ cat .gitattributes
-**/build/* linguist-vendored
+```
+$ cat .gitattributes
+build/* linguist-vendored     # will match build/* but not dist/build/*
+**/build/* linguist-vendored  # will match build/* and dist/build/*
 ```
 
 #### Vendored code
--- a/README.md
+++ b/README.md
@@ -131,6 +131,12 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
+Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that `.gitattributes` patterns are not equivalent to `.gitignore` ones. You may need to use wildcards to match files in subdirectories:
+
+```$ cat .gitattributes
+**/build/* linguist-vendored
+```
+
 #### Vendored code
 
 Checking code you didn't write, such as JavaScript libraries, into your git repo is a common practice, but this often inflates your project's language stats and may even cause your project to be labeled as another language. By default, Linguist treats all of the paths defined in [`vendor.yml`](/lib/linguist/vendor.yml) as vendored and therefore doesn't include them in the language statistics for a repository.


Pawamoy:
Also english is not my natural language so please correct my phrasing if needed!

Pawamoy:
Okay I think I got it:

> For example, "Documentation/*.html" matches "Documentation/git.html" but not "Documentation/ppc/ppc.html" or "tools/perf/Documentation/perf.html".

Since Linguist requires us to match files and not directories, we are tempted to use `licenses/*` just like we would ignore any directory called `licenses` in a `.gitignore` with `licenses/`. But adding that trailing asterisk makes so that subdirectories called `licenses` and their files are not matched. Adding that trailing asterisk also requires to add the leading double-asterisk if we want to match subdirectories called `licenses` and not only a toplevel `licenses` directory.

Will try to update the note accordingly.

README.md
```diff
@@ -131,6 +131,14 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
+Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that Linguist requires that you match files and not directories in `.gitattributes`. You might need to use wildcards to match files deeper than top-level ones.
```
lildude: You changed this before I could point out the pattern matching is identical for `gitattributes` and `gitignore` with the exception that `gitattributes` doesn't support negation 😄.

> Also note that Linguist requires that you match files and not directories...

... doesn't sound correct to me.

From `gitignore`:

```
 o   If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other
     words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the
     way how pathspec works in general in Git).

 o   If the pattern does not contain a slash /, Git treats it as a shell glob pattern and checks for a match against the pathname relative to the location
     of the .gitignore file (relative to the toplevel of the work tree if not from a .gitignore file).

 o   Otherwise, Git treats the pattern as a shell glob suitable for consumption by fnmatch(3) with the FNM_PATHNAME flag: wildcards in the pattern will not
     match a / in the pathname. For example, "Documentation/*.html" matches "Documentation/git.html" but not "Documentation/ppc/ppc.html" or
     "tools/perf/Documentation/perf.html".
```

So you should find `dir` will work just as well as `dir/*`, however `dir/` will _not_ because of the trailing `/` which makes git treat it as a directory and not a glob.

Pawamoy: What if I simply remove the `Linguist requires that you match files and not directories` part then?

lildude: Hmmm, ignore my comment about `dir` being equivalent to `dir/*`... just tested it and seems it doesn't work, which is news to me. 🤔

Pawamoy: And really for my use-case, `licenses linguist-vendored` does not work. I found only `**/licenses/* linguist-vendored` to be working :confused:

Pawamoy:
I think the simple example I used is enough for users to understand how it works. I think we should not lose ourselves in complicated explanations about gitignore and gitattributes pattern matching mechanism :smile: What do you think?

lildude:
> I think the simple example I used is enough for users to understand how it works. I think we should not lose ourselves in complicated explanations about gitignore and gitattributes pattern matching mechanism 😄 What do you think?

I agree. I think we should however change "Linguist requires" as this is not specific to Linguist and appears to be a trait of gitattributes.

Pawamoy:
Yup already removed it :)

lildude:
The overall content looks good though I'll need run over the wording a bit in my head when I've a bit more time to consider it in the context of the rest of the doc.

One of the other maintainers may beat me to it when they come to review this PR.

lildude:
Whhooops. I've just realised I've inadvertently mostly obsoleted this PR by merging https://github.com/github/linguist/pull/4115. Sorry. 😊 

The only thing remaining is the example and comment about wildcards. I'm not sure that we should get into the realms of `.gitattributes` usage within the Linguist documentation. What do you think @Alhadis and @pchaigno?

README.md
```diff
@@ -131,6 +131,14 @@ $ cat .gitattributes
 *.rb linguist-language=Java
 ```
 
+Note that, when using Linguist locally, `.gitattributes` must be committed again each time it is modified for Linguist to take it into account. Also note that you might need to use wildcards to match files deeper than top-level ones.
```
Alhadis: Maybe amend this so it reads:

> **Note:** When run locally, Linguist will ignore uncommitted changes to `.gitattributes`. Remember to commit each modification before rerunning `linguist`.

README.md
```diff
@@ -131,6 +131,14 @@ $ cat .gitattributes
...
+
+```
```
Alhadis: If you add `console` to this fenced code-block, it'll add spiffy highlighting to the output. 😉

~~~console
$ cat .gitattributes
build/* linguist-vendored     # will match build/* but not dist/build/*
**/build/* linguist-vendored  # will match build/* and dist/build/*
~~~

README.md
Alhadis: And yes, make sure you get rid of this fluff:

>  Also note that you might need to use wildcards to match files deeper than top-level ones.

Pawamoy: Since `When testing with a local installation of Linguist, take note that the added attributes will not take effect until the .gitattributes file is commited to your repository.` has already been added with https://github.com/github/linguist/pull/4115, I should simply remove this paragraph :slightly_smiling_face:

Alhadis: Ah. Erm... yes.

So, uh, I guess that reduces this PR down to amending the fenced code-block with `console`. 😅

Pawamoy:
If you want to keep the example I add, should I merge it with the one above? If not, simply close this PR :slightly_smiling_face:

Alhadis:
Honestly, this all feels so absurd, we've crushed your attempt to contribute into little pieces and I feel kinda bad.

@lildude should we at least get this guy on the contributors list? He's been patient enough to deserve it. 😅

Pawamoy:
@Alhadis I appreciate that, thanks :smile: But really, don't worry, it was a small PR I wasn't expecting great rewards :slightly_smiling_face:

Alhadis:
Alright, uh, guess you can close this... 😞

This has been the most awkward PR I've encountered in recent memory... 🥇

PR #4058, opened 2018-03-04
Add Hive QL and q to list of languages

yznima:
## Checklist:

Q
----
     -  https://github.com/search?utf8=%E2%9C%93&q=extension%3Aq+each+NOT+nothack&type=Code
      -  https://github.com/search?utf8=%E2%9C%93&q=extension%3Aq+flip+NOT+nothack&type=Code
      - https://github.com/search?utf8=%E2%9C%93&q=extension%3Aq+reverse+NOT+nothack&type=Code
    - Sample source(s):
      - ml.q https://github.com/jlas/ml.q/blob/master/ml.q
      - tq.q https://github.com/KxSystems/kdb-taq/blob/master/tq.q
    - Sample license(s):
      - ml.q MIT
      - tq.q Apache-2.0
  - [X] I have included a syntax highlighting grammar.

Hive QL
----
     -  https://github.com/search?utf8=%E2%9C%93&q=extension%3Aq+select+NOT+nothack&type=Code
    - Sample source(s):
      - mi.q https://github.com/apache/hive/blob/master/ql/src/test/queries/clientpositive/mi.q
    - Sample license(s):
      - mi.q  Apache-2.0
  - [X] I have included a syntax highlighting grammar.


lildude:
This is a duplicate of an already open and blocked PR https://github.com/github/linguist/pull/3820 and doesn’t provide a solution to the blocker either.

Please review that PR and if you can find a way to address the issues blocking that PR, please feel free to update your PR to do so and we’ll look into it.

yznima:
@lildude It seems like the other PR is blocked since Hive Ql also uses `.q` extension. Is the solution to add Hive QL also to the list of languages and write a heuristic to resolve the two files language?

Also regarding the heuristic regex, I know that any line that starts with `/` is a comment. And almost 90 percent of files set the namespace in q file using `\d` in their files. However, the comment in Hive QL starts with `--`. Would this be a good enough identifier?

lildude:
> Is the solution to add Hive QL also to the list of languages and write a heuristic to resolve the two files language?

You'll need to add associations for as many languages as possible that use a `.q` extension (which looks like it may include SQL too as pointed out by @pchaigno at https://github.com/github/linguist/pull/3820#issuecomment-338375928) and a heuristic to differentiate between then all.

As for the heuristic itself, I'm not familiar with Q or Hive QI so I can't really comment on an appropriate regex to use other than to say, the more cases it matches, the better. If that means using a pretty complex regex, then so be it. We've got quite a few complex regexes already.  The only thing to be mindful of is the performance of the regex.

yznima:
@lildude The only other language that I could find that uses `.q` is a Hive QL script which is a SQL-like syntax. I added the language Hive to the list of languages. I also added a heuristic to recognize the q language by one of the following

1. Function definition and assignment. `....:{`  (Note that in Kdb Q, functions don't have to specify the parameters, so the `[...]` is not necessary)
2. If the file or a line starts with `/`. In Kdbq, this is single line comment. in Hive QL a comment always starts with `--`.
3. If the file has `\d .`.

I had two questions, 
1. Is it ok that my heuristic doesn't set the language to `Hive` ever? My logic was that if anyone would like to add syntax highlighting for Hive QL or language recognition, they could in future.
2. I am tempted to also remove the `language-hql` syntax highlighter. Is it ok to only add the language Hive to list of languages without a syntax highlighter?

Finally, how do I verify the correctness of the heuristic and performance of it?

yznima:
@lildude Could you re-review?

lildude:
> @lildude Could you re-review?

I was kinda hoping you'd resolve at least these two failing tests first as we won't merge
 without passing tests 😄:

```
  2) Failure:
TestPedantic#test_heuristics_are_sorted [/home/travis/build/github/linguist/test/test_pedantic.rb:38]:
    disambiguate ".x" do |data|
 should come after     disambiguate ".q" do |data|
  3) Failure:
TestColorProximity#test_color_proximity [/home/travis/build/github/linguist/test/test_color_proximity.rb:21]:
The following 1 languages have failing color thresholds. Please modify the hex color.
- Parrot (#f3ca0a) is too close to ["ffdd00", "f3ca0a"]
```

The license failure will be resolved once https://github.com/github/linguist/pull/3982 is merged, which I hope to do :soon:.

I also won't give my final review until one of the non-GitHub maintainers has reviewed the PR.

To answer your questions:

> Is it ok that my heuristic doesn't set the language to Hive ever? My logic was that if anyone would like to add syntax highlighting for Hive QL or language recognition, they could in future.

Yeah. We've got this in several other places too so I don't think there will be too many problems, but I could be missing something which I'm sure one of the other maintainers will point out if I am 😉. As for someone else adding the grammar etc... you've done it already.

> I am tempted to also remove the language-hql syntax highlighter. Is it ok to only add the language Hive to list of languages without a syntax highlighter?

It is, but you may as well keep the grammar and help out the Hive community whilst you're here.

yznima:
@lildude Thanks you. All test failures except the following are addressed.

```
Warnings:
vendor/licenses/grammar/language-hql.txt:
  - license needs reviewed: other.
303 dependencies checked, 1 warnings found.
```

lildude:
https://github.com/github/linguist/pull/3982 has now been merged. You should find the license test will now pass once you've updated this branch in your fork.

yznima:
@lildude Just a friendly reminder that this is waiting for review

lildude:
From a quick look, this looks good, but I'll reserve my final review until I've had a chance to take another closer look, especially in light of the comments in https://github.com/github/linguist/pull/3820. I also won't give my final :+1: until one of the other core maintainers has given their review.

lib/linguist/heuristics.rb
```diff
@@ -400,6 +400,14 @@ def call(data)
       end
     end
 
+    disambiguate ".q" do |data|
+      if /[A-Za-z\.][A-Za-z0-9\.]*:{/.match(data) || /(^|\n)\\(cd?|d|l|p|ts?) /.match(data)
+        Language["q"]
+      else
```
pchaigno: Couldn't we restrict this to files that contain a line starting with `--` (since you mentioned these are comments in HiveQL)?

pchaigno: Also, this is basically SQL. We could probably use a number of heuristic rules specific to SQL (e.g., `SELECT \*` or `ALTER (table|database)`).

yznima: I took the second approach.

yznima:
> I don't see these last two rules being used in the heuristic rule you implemented for q.

Sorry, I forgot to update the comments, I basically thought that relying on the comment would not be very useful hence I removed the second rule.

I also updated the third rule to not only account to `\d` but also for other most frequently used system functions such as `\t`, `\ts` and `\l`.  (BTW, Full list of kdbq system commands can be found [here](https://code.kx.com/wiki/Reference/SystemCommands)).



> What makes you think that file is under Apache 2.0 license?

You are right. I miscopied the `tq.q`. I replaced it with the correct [tq.q](https://github.com/KxSystems/kdb/blob/master/tq.q) with the `Apache-2.0` [license](https://github.com/KxSystems/kdb)

pchaigno:
Going through [the search results that do not seem to be Hive QL or q](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aq+NOT+each+NOT+flip+NOT+reverse+NOT+select&type=Code), I found a [large number of `.uosl.q` files](https://github.com/search?p=96&q=in%3Afilename+uosl.q&type=Code). It looks like scripts for an MMORPG, [Ultima Online](https://en.wikipedia.org/wiki/Ultima_Online). I used [Harvester](https://github.com/Alhadis/Harvester) to download 2999 of those files and found that they are distributed across 1123 repositories and 1079 users.
I think it warrants support in Linguist with the other `.q` languages. The good news is that, since these files are using a nested file extension, we don't need a new heuristic rule to identify them; adding the language to `languages.yml` with `.uosl.q` as a file extension should be enough.

pchaigno:
Actually, there aren't near far enough `.uosl.q` files to warrant support. Only 2 users with 2 repositories. Looks like searching for `in:filename uosl.q` doesn't return only files with `uosl.q` in the filename and I didn't bother to verify. Sorry about this @yznima.

yznima:
@lildude Is there anything left to be done for this?

zak-b2c2:
Is this correct usage? And do I need to merge to master?
$ cat .gitattributes 
*.q linguist-language=q

pchaigno:
@zak-b2c2 This will only work after the next version of Linguist is released. Your `.gitattributes` appears to be correct :+1:

zak-b2c2:
Cool. When can we expect the next release? :)

lildude:
> Cool. When can we expect the next release? :)

There's no set date, but I'm trying to get into a habit of making a release at least once a month, so I anticipate probably somewhere around the middle of the month.

yznima:
@lildude I was wondering if you are still planning to do a release this month?

zak-b2c2:
Looking good, thanks guys !

lildude:
@zak-b2c2 snuck in and stole my thunder 😜 - this is now live on GitHub.com 🎉

PR #2005, opened 2015-01-18
Adds Rexfile

JJ:
That's a Vagrantfile-like file for the [Rex provisioning too](http://rexify.org).

PR #4066, opened 2018-03-13
Additions to the Perl family of languages

JJ:
Adding some extensions and sample files for Perl

## Description

- Moves Rexfile to the correct language
- Adds extensions to Perl and Perl6. Some sample files, too.

## Checklist:

- [x] **I am associating a language with a new file extension.** (but search for pod6 does not seem to work very well)
    - Sample source(s):
      - https://github.com/perl6/doc/blob/master/doc/Type/Any.pod6
    - Sample license(s): ARTISTIC https://github.com/perl6/doc/blob/master/LICENSE


lib/linguist/languages.yml
```diff
@@ -3369,6 +3369,8 @@ Perl:
   - ".psgi"
   - ".t"
   filenames:
+  - Makefile.PL
```
Alhadis: This is redundant, because it'll be matched against the `*.pl` file extension.

CONTRIBUTING.md
```diff
@@ -6,11 +6,15 @@ Contributions to this project are [released](https://help.github.com/articles/gi
 
 This project adheres to the [Contributor Covenant Code of Conduct](http://contributor-covenant.org/). By participating, you are expected to uphold this code.
 
-The majority of contributions won't need to touch any Ruby code at all.
```
Alhadis: I see you first forked Linguist in [November 2015](https://github.com/github/linguist/pull/4066/commits) (!!). You're overwriting unrelated changes that have been made to the docs since then. Please update `CONTRIBUTING.md` with the [latest version](https://github.com/github/linguist/blob/4f0f9bd51d9bf54e05d1b1456dfdc3df54b3360d/CONTRIBUTING.md).

lib/linguist/languages.yml
```diff
@@ -3369,6 +3369,8 @@ Perl:
...
+  - Rexfile
```
Alhadis: In-the-wild usage is a little questionable... only [~155 results](https://github.com/search?q=filename%3ARexfile+NOT+nothack&type=Code), although they appear to be well-distributed across users/repositories (which technically meets the criteria of "being used in hundreds of repositories"). I think it's okay, particularly since it's an uncommon filename unlikely to cause conflict. @pchaigno, thoughts?

lib/linguist/languages.yml
```diff
@@ -3387,9 +3389,8 @@ Perl 6:
   - ".pl6"
   - ".pm"
   - ".pm6"
+  - ".pod6"
```
Alhadis: This was discussed before, and should be omitted. See #3366.

lib/linguist/languages.yml
pchaigno: @Alhadis That extension is shared between Perl, Perl 6, and Prolog. Adding the `Makefile.PL` might improve detection for that particular case. Are we sure `Makefile.PL` is always Perl and not Perl 6 though?

CONTRIBUTING.md
pchaigno: @JJ You might want to update your fork...

lib/linguist/languages.yml
pchaigno: I'll defer to @lildude :smile:

CONTRIBUTING.md
JJ: I'll do that :+1:

lib/linguist/languages.yml
Alhadis: I stand corrected, then. :wink: @pchaigno knows more about the Bayesian classifier than I do.

I don't recall ever seeing a Perl 6 version of `Makefile.PL`... it's always ever been used by Perl 5 modules downloaded/installed from CPAN. I could be wrong though.

JJ: Pretty sure. Perl6 has pretty much normalized to .pl6 or .p6 now. It's not using this particular file, either. This is the repo for all modules https://github.com/moritz/perl6-all-modules/search?utf8=%E2%9C%93&q=%22Makefile.PL%22&type= and it's not there even once. Makefile.PL is kind of a Rakefile used for some CPAN distributions; Perl6 uses zef and META6.json

lib/linguist/languages.yml
JJ: Checking it out... And I would kindly ask you to reconsider. What brought me (back) here was github/markup#1173, where I spent quite a long time until I discovered .pod6 files were not recognized as Perl6, which had me do a workaround... In fact, I fear .pod6 files will not be recognized as such, even by markup, if that extension is not included here. pod6 is used extensively in https://github.com/perl6/doc, if nowhere else, and as indicated in #3366, it would help adoption of the language. I can wait for that one to be accepted, though... Or just eliminate that change and defer to that one.

lib/linguist/languages.yml
pchaigno: > @pchaigno knows more about the Bayesian classifier than I do.

I don't know much about the classifier either, but in this particular case, there's a [specific filename strategy](https://github.com/github/linguist/blob/dd3b1eec911072ca051ecccabc03a47bafa2c76c/lib/linguist.rb#L62) before we reach the classifier.

PS: Not sure why I keep using "particular case" today...

lib/linguist/languages.yml
JJ: Problem here is that Rexfile was actually in the Perl6 field (my fault: https://github.com/github/linguist/pull/2005/files). It should better be here, Rexfile is in no way Perl6 code.

lib/linguist/languages.yml
Alhadis: Sounds good to me, then. :+1: The bulk of my familiarity with `Makefile.PL` is limited to seeing those files whoosh past the terminal when installing CPAN modules... :grinning:

CONTRIBUTING.md
JJ: I've restored the HEAD copy. In fact, advice to install CMAKE, which was the only actual change, is already in the README.md file.

lib/linguist/languages.yml
JJ: Is there any other change you want me to do here?

JJ: Ping... Is there anything missing?

lib/linguist/languages.yml
Alhadis: Personally, I'm okay with adding `.pod6` as an extension. But since there's been no official consensus on the matter, it might be easier to omit this particular addition and defer to #3366 instead. =)

lib/linguist/languages.yml
Alhadis: No, it looks fine to me. For the sake of quicker review, it might be better to omit `.pod6` since #3366 is more appropriate for discussion its inclusion.

JJ: Eliminated `.pod6`. I can try and work on that one, although I can't add or modify that pull request...

Alhadis: Don't worry about #3366. :) I believe that PR needs formal approval from GitHub staff more than anything.

JJ:
Thanks!

(But let me worry about that one, it's really important for the community)

lib/linguist/languages.yml
lildude: This is fine. It slipped under the radar before when it was initially added so no point removing it now.

samples/Perl 6/Any.pod6
```diff
@@ -0,0 +1,987 @@
+=begin pod
```
lildude: This file needs to be removed too as you've removed `.pod6` from this PR.

JJ:
OK.

samples/Perl 6/Any.pod6
JJ: Done.

lildude:
Pulling in the latest master into your fork will address the CI failure.  You'll also need to address the merge conflict.

JJ:
Will do that. Thanks.

JJ:
Please check the latest change that addresses conflict.

JJ:
Fixes CI error about alphabetical order on the files.

JJ:
Sorry, what else is left to do here?

Alhadis:
> Sorry, what else is left to do here?

Wait for this long-ass harvest to finish over my primitive, non-NBN-enabled Australian Internet connection, and be patient.

As an alternative to patience, you could entertain yourself by racing it against the following Perl script and seeing which one finishes in your lifetime:

~~~perl
$_ = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
/a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*[nope]/;
~~~

JJ:
I can do that in Perl 6, where infinity is just another data structure... Or I can wait for the [latest CI to finish with flying colors](https://travis-ci.org/github/linguist/builds/364160795?utm_source=github_status&utm_medium=notification), which it did! :penguin:

Alhadis:
Better idea! Let's all figure out why negative zero exists, and what sick human being thought a zero needed a sign-bit, and wonder why the hell -0 is different to 0 in any fathomable universe.

We can't divide by zero, but we can pointlessly sign it as +0 or -0 instead. Mmmm'kay, IEEE 754, whatever you reckon...

I don't know why I'm flagellating myself thinking about the real-world difference of positive and negative zero when I should be [making wisecracks in commit logs](https://github.com/Alhadis/language-sed/commit/cfbb39480e6577dc46cbaf48399dedd142ac3d8c) like last night.

samples/Perl 6/chromosome.pl
```diff
@@ -0,0 +1,9 @@
+class Chromosome {
+  has Seq $.chromosome is rw;
+  has $.fitness is rw;
+
+}
+
+my $len = 32;
+my $this-chromosome = Chromosome.new( chromosome => map( { rand >= 0.5 ?? True !! False }, 1..$len )  );
+say $this-chromosome.chromosome();
```
lildude: Is this file somehow uniquely Perl 6 such that it can clearly be differentiated from the exact same code written in Perl 5, or should this be added to the `samples/Perl/`?

The reason I ask is we use these samples to train the classifier and if this is identical to Perl 5, it's likely to add noise to the Perl 6 side of the classifier when it comes down to a Perl 5 vs Perl 6 comparison.

pchaigno: It could also go into `tests/fixtures/Perl 6/` if we don't want it as a sample.

JJ: On a superficial level, it could look similar to Perl. It's been a long time, but if I remember correctly I included it here precisely for that reason. It could go to fixtures alright. Just let me know what to do, and I'll do it straight away.

lildude: Ok. If this is just to keep the tests passing, then yup, `tests/fixtures/Perl 6/` is definitely the best place for it.

JJ: Done. I'm getting some local errors which apparently have nothing to do with moving stuff around. Let me know if there's something I can do about that.

JJ:
Yay! You made my day. Thanks!

PR #4067, opened 2018-03-13
Register "cperl" as an alias of Perl 5

Alhadis:
This PR enables recognition of Perl files containing cperl modelines:

~~~perl
#!/usr/bin/env perl
# -*- cperl -*-
~~~

## Description
Emacs ships with an enhanced major-mode for editing Perl 5 with embedded C sections (called `cperl-mode`, appropriately enough). One of the [sample files](https://github.com/github/linguist/pull/4066/files#diff-bfce7097cfedb19b250fbcbf3ad47aa6R1) added in #4066 reminded me about the mode, and I noticed it isn't being picked up by Linguist.

Since the mode has to be manually activated per file (since it isn't the default), the odds of it being used in embedded modeline are pretty high.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3371,6 +3371,7 @@ Perl:
   filenames:
   - cpanfile
   interpreters:
+  - cperl
   - perl
   aliases:
   - cperl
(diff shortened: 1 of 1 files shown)


Alhadis:
Oh right, I forgot there was a fork of Perl called "CPerl". :grinning: (I remember being confused by timestamps before realising the major-mode was unrelated to the fork...)

Will update.

PR #4068, opened 2018-03-15
Exclude R package doc directory

jsta:
A standard path for documentation in R packages is `inst/doc`. `linguist` is picking up files in this directory and mis-classifying R packages as html. This PR excludes `inst/doc` from linguist analysis.

## Description
There are numerous examples at https://github.com/cran/. For example:

* https://github.com/cran/rtrim
* https://github.com/cran/wheatmap

### DIFF
--- a/lib/linguist/documentation.yml
+++ b/lib/linguist/documentation.yml
@@ -15,6 +15,7 @@
 - ^[Mm]an/
 - ^[Ee]xamples/
 - ^[Dd]emos?/
+- ^inst/doc/
 
 ## Documentation files ##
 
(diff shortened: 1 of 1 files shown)


jsta:
I opened this PR at the suggestion of @pchaigno

lib/linguist/documentation.yml
```diff
@@ -15,6 +15,7 @@
 - ^[Mm]an/
 - ^[Ee]xamples/
 - ^[Dd]emos?/
+- ^inst/doc/
```
pchaigno: I think we need `(^|/)` here.

Issue #3746, opened 2017-07-31
Specify disambiguation rules in YAML


smola:
## What?

Disambiguation rules in `heuristics.rb` are specified with a DSL in a very structured way. It should be possible to move this to YAML instead of Ruby.

## Why?

We work on a linguist port to Go: [enry](https://github.com/src-d/enry). Most of the Go code is generated from the YAML files in the linguist project. The disambiguation rules are also generated, but using a custom parser, written ad hoc for the `heuristics.rb` file. Having these rules specified in YAML would make it easier to reuse them in other languages.

This is something we've been thinking about for a long time, but I didn't make the proposal yet since I was not sure there is any benefit for the linguist project itself. However, it came to my knowledge that there is other people interested in porting linguist to other languages (e.g. C++), so it would be also good for them to have something like this.

The way I see it, the biggest value of linguist is not the code itself, but all the knowledge about language detection that is included in it, and making this knowledge more accessible from other software would benefit everyone by making it easier to collaborate.

## How?

If you think this is worth it, I can work on a specific implementation proposal and PR.

smola:
cc @vmg

vmg:
Hi @smola!

Sorry for the late reply, you caught me on holidays. I really like this idea. It should make contributing to Linguist easier for people who don't know Ruby, and it will let you kill the custom parser in `enry`, which I know you really want to do. :)

> If you think this is worth it, I can work on a specific implementation proposal and PR.

Absolutely, I'll merge a PR like that if it has identical functionality to the existing code. Let' see it!

smola:
I'm implementing this in https://github.com/github/linguist/pull/4087. It's still work in progress, but the YAML syntax should already be pretty stable.

smola:
Merged.

PR #4087, opened 2018-04-02
[WIP] Specify disambiguation rules in YAML

smola:
Specify disambiguation rules in a YAML file. See rationale in #3746.

## Description
All disambiguation rules are moved from `heuristics.rb` into `heuristics.yml`.

## Checklist:


smola:
This is now ready for review, although it could use some more testing and cleanup of `heuristics.rb`.

vmg:
@smola: Looks like the heuristics got a bit messed up during the translation, as two language identification tests are failing. Can you look at them?

smola:
Fixed negative matches.
Also rebased this on top of https://github.com/github/linguist/pull/4189 to improve verification of the change. Ideally #4189 would be merged first.

lib/linguist/heuristics.rb
```diff
@@ -29,33 +32,52 @@ def self.call(blob, candidates)
...
+      end
     end
 
     # Internal: Array of defined heuristics
     @heuristics = []
 
     # Internal
-    def initialize(exts_and_langs, &heuristic)
+    def initialize(exts_and_langs, rules)
       @exts_and_langs, @candidates = exts_and_langs.partition {|e| e =~ /\A\./}
```
pchaigno: It looks like the YAML file doesn't support candidates (which is fine since we don't use that anymore). Thus, the above line can be simplified, as well as the additional logic in `matches?` to handle candidates.

lib/linguist/heuristics.yml
```diff
@@ -0,0 +1,385 @@
...
+#                       Pattern can be a string with a single regular expression
+#                       or an array of strings that will be merged in a single
+#                       regular expression (with union).
+# negative_pattern    - Same as pattern, but checks for absence of matches.
+# named_pattern       - A pattern can be reused by specifying it in the
+#                       named_patterns section and referencing it here by its
+#                       key.
+# named_patterns      - Key-value map of reusable named patterns.
+#
+# Please keep this list alphabetized.
```
pchaigno: Actually, there's a test for that in `test_pedantic.rb`, [`test_heuristics_are_sorted`](https://github.com/github/linguist/blob/51feda9e1f3bd1d41f19425e663e9f6f2cb640ed/test/test_pedantic.rb#L35-L39); could you update it to check the YAML file instead? We poor humans have a hard time keeping things alphabetized (speaking from experience :p).

lib/linguist/heuristics.yml
```diff
@@ -0,0 +1,385 @@
...
+  - '^[ \t]*catch\s*\('
+  - '^[ \t]*(class|(using[ \t]+)?namespace)\s+\w+'
+  - '^[ \t]*(private|public|protected):$'
+  - 'std::\w+'
+  fortran: '^(?i:[c*][^abd-z]|      (subroutine|program|end|data)\s|\s*!)'
+  objectivec: '^\s*(@(interface|class|protocol|property|end|synchronised|selector|implementation)\b|#import\s+.+\.h[">])'
+  perl5: '\buse\s+(?:strict\b|v?5\.)'
+  perl6: '^\s*(?:use\s+v6\b|\bmodule\b|\b(?:my\s+)?class\b)'
+#TODO: add tests for missing heuristics
+#TODO: add support to return multiple languages
```
pchaigno: Could you remove these? I'll open an issue for the first one so we don't forget and I don't think we supported the second one before anyway. We're very likely to forget TODOs at the end of the file, even if we later fix them.

smola: Oh, I forgot about these.

Actually, there is currently a disambiguation rule that returns two languages: `[Language["Linux Kernel Module"], Language["AMPL"]]` for `.mod`.

lib/linguist/heuristics.yml
smola: Updated the test.

lib/linguist/heuristics.rb
smola: I'll perform this change next time I get to work on this.

smola:
There is still one case I couldn't convert to a plain regular expression:

https://github.com/github/linguist/blob/ed44c0b86d317b753c607742b137dce9594bed97/lib/linguist/heuristics.rb#L326

However, I found no file that is misclassified with something simpler such as:

```
(?<!\S)\.(include|globa?l)\s|(?<!\/\*)(\A|\n)\s*\.[A-Za-z][_A-Za-z0-9]*:
```

According to @Alhadis here https://github.com/github/linguist/commit/06c049b8c0533add03d2412616bef316da770d8e this is the rationale:

> This commit modifies the heuristic to assume the following flow:
>  1. If a line contains ".include" or ".global"/".globl" which doesn't
>      follow a non-whitespace character, assume GAS.
>  2. Otherwise, if the line starts with a command like ".LG7E0" with a
>     possible string of whitespace before it, assume it's also GAS.
>
>    UNLESS either of the following conditions are true:
>    2a. The token is enclosed by a string or /* multiline comment */
>    2b. The previous line ends with a backslash to denote a statement
>          broken between lines, with possible whitespace and/or comment
>          sequences between the backslash and the actual newline.
>
>    3. If neither of the above are met, assume the file is MAXScript.

It seems usage of multiline comments with `/* */` syntax are rare in Unix Assembly. Supported by [HLA Syntax](https://en.wikibooks.org/wiki/X86_Assembly/HLA_Syntax) but I still didn't see any file with `.ms` extension using them.

If we assumed this kind of comments are not there, we can just simplify the expression to look for a label without any preceding `/*` (in multi-line mode). What do you think?

Alhadis:
> If we assumed this kind of comments are not there, we can just simplify the expression to look for a label without any preceding `/*` (in multi-line mode). What do you think?

That would work. I was probably being overly cautious with accuracy considering the PR was my first real contribution to Linguist (and the open-ended fallback to MAXScript convinced me that keeping matches airtight was a wise thing to do).

I don't remember seeing any GAS files with multi-line `/* */` comments, however.

smola:
@Alhadis :+1: Created a PR simplifying it: https://github.com/github/linguist/pull/4224

I'll change this PR to support combining negative and positive patterns at the same time. And also supporting multiple patterns with `&&` instead of `||`. The later also affects QMake, which was originally `data.include?("HEADERS") && data.include?("SOURCES")` but was effectively changed to `data.include?("HEADERS") || data.include?("SOURCES")`

smola:
Now with the support of `and:` blocks in `heuristic.yml`, functionality matches `heuristics.rb` behavior.
Extensions for QMake and `.ms` were the remaining ones, fixed in this iteration.

pchaigno:
And, just like that, I broke @Alhadis' pull request (cf. #4258). Sorry :stuck_out_tongue:

Alhadis:
Lovely. Now what am I gonna do? 😕 Change the YAML parsing logic to accept a filename property?

pchaigno:
@Alhadis I was thinking we could add something along the lines of `filename_pattern:`?

Alhadis:
Alright, I'll add it. But this really needs to be more object-oriented, IMHO.

For example, rather than specifying `filename_pattern` or `negative_pattern` as separate fields, we should have them as properties of a single `Pattern` object:

~~~yml
pattern:
- source: '/^foo/'
  target: data
- source: '/^bar'
  target: name
- source: '^qux'
  negated: true
~~~

Strings could still be used; simply treated as shorthand for an object with only the `source` field defined:

~~~yml
pattern: '/^foo/'

# Which is shorthand for:
pattern:
- source: '/^foo/'
  negated: false
  target: data
~~~

That way, we could adopt a cleaner implementation in `heuristics.rb` by subclassing `Regexp` and storing rule-specific properties on instances (such as whether to test against a file's data or its name).

This would also solve the problem of not being able to specify modifier flags for the regular expression, such as `/i` or `/x` (the latter would be useful if very lengthy expressions need to be defined in the YML file.

PR #4099, opened 2018-04-14
Strategies take result from previous strategy into account

### LINKED ISSUE #4090: Perl 5 vs Perl 6 mis-detection
## Preliminary Steps

Please confirm you have...

## Problem Description

Perl 5 and Perl 6 are different languages (despite similar names; like C vs C++). However, the heuristic fails in some cases. This was previously opened in https://github.com/github/linguist/issues/3906 and the cases covered in that issue were resolved.

https://github.com/zoffixznet/Music-ScaleNote

As far as I can see, the heuristic is failing because the heuristic looks for `use strict`, but the module uses `use strictures` instead.

### Last modified on:
No idea

### Expected language:
Perl 5 (aka "Perl")

### Detected language:
Perl 6

### Details

Since the previous times we had ambiguity issues, the Perl 6 language has now decided [more strongly](https://github.com/perl6/doc/commit/483b883c7e1bd7608489cf3dfa81588e09e1bbe2) that its extensions are `.pm6` for modules, `.p6` for files and `.t` for scripts.

Thus, I'd argue that the [heuristic for `.pm` and `.pl` files](https://github.com/github/linguist/blob/f4526126661d6f6ea15c967144454b088d7cd96f/lib/linguist/heuristics.rb#L363-L381) should prefer Perl 5 much more strongly. Likely just checking for Perl 6 first, with `use\s+v6\b` only, not the current regex, and marking code as Perl 6 if detected, otherwise, going to Perl 5. (personally, I'd say it shouldn't be disambiguated with Perl 6 at all and just assume those extensions are always Perl 5 or other lang). 

That leaves `.t` disambiguation. I'm less certain of this one, but just thinking that it can be improved if Perl 6 is tested first and checked against `use\s+v6\b`, along with some testing subs that have a dash in their names `done-testing|is-deeply|isa-ok|throws-like`. If that doesn't match, then assume the code's Perl 5 or other lang.

If someone submits a PR, is there a way to somehow preview that the changes give right results? Is [installing Linguist locally sufficient](https://github.com/github/linguist#usage)?

### LINKED ISSUE #4094: Perl file containing :- misidentified as Prolog
## Preliminary Steps

Please confirm you have...

## Problem Description

The file [`gdown.pl`](https://github.com/circulosmeos/gdown.pl/blob/master/gdown.pl) is being misidentified as Prolog when it is clearly Perl with a perl shebang.

Shebang:
```perl
#!/usr/bin/env perl
```
https://github.com/circulosmeos/gdown.pl/blob/178fa1061c858da375c4d4c2452b9ecb84789139/gdown.pl#L1

The line that I suspect causes it to be identified as Prolog (note the `:-` in the `:-(`):
```perl
    die "Couldn't download the file :-(\n" if ($check==0);
```
https://github.com/circulosmeos/gdown.pl/blob/178fa1061c858da375c4d4c2452b9ecb84789139/gdown.pl#L57

We'll add a manual file type override for now, but this seems like an issue in the detector.

https://github.com/circulosmeos/gdown.pl

### Last modified on:

2018-04-10

### Expected language:

Perl

### Detected language:

Prolog

pchaigno:
With this pull request, each strategy takes as candidates the language outputted by the previous strategy, if any. This was already the case for the Classifier and Heuristic strategies as these couldn't generate new candidate languages (as opposed to the Modeline, Filename, Shebang, and Extension strategies).

In practice, this signifies that if, for example, the Shebang strategy finds two possible languages for a given file (as is currently possible with the `perl` interpreter), the next strategy, the Extension strategy, will use that information and further reduce the set of possible language.
Currently, without this commit, the Extension strategy would discard the results from the previous strategy and start anew, possibly returning a different language from those returned by the Shebang strategy.

The first commit adds a fixture file to show the error happening (see [Travis CI](https://travis-ci.org/github/linguist/builds/366515803?utm_source=github_status&utm_medium=notification)) when a shebang (in this case the `perl` shebang) is owned by several languages (Perl and Pod). According to [this small and dirty script](https://gist.github.com/pchaigno/616aecceeb759e3ad19730439e146bea), 2 other interpreters are shared by 4 languages:
```
$ ./find-shared.py interpreters
OCaml and Reason share set(['ocaml'])
Terra and Lua share set(['lua'])
Parrot Internal Representation and Parrot Assembly share set(['parrot'])
Parrot Assembly and Parrot Internal Representation share set(['parrot'])
Reason and OCaml share set(['ocaml'])
Perl and Pod share set(['perl'])
Lua and Terra share set(['lua'])
Pod and Perl share set(['perl'])
``` 

Fixes #4090.
Fixes #4094.

### DIFF
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -3,13 +3,15 @@ module Linguist
     # Public: Use shebang to detect language of the blob.
     #
     # blob               - An object that quacks like a blob.
+    # candidates         - A list of candidate languages.
     #
     # Examples
     #
     #   Shebang.call(FileBlob.new("path/to/file"))
     #
-    # Returns an Array with one Language if the blob has a shebang with a valid
-    # interpreter, or empty if there is no shebang.
+    # Returns an Array of Languages from the candidate list for which the
+    # blob's shebang is valid. Returns an empty list if there is not shebang.
+    # If the candidate list is empty, any language is a valid candidate.
     def self.call(blob, candidates)
       return [] if blob.symlink?
 
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -10,10 +10,11 @@ module Linguist
     #
     # Returns an Array with one Language if the blob has a shebang with a valid
     # interpreter, or empty if there is no shebang.
-    def self.call(blob, _ = nil)
+    def self.call(blob, candidates)
       return [] if blob.symlink?
 
-      Language.find_by_interpreter interpreter(blob.data)
+      languages = Language.find_by_interpreter interpreter(blob.data)
+      if candidates.any? then candidates & languages else languages end
     end
 
     # Public: Get the interpreter from the shebang
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -9,8 +9,8 @@ module Linguist
     #
     #   Shebang.call(FileBlob.new("path/to/file"))
     #
-    # Returns an Array of Languages from the candidate list for which the
-    # blob's shebang is valid. Returns an empty list if there is not shebang.
+    # Returns an array of languages from the candidate list for which the
+    # blob's shebang is valid. Returns an empty list if there is no shebang.
     # If the candidate list is empty, any language is a valid candidate.
     def self.call(blob, candidates)
       return [] if blob.symlink?
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -16,7 +16,7 @@ module Linguist
       return [] if blob.symlink?
 
       languages = Language.find_by_interpreter interpreter(blob.data)
-      if candidates.any? then candidates & languages else languages end
+      candidates.any? ? candidates & languages : languages
     end
 
     # Public: Get the interpreter from the shebang
--- a/lib/linguist/strategy/extension.rb
+++ b/lib/linguist/strategy/extension.rb
@@ -2,8 +2,9 @@ module Linguist
   module Strategy
     # Detects language based on extension
     class Extension
-      def self.call(blob, _)
-        Language.find_by_extension(blob.name.to_s)
+      def self.call(blob, candidates)
+        languages = Language.find_by_extension(blob.name.to_s)
+        if candidates.any? then languages & candidates else languages end
       end
     end
   end
--- a/lib/linguist/strategy/extension.rb
+++ b/lib/linguist/strategy/extension.rb
@@ -16,7 +16,7 @@ module Linguist
       # in which case any language is a valid candidate.
       def self.call(blob, candidates)
         languages = Language.find_by_extension(blob.name.to_s)
-        if candidates.any? then languages & candidates else languages end
+        candidates.any? ? candidates & languages : languages
       end
     end
   end
--- a/lib/linguist/strategy/extension.rb
+++ b/lib/linguist/strategy/extension.rb
@@ -2,6 +2,18 @@ module Linguist
   module Strategy
     # Detects language based on extension
     class Extension
+      # Public: Use the file extension to detect the blob's language.
+      #
+      # blob               - An object that quacks like a blob.
+      # candidates         - A list of candidate languages.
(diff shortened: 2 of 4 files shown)


lib/linguist/shebang.rb
```diff
@@ -10,10 +10,11 @@ class Shebang
     #
     # Returns an Array with one Language if the blob has a shebang with a valid
```
lildude: We should change the comment here too to reflect the change in behaviour.

lib/linguist/shebang.rb
```diff
@@ -10,10 +10,11 @@ class Shebang
...
     # interpreter, or empty if there is no shebang.
-    def self.call(blob, _ = nil)
+    def self.call(blob, candidates)
       return [] if blob.symlink?
 
-      Language.find_by_interpreter interpreter(blob.data)
+      languages = Language.find_by_interpreter interpreter(blob.data)
+      if candidates.any? then candidates & languages else languages end
```
pchaigno: By the way @lildude, do you know if there's a more Rubyist way to write this?

lildude: Yup: 

```Ruby
candidates.any? ? candidates & languages : languages
```

pchaigno: I think it looks fine with syntax highlighting, but if you'd prefer stick with the simpler syntax, tell me and I'll remove the last commit.

lib/linguist/shebang.rb
pchaigno: I update this comment and added one for both the Filename and the Extension strategies. The wording's not great, so if you have any suggestion, I'll be happy to update!

lib/linguist/shebang.rb
```diff
@@ -3,17 +3,20 @@ class Shebang
...
     #
     # Examples
     #
     #   Shebang.call(FileBlob.new("path/to/file"))
     #
-    # Returns an Array with one Language if the blob has a shebang with a valid
-    # interpreter, or empty if there is no shebang.
-    def self.call(blob, _ = nil)
+    # Returns an Array of Languages from the candidate list for which the
+    # blob's shebang is valid. Returns an empty list if there is not shebang.
```
lildude: `s/not/no/`

lib/linguist/shebang.rb
```diff
@@ -3,17 +3,20 @@ class Shebang
...
+    # candidates         - A list of candidate languages.
     #
     # Examples
     #
     #   Shebang.call(FileBlob.new("path/to/file"))
     #
-    # Returns an Array with one Language if the blob has a shebang with a valid
-    # interpreter, or empty if there is no shebang.
-    def self.call(blob, _ = nil)
+    # Returns an Array of Languages from the candidate list for which the
```
lildude: I’d lowercase Array and Languages here and elsewhere they appear capitalised.

lib/linguist/strategy/extension.rb
```diff
@@ -2,8 +2,21 @@ module Linguist
...
+      # Public: Use the file extension to detect the blob's language.
+      #
+      # blob               - An object that quacks like a blob.
+      # candidates         - A list of candidate languages.
+      #
+      # Examples
+      #
+      #   Extension.call(FileBlob.new("path/to/file"))
+      #
+      # Returns an Array of Languages associated blob's file extension.
```
lildude: As above regarding case, and we need a few more words… “...associated with a blob’s file extension.”

lib/linguist/strategy/filename.rb
```diff
@@ -2,9 +2,22 @@ module Linguist
...
+      # Public: Use the filename to detect the blob's language.
+      #
+      # blob               - An object that quacks like a blob.
+      # candidates         - A list of candidate languages.
+      #
+      # Examples
+      #
+      #   Filename.call(FileBlob.new("path/to/file"))
+      #
+      # Returns an Array of Languages associated blob's filename. Selected
```
lildude: As above for case and “with a”

pchaigno:
Fixed. (Sorry about these messy sentences; I must have been more tired than I though yesterday :-/)

PR #4101, opened 2018-04-15
Improve error handling when registering new grammars

### LINKED ISSUE #4056: Scripts should handle Docker-related errors more gracefully
## Problem Description (copied from [`#4054`](https://github.com/github/linguist/pull/4054#issuecomment-369879875))

I ran into a few issues when testing an update to the `add-grammar` script:

1. Docker wasn't installed, so I had to `sudo apt-get install docker`

2. Docker *still* wasn't installed, so I had to double-check what the hell the `docker` package was, and found out the correct name is actually `docker-ce` (??!)

3. The correct Docker was installed, but it complained about inadequate permissions:

~~~
Warning: failed to get default registry endpoint from daemon (Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.35/info: dial unix /var/run/docker.sock: connect: permission denied). Using system default: https://index.docker.io/v1/
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.35/images/create?fromImage=linguist%2Fgrammar-compiler&tag=latest: dial unix /var/run/docker.sock: connect: permission denied
~~~

This makes for some pretty unfriendly UX; the scripts should ideally be updated to deal with these hurdles. Even bailing early in `add-grammar` would be preferable to bailing halfway, because each of the above three errors left the user with half-registered grammar modules...

Bear in mind I've never used Docker before, and I'm largely ignorant to how/why it's used. So this presents a pretty good example of how The Average User might feel when preparing a PR for adding a new language to Linguist.

Alhadis:
This PR adds rollback functionality to the `add-grammar` script to ensure all changes are reverted in the case of an error. Resolves #4056.

## Description
When an error happens for whatever reason, users are left with a partially-registered submodule in their working tree, which won't always be apparent. The script now encloses every command it executes inside a `begin…rescue` block to catch raised exceptions and perform cleanup before exiting.

@lildude, @pchaigno One of you guys will need to test this on your end (specifically the `--replace` option) because somehow the following line didn't update `vendor/README.md` correctly:

~~~console
$ script/add-grammar --replace language-javascript Alhadis/language-js
~~~

I'm assuming it may have something to do with the Docker-related stuff I'm unable to run...


Alhadis:
Thank you, @Popsoywhip, for your enlightening review and comprehensive feedback.

script/add-grammar
```diff
@@ -43,24 +43,32 @@ def log(msg)
...
   if !status.success?
-    output.each_line do |line|
-      log "  > #{line}"
+    output = output.each_line { |line| "  > #{line}" }
+    unless $aborted
+      $aborted = true
+      warn "Command failed. Aborting."
+      raise output
+    else
+      warn output
```
lildude: This seems overly complicated on quick inspection. 

Why do we need to differentiate behaviour here? Wouldn't we want to _always_ exit and clean up on failure?

script/add-grammar
```diff
@@ -81,41 +89,71 @@ $url = ARGV[0]
...
-  log "Deregistering: #{repo_old}"
-  command('git', 'submodule', 'deinit', repo_old)
-  command('git', 'rm', '-rf', repo_old)
-  command('script/grammar-compiler', 'update', '-f')
-end
+  # Ensure the given URL is an HTTPS link
+  parts     = parse_url $url
+  $https    = "https://#{parts[:host]}/#{parts[:user]}/#{parts[:repo]}"
+  $repo_new = "vendor/grammars/#{parts[:repo]}"
+  $repo_old = parse_submodule($replace) if $replace
```
lildude: These shouldn't need to be global as they're defined within the `begin`

script/add-grammar
Alhadis: If the command raises an exception, the output is discarded entirely. The globals appear scoped to the containing block, as they weren't showing up when output in the `rescue` part.

script/add-grammar
Alhadis: See above. They appeared to be block-scoped or something...

I'm currently on macOS so I can't rerun the script to reproduce (owing to that same `charlock_holmes` error 😞).

lildude:
Just tested this quickly as this PR stands and it doesn't do as I understand you'd hope on macOS, or at least how I'm understanding it should behave:

```console
$ git status
On branch script-fixes
Your branch is up to date with 'origin/script-fixes'.

nothing to commit, working tree clean
$ script/add-grammar --replace language-javascript Alhadis/language-js
Checking Docker is installed and running
$ docker ps
Deregistering: vendor/grammars/language-javascript
$ git submodule deinit -f vendor/grammars/language-javascript
$ git rm -rf vendor/grammars/language-javascript
$ script/grammar-compiler update -f
Registering new submodule: vendor/grammars/language-js
$ git submodule add -f https://github.com/Alhadis/language-js vendor/grammars/language-js
$ script/grammar-compiler add vendor/grammars/language-js
Confirming license
$ script/licensed
Updating grammar documentation in vendor/README.md
$ bundle exec rake samples
$ script/sort-submodules
$ script/list-grammars
script/list-grammars:81:in `block in to_markdown': undefined method `chomp' for nil:NilClass (NoMethodError)
	from script/list-grammars:65:in `each'
	from script/list-grammars:65:in `to_markdown'
	from script/list-grammars:96:in `update_readme'
	from script/list-grammars:105:in `<main>'
$ git status
On branch script-fixes
Your branch is up to date with 'origin/script-fixes'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   .gitmodules

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   .gitmodules
	modified:   grammars.yml

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	vendor/licenses/grammar/language-js.txt

$
```

Admittedly, this isn't the cleanest of clones and `script/list-grammars` behaves correctly on a clean clone. It does however illustrate that the failure is still leaving the repo in a messy state.

Sidenote: I've found running `bundle install --force` commonly solves my `charlock_holmes` library issues. I also use rbenv.

Alhadis:
You're right. Think I'll close it, because the complexity it's introducing is probably going to *add* to future headaches, rather than solve them. 😞

Alhadis:
> I also use rbenv.

@lildude Uh, crap. I just remembered I disabled something like that ages ago. Except it was called `rvm`, and I don't remember how it got onto my MacBook (or whether it was installed by default). I only remember a knee-jerk deletion when I noticed it had overridden the shell's `cd` with some arbitrary command. 😅

Would that have something to do with Charlock Holmes hating my guts?

(BTW, sorry for wasting your time with the PR, I wrote most of it while sleep-deprived 😢)

Issue #3808, opened 2017-09-02
Perl's xs ppport.h is detected as C++


Mons:
Many XS Perl libraries are detected as C++, since all of them include generated tool file/header `ppport.h` (ex: https://github.com/Mons/EV-Tarantool/search?l=C%2B%2B)
It is very big and with a great probability will become top1 language in repo.

I think it would  be good to include it into vendor.yml

lildude:
The comment in the file "Automatically created by..." suggests this may be better classified as "generated". It could also be classified as vendored as you suggested, but it really depends if you want changes to it to appear in PR diffs or not (vendored do, generated don't).

Please feel free to review https://github.com/github/linguist/blob/master/CONTRIBUTING.md and submit a PR.

lildude:
Flagging as stale. Linguist is behaving as expected as this file is technically a C-family header file, and we've not received a PR to change the behaviour.

PR #4106, opened 2018-04-18
[#3808] flag ppport.h files created by Devel::PPPort perl module as generated

ioanrogers:
Marking `ppport.h` files which are generated by [Devel::PPPort](https://metacpan.org/pod/Devel::PPPort) as generated.

## Checklist:


lib/linguist/generated.rb
```diff
@@ -533,5 +534,14 @@ def generated_grpc_cpp?
       return false unless lines.count > 1
       return lines[0].start_with?("// Generated by the gRPC")
     end
+
+    # Internal: Is the file a generated Perl/Pollution/Portability header file?
+    #
+    # Returns true or false.
+    def generated_perl_ppport_header?
+        return false unless name.match(/ppport\.h$/)
+        return false unless lines.count > 10
```
pchaigno: Why 10? Is that arbitrary?

ioanrogers: The string match in the next test is on line 9, so it's just to make sure the file is at least that long.

lib/linguist/generated.rb
```diff
@@ -533,5 +534,14 @@ def generated_grpc_cpp?
...
+        return lines[8].include?("Automatically created by Devel::PPPort")
```
lildude: Will this _always_ be on line 9?

ioanrogers: AFAICT, yes
https://github.com/mhx/Devel-PPPort/blame/d7849d96015b66d69606488aa65877dc2ba737a4/PPPort_pm.PL#L585

The git history goes back to 2012.

lildude: I think in that case we can probably skip the `lines.count > 10` check.

ioanrogers: In Perl, if you try to perform a match against something which is undefined you will get a warning.
I don't know Ruby - does the `include` call handle array elements which don't exist?

ioanrogers: Okay, I just checked with a truncated file and it seemed ok.

lildude: Ah, good point. 👍

lib/linguist/generated.rb
pchaigno: Ah, ok. I guess I was expecting `> 9` or a higher number for the whole header comment.

lildude:
This is looking good from my perspective, except for the heeeaaawwg `ppport.h` file. As this is only being used for testing, I think we can safely truncate it to about 30 lines with a comment at the point of truncation along the lines of "linguist: deliberately truncated for test purposes" or something along those lines.

Thoughts?

ioanrogers:
Done

PR #4118, opened 2018-04-27
Add compiled graphql relay to be considered generated

vkmita:
Marking `.graphql.js` files which are generated by [GraphQL relay-compiler](https://github.com/facebook/relay/tree/master/packages/relay-compiler) as generated.

Checklist:
- [x] I am adding new or changing current functionality


lib/linguist/generated.rb
```diff
@@ -533,5 +534,12 @@ def generated_grpc_cpp?
       return false unless lines.count > 1
       return lines[0].start_with?("// Generated by the gRPC")
     end
+
+    # Internal: Is this a relay-compiler generated graphql.js file?
+    #
+    # Return true or false
+    def generated_graphql_relay_js?
+      !!name.match(/__generated__\/[_A-Za-z][_0-9A-Za-z]+\.graphql\.js$/)
```
pchaigno: Shouldn't we mark the whole `__generated__` directory as generated?

pchaigno: Isn't the `__generated__` directory created by relay-compiler?

pchaigno: I'd probably mark the whole `__generated__` directory as generated, but then I don't know much about GraphQL and it looks like you have some reservations about it. Anything specific that makes you think this could be "unsafe" (i.e., not the expected user behavior)?

vkmita: That sounds good to me. In a GraphQL relay project it is definitely safe to mark all files under the `__generated__` directory as generated. I'm not sure about other languages/frameworks, and if they ever would use a `__generated__` directory for some other purpose... yeah probably never...

One sec will make the change

pchaigno: From what I could see in the search results, it looks like it's not really used outside of GraphQL. Worst-case scenario, we'll reverse if we have users complaining :/

lib/linguist/generated.rb
```diff
@@ -54,6 +54,7 @@ def generated?
       xcode_file? ||
       cocoapods? ||
       carthage_build? ||
+      generated_graphql_relay_js? ||
```
pchaigno: Maybe remove `_js` at the end of the name, now that we're marking the whole directory as generated?

vkmita: done!

Issue #4129, opened 2018-05-14
GML Projects get marked as Yacc because of .yy files


RobQuistNL:
Newer Game Maker Studio 2 projects will contain .gml files with GameMakerLanguage code in it, and corresponding .yy files - which contain metadata in a JSON structure. They get falsely detected as Yacc files.

We (in the community) have been working around this problem by adding ``*.yy linguist-language=Game Maker Language`` to our .gitattributes file.

We would prefer to have .GML detected as GML, and .yy as JSON. (Maybe even GML because its a part of the .gml files).

Is this something I could create an MR for? I'm not sure how to implement this case into this repository, but I'd love to.

> If there's a .yyp, a bunch of .gml and a bunch of .yy - I'm 99% sure its a GMS2 project, with .yyp and .yy being JSON files and .gml being GML.

### URL of the affected repository:
Example with override: https://github.com/GameMakerDiscord/handytools
Example without override: https://github.com/RobQuistNL/handytools

### Expected language:
GML, JSON

### Detected language:
GML, Yacc

pchaigno:
You could follow [the contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language) to add `.yy` as an extension to JSON. [It looks like the in-the-wild requirement is met](https://github.com/search?p=99&q=modelname+extension%3Ayy&type=Code&utf8=%E2%9C%93).

`.gml` should already be [recognized as Game Maker Language files](https://github.com/github/linguist/blob/59f5b05161e7b17236f90a6aa8f9085389d64b1f/lib/linguist/languages.yml#L1494). The detection is not 100% accurate though. One way to improve it would be to add new [heuristic rules](https://github.com/github/linguist/blob/59f5b05161e7b17236f90a6aa8f9085389d64b1f/lib/linguist/heuristics.rb) for `.gml`. Do you have an easy way to differentiate between Game Maker Language, Graph Modeling Language, and XML (I'm guessing not the three, but if you have something for Game Maker Language, that's already something we might be able to use)?

RobQuistNL:
Ah, the heuristic rules is probably something I'll look in to. It seemed like there were certain keywords and built-in functions missing, so I'll see if I can open up a PR for that. I'll open up a PR for the .yy as well.

kt-altyn:
has been over 5 years and not fixed lol

lildude:
[...]

PR #4130, opened 2018-05-14
Adding .yy and .yyp (GMS2 Project / metadata files) as JSON, and GML heuristics

RobQuistNL:
Fixes https://github.com/github/linguist/issues/4129

This change makes sure the .yy and .yyp files are correctly detected as JSON files within a Game Maker project.

I'm having difficulties running this tool in an example project - I could use some help with running my branch of linguist inside of an example project. (I get an error message ``Could not locate Gemfile or .bundle/ directory``)

## Description
Not sure how the Yacc language responds to this. Maybe I should also add some heuristic rules for .yy and .yyp files? They are pretty easily distinguishable from Yacc files.

## Checklist:
      
      - https://github.com/search?p=99&q=modelname+extension%3Ayy&type=Code&utf8=%E2%9C%93
      - https://github.com/search?p=99&q=modelname+extension%3Ayyp&type=Code&utf8=%E2%9C%93
    - Sample source(s):
      - https://github.com/GameMakerDiscord/depth-system/blob/master/Depth_Tutorial.yyp
      - https://github.com/GameMakerDiscord/depth-system/blob/master/views/2ea73365-b6f1-4bd1-a454-d57a67e50684.yy
      - https://github.com/RobQuistNL/GMOculus/blob/master/GMOculus.project.gmx
      - https://github.com/RobQuistNL/GMOculus/blob/master/objects/obj_control.object.gmx
      - https://github.com/zcbenz/yacci/blob/master/examples/funccal/calc.y
    - Sample license(s): MIT


lib/linguist/languages.yml
```diff
@@ -2079,6 +2079,8 @@ JSON:
   - ".topojson"
   - ".webapp"
   - ".webmanifest"
+  - ".yyp"
```
pchaigno: The file extensions (except the first) need to be sorted.

lib/linguist/heuristics.rb
```diff
@@ -541,5 +541,13 @@ def call(data)
       end
     end
 
+    disambiguate ".yy" do |data|
+      Language["JSON"] if /\"modelName\"\:\s?\"/.match(data)
+    end
```
pchaigno: We'll need to add a case for Yacc here.

lib/linguist/heuristics.rb
```diff
@@ -541,5 +541,13 @@ def call(data)
...
+
+    disambiguate ".yyp" do |data|
+      Language["JSON"] if /\"modelName\"\:\s?\"GMProject\"/.match(data)
+    end
```
pchaigno: Is this required? Did you find other languages using `.yyp`?

Alhadis: Just curious: is there a guarantee that only one or zero whitespace characters follow `modelName:`? If not, you should replace `\s?` with `\s*` so it'll match 2+ blanks instead of 1 or none. 😉

RobQuistNL:
Would it be OK to add a .gml check like this;
```rb
disambiguate ".gml" do |data|
      Language["Game Maker Language"] if /^\s*(var|sprite_|vk_|image_|draw_|visible|sprite_)/.match(data)
    end
```

to this PR as well? I suppose I will have to add checks for Graph Modeling Language and XML as well?

I'm sorry this is a bit slow & sluggish - I'm not familiar with Ruby. @pchaigno is there a IRC / slack / discord channel where I can ask for some guidance or this is place OK?

lib/linguist/heuristics.rb
pchaigno: Right, my *need* above might be a bit strong. It's *better* if we have a case for Yacc :smile: 
Is the presence of the `modelName` key sufficient to identify GML JSON files? (i.e., is it mandatory in those files?)

RobQuistNL: Pretty much. All the .yy files I've seen have a modelName. As does the .yyp file. But like you said - the .yyp file isn't documented in the languages.rb file so I've removed it in the latest commit. I've extended this testcase to first go for XML, then Graph Markup Language, and then a GameMaker file. The latter is harder because its a generic programming language and has no "required" keywords in the file. XML does, and it looks like Graph Markup Language does too (http://www.fim.uni-passau.de/fileadmin/files/lehrstuhl/brandenburg/projekte/gml/gml-technical-report.pdf) - the ``graph [`` and potentially the ``node [``.

lib/linguist/heuristics.rb
RobQuistNL: Good one, the * would be better. They are automatically generated files though, so I'm sure its usually 1 whitespace, but the * is better. Committing as we speak!

RobQuistNL: I've extended it since I've checked pretty much all .yy files, and they all contain a modelName element with a value that starts with ``GM``.

``"modelName": "GM`` a match on that one is what I'm doing for this rule.

RobQuistNL:
So the test is failing because;

> TestSamples#test_Yacc_has_samples [/home/travis/build/github/linguist/test/test_samples.rb:86]: 
> Missing samples in "samples/Yacc/*.yy".

I'm not sure where I can pull an example file from - I can find some example files online but I'm not sure about the licenses on that one

> TestSamples#test_JSON_has_samples [/home/travis/build/github/linguist/test/test_samples.rb:86]:
> Missing samples in "samples/Yacc/*.yy".
I suppose this is to check if Yacc files don't get detected as JSON, while GM files do? So thats 2 birds one stone

> TestClassifier#test_classify_ambiguous_languages:
> NoMethodError: undefined method `[]' for nil:NilClass
I'm not sure what to do with this one :(

RobQuistNL:
I've also taken this opportunity to add .gmx files. They are from GameMaker:Studio 1 (EOL). Haven't found any other .gmx files in the languages.yml so this should do just fine. 

GameMakerStudio 2 uses the .gml, .yy and .yyp files.

RobQuistNL:
Hey @pchaigno, are the changes I made valid like this?

RobQuistNL:
Ping @pchaigno :x sorry for my impatience :P

samples/Yacc/words2.yy
```diff
@@ -0,0 +1,53 @@
+/* File taken from http://home.adelphi.edu/~sbloch/class/271/examples/lexyacc/words2.y . */
```
pchaigno: Couldn't find the license for this. Is it indicated anywhere on the website?

RobQuistNL: Me neither, I find it very hard to find an example file for the .y files with a license - Since it was published like this online on an .edu site I was assuming. As we all know, assumptions are the mother of all f*ckups, so I'll have to find something else for this.

The GMX files are my own, I'll add a MIT license to the repository there.

RobQuistNL:
@pchaigno Sorry for the inconvenience.

I've included a Yacc file from an MIT repository (https://github.com/zcbenz/yacci/blob/master/examples/funccal/calc.y) and I've added the MIT license to my own repository for the 2 GMX files.

lildude:
The test failures appear to be caused by recent changes to licensee which has resulted in a slightly modified hash for the four licenses in question. I'll address this in https://github.com/github/linguist/pull/4152

lildude:
Merging master into your branch should fix the failing tests.

RobQuistNL:
Thanks @lildude !

@pchaigno We should be good to go now I think :)

lib/linguist/heuristics.rb
```diff
@@ -226,6 +226,16 @@ def call(data)
...
     end
 
+    disambiguate ".gml" do |data|
+      if /^\s*(\<\?xml|xmlns)/i.match(data)
+        Language["XML"]
+      elsif /^\s*(graph\s?\[|node\s?\[)/i.match(data)
+        Language["Graph Modeling Language"]
+      else
+        Language["Game Maker Language"]
+      end
```
pchaigno: I don't think we can be sure we've eliminated the two other candidate languages here. Are we even sure `/^\s*(graph\s?\[|node\s?\[)/i` will match all Graph Modeling Language files?

lib/linguist/heuristics.rb
```diff
@@ -531,5 +541,13 @@ def call(data)
       end
     end
 
+    disambiguate ".yy" do |data|
+      if /\"modelName\"\:\s*\"GM/.match(data)
```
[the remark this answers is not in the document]
RobQuistNL: They are generated, but I've only seen them with the double quotes. I put the ``\s`` in there just in case

lib/linguist/heuristics.rb
RobQuistNL: As far as I could tell the Graph modeling Languages always have a ``graph [`` or ``node [`` element. 

http://www.fim.uni-passau.de/fileadmin/files/lehrstuhl/brandenburg/projekte/gml/gml-technical-report.pdf

I haven't found any docs, but I'm quite sure that it then will not be GML (``text [`` is invalid syntax), unless its in a comment ofcourse.

Maybe that would be a nicer test?

``([a-zA-Z]?\s?\[)`` being Graph Modeling Language?

RobQuistNL: Here there is some more documentation. All files I've seen so far have that ``graph [`` and ``node [`` stuff.

http://www.fim.uni-passau.de/index.php?id=17297&L=1

RobQuistNL:
I've edited the Graph Modeling Language regex - it will now match most of the files found here;

https://github.com/search?p=9&q=extension%3Agml+graph&ref=searchresults&type=Code&utf8=%E2%9C%93

I can't find any hard evidence that ``graph [`` or ``node [`` _has_ to be in a graph modeling language file - but all the examples I've found have that setup, and so do the files in the search results.

lib/linguist/heuristics.rb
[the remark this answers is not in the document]
pchaigno: Ok, thanks for the precision. We should probably add a case in [`generated.rb`](https://github.com/github/linguist/blob/a5df9a00ab7c9828bd7038bb9f9bd5e56d325dc9/lib/linguist/generated.rb) for those files then.

pchaigno:
> Hmm, might be an issue, since ``graph [4] = 20;`` is actually valid GML and it will match. Most people will write array accessors like ``graph[4] = 20;`` without the space though.. :/ 

We could anchor the `[` to the end of the line?

RobQuistNL:
@pchaigno I've edited the regex so that the whitespace between ``graph`` and ``[`` _has_ to be there. Anchoring to the end of the line wont work, because its valid graph language to write it in 1 line.

I've tried messing around with whitespaces / newlines and number of values within the tags, but its hard. I think this regex gives a good cover. In a very small chance that you write
```
graph [23] = 12;
```
in GML, it might get triggered as being the Graph Language.

Its still better than what it is now :) But its not perfect.

pchaigno:
> Anchoring to the end of the line wont work, because its valid graph language to write it in 1 line.

Is that common though?

> In a very small chance that you write `graph [23] = 12;` in GML, it might get triggered as being the Graph Language.

It *will* be recognized as Graph Modeling Language. And it doesn't look that small a chance to me; I've already seen white spaces in that position before.

RobQuistNL:
Allright, I've added the end string delimiter - this is pretty much sure to not cause game maker language to be recognized as Graph Markup Language;

image

There is only 1 extreme case in Game Maker Language in which it will trigger, and 1 case (the onliner) in which Graph won't trigger;

```
graph [
	i*2 + offset + 28
] ;
```

But I think thats much less common..

pchaigno:
> But I think thats much less common..

:laughing: I think so too. If issues are reported with the current heuristic rule, we'll try to add one for Game Maker Language.

lib/linguist/heuristics.rb
```diff
@@ -226,6 +226,16 @@ def call(data)
       end
     end
 
+    disambiguate ".gml" do |data|
+      if /^\s*(\<\?xml|xmlns)/i.match(data)
+        Language["XML"]
+      elsif /^\s*(graph[\s\n\r]{1,}\[$)|(node[\s\n\r]{1,}\[$)/i.match(data)
```
pchaigno: `{1,}` can be written `+`, which is more widely supported by regex engines (Oniguruma, PCRE), even though Ruby uses Oniguruma anyway. We can also simplify to `^\s*(graph|node)[\s\n\r]+\[$`.

RobQuistNL: Thanks! It was quite a lame regex indeed. Updated :)

lib/linguist/heuristics.rb
```diff
@@ -226,6 +226,16 @@ def call(data)
       end
     end
 
+    disambiguate ".gml" do |data|
+      if /^\s*(\<\?xml|xmlns)/i.match(data)
+        Language["XML"]
+      elsif /^\s*((graph|node)+[\s\n\r]+\[$)/i.match(data)
```
pchaigno: We don't need the outer parentheses here. Also, is the `+` after `(graph|node)` an error?

pchaigno:
Could you also add a new test case in [`test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb) for the `.gml` extension?

lib/linguist/heuristics.rb
RobQuistNL: Ah, you're right ^_^

RobQuistNL:
Here goes :crossed_fingers: 

Thanks a ton for your quick responses and your patience @pchaigno! I really appreciate it.

lib/linguist/heuristics.rb
```diff
@@ -226,6 +226,16 @@ def call(data)
       end
     end
 
+    disambiguate ".gml" do |data|
+      if /^\s*(\<\?xml|xmlns)/i.match(data)
+        Language["XML"]
+      elsif /^\s*(graph|node)\s+\[$/i.match(data)
```
pchaigno: We still want to match on:
```
graph
[
    foobar
```
right?
Any reason you changed the `[\s\n\r]`?

RobQuistNL: I thought \s would include \n and \r, hence I removed it

RobQuistNL: ``\s matches any whitespace character (equal to [\r\n\t\f\v ])``

RobQuistNL: This format is in sample.gml and now also in sample3.gml. I think the succeeding tests should prove that the regex works, right? :)

test/test_heuristics.rb
```diff
@@ -139,6 +139,13 @@ def test_fs_by_heuristics
     })
   end
 
+  def test_gml_by_heuristics
+    assert_heuristics({
+      "Game Maker Language" => all_fixtures("Game Maker Language", "*.gml"),
+      "Graph Modeling Language" => all_fixtures("Graph Modeling Language", "*.gml")
```
pchaigno: Could we add a case for XML too?

lib/linguist/heuristics.rb
pchaigno: Oh, right! my bad!

test/test_heuristics.rb
RobQuistNL: Good catch!

lib/linguist/heuristics.rb
RobQuistNL: My bad for initially writing \s\r\n ;)

RobQuistNL:
@pchaigno Ping! Congratulations on winning the finals BTW :) Hope your hangover is not too bad ;)

lib/linguist/heuristics.rb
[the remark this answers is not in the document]
pchaigno: > do you think we need the same check in generated.rb?

Yes, I think it's worth it. Is that the only indication that the file is generated? We often have clear comments such as `Generated by XXXX`.

RobQuistNL:
@pchaigno Sorry :x ping again

lib/linguist/heuristics.rb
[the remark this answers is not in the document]
RobQuistNL: Unfortunately GM doesn't do that. Its plain JSON, but at least it has a consistent format. I'll add a case to generated.rb.

lib/linguist/generated.rb
```diff
@@ -577,5 +578,17 @@ def generated_perl_ppport_header?
...
+
+    # Internal: Is this a generated Game Maker Studio (2) metadata file?
+    #
+    # All Game Maker Studio 2 generated files will be JSON, .yy or .yyp, and have
+    # a part that looks like "modelName: GMname" on the 3rd line
+    #
+    # Return true or false
+    def generated_gamemakerstudio?
+      return false unless extname == '.yy' || extname == '.yyp'
+      return false unless lines.count > 5
```
pchaigno: Why require min. 6 lines if we only look at the third?

lib/linguist/generated.rb
```diff
@@ -577,5 +578,17 @@ def generated_perl_ppport_header?
...
+      return /\"modelName\"\:\s*\"GM/.match(lines[3])
```
pchaigno: In the comment you mention the third line, but here you match on the fourth. Typo?

RobQuistNL: Sorry, yeah, was already on it ^_^

RobQuistNL: (can't run the tests locally so abusing Travis for it :x )

RobQuistNL:
@pchaigno The generated test should work now ^_^

RobQuistNL:
Oh, if anything I should thank _you_ for your patience @pchaigno !

RobQuistNL:
Awesome! Thanks :)

PR #4131, opened 2018-05-14
Consider Gopkg.lock and glide.lock as generated

burke:
Recognize [`Gopkg.lock`](https://golang.github.io/dep/) and [`glide.lock`](https://glide.sh/) — two common Go package manager lockfiles — as generated.

## Checklist:
  

samples/Go/filenames/glide.lock
```diff
@@ -0,0 +1,10 @@
+hash: e7e387e385240ea7d48e6de6a0c58fe523bfefa1d7895a2c25c100feb978774d
```
pchaigno: Is this format YAML?

samples/Go/filenames/Gopkg.lock
```diff
@@ -0,0 +1,21 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
pchaigno: What format is this file?

burke:
Ah, I completely misunderstood how the tests worked. Revised.

lib/linguist/languages.yml
```diff
@@ -5273,6 +5274,7 @@ YAML:
   - ".clang-format"
   - ".clang-tidy"
   - ".gemrc"
+  - "glide.lock"
```
pchaigno: I'm clearly nitpicking here, but the double quotes are not required when the name doesn't start with a dot (we try to follow Ruby's YAML format).

PR #3650, opened 2017-05-27
Add recognition and classification of WebAssembly

Alhadis:
This PR adds support for WebAssembly's textual representation on GitHub (`.wast` and `.wat`) files. Use of these files has spread substantially since the working group announced [cross-browser consensus](http://webassembly.org/roadmap/). In usual fashion, I've done what I could to collate in-the-wild examples:

* [`.wast`](https://github.com/Alhadis/Linguist-Silo/tree/wast): 4,705 files
* [`.wat`](https://github.com/Alhadis/Linguist-Silo/tree/wat) 25,812 files

The `wast` results were very promising; virtually every result was WebAssembly. The `wat` search, while offering a healthy amount of WebAssembly too, tossed up an unusual number of Makefiles and other junk related to "Watcom" (probably [Watcom C](https://en.wikipedia.org/wiki/Watcom_C/C%2B%2B0)…?) The biggest offender here are `makefile.wat` files, [which as you'll see](https://github.com/Alhadis/Linguist-Silo/tree/wat/files/Makefile), add up to nearly 800 unique variations from collated results (that's with duplicates removed).

## Licenses of example files
From **[WebAssembly/binaryen](https://github.com/WebAssembly/binaryen/)** ⇒ [Apache 2.0](https://github.com/WebAssembly/binaryen/blob/master/LICENSE)
+ [`imported-min.wast`](https://raw.githubusercontent.com/WebAssembly/binaryen/0b8acfc0139652d1c0f284c7650812a309574586/test/ctor-eval/imported-min.wast)
+ [`local-cse.wast`](https://raw.githubusercontent.com/WebAssembly/binaryen/1f1018564c472b75335bf4bddbf65cfebb226fb0/test/passes/local-cse.wast)
+ [`remove-unused-brs_shrink-level%3D1_ignore-implicit-traps.wast`](https://raw.githubusercontent.com/WebAssembly/binaryen/1f1018564c472b75335bf4bddbf65cfebb226fb0/test/passes/remove-unused-brs_shrink-level%3D1_ignore-implicit-traps.wast)

From **[bzar/wasm-intro](https://github.com/bzar/wasm-intro)** ⇒ [MIT/Expat license](https://github.com/bzar/wasm-intro/blob/master/LICENSE)
+ [`add.wat`](https://raw.githubusercontent.com/bzar/wasm-intro/d53c653b35649f9433ef5e7f5fff9da49483221b/wat/add.wat)
[...]

PR #4133, opened 2018-05-16
Add support and recognition for "X BitMap" images

Alhadis:
This PR introduces support for a text-based bitmap image format called [**X BitMap**](https://en.wikipedia.org/wiki/X_BitMap). (Not to be confused with the currently-supported and superficially similar format, [**X PixMap**](https://en.wikipedia.org/wiki/X_PixMap).

## Description
This PR also adjusts XPM's `languages.yml` entry so it's technically correct and uses its longer, human-readable name (since that's how those images get referred to everywhere else).

Both formats use C syntax, so I've lumped them under the `C` group whilst leaving their language-types to `data` (since that's ultimately what they are).

## In-the-wild usage
[~364,280 results](https://github.com/search?q=extension%3Axbm+NOT+nothack&type=Code) for `xbm` files, none of them shared with a recognised language. From what I've gone through, they're all XBM data.

## Sample files
**Update:**  
Replaced the sample images with a [manually-converted one](https://github.com/github/linguist/pull/4133#issuecomment-390923936) that's free from licensing complications.
~~Both of the `xbm` samples I've added were hocked from [`w3c/libwww`](http://github.com/w3c/libwww):~~
* ~~[`image.xbm`](https://github.com/w3c/libwww/blob/315b6faf1d4ab0af71cec9b41ca9aa23c11c12ee/Icons/internal/image.xbm)~~
* ~~[`doc.xbm`](https://github.com/w3c/libwww/blob/315b6faf1d4ab0af71cec9b41ca9aa23c11c12ee/Icons/internal/doc.xbm)~~

~~Licensing-wise, they're released under the [**W3C Software Notice and License**](https://en.wikipedia.org/wiki/W3C_Software_Notice_and_License), which is new to me. Wikipedia's page for the W3C's license states that it's acceptably permissive and non-copyleft, though it mentions GPIL compatibility.~~

~~The full license text [can be found here](https://www.w3.org/Consortium/Legal/2002/copyright-software-20021231). Should we summon the wisdom of our venerable LawGuruGuy™, @mlinksva? Even though this license *looks* okay, I dare not tread the swampy waters of *"Legal-Looking Stuff That Will Actually Get You Life-Jailed If You Read It Wrong"*. Hereafter known as `LLSTWAGYLJIYRIW`, my new favourite Perl motto.~~


pchaigno:
> Should we summon the wisdom of our venerable LawGuruGuy™, @mlinksva? Even though this license *looks* okay, I dare not tread the swampy waters of *"Legal-Looking Stuff That Will Actually Get You Life-Jailed If You Read It Wrong"*. Hereafter known as `LLSTWAGYLJIYRIW`, my new favourite Perl motto.

Didn't we discuss copyleft licenses not being an issue for sample files not so long ago? (I can't find the issue/PR again :-/)

lib/linguist/languages.yml
```diff
@@ -5022,6 +5022,31 @@ World of Warcraft Addon Data:
...
+  aliases:
+  - xbm
+  extensions:
+  - ".xbm"
+  ace_mode: c_cpp
+  tm_scope: source.c
+  codemirror_mode: clike
+  codemirror_mime_type: text/x-csrc
+  language_id: 782911107
+X PixMap:
```
lildude: Aaaah, man. This is a PITA for me. I need to relearn Rails migrations if we do this. If I don't put in a migration, we end up with the same problems we had with the Pelr6 -> Perl 6 change and the DB and search gods frown upon me.

Alhadis: Well, erh, should we then start thoroughly reviewing the hell out of every single grammar-name to make any other potential future change happen in one migration? 😅

I didn't want to rename the entry, as I *am* conscious of how much of a pain it is database-wise. 😞 Unfortunately, I can't change fact. :( That's a fact.

Alhadis: If the KiCad incident last year proved anything, it's that inaccuracies in `language.yml` are better dealt with much sooner than a hell of a lot later. 😥

mlinksva:
[W3C](https://spdx.org/licenses/W3C.html) is fine. I don't see a copy of it in this PR though.

Alhadis:
Oh. No, I didn't include a copy as I was asking about licenses for sample files.

> Didn't we discuss copyleft licenses not being an issue for sample files not so long ago? (I can't find the issue/PR again :-/)

So I'm guessing @pchaigno is right?

mlinksva:
@Alhadis if we're relying on a license to distribute sample material from this repo, we need to comply with the license we're relying on. The W3C license requires keeping a copy of the license. I recall this coming up before regarding samples: https://github.com/github/linguist/pull/3650#issuecomment-304416394

mlinksva:
@Alhadis :cool:. Side note: slightly better, or additional, link in your comment above would be to https://commons.wikimedia.org/wiki/File:Cc-public_domain_mark_white.svg (the page about the file, which on Wikimedia Commons always includes copyright/license/public domain info). It isn't a given that an image about the public domain will itself be public domain 😬 though thankfully this one is.

PR #4134, opened 2018-05-16
Silence warnings being emitted when running tests

Alhadis:
I saw these warnings as I was running tests for #4133:

~~~
lib/linguist/heuristics.rb:414: warning: character class has duplicated range: /[A-Z\.][\w\d\.]*:{/
lib/linguist/language.rb:504: warning: assigned but unused variable - filenames
~~~

They're trivial but still unrelated to what #4133 is doing, so I siphoned these changes off to a separate PR. :-) Explanation of changes:

> warning: character class has duplicated range: `/[A-Z\.][\w\d\.]*:{/`

That's happening because `\w` already covers the character ranges that `\d` does. Using them together in a bracketed character class is therefore redundant.

I also removed the backslash from the dots too, because hold no special meaning when used within a character class. It wouldn't make sense for it to match any single character in a defined subset of specific characters you're trying to match. =)

> warning: assigned but unused variable - filenames

Admittedly, I don't even know what this line is doing:

~~~rb
filenames = Samples.cache['filenames']
~~~

But I didn't wanna remove it in case it was there to deliberately invoke a side-effect by accessing `Samples.cache['filenames']`. But I wanted it to shut up so I found a way to make it "used" by sprucing up a conditional assignment:

~~~rb
-    if interpreters == nil	+    interpreters = {} if interpreters.nil?
-      interpreters = {}	+    filenames    = {} if filenames.nil?
-    end
+    interpreters = {} if interpreters.nil?
+    filenames    = {} if filenames.nil?
~~~

Which inevitably lead to me polishing whitespace elsewhere in the file, as you'll see by the diffs. :grin: I'd apologise for being a neat-freak, but you guys know I'm not actually sorry. :heart:

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -27,9 +27,9 @@ module Linguist
     @alias_index        = {}
     @language_id_index  = {}
 
-    @extension_index          = Hash.new { |h,k| h[k] = [] }
-    @interpreter_index        = Hash.new { |h,k| h[k] = [] }
-    @filename_index           = Hash.new { |h,k| h[k] = [] }
+    @extension_index    = Hash.new { |h,k| h[k] = [] }
+    @interpreter_index  = Hash.new { |h,k| h[k] = [] }
+    @filename_index     = Hash.new { |h,k| h[k] = [] }
 
     # Valid Languages types
     TYPES = [:data, :markup, :programming, :prose]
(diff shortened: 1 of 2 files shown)


lib/linguist/language.rb
```diff
@@ -499,12 +499,12 @@ def inspect
     end
   end
 
-  extensions = Samples.cache['extnames']
+  extensions   = Samples.cache['extnames']
   interpreters = Samples.cache['interpreters']
-  filenames = Samples.cache['filenames']
-  popular = YAML.load_file(File.expand_path("../popular.yml", __FILE__))
+  filenames    = Samples.cache['filenames']
```
pchaigno: This is safe to remove. I forgot to do it in 0bf4b8a4826356a882c2baca16722991aa4c4b0c.

Alhadis: Perfect! Works for me. :grinning: Consider it gone, and my drastic spike of Ruby fluency to be concerning

(for me)

Alhadis: And don't worry, I *did* double-check the `||=` operator to guarantee it did [what I thought it does](http://www.rubyinside.com/what-rubys-double-pipe-or-equals-really-does-5488.html). :wilted_flower:

There's no good reason I should be this good at Ruby, I'm always bitching about why its ecosystem are the antichrist :cry:

Alhadis:
Now we just need to get Charlock Holmes to STFU about that encoding variable not being initialised. 😂👌

PR #4147, opened 2018-05-30
Revert "Add Pod 6 to the languages.yml"

lildude:
Reverts github/linguist#4083

As per my comment at 

> Unfortunately, I'm going to have to revert this and open a new PR for it for the moment as it requires Pod 6 rendering support to be implemented in github/markup#1173 or github/markup#1185 and GitHub.com updated. This needs to coincide with the release of a Linguist release that introduces support for Pod 6 so we don't end up leaving people scratching their heads trying to work out why one pod document is being rendered and not another.
> 
> I don't currently have the bandwidth to look into the markup side of things either.

... and we need to get a new Linguist release out.

I'll open a new PR to revert this revert once this has been merged.  That new PR will be dependent on Pod 6 rendering support being added to markup.


lildude:
Whoops. That's not what I wanted. Lets try that again manually this time.

PR #4149, opened 2018-05-30
Add Pod 6 to the languages.yml [Pending markup support]

lildude:
This is a revert of the revert in https://github.com/github/linguist/pull/4148 of https://github.com/github/linguist/pull/4083 as explained at https://github.com/github/linguist/pull/4148#issue-191418902:

> As per my comment at https://github.com/github/linguist/pull/4083#issuecomment-393121096
> 
>> Unfortunately, I'm going to have to revert this and open a new PR for it for the moment as it requires Pod 6 rendering support to be implemented in github/markup#1173 or github/markup#1185 and GitHub.com updated. This needs to coincide with the release of a Linguist release that introduces support for Pod 6 so we don't end up leaving people scratching their heads trying to work out why one pod document is being rendered and not another.
>> 
> I don't currently have the bandwidth to look into the markup side of things either.
>
> ... and we need to get a new Linguist release out.
>
>I'll open a new PR to revert this revert once this has been merged.  That new PR will be dependent on Pod 6 rendering support being added to markup. 

This PR can only be merged when markup has support for rendering Pod 6 via https://github.com/github/markup/pull/1173 or https://github.com/github/markup/pull/1185. A combined Linguist and markup update can then be made on GitHub.com to combine the two.


Tyil:
Any updates on this? Anything that still needs to be done? We're currently a couple months further down the road, and the latest comment we had was that someone else needs to look at it, which seems to have never happened. The latest ping from @JJ in https://github.com/github/markup/pull/1173 got no response either.

lildude:
Nope. This is still blocked by the markup side of things.

Tyil:
I understand that, but we're getting no response in that repository for a couple of months there too.

lib/linguist/heuristics.rb
```diff
@@ -380,6 +380,16 @@ def call(data)
       end
     end
 
+    disambiguate ".pod" do |data|
+      if /^\s*\/\* XPM \*\//.match(data)
```
kivikakk: This was copied from `.pm`, but I think isn't necessary for pod.

kivikakk: Have removed from `heuristics.yml` on the tip of this branch. Haven't regenerated a gem for testing because it won't actually come up.

kivikakk:
Brought up to date.

lildude:
Thanks @kivikakk. Merging this so chain of PRs can progress.

JJ:
Thanks all!

Issue #3628, opened 2017-05-16
Dockerfile showing as shell in language stats bar


esalter-va:
I recently created [esalter-va/shairport-sync-docker-pi](https://github.com/esalter-va/shairport-sync-docker-pi) and noticed that the language stats bar says it's 100% shell.

I have 2 files other than `README.md` and `LICENSE`, and they are as follows:

|Name|Bytes|
|---|---|
|`Dockerfile`|673|
|`start.sh`|74|

So the bar should be closer to 90% Dockerfile, 10% Shell.  It might be because the Dockerfile doesn't have an extension, but it would be nice if it was recognized in the language stats.

lildude:
This is expected. Dockerfile, like SQL, is not counted in statistics because it's considered a data language by Linguist.

pchaigno:
@lildude It seems weird to categorize Dockerfiles as data. In the end, they're just scripts. Should we reconsider this?

lildude:
Sure, I thought it was odd too, but it's addition pre-dates me. 

@vmg was there any particular reason you marked Dockerfile as data when adding it in https://github.com/github/linguist/pull/1714?

(Re-opening to continue the discussion)

ElectricWarr:
Coming across this issue almost a year on - I also would have expected to see Docker or Dockerfile in the languages bar by now.

Seems all it would take would be to replace the `type` here with `programming`:
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1060

@lildude / @vmg any further? It would be really interesting to see the effect this could have on the site - take this search, for example: https://github.com/search?utf8=%E2%9C%93&q=Dockerfile&type=

PR #4169, opened 2018-06-21
Make Dockerfile a programming language

lildude:
## Description
Dockerfiles have grown hugely in popularity since first added to Linguist and are essentially scripts. To quote https://docs.docker.com/engine/reference/builder/:

> A Dockerfile is a text document that contains all the commands a user could call on the command line to assemble an image. 

... which is essentially what a shell script it.  This PR flips the categorisation from `data` to `programming` so these files are now included in the language stats.

## Checklist:
  
No need for new or different tests as we're just changing how a language is classified.

#### TODO: 
- [x] Find a colour that is representative, but doesn't conflict - gonna be tough as Docker ❤️ blue, like everyone else.

Fixes https://github.com/github/linguist/issues/3628

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1057,7 +1057,7 @@ Diff:
   codemirror_mime_type: text/x-diff
   language_id: 88
 Dockerfile:
-  type: data
+  type: programming
   tm_scope: source.dockerfile
   extensions:
   - ".dockerfile"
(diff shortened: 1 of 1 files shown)


lildude:
Wow! First try at picking the colour and it's a 🎯, courtesy of https://brandcolors.net/b/docker 🎉

PR #4175, opened 2018-06-25
Fix up macOS build

kivikakk:
Fix the macOS build by not defining `strndup` for it; it's already in the standard library.

/cc @lildude


kivikakk:
Thank you!

PR #4182, opened 2018-07-03
Add color for the GDScript language (#355570)

aaronfranke:
## Description

I've added coloring (`#355570`) to the GDScript programming language, since I got tired of seeing gray. `#478cbf` is the hex code for the color used in the Godot logo and icon, which would be the best to use, but conflicts. However this color (`#355570`) also works well, it's the background color for the icon:

icon

I would make it brighter, but if I did, then it would conflict with other colors.

## Checklist:

I removed the checklist as none of the items applied.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1450,6 +1450,7 @@ GDB:
   language_id: 122
 GDScript:
   type: programming
+  color: "#355570"
   extensions:
   - ".gd"
   tm_scope: source.gdscript


aaronfranke:
> - GDScript (#478cbf) is too close to ["6594b9", "3F85AF", "478cbf", "0096D8", "198CE7"]

These colors seem fairly distinct to me, but if GDScript is too similar to them, why aren't they listed as too similar to each other?

lildude:
> These colors seem fairly distinct to me, but if GDScript is too similar to them, why aren't they listed as too similar to each other?

I don’t fully understand the [colour mathematics used to determine the proximity of colours](http://www.ece.rochester.edu/%7Egsharma/ciede2000/ciede2000noteCRNA.pdf) but my vague understanding is the proximity calcs are detecting that the current colours are sufficiently different, but as soon as you add your colour, it ends up at a point within the others that has a knock on effect on the proximity calcs for the others once your colour is added to the mix.

In simple terms take: `A xxxxxxxxxx B` vs `A xxxx C xxxx B` and a “threshold” of 5.

A is 8 x’s away from B, but as soon as we put C in, there’s only a “distance” of 4 between C and another letter, thus it no longer meets our threshold requirement.

This is something the maintainer of the [color-proximity gem](https://github.com/gjtorikian/color-proximity) may have more knowledge about. We use that gem for the proximity test.

> In any case, none of those languages would ever be used in combination with GDScript. Godot only supports GDScript, C#, C++, visual scripting, and in the future perhaps F#.

_You_ might not, but others may 😉

> I assume there's a way to tell it to skip comparing GDScript to those languages, since it obviously doesn't mind that Clean is too similar to the others.

Nope as you can not predict how peeps will mix and match the content of their repos. Sadly, you’ve gone for a shade of blue and this is by far the most popular colour. You’re going to need to keep trying until you find another colour. 

And no, we can’t decrease the threshold used in the test as this brings differences down too low for most peeps to distinguish.

aaronfranke:
@lildude I think there's something here that's causing the proximity checker to ignore the existing colors. I just pushed a test commit to make it equal to Clean's color, and it does not give any warnings about proximity to Clean's color despite the colors being equal - the test passes.

It would be preferable to use Godot's official color of `#478cbf`, but if the requirement is that the Travis CI tests need to pass, then this color is close enough and I can squash the commits so we can merge.

lildude:
Hmmm, I thought we had a test for duplicate colours. Guess not (I’ll add one soon). We can’t really have duplicates for anything other than the default fallback grey. Everything else should be unique.

Don’t worry about squashing commits; we squash on merge anyway.

aaronfranke:
Looks like my patch is live, and it looks great. Thanks everyone!

gds-color

PR #4185, opened 2018-07-09
fix php disambiguation regex

smola:
## Description
Because of an error in the regular expression, disambiguation rule for the .php
matched any non-empty file as PHP, except for Hack files.

Note that this means that now files with .php but containing no `<?` will be not classified as PHP unless other strategy kicks in. If that is not the desired behaviour, we should just remove the PHP regex and make it explicit that it is a catch-all fallback.

## Checklist:
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -355,7 +355,7 @@ module Linguist
     disambiguate ".php" do |data|
       if data.include?("<?hh")
         Language["Hack"]
-      elsif /<?[^h]/.match(data)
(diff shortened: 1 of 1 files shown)


smola:
Not really. A real world example would involve a file with `.php` extension that is not PHP (or Hack) at all.

pchaigno:
Hm, it's probably not worth it to add a new *fixture file* (files used for tests only, as opposed to *sample files*) then.

LGTM!

PR #4186, opened 2018-07-09
move pure C headers to C samples directory

smola:
## Description
Move pure C header samples from C++ directory to C. In the absence of C++ specifics, it is more reasonable to expect classification of C, which is technically correct.

See https://github.com/github/linguist/issues/1626 for more discussion about the topic.

## Checklist:


smola:
This are as much valid C++ as valid C. Its only effect seems biasing the classifier towards C++.

smola:
Closing. I might add another proposal if this is meaningful to improve classifier performance.

PR #4187, opened 2018-07-09
classify XML files with .ncl extension as XML

smola:
## Description
When NCL support was added (see #2440), XML files with `.ncl` extension were classified as Text. This PR moves them to XML adding a simple heuristic.

Tested on https://gist.github.com/pchaigno/2478bc86dd831382bfba by @pchaigno and, at least for the files that still exist today, this seems to be working correctly.
XML files with .ncl extensions were being classified as text.

## Checklist:


smola:
:+1: Done.

PR #4195, opened 2018-07-12
Use a tempfile instead of building our own.

tenderlove:
This commit changes cache writing to use the Tempfile class rather than
constructing our own tempfile.  It means we don't have to deal with
unlinking the tempfile in case there's an exception, and it also fixes
this code on Ruby 2.5 (`Dir::Tmpname.make_tmpname` is no longer
available)

Before this change:

```
[aaron@TC linguist (master)]$ ruby -I lib:~/git/charlock_holmes/lib:~/github/rugged/lib bin/git-linguist -c 15885701cd883b633847979583f128e4ba1beac9 breakdown
{"Ruby":["Gemfile","Rakefile","bin/git-linguist","bin/linguist","github-linguist.gemspec","lib/linguist.rb","lib/linguist/blob.rb","lib/linguist/blob_helper.rb","lib/linguist/classifier.rb","lib/linguist/file_blob.rb","lib/linguist/generated.rb","lib/linguist/grammars.rb","lib/linguist/heuristics.rb","lib/linguist/language.rb","lib/linguist/lazy_blob.rb","lib/linguist/md5.rb","lib/linguist/repository.rb","lib/linguist/samples.rb","lib/linguist/shebang.rb","lib/linguist/strategy/extension.rb","lib/linguist/strategy/filename.rb","lib/linguist/strategy/modeline.rb","lib/linguist/tokenizer.rb","lib/linguist/version.rb","script/add-grammar","script/convert-grammars","script/fast-submodule-update","script/licensed","script/list-grammars","script/prune-grammars","script/set-language-ids","test/helper.rb","test/test_blob.rb","test/test_classifier.rb","test/test_color_proximity.rb","test/test_file_blob.rb","test/test_generated.rb","test/test_grammars.rb","test/test_heuristics.rb","test/test_instrumentation.rb","test/test_language.rb","test/test_md5.rb","test/test_modelines.rb","test/test_pedantic.rb","test/test_repository.rb","test/test_samples.rb","test/test_shebang.rb","test/test_tokenizer.rb"],"Shell":["script/bootstrap","script/cibuild","script/travis/before_install"]}
Traceback (most recent call last):
	4: from bin/git-linguist:150:in `<main>'
	3: from bin/git-linguist:135:in `git_linguist'
	2: from bin/git-linguist:37:in `linguist'
	1: from bin/git-linguist:50:in `save_language_stats'
bin/git-linguist:80:in `write_cache': uninitialized constant GitLinguist::Tempfile (NameError)
```

[...]

PR #4199, opened 2018-07-18
drop cache tempfile in same dir

kivikakk:
See https://github.com/github/linguist/pull/4195#issuecomment-405813272. We create the tempfile in the same directory as the cache file so when we do the rename it's not cross-device.

/cc @tenderlove @lildude


kivikakk:
Confirmed this fixes the issue seen 👍

PR #4202, opened 2018-07-19
Update YARA language config

wesinator:
## Description
* Changes YARA language ace mode to c_cpp
* Change type to programming

These are more appropriate for the syntax

## Checklist:
  

pchaigno:
@lildude Are you making progress on Lightshow? It would be good if we could test this.

lildude:
> @lildude Are you making progress on Lightshow? It would be good if we could test this.

Unfortunately not. It's out of my hands at the moment.

lib/linguist/languages.yml
```diff
@@ -5303,12 +5303,12 @@ YANG:
...
   language_id: 408
 YARA:
-  type: data
-  ace_mode: text
+  type: programming
   extensions:
   - ".yar"
   - ".yara"
   tm_scope: source.yara
+  ace_mode: c_cpp
```
pchaigno: @wesinator The syntax looks very different from C++. Did you check the output this mode produces for YARA? Could you add a screenshot?

wesinator: The comments, indentation, and bracketing are similar, but if it adds C/Cpp keywords then I'll change it back.

pchaigno: @lildude Any way to test this?

lildude: Nope, none what-so-ever, as as I've just discovered _all_ the editors on the GitHub.com interface are now using CodeMirror (as of Sep 2016 and GitHub Enterprise 2.8 onwards) and reflected as such in the docs at https://help.github.com/articles/editing-files-in-your-repository/ and https://help.github.com/articles/about-gists/ and https://help.github.com/articles/using-keyboard-shortcuts/.

https://github.com/github/linguist/pull/3243 was the Linguist side of things that put the finishing touches in place.

So this line will have absolutely _no_ effect on how things look in GitHub.com right now, though may in future as we may switch back to Ace, or other Ace-compatible editor, if it's deemed necessary to do so at some time in the future.

@wesinator If you want to make an effective change, you'll need to implement the `codemirror_mode` and `codemirror_mime_type` equivalents too.

pchaigno: > Nope, none what-so-ever, as as I've just discovered all the editors on the GitHub.com interface are now using CodeMirror

Oh, right. I forgot that!

wesinator: Ok, codemirror doesn't have a builtin YARA mode.

linguist doesn't use the language grammars for editing ?

pchaigno: No, it uses CodeMirror.

wesinator:
@pchaigno Done - that was not why the tests were failing though, I don't see why they would be failing.

PR #4205, opened 2018-07-20
Release v6.3.1

lildude:
This only pulls in an update to the Elixir grammar. We happened to catch it with a bad commit that was then fixed the day we released Linguist v6.3.0.

Fixes github/linguist#4203

Template omitted as it's not relevant for a release.

/cc @kivikakk


lildude:
This is now live and all the Elixir files are pretty again. 🌈🌈🌈

PR #4215, opened 2018-07-31
Compiler fixes

kivikakk:
Some fixes for the grammar compiler:

* `-linguist-path` wasn't being used
* some repositories have syntax files that don't declare a scope name (see [`ballerina.monarch.json`](https://github.com/ballerina-platform/ballerina-grammar/blob/b7372c23154b3ccf9a55ef14e1abdbf753f4e2f2/syntaxes/ballerina.monarch.json)) — these were being written to `.json` in the output directory, which will in turn crash prettylights when it tries to read `.json` from the grammar directory and can't find a `scopeName` in it.

/cc @vmg


kivikakk:
Added:

* Write a `version` file out with the version as stated in `lib/linguist/version.rb` so consumers of the grammars know what version of Linguist they correspond to.

tools/grammars/compiler/converter.go
```diff
@@ -248,13 +257,25 @@ func (conv *Converter) Report() error {
 	return nil
 }
 
+var VERSION_RE *regexp.Regexp = regexp.MustCompile("VERSION = \"(.+)\"")
```
jhawthorn: Might be a little cleaner to change `version.rb` to do `VERSION = File.read("#{__dir__}/VERSION")` to simplify this, but this is clever and works 👍

kivikakk: I prefer your suggestion 100%. Thank you!

lib/linguist/version.rb
```diff
@@ -1,3 +1,3 @@
 module Linguist
-  VERSION = "6.4.0"
+  VERSION = File.read(File.expand_path("../VERSION", __FILE__)).strip
```
lildude: Nice. One less hack for the release script I've been slowly working on.

kivikakk: Sweet!

PR #4218, opened 2018-08-03
Clarify adding a language section

wesinator:
Update contributing guide adding a language section
* step for list-grammars

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,8 +61,9 @@ To add support for a new language:
   This command will analyze the grammar and, if no problems are found, add it to the repository. If problems are found, please report them to the grammar maintainer as you will not be able to add the grammar if problems are found.  
   **Please only add grammars that have [one of these licenses][licenses].**
 1. Add samples for your language to the [samples directory][samples] in the correct subdirectory.
-1. Add a `language_id` for your language using `script/set-language-ids`.  
-  **You should only ever need to run `script/set-language-ids --update`. Anything other than this risks breaking GitHub search :cry:**
+1. Add a `language_id` for your language using `script/set-language-ids --update`.  
+  **You should only ever need to run with `--update`. Anything other than this risks breaking GitHub search :cry:**
+1. Add an entry to the [vendor README](vendor/README.md) using `script/list-grammars`.
 1. Open a pull request, linking to a [GitHub search results](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aboot+NOT+nothack&type=Code&ref=searchresults) showing in-the-wild usage.  
   Please state clearly the license covering the code in the samples. Link directly to the original source if possible.
 


CONTRIBUTING.md
```diff
@@ -61,8 +61,9 @@ To add support for a new language:
   This command will analyze the grammar and, if no problems are found, add it to the repository. If problems are found, please report them to the grammar maintainer as you will not be able to add the grammar if problems are found.  
   **Please only add grammars that have [one of these licenses][licenses].**
 1. Add samples for your language to the [samples directory][samples] in the correct subdirectory.
-1. Add a `language_id` for your language using `script/set-language-ids`.  
-  **You should only ever need to run `script/set-language-ids --update`. Anything other than this risks breaking GitHub search :cry:**
+1. Add a `language_id` for your language using `script/set-language-ids --update`.  
+  **You should only ever need to run with `--update`. Anything other than this risks breaking GitHub search :cry:**
+1. Add an entry to the [vendor README](vendor/README.md) using `script/list-grammars`.
```
pchaigno: This step shouldn't be necessary; `script/add-grammar` already calls `script/list-grammar`.

CONTRIBUTING.md
```diff
@@ -61,8 +61,9 @@ To add support for a new language:
   This command will analyze the grammar and, if no problems are found, add it to the repository. If problems are found, please report them to the grammar maintainer as you will not be able to add the grammar if problems are found.  
   **Please only add grammars that have [one of these licenses][licenses].**
 1. Add samples for your language to the [samples directory][samples] in the correct subdirectory.
-1. Add a `language_id` for your language using `script/set-language-ids`.  
-  **You should only ever need to run `script/set-language-ids --update`. Anything other than this risks breaking GitHub search :cry:**
+1. Add a `language_id` for your language using `script/set-language-ids --update`.  
+  **You should only ever need to run with `--update`. Anything other than this risks breaking GitHub search :cry:**
```
pchaigno: Maybe we should make the `update` flag implicit rather than adding warnings in the documentation. @lildude What do you think?

Alhadis: My advice:

*	Rename the file `script/update-ids` to clarify the script's responsibility

*	Rename the option `--force` to `--reset-all`; better yet, axe it entirely. The output which the `--force` option emits [even suggests](https://github.com/github/linguist/blob/ed44c0b86d317b753c607742b137dce9594bed97/script/set-language-ids#L54-L56) removal is a good idea:
	> You're regenerating all of the language_id attributes for all Linguist languages defined in languages.yml. This is almost certainly NOT what you meant to do!

	There's absolutely no good reason that something this destructive would ever be needed by a commonly-run script, especially since the code it invokes could easily be typed into an `irb` session:
	~~~rb
	languages.each do |name, vals|
	  vals.merge!('language_id' => generate_language_id(name))
	end
	~~~

*	Add support for a dry-run option to exit with a non-zero status if any of the IDs need updating:
	~~~console
	$ script/update-ids --check
	languages.yml:5017: missing ID for "Windows Registry"
	languages.yml:5058: missing ID for "XYZ Language"
	2 languages were found with missing IDs. Run `script/update-ids` to generate them.
	$ echo $?
	1
	~~~

	This would help CI for testing whether an ID is missing (or even duplicated...)

Alhadis: I'd also have the `languages.yml` header extracted from the file itself, rather than [embedding](https://github.com/github/linguist/blob/ed44c0b86d317b753c607742b137dce9594bed97/script/set-language-ids#L7) a copy of it inside the `script/update-ids` file. We've had issues with out-of-sync header content in the past, and having only one source of truth will make future amendments to it much less burdensome.

lildude: I can't argue with @Alhadis here. Good ideas there. Lets do it.

Alhadis: Assuming I can get Ruby and Bundler working, I'll happily take a stab at it. =)

CONTRIBUTING.md
wesinator: Hmm, I've always had to run `list-grammar` manually after adding or else vendor/README.md isn't updated and the tests fail (e.g. https://travis-ci.org/github/linguist/builds/405880762)

pchaigno: @wesinator Could you file a separate issue for that with example steps to reproduce?

PR #4222, opened 2018-08-04
XML strategy

### LINKED ISSUE #2780: Auto-detecting XML files based on the header
There are just way too many XML files out there. Instead of the 50+ file extensions we have registered in [languages.yml](https://raw.githubusercontent.com/github/linguist/master/lib/linguist/languages.yml), why not just recognize it if the top of the file starts with `<?xml version="1.0" encoding="UTF-8"?>` and isn't one of the other known extensions? Similar with what we do for the shebang in Bash/Perl.

pchaigno:
Following discussion in #2780, this pull request adds a new XML Strategy to detect XML files using their expected root tag `<?xml version=`.

## Description

The XML strategy runs after all other generative strategies (strategies that can generate new candidate languages, such as the Extension strategy) to avoid misclassifying files from languages that are subsets of XML (we have a few, such as [XSLT](https://github.com/github/linguist/blob/ed44c0b86d317b753c607742b137dce9594bed97/lib/linguist/languages.yml#L5266)). To that effect, the XML strategy only runs if previous strategies were unable to identify candidate languages.

This pull request will have the side effect of making Linguist read the content (which I limited to the 2 first lines) of all files whose language we previously weren't identifying. @lildude Not sure if that could be an issue on github.com's side?

The new fixture file is from [danpal/OpenSAML](https://github.com/danpal/OpenSAML/blob/8297262c7ef649249f84750a6c197f4359669f97/src/test/resources/data/org/opensaml/saml2/core/impl/.svn/text-base/AssertionIDRequestOptionalAttributes.xml.svn-base) and is licensed under Apache v2.0.

I've also merged all tests for strategies (files under `lib/linguist/strategies/`) into a single file. Each strategy only has one or two tests anyway.

Fixes #2780.

/cc @jamesqo @Alhadis


pchaigno:
> Personally, I'd broaden the strategy a little so it leaves room for future additions, if needed. E.g., naming it `Strategy::Signature` might be better than making it XML-specific.

I'd prefer to start with something small (like the 132M XML files :stuck_out_tongue_closed_eyes: ) and extend it later if it works well. We can always change the name then since it's not exposed in the API anyway.

I can see the benefit of doing the same for HTML and Roff in the future, but are there really that many PHP extensions we don't recognize?

Alhadis:
> We can always change the name then since it's not exposed in the API anyway.

Ah, that changes things then. 😉

> ... but are there really that many PHP extensions we don't recognize?

Not really, and not for PostScript either. But any unrecognised text file which begins with an unmistakable signature like `<?php` or `%!PS` can be reasonably assumed to be the format in question, IMHO.

lildude:
> @lildude Not sure if that could be an issue on github.com's side?

I shouldn't think there'd be a problem... it's few lines than we currently read when searching for the modeline 😄

pchaigno:
@lildude Okay to merge then?

pchaigno:
@lildude Done.

pchaigno:
@lildude The test failure were caused by a few new sample files which did not have a XML root tag (`XML/GMOculus.project.gmx` and `XML/obj_control.object.gmx`); I added those to the whitelist.

PR #4224, opened 2018-08-06
simplify .ms heuristic

smola:
Simplify .ms heuristic for Unix Assembly by just checking that there is no
`/*` string in the file. This is simpler and seems to work with all Unix Assembly
files in GitHub. A synthetic MAXScript file was also tested by adding assembly inside
a multiline comment.

## Checklist:

None applies?

### DIFF
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -323,7 +323,7 @@ module Linguist
     disambiguate ".ms" do |data|
       if /^[.'][a-z][a-z](\s|$)/i.match(data)
         Language["Roff"]
-      elsif /(?<!\S)\.(include|globa?l)\s/.match(data) || /(?<!\/\*)(\A|\n)\s*\.[A-Za-z][_A-Za-z0-9]*:/.match(data.gsub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'|\\\s*(?:--.*)?\n/, ""))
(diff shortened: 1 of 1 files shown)


lib/linguist/heuristics.rb
```diff
@@ -323,7 +323,7 @@ def call(data)
     disambiguate ".ms" do |data|
       if /^[.'][a-z][a-z](\s|$)/i.match(data)
         Language["Roff"]
-      elsif /(?<!\S)\.(include|globa?l)\s/.match(data) || /(?<!\/\*)(\A|\n)\s*\.[A-Za-z][_A-Za-z0-9]*:/.match(data.gsub(/"([^\\"]|\\.)*"|'([^\\']|\\.)*'|\\\s*(?:--.*)?\n/, ""))
+      elsif !/\/\*/.match(data) && /^\s*\.(include\s|globa?l\s|[A-Za-z][_A-Za-z0-9]*:)/.match(data)
```
Alhadis: It wouldn't hurt to make this a non-capturing group:

~~~
        ↓↓
/^\s*\.(?:include\s|globa?l\s|[A-Za-z][_A-Za-z0-9]*:)
        ↑↑
~~~

smola:
@Alhadis could you have a look?

Alhadis:
Looks good to me. 👍

PR #4232, opened 2018-08-16
Extend aliases and add `.mdoc` as a supported Roff extension

Alhadis:
This PR adds `.mdoc` to GitHub's list of Roff extensions, and adds several aliases to make tagging fenced code-blocks easier.

## Description
The `.mdoc` extension was brought to my attention by [`github/markup#1196`](https://github.com/github/markup/pull/1196#discussion_r186881929), and I was surprised it was missing. In-the-world usage is in the neighbourhood of [9,374 results](https://github.com/search?q=extension%3Amdoc+NOT+nothack&type=Code). The sample being added here is [taken from `libzip`](https://github.com/aburgh/libzip/tree/b1c3e4d104cd3c2d3be532d212dd960ad3a115e7), released under [MIT](https://github.com/aburgh/libzip/blob/b1c3e4d104cd3c2d3be532d212dd960ad3a115e7/LICENSE).

In addition, I've added several commonly-used synonyms for Roff to facilitate the tagging of fenced code-blocks on GitHub. There's a good chance users might use `~~~man` or `~~~groff` without knowing the correct name/spelling.

I've also replaced `.man` with `.roff` as the language's primary extension. Roff doesn't have a canonical file extension, but if it did, `.roff` would make more sense than `.man`. =)

## Checklist:
    - [Search results for each extension](https://github.com/search?q=extension%3Amdoc+NOT+nothack&type=Code)
    - [Sample source(s)](https://github.com/aburgh/libzip/blob/b1c3e4d104cd3c2d3be532d212dd960ad3a115e7/man/zip_file_add.mdoc)
    - [Sample license(s)](https://github.com/aburgh/libzip/blob/b1c3e4d104cd3c2d3be532d212dd960ad3a115e7/LICENSE)


Alhadis:
I wonder why GitHub allows self-assignment to merged/closed pull-requests...

Alhadis:
@RalphCorderoy Yes, but unfortunately, the `.tr` extension has [very little usage in-the-wild](https://github.com/search?q=extension%3Atr+NOT+nothack&type=Code); and that's a binding factor when determining which extensions we can and can't register on GitHub. =)

My point was that there isn't a canonical extension, but `.man` probably isn't the best compromise... 😕

Alhadis:
See [`CONTRIBUTING.md`](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language). =) It pretty much explains the prerequisites; and yes, to some extent, I'm afraid it is largely down to a popularity contest, so to speak...

Think of it this way: GitHub certainly don't want to be indexing literally every single file-extension that has less than a hundred uses, and that's where they'd find themselves eventually. If every user felt their hobbyist side-project-programming-languages were eligible for registration on GitHub, think of the mess we'd find ourselves in. 😉  It'd be an endless deluge of conflicting extensions and misclassifications as the Bayesian classifier struggles to cope with 430 `.gs` extensions that were used about 10-30 times on GitHub.

I'm not site staff, so I don't claim to know the exact reasoning behind their prerequisites... but if I were in their shoes, I'd be doing the same thing. A line has to be drawn somewhere.

pchaigno:
> So it's a land-grab of file extension, first takes all?

To complete @Alhadis's answer, several languages can be associated to the same extension of course. Even if we associated extension `.stuff` only to language X in the past, language Y can be associated to `.stuff` at a later time. Both languages will have to meet the prerequisite @Alhadis mentioned of course.

PR #4233, opened 2018-08-17
Dockerfile to quick language detection

coldner:
It should help Windows users because it's easier to get dependencies on Linux.


pchaigno:
We'll need to document this. The contribution guidelines are probably the best place at this point.

pchaigno:
Is this blocked because of #4234?

pchaigno:
For the Travis CI build failures, they should go away if you update your branch.

You'll still need to document this change ;)

pchaigno:
Flagging this as stale.

PR #4243, opened 2018-08-21
Add Slice language (.ice)

externl:
Add new language syntax for Slice

## Checklist:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Aice+module+NOT+nothack
    - Sample source(s):
      - [samples/Slice/Test.ice](samples/Slice/Test.ice)
    - Sample license(s): bsd-3-clause
  - [x] lightshow example: [see here](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fzeroc-ice%2Fvscode-slice%2Fmaster%2Fsyntaxes%2Fslice.tmLanguage.json&grammar_text=&code_source=from-text&code_url=&code=%23pragma+once%0D%0A%0D%0A%23ifndef+SOME_TEST%0D%0A%5B%5B%22java%3Apackage%3Alinguist%22%5D%5D%0D%0A%23endif%0D%0A%0D%0Amodule+Linguist%0D%0A%7B%0D%0A++++enum+MyEnum%0D%0A++++%7B%0D%0A++++++++One%2C%0D%0A++++++++Two%2C%0D%0A++++++++Three%0D%0A++++%7D%0D%0A%0D%0A++++struct+MyStruct%0D%0A++++%7B%0D%0A++++++++%2F%2F+An+int%0D%0A++++++++int+a%3B%0D%0A++++++++%2F*+string+*%2F%0D%0A++++++++string+b%3B%0D%0A%0D%0A++++++++MyEnum+e%3B%0D%0A++++%7D%0D%0A%0D%0A++++exception+MyException+%7B%0D%0A++++++++string+e%3B%0D%0A++++%7D%0D%0A%0D%0A++++dictionary%3Cstring%2C+string%3E+MyDict%3B%0D%0A%0D%0A++++sequence%3CMyEnum%3E+MyEnumSeq%3B%0D%0A%0D%0A++++class+BaseClass+%7B%0D%0A++++++++int+value+%3D+-1%3B%0D%0A++++%7D%0D%0A%0D%0A++++class+MyClass+extends+BaseClass%0D%0A++++%7B%0D%0A++++++++MyDict+info%3B%0D%0A%0D%0A++++++++optional%281%29+string+op%3B%0D%0A++++%7D%0D%0A%0D%0A%0D%0A++++interface+MyInterface%0D%0A++++%7B%0D%0A++++++++void+operationA%28out+bool+valid%29%3B%0D%0A++++++++idempotent+void+operationB%28int+a%29%3B%0D%0A++++++++MyEnumseq+getEnum%28%29%3B%0D%0A%0D%0A++++++++%5B%22cpp%3Aconst%22%2C+%22cpp%3Anoexcept%22%5D+string+getName%28%29%3B%0D%0A++++%7D%0D%0A%7D)
  - [x]  I have updated the heuristics to distinguish my language from others using the same extension.
       According to the contributing guide this was not needed as there are no other languages using `.ice`.


externl:
The errors seem unrelated, or maybe I'm mistaken?

externl:
Thanks, @pchaigno. I've added that.

pchaigno:
Are [these files](https://github.com/search?p=1&q=extension%3Aice+NOT+module&type=Code) also Slice? They look like JSON.

Could you try to update your branch with the latest master to address the Travis CI issue?

lildude:
> @lildude Any idea of what's happening here?

Looking now. My suspicion is licensed has changed again 😞

pchaigno:
@lildude Actually, it's Licensee this time. I filled an issue: benbalter/licensee#328.

lildude:
Yeah, I was just about to confirm this.

externl:
> Are these files also Slice? They look like JSON.
Yeah, most of those are definitely JSON. Some binary files too.

pchaigno:
> Yeah, most of those are definitely JSON.

We'll need to find a way to distinguish JSON and Slice files then.

externl:
Ok, I'll add a heuristic for Slice to `heuristic.rb`

externl:
Ok, so I was going to add:
```
disambiguate ".ice" do |data|
      if /^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)/i.match(data)
        Language["Slice"]
      end
    end
```
This should be a good check for `Slice`. I don't really see a good check for JSON.

pchaigno:
@externl Could you update your branch with the latest master? It includes a fix for the Travis CI tests.

> I don't really see a good check for JSON.

Would checking for the root object of the JSON document (`{` or `[`) be enough? /cc @Alhadis who, with a bit of luck, knows more about the JSON spec. than I do.

externl:
@pchaigno I've merged in the latest. Maybe the best is to just not check for JSON, leave it non-highlighted?

pchaigno:
Unfortunately, that's not an option our classifier has at the moment. With this pull request all `.ice` files will be classified as either JSON or Slice, so we need to do a good job distinguishing the two.

externl:
What happens to files that don't match the heuristic?

```
disambiguate ".ice" do |data|
    if /^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)/i.match(data)
        Language["Slice"]
    end
end
```

vs

```
disambiguate ".ice" do |data|
    if /^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)/i.match(data)
        Language["Slice"]
    elsif SOME_JSON_CHECK
        Language["JSON"]
    end
end
```
?

pchaigno:
They go to the naïve Bayesian classifier (see [*How Linguist Works*](https://github.com/github/linguist#how-linguist-works)) which assigns a score to each language based on the tokens identified in the file.

externl:
@pchaigno  I've rebased my branch on the latest. Do you still want me to look into adding a JSON disambiguations check for these other `.ice` files?

pchaigno:
> Do you still want me to look into adding a JSON disambiguations check for these other `.ice` files?

Yes, that would be much better.

> Would checking for the root object of the JSON document (`{` or `[`) be enough?

@lildude @Alhadis Any ideas on the above?

Alhadis:
> Any ideas on the above?

Yep. Don't forget to check for leading whitespace too (either vertical or horizontal). E.g, something like this should be enough:

~~~regexp
/\A\s*[{[]/
~~~

pchaigno:
@externl Could you implement the above heuristic rule with the corresponding test in [`test/test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb)?

vendor/licenses/grammar/vscode-slice.txt
```diff
@@ -31,4 +32,4 @@ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
 SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
\ No newline at end of file
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
Alhadis: What happened here?

externl: What do you mean? In my recent commit I added the version.

Alhadis:
~~~
  1) Error:
TestFileBlob#test_include_in_language_stats:
RegexpError: empty char-class: /\A\s*[{[]/
~~~

Ah whoops, my bad. Sorry, that expression should have been:

~~~regexp
~~~

externl:
Is this correct?
```
def test_ice_by_heuristics
    assert_heuristics({
      "Slice" => all_fixtures("Slice", "*.ice"),
      "JSON" => all_fixtures("JSON", "*.ice")
    })
  end
```

Alhadis:
Looks fine to me.

vendor/licenses/grammar/vscode-slice.txt
Alhadis: This line appears to have been modified, although GitHub's diff-view isn't showing what changed... Strange.

A trailing newline may have been added or removed when you edited the license, but GitHub normally indicates that with an 🚫 icon.

externl:
How do I fix:
```
/home/travis/build/github/linguist/vendor/licenses/grammar/vscode-slice.txt:
  - cached license data out of date0m
```

vendor/licenses/grammar/vscode-slice.txt
externl: Yeah my local diff shows that I added a trailing newline
```
diff --git a/vendor/licenses/grammar/vscode-slice.txt b/vendor/licenses/grammar/vscode-slice.txt
index b1e84f5c..0103ec63 100644
--- a/vendor/licenses/grammar/vscode-slice.txt
+++ b/vendor/licenses/grammar/vscode-slice.txt
@@ -1,6 +1,7 @@
 ---
 type: grammar
 name: vscode-slice
+version: 6a28c2125bb442fa6a7f31c995ce423b1c9bbfb7
 license: bsd-3-clause
 ---
 BSD 3-Clause License
@@ -31,4 +32,4 @@ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
 SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
\ No newline at end of file
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

lildude:
> How do I fix:
> 
> ```
> /home/travis/build/github/linguist/vendor/licenses/grammar/vscode-slice.txt:
>   - cached license data out of date0m
> ```

You need to ensure that the SHA in the version is the same as the SHA for the committed grammar submodule. `script/licensed` should have done this for you if your local repo was on the same SHA as that that you've committed in this PR.

Committed submodule SHA: 635d41ba3924f168ba5a54101bf71fbf8aef3d20
License version: 6a28c2125bb442fa6a7f31c995ce423b1c9bbfb7

As you can see, they're not the same.

externl:
@pchaigno @Alhadis Thanks for guiding me through the process.

> LGTM. I'm mortified that I didn't think to make an "Ice-ense" pun somewhere, though. 😓

 I'm going to start using this 😄

lib/linguist/languages.yml
```diff
@@ -4496,6 +4496,13 @@ Slash:
   tm_scope: text.html.slash
   ace_mode: text
   language_id: 349
+Slice:
+  type: programming
+  tm_scope: source.slice
+  ace_mode: text
+  extensions:
+  - ".ice"
+  language_id: 894641667
```
lildude: Errm, don't you want a colour associated with this language? At the moment it's going to fallback to the default of `#cccccc`.

Looking at https://doc.zeroc.com/ice/latest/the-slice-language, it seems blue is the colour of choice so it's probably going to be a PITA to find one that doesn't conflict with one already associated with a language on GitHub.com - blue is really really popular.

externl: Whoops, would definitely like a colour. Can we use `#477ed0`?

lildude: I doubt it. Various shades of blue are already highly used. Try it. Add the `color:` entry and then run `bundle exec ruby test/test_color_proximity.rb` (I'm assuming you've already run `script/bootstrap`). Rinse and repeat until you find a colour the test likes, then commit your colour and push.

externl: Thanks @lildude, I found a blue! `#003fa2`

lildude: 🎉 nice!!

lib/linguist/heuristics.yml
```diff
@@ -149,6 +149,12 @@ disambiguations:
     named_pattern: objectivec
   - language: C++
     named_pattern: cpp
+- extensions: ['.ice']
+  rules:
+  - language: Slice
+    pattern: '^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)'
+  - language: JSON
+    pattern: '/\A\s*[{\[]/'
```
bzz: I'm sorry for digging this out, but I'm not sure if such patter with 2 slashes works with [`Regexp.new`](https://github.com/github/linguist/pull/4087/files#diff-e56a8d50b1a75fc4212ffb3504faa007R79).

AFAIK it should not match, here is [REPL example](https://repl.it/repls/GeneralResponsibleDesign) where it indeed does not.

Saw @Alhadis https://github.com/github/linguist/pull/4243#issuecomment-423562417 with pattern suggestion and wanted to verify my understanding first here, before opening a new PR with removal of the slashes.

WDYT?

externl: Yeah, you're right. The JSON detection doesn't seem to be working right. I guess a new issue should be opened?

Alhadis: Oops, you're right. Yes, those should be removed.

pchaigno: @bzz Go ahead for the pull request!

PR #4248, opened 2018-08-24
Update license hashes following Licensee update

pchaigno:
While we're trying to handle this better in the long term (cf. benbalter/licensee#328), this pull request updates the hardcoded license hash values in the tests and fixes the Travis CI builds.

See benbalter/licensee#328 for information on why the hash values changed. Basically, the hash values are computed over the *normalized license text*, so every time the normalization process changes, the values will change.

This pull request also removes two hardcoded license values that are not needed anymore.

### DIFF
--- a/test/test_grammars.rb
+++ b/test/test_grammars.rb
@@ -11,29 +11,27 @@ class TestGrammars < Minitest::Test
   ].freeze
 
   HASH_WHITELIST = [
-    "2edac46b0a63309c96442d2826321a442217472f", # Agda.tmbundle
-    "4da01d631a29c76456fd0bd16749c71e8d5f6dbf", # ant.tmbundle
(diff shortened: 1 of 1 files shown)


Alhadis:
What's involved in normalising license text? What does it mean?

(I wanted to ask at benbalter/licensee#328, but I don't have permissions to post a comment for some reason. 😕)

pchaigno:
> What's involved in normalising license text? What does it mean?

Before comparing licenses, they remove text that doesn't have any particular semantic (such as bullet points) and text whose semantic doesn't matter to the recognition of the license (such as the Copyright line that includes the author's name). 

> but I don't have permissions to post a comment for some reason.

uh, that's very weird.

Alhadis:
Ah, now I see. 😀 Thanks!

Yeah, it is strange. 🤔 Maybe the repository's activity is currently limited to past contributors only.

PR #4258, opened 2018-08-31
Associate `*.[1-9]` file extensions with Text

Alhadis:
This is an attempt to fix widespread misclassification of any file whose name ends in a dot-separated numeric suffix:

<a name="description"></a>
## Description
Many files on GitHub receive an incorrect classification of "Roff", simply because their filenames end in a numeric suffix: 

	Gemfile-3.0.3
	GFDL-1.2
	AFL-1.2
	mkfontscale-1.1.1

Change-logs and version-numbers are particularly problematic examples of these. The [`ValveSoftware/steam-runtime`](https://github.com/ValveSoftware/steam-runtime), for instance, was shown as 72.8% Roff (until [I fixed it](https://github.com/ValveSoftware/steam-runtime/pull/143)), because of [three license files](https://github.com/ValveSoftware/steam-runtime/search?l=roff) with version-numbers. [`haskell-CI/haskell-ci`](https://github.com/haskell-CI/haskell-ci/search?l=roff) was another example (until [fixed](haskell-CI/haskell-ci#223)). You can find those and more by checking GitHub's list of [trending Roff repositories](https://github.com/trending/roff?since=daily).

<a name="solution"></a>
### Solution
To address this matter, I'm adding `.{1..9}` to the list of Text file extensions. We don't have `no language` as an option here, so using `Text` as an open-ended fallback for numeral-suffixed filenames (that aren't Roff) feels like a decent compromise.

The heuristics I've added basically go like this:

1.	**Choose "Roff" if file contains a well-formed title macro:**  
	~~~roff
	.TH PAGE_TITLE 1  \" man(7)
	.Dt PAGE_TITLE 3  \" mdoc(7)
	~~~
2.	**Choose "Text" if file has NO lines which match a well-formed Roff command.**  
 	The commands I check for are those used for [man-page authoring](https://github.com/github/linguist/blob/baa4cad2925e4ad82aee4a8614bf0e9277fbc3d7/lib/linguist/heuristics.rb#L95-L107): `man` and `mdoc`. Other macro packages aren't accounted for because by convention, documents using the `me` or `ms` macros have file extensions of that name.
3.	**If neither of the above conditions are met, fall back to the Bayesian classifier.**

**Update:** I've also added two `Text` aliases to accommodate change-logs. Emacs uses `-*- change-log -*-`, and Vim uses `changelog`.

<a name="checklist"></a>
## Checklist:

		| Sample                   | Source                                  | License  |
		|--------------------------|-----------------------------------------|-----------
		| `atom.1`                 | [Alhadis/.files][1]                     | ISC
		| `gather_profile_stats.1` | [mozilla/home-snippets-server-lib][1a]  | BSD (mentioned in [page source][1al])
		| `socket.2`               | [OpenBSD's source tree][2]              | ISC
		| `pcre32.3`               | [PCRE's manpages][3]                    | [BSD][3l]
		| `textmate.5`             | [Alhadis/.files][5]                     | ISC
		| `dunelegacy.6`           | [spellbindguy/my_slackbuilds][6]        | Unclear, but *["it may be freely used by anyone for any purpose"][6l]*
		| `utf.7`                  | [Plan9's `libutf` library][7]           | [Lucent](http://akaros.cs.berkeley.edu/files/Plan9License) ([Permissive](http://akaros.cs.berkeley.edu/files/Plan9License))
		| `rc.8`                   | [OpenBSD's source tree][8]              | ISC
		| `cpumem_get.9`           | [OpenBSD's source tree][9]              | ISC
		| `mkfontscale-1.1.1`      | [X11 App ports][1t]                     | X Consortium
		| `AFL-1.2`                | [mmoselhy/poky-ml507][2t]               | N/A? (Do literal licenses have their own licenses?)
		| `GFDL-1.2`               | [ValveSoftware/steam-runtime][2ta]      | See above
		| `ChangeLog-1.9.3`        | [johnl/deb-ruby1.9.1][3t] (truncated)   | BSD (assumed based on [`COPYING`][3tl]
		| `_md5sums-7.0.4`         | [qrux/xlfs][4t]                         | N/A? (Data generated from a public list of files)
		| `CC-BY-2.5`              | [mmoselhy/poky-ml507][5t]               | N/A; assuming Creative Commons (?)
		| `hierarchyThreshold.6`   | [apache-log4j-1.2.16][6t]               | [Apache][6tl]
		| `NEWS-1.8.7`             | [ruby/ruby][7t]                         | BSD
		| `random-numbers.8`       | Myself                                  | ISC
		| `pathological.9`         | Myself                                  | ISC

[1]:    https://github.com/Alhadis/.files/blob/23fdf91a6766dd4d3e353c58fe896ecedcaee81b/share/man/man1/atom.1
[1a]:   https://github.com/mozilla/home-snippets-server-lib/blob/3d7b54032123a7d1615f1af21af40358f1d7f93b/packages/Django/docs/man/gather_profile_stats.1
[1al]:  https://github.com/mozilla/home-snippets-server-lib/blob/3d7b54032123a7d1615f1af21af40358f1d7f93b/packages/Django/docs/man/gather_profile_stats.1#L23-L25
[2]:    https://github.com/openbsd/src/blob/f9db39a7df472c4a1ab736b6ca3d012a0910b2c4/lib/libc/sys/socket.2
[3]:    https://github.com/vmg/pcre/blob/019468890a4000d1e12cdd775b6e094bcd338415/doc/pcre32.3
[3l]:   https://github.com/vmg/pcre/blob/019468890a4000d1e12cdd775b6e094bcd338415/LICENCE
[5]:    https://github.com/Alhadis/.files/blob/752c89dbaa2716f62f7cf5aa8e16090b55ac91c7/share/man/man5/textmate.5
[6]:    https://github.com/spellbindguy/my_slackbuilds/blob/e7050e8d895077e19809a9132f9715fc51d3a5f3/games/dunelegacy/dunelegacy.6
[6l]:   https://github.com/spellbindguy/my_slackbuilds/blob/e7050e8d895077e19809a9132f9715fc51d3a5f3/games/dunelegacy/dunelegacy.6#L173-L175
[7]:    https://github.com/adriencarbonne/libutf/blob/master/utf.7
[8]:    https://github.com/openbsd/src/blob/master/share/man/man8/rc.8
[9]:    https://github.com/openbsd/src/blob/master/share/man/man9/cpumem_get.9
[1t]:   https://github.com/matijaskala/ports-2013/blob/ef0285ec892670db261386296bc40e95fa037d79/metadata/md5-cache/x11-apps/mkfontscale-1.1.1
[2t]:   https://github.com/mmoselhy/poky-ml507/blob/84f45b233be0ea263eea20d4b5adfa88e3936032/meta/files/common-licenses/AFL-1.2
[2ta]:  https://github.com/ValveSoftware/steam-runtime/blob/0e0384f4d7b8ee256096ca56c91866464edb397b/runtime/common-licenses/GFDL-1.2
[3t]:   https://github.com/johnl/deb-ruby1.9.1/blob/e0239ee319580935ef846b18bf97c8d419f74514/doc/ChangeLog-1.9.3
[3tl]:  https://github.com/johnl/deb-ruby1.9.1/blob/e0239ee319580935ef846b18bf97c8d419f74514/COPYING
[4t]:   https://github.com/qrux/xlfs/blob/3d9e30774218ebb6dea332da42e911b47337c1c3/src/_md5sums-7.0.4
[5t]:   https://github.com/mmoselhy/poky-ml507/blob/84f45b233be0ea263eea20d4b5adfa88e3936032/meta/files/common-licenses/CC-BY-2.5
[6t]:   https://github.com/tmfev/MDR/blob/c681884643ad961d78367b189533f4a67964bdbc/tools/apache-log4j-1.2.16/tests/witness/hierarchyThreshold.6
[6tl]:  https://github.com/tmfev/MDR/blob/c681884643ad961d78367b189533f4a67964bdbc/tools/apache-log4j-1.2.16/LICENSE
[7t]:   https://github.com/ruby/ruby/blob/117c384560bf40fa062f2cac14fc510184a4006a/doc/NEWS-1.8.7

/cc @smola


pchaigno:
Do non-Roff files ending with a number always (or most of the time) follow the pattern `.*-(\d+\.)+\d`?

Alhadis:
Release versions like those above do, yeah.

But it's also pretty common for somebody to track a duplicate download, which gets saved with a numeric extension (`wget` does this, for instance).

pchaigno:
> But it's also pretty common for somebody to track a duplicate download, which gets saved with a numeric extension (`wget` does this, for instance).

Is that something we should even try to detect though? Those files could be anything, right?

Alhadis:
Yes, as I explained under the [**Solution** section](https://github.com/github/linguist/pull/4258#solution):

> We don't have `no language` as an option here, so using `Text` as an open-ended fallback for numeral-suffixed filenames (that aren't Roff) feels like a decent compromise.

Classifying them as text is the safest thing to do, since the files in question *are* literally text. It's as best as we can get to an "unspecified".

pchaigno:
I've read that. It's fine if the files' content is text from a human language (as in licenses for example). We cannot, however, add samples whose contents match another language under the Text entry because that would screw the Bayesian classifier, making it unable to distinguish properly between Text and the other language and lessening its ability to recognize Text.

I think Roff is one of these cases that we are not going to be able to solve entirely without a real no-a-language option. Since you identified that many Text files with a number as extension have filenames ending with a version number, I was thinking we could use that as a much simpler heuristic rule. We would have to feed the filename as input to the Heuristic strategy though. What do you think?

PS: The Text samples already contain two files whose contents match other languages. I'll make a pull request to move those.

Alhadis:
> We would have to feed the filename as input to the Heuristic strategy though. What do you think?

Wait, we can do that? 💚 God I wish I'd known that sooner. Yes, deffo.

pchaigno:
@Alhadis Do you want to take a stab at it?

Alhadis:
Sure, I'll have a crack. :)

smola:
Something like this would be great, see for my results while getting Roff files: https://github.com/smola/language-dataset/tree/master/samples/Roff

Maybe detecting `[-.][0-9]+(\.[0-9]+)+` suffix would do a decent job?

Alhadis:
That sounds like it'd work. 👍

On it.

Alhadis:
Good point.

`[0-9]` is probably faster than `\d`, BTW. Oniguruma is Unicode-aware by default, meaning `\d` matches a wider range of codepoints than simple ASCII.

pchaigno:
Oh, right. I forgot that!

Alhadis:
Should I remove the current `Text` heuristic, or have it run after the filename is checked for version numbers?

Alhadis:
Actually, what I *should* be asking: how do I access the filename from within the heuristic's body?

~~~rb
elsif /[-.](?>[0-9]+\.)++[0-9]$/.match(WELL_OOPS) || !roff_match.match(data)
  Language["Text"]
end
~~~

Is the `data` variable just a `String` object, or does it have some additional file-specific properties that I can access?

pchaigno:
> Is the `data` variable just a `String` object, or does it have some additional file-specific properties that I can access?

`data` is just a String (or the Ruby equivalent), but all strategies have the blob. `blob.name` is the filename, which you'll have to pass to `heuristic.call` in addition to `data`. Or you can pass the blob to `heuristic.call`.

https://github.com/github/linguist/blob/b5e2687b2ebe02a6f30943b30ddd8ee08322881c/lib/linguist/heuristics.rb#L21-L24

Alhadis:
Okay, heuristic updated. I've amended the OP with the sources of the new sample files too.

I've also added two aliases for change-log modelines. Emacs uses a hyphenated `change-log`, while Vim uses `changelog`.

lib/linguist/heuristics.rb
```diff
@@ -70,7 +71,7 @@ def matches?(filename, candidates)
     end
 
     # Internal: Perform the heuristic
-    def call(data)
+    def call(data, filename = nil)
       @heuristic.call(data)
```
pchaigno: Are you sure this works? It's weird that `filename` is not used here, no?

Alhadis: ... I'm impressed at how I managed to overlook that, what the hell. 😂

I should really add a test for this.

Alhadis: Will this suffice?

~~~rb
  def test_1to9_by_heuristics
    assert_heuristics({
      "Roff" => all_fixtures("Roff", "*.{1..9}"),
      "Text" => all_fixtures("Text", "*.{1..9}")
    })
  end
~~~

Alhadis:
Okay, merged and updated to accommodate our new heuristic format.

@smola Could you review this, please?

larsbrinkhoff:
Note: Some Forth files use `.4`.  Maybe not many enough to care about.  Then again, maybe it's easy to do it right?

Alhadis:
I wouldn't say so. Going by [your own SO answer](https://stackoverflow.com/questions/18246756/what-is-the-most-common-filename-extension-of-a-forth-source-code-file), there were only 55 hits for `.4` FORTH files in 2013, and between the two languages, only Roff continues to have modern-day usage of that extension.

Using `.4` as a Forth extension probably isn't common even historically, since man-pages have been around almost as long as Forth has.

lib/linguist/heuristics.rb
```diff
@@ -98,9 +101,13 @@ def matches?(filename, candidates)
     end
 
     # Internal: Perform the heuristic
-    def call(data)
+    def call(data, filename)
       matched = @rules.find do |rule|
-        rule['pattern'].match(data)
+        unless rule['filename_pattern'].nil?
```
lildude: Eeeeew 🤢 This feels wrong to me. How about flipping it to if / else:

```suggestion
        if rule['filename_pattern']
```

Alhadis: Sorry, I must've been thinking backwards. 😂 Again.

Fixed.

Alhadis:
@pchaigno Do you think we should close this in favour of a dedicated manpage strategy? Because that would enable us to axe the numeric file extensions from Roff's language entry, and make unrelated `.1` files appear to have *no* language, which I feel is better than lumping under `Text`.

And let's face it, none of us want this mess in our heuristics. :wink:

pchaigno:
> Do you think we should close this in favour of a dedicated manpage strategy?

The dedicated strategies run on almost all files, so we're probably going to be very limited in terms of what we can do without degrading performance. Maybe we could keep this heuristic but without the `negative_pattern`?

Taking a look at this again, I'm surprised the Bayesian classifier isn't giving us better results. Roff files should be fairly easy to distinguish from Text files since they contain a small set of recurring keywords.

Alhadis:
> Maybe we could keep this heuristic but without the `negative_pattern`?

I'd rather not. If github/markup#1196 is to go through, it needs to avoid rendering misclassified `.1` files as markup. Doing so will likely render their contents unintelligible, and users will voice confusion on why their log files have suddenly become mangled when viewed on GitHub...

pchaigno:
Ok. Do we have an evaluation of the Bayesian classifier's accuracy for these files?

lildude:
I guess this PR is going to need reworking due to https://github.com/github/linguist/pull/4393, right?

Alhadis:
I have no idea what this means. 😭

~~~
  1) Failure:
TestHeuristics#test_1to9_by_heuristics [/home/travis/build/github/linguist/test/test_heuristics.rb:23]:
no fixtures for Roff *.{1..9}
~~~

There *are* fixtures in `samples/Roff/*.{1.9}`... :confused:

test/test_heuristics.rb
```diff
@@ -61,6 +61,13 @@ def test_detect_still_works_if_nothing_matches
     assert_equal Language["Objective-C"], match
   end
 
+  def test_1to9_by_heuristics
+    assert_heuristics({
+      "Roff" => all_fixtures("Roff", "*.{1..9}"),
```
lildude: Your test is failing because `*.{1..9}` isn't a valid glob pattern for a set:

```
irb(main):001:0> Dir.glob("./samples/Roff/*.{1..9}")
=> []
```

```suggestion
      "Roff" => all_fixtures("Roff", "*.[1-9]"),
```

... should do the trick.

```
irb(main):002:0> Dir.glob("./samples/Roff/*.[1-9]")
=> ["./samples/Roff/as.1", "./samples/Roff/fstat.2", "./samples/Roff/moo.6", "./samples/Roff/ino.8", "./samples/Roff/cat.4", "./samples/Roff/sample.4", "./samples/Roff/vs.9", "./samples/Roff/pow.3", "./samples/Roff/core.5", "./samples/Roff/greek.7", "./samples/Roff/soelim.8"]
irb(main):003:0>
```

Same below.

Alhadis: > *.{1..9} isn't a valid glob pattern for a set:

Ugggh, I always get `*.[1-9]` confused with `1..2`. 😰 Blame `git log`'s syntax for showing a range of commits.

I was about to ask, *"how did this not fail the first time I submitted this?"* but noticed the assertion responsible for this was only [added recently](https://github.com/github/linguist/pull/4410/files#diff-952df3fccfc8a6df1579b4554d15c6b8R23). Will update.

lib/linguist/heuristics.yml
```diff
@@ -18,6 +18,7 @@
 #                       regular expression (with union).
 # and                 - An and block merges multiple rules and checks that all of
 #                       of them must match.
+# filename_pattern    - Same as pattern, but tests filenames instead of content.
```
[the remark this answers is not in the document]
Alhadis: Oh, that. This was an addition to the heuristics format to accommodate `Text` files which ended with version numbers (stuff like `GPL-2.2` and stuff). I must've overlooked the heuristic that used it when resolving merge conflicts, but on second thought, I'm not sure it's even needed...

I recall it was @pchaigno's suggestion. @pchaigno, do we still need this? If not, we can simplify the format by removing the `filename_pattern` feature, but it might come in handy in future...

lib/linguist/languages.yml
```diff
@@ -5278,8 +5278,20 @@ Text:
   wrap: true
   aliases:
   - fundamental
+  - change-log
+  - changelog
+  - outline
   extensions:
   - ".txt"
+  - ".1"
```
smola, line 5286: What about `.0`?

Alhadis: We don't associate that extension with Roff, because there's never been a `man0` section…

Alhadis:
Revisiting this: I'm wondering if we shouldn't just remove the problematic suffixes altogether and entrust [our man-page strategy](https://github.com/github/linguist/pull/4433) to identify numeric file-extensions. @lildude, what do you think?

lildude:
> Revisiting this: I'm wondering if we shouldn't just remove the problematic suffixes altogether and entrust [our man-page strategy](https://github.com/github/linguist/pull/4433) to identify numeric file-extensions. @lildude, what do you think?

Makes sense to me if you think that'll do the trick. I'm also happy to close this if you prefer as I don't think anyone has raised an issue about the current behaviour 😉

Alhadis:
Agreed. Probably better to start anew than turn this PR inside-out. 😂

PR #4266, opened 2018-09-02
Entry for grammar updates in pull request template

pchaigno:
I'm hoping having an explicit entry for it in the template will reduce the number of grammar maintainers opening pull requests for that.


RobQuistNL:
In the extent of this PR; is there a known release schedule? Is a new version also immediately deployed to GitHub? I'd like to know when I can expect certain features to get live ^_^

pchaigno:
There's no strict schedule, but @lildude does a new release per month on average I think.

lildude:
> In the extent of this PR; is there a known release schedule? Is a new version also immediately deployed to GitHub? I'd like to know when I can expect certain features to get live ^_^

What @pchaigno said 😁 I plan to make a release this week or early next.

Issue #4265, opened 2018-09-02
Running script/licensed creates a mess


pchaigno:
Running `./script/licensed` on an unmodified clone of Linguist seems to create a huge mess (see below). It removes newlines at the end of files, adds lines to license texts, and even deletes a few files. Is that expected? Should we be running `./script/licensed` from time to time??

```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   vendor/licenses/grammar/ABNF.tmbundle.txt
	modified:   vendor/licenses/grammar/Agda.tmbundle.txt
	modified:   vendor/licenses/grammar/Alloy.tmbundle.txt
	modified:   vendor/licenses/grammar/Assembly-Syntax-Definition.txt
	modified:   vendor/licenses/grammar/AutoHotkey.txt
	modified:   vendor/licenses/grammar/BrightScript.tmbundle.txt
	modified:   vendor/licenses/grammar/ColdFusion.txt
	modified:   vendor/licenses/grammar/Docker.tmbundle.txt
	modified:   vendor/licenses/grammar/EBNF.tmbundle.txt
	modified:   vendor/licenses/grammar/Elm.txt
	modified:   vendor/licenses/grammar/Handlebars.txt
	modified:   vendor/licenses/grammar/IDL-Syntax.txt
	modified:   vendor/licenses/grammar/Isabelle.tmbundle.txt
	modified:   vendor/licenses/grammar/Lean.tmbundle.txt
	modified:   vendor/licenses/grammar/LiveScript.tmbundle.txt
	modified:   vendor/licenses/grammar/MQL5-sublime.txt
	modified:   vendor/licenses/grammar/MagicPython.txt
	modified:   vendor/licenses/grammar/NSIS.txt
	modified:   vendor/licenses/grammar/NimLime.txt
	modified:   vendor/licenses/grammar/PHP-Twig.tmbundle.txt
	modified:   vendor/licenses/grammar/PogoScript.tmbundle.txt
	modified:   vendor/licenses/grammar/Racket.txt
	modified:   vendor/licenses/grammar/SCSS.tmbundle.txt
	modified:   vendor/licenses/grammar/SMT.tmbundle.txt
	modified:   vendor/licenses/grammar/Scalate.tmbundle.txt
	modified:   vendor/licenses/grammar/Slash.tmbundle.txt
[...]

PR #4267, opened 2018-09-06
WIP: Use Licensed manifest method for caching license metadata

lildude:
Floating this WIP early to ease getting feedback as I work this. I'll add in the apt parts of the template later.

For the moment, I'm working on using the latest `licensed` directly as I iterate on the configuration:

```
$ bundle exec licensed cache -c vendor/licenses/config.yml
Caching licenses for grammars:
  manifest dependencies:
License caching complete!
* grammars manifest dependencies: 0
$
```

Commented out parts are things I've considered and will need to reconsider later too, so feel free to ignore those parts or comment as necessary.

I'll update the `script/licensed` and `script/add-grammar` later.

/cc @jonabc https://github.com/github/linguist/issues/4265

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -27,6 +27,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rake'
   s.add_development_dependency 'yajl-ruby'
   s.add_development_dependency 'color-proximity', '~> 0.2.1'
-  s.add_development_dependency 'licensed', '~> 1.1.0'
+  s.add_development_dependency 'licensed', '~> 1.3.0'
   s.add_development_dependency 'licensee'
 end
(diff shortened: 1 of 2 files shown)


lildude:
After a fair bit of experimentation over the weekend, I'm going to defer on this idea in favour of updating our current approach. This is primarily because using the manifest requires monkey patching the current behaviour and implementing functionality necessary to automatically add a new line to the config.yml everytime a new grammar is added. This is far more complex than simply adding a version to all our current license files.

jonabc:
> This is far more complex than simply adding a version to all our current license files.

👍 let me know if theres anything else I can help with.  I _think_ you should still be able to leverage `licensed` to test for allowed licenses, and remove license tests from https://github.com/github/linguist/blob/8bf9efa3702a1a43df85dc5cd72b63f3ff36871f/test/test_grammars.rb.  Though of course that choice is up to `linguist` peeps 😄

PR #4268, opened 2018-09-11
Store grammar git version within license file when caching

lildude:
## Description
As discovered by @pchaigno in https://github.com/github/linguist/issues/4265, running `script/licensed` results in some unexpected changes to the cached licenses. Some of this was because the grammars have been updated, but not the licenses (🙋‍♂️ my fault), and some due to stripping of newlines etc.

As pointed out by @jonabc at https://github.com/github/linguist/issues/4265#issuecomment-418800281:

> licensed uses logic based on version metadata to detect whether changes are needed. The issue here is that linguist uses a custom enumeration source that doesn't provide version metadata.
> 
> The solutions to this are
> 
> - update the Filesystem source to give `"version" => <version>` metadata for each dependency
> - switch to licensed's built-in manifest dependency source

I looked into the latter in https://github.com/github/linguist/pull/4267, and this proved way too complex for our needs so I've gone with the former.

At the same time, I've added `script/licensed` to be run just before `script/licensed status` in CI. This should cause the latter to flag any issues.

I've also updated the contributing guidelines so other Hubbers will remember to perform the `script/licensed` step each time the grammars are updated.

And of course, I've updated _all_ of the licenses so they have a version flag. The only changes to actual licenses are where the license has legitimately changed within the grammar. Mr @Alhadis has done this a few times 😉

The `curated` flag has also been removed from the license files as this isn't honoured by anything and the version check should effectively lock us to an approved version. The `script/licensed && script/licensed status` should pick up any changes that'll need re-evaluating as-and-when the grammar is updated.

As licensed is performing a license check on our cached licenses and these are now tied to a version, I'm tempted to ditch the checks we're performing directly with licensee at https://github.com/github/linguist/blob/8bf9efa3702a1a43df85dc5cd72b63f3ff36871f/test/test_grammars.rb#L91-L142

Thoughts? @jonabc, I think the `script/licensed status` is effectively performing what ☝️ tests are. Am I right?

## Checklist:
  
Fixes https://github.com/github/linguist/issues/4265


script/licensed
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env ruby
 
-# TODO: push these changes to licensor gem
-
```
lildude: We're not going to do this, so no need for this comment.

script/licensed
```diff
@@ -24,10 +22,10 @@ module Licensed
 
       def dependencies
         Dir.glob(@glob).map do |directory|
-          puts "caching #{directory}"
```
lildude: This is duplicating info that is already printed by Licensed. Licensed will print:

```
Caching licenses for linguist:
  grammar dependencies:
    Using sublime-rexx (2e304a34d3493e7453a430735191d91f356faa5a)
    Using Modelica (c841b17b0f441451b43a0b0e62bc6201a3456b83)
    Using atom-language-purescript (cc2cb471968eb331af7259aac219feb3017d709e)
[...]
    Caching Sublime-Coq (94e43bdc194535cf5bae9306624039af9f38c014)
[...]
```

`Using` signifies nothing has changed, `Caching` signifies a change.

vendor/licenses/config.yml
```diff
@@ -23,7 +23,9 @@ reviewed:
   rubygem:
     - mime-types
   grammar:
-    - Sublime-Lasso
+    - blitzmax        # No license file. License in README
+    - creole          # License filename is not LICENSE(.*) but rather "The MIT License (MIT)"
+    - Sublime-Lasso   # No license file. License in README
```
lildude: Equivalent of what we do in test_grammars.rb

jonabc:
> I think the script/licensed status is effectively performing what ☝️ tests are. Am I right?

@lildude `licensed status` covers the first three tests, but doesn't particularly care about the last three that check for unused values since that's related to the local repo and not the dependencies.

> At the same time, I've added script/licensed to be run just before script/licensed status in CI. This should cause the latter to flag any issues.

I'm not sure this is entirely necessary.  Now that you've built in versioning, `licensed status` will complain if a license is out of date, as a prompt to re-run `licensed cache`.

vendor/licenses/grammar/ABNF.tmbundle.txt
```diff
@@ -1,9 +1,10 @@
 ---
 type: grammar
 name: ABNF.tmbundle
+version: 0042b3d7ee21b2f800a8f80d42761513f16d9a52
 license: apache-2.0
 ---
-                                 Apache License
+Apache License
```
jonabc: 🤦‍♂️ ok `licensed` is too aggressive in stripping leading empty space

lildude: Oops. I meant to update our Licensed dependency too. If you're going to address this 🔜, I can hold off merging this PR and update the licenses again once a newer version of Licensed is available.

lildude:
> I'm not sure this is entirely necessary. Now that you've built in versioning, `licensed status` will complain if a license is out of date, as a prompt to re-run `licensed cache`.

Ah, cool. I'll remove that then. 🙇‍♂️

vendor/licenses/grammar/ABNF.tmbundle.txt
lildude: Hmm, I think you may not need to do anything @jonabc. Just updated the gemspec and I'm not seeing that over-stripping of whitespace with Licensed 1.3.3.

lildude: Hmmm, but now the version has changed 😕 Issue in Licensed coming up.

lildude:
After a little back-and-forth with @jonabc, I think I've come to a good solution which ties our cached license versions to the submodule SHAs. We don't update these independently of each other so it makes sense to use this tie.

Licensed may gain support for a submodule source which may allow us to simplify or even remove our Filesystem implementation in future. Until then, I think this does the trick.

Ready for reviews.

jonabc:
> > I think the script/licensed status is effectively performing what ☝️ tests are. Am I right?

> @lildude `licensed status` covers the first three tests, but doesn't particularly care about the last three that check for unused values since that's related to the local repo and not the dependencies.

@lildude I'm curious whether removing the tests is still being thought about.  I'm curious whether licensed is able to cover those test cases or if there's something that it's not checking that it could and/or should.

lildude:
Whoops. I meant to come back and remove those three tests. I think `licensed status` should be all we need for those. Thanks for the reminder.

jonabc:
Are the remaining tests using `PROJECT_WHITELIST`, `HASH_WHITELIST` and `LICENSE_WHITELIST` still useful if license verification is done by `licensed`?

It looks like those tests only exist to ensure the hashes are valid for use in the tests that you removed in https://github.com/github/linguist/pull/4268/commits/7cf1233a23a31c99f81335e35d067f111601e88c

lildude:
> Are the remaining tests using `PROJECT_WHITELIST`, `HASH_WHITELIST` and `LICENSE_WHITELIST` still useful if license verification is done by `licensed`?

🤔 good question. I believe they were originally written to keep track of when a whitelisted project gained a license and `licensed status` should definitely pick this up. The last test is skipped too.  I think we can safely 🔥 those too, and remove the last of the licensee-based tests.

lildude:
@pchaigno or @Alhadis can one of your please give this PR a review to be sure I'm not missing anything. I'd like to get this in 🔜  so I can get a new release out. The next release is going to include quite a few changes so make take a bit longer to test.

🙇

lildude:
I'm merging a few PR which will require this PR to be updated before merging too, but better they're in and only waiting on me to fix this PR 😄

lildude:
> which will require this PR to be updated before merging too

Done.

lildude:
> LGTM. 👍 I ran the tests after adding a grammar for a new language entry, and they all passed (the grammar is [`language-file-magic`](https://github.com/Alhadis/language-file-magic)).

🎉  I've also tested with updating grammars and it does a good job of flagging changes there too. Gonna merge and start work on a new release.

PR #4280, opened 2018-09-30
Added support for Pest grammars.

dragostis:
I'm adding support for [Pest](https://pest-parser.github.io/) grammars.

Usage:
https://github.com/search?q=extension%3Apest+NOT+nothack&type=Code

[Examples](https://github.com/pest-parser/pest/tree/master/grammars/src/grammars) have been added from the main repo, but here are a few others:

* [comrak](https://github.com/kivikakk/comrak/blob/master/src/lexer.pest)
* [handlebars-rust](https://github.com/sunng87/handlebars-rust/blob/master/src/grammar.pest)
* [Huia](https://gitlab.com/jimsy/huia/blob/master/parser/src/grammar.pest)
* [sql](https://github.com/gwenn/sqlpest/blob/6da014ecce33cac13c857641b0dbfd5655ac22a6/src/sql.pest)
* [tera](https://github.com/Keats/tera/blob/master/src/parser/tera.pest)

I don't have a TextMate grammar yet, but I can work on one if this PR is fine.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3511,6 +3511,13 @@ Perl 6:
   codemirror_mode: perl
   codemirror_mime_type: text/x-perl
   language_id: 283
+Pest:
+  type: programming
+  color: "#e66448"
+  extensions:
+  - ".pest"
+  ace_mode: text
+  language_id: 432
 Pic:
   type: markup
   group: Roff
--- a/samples/Pest/json.pest
+++ b/samples/Pest/json.pest
@@ -1,0 +1,32 @@
+// pest. The Elegant Parser
+// Copyright (c) 2018 Dragoș Tiselice
+//
+// Licensed under the Apache License, Version 2.0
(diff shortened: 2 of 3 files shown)


pchaigno:
Thanks for the contribution! Any reason you removed the template?

pchaigno:
The template is easier to read to us because we're used to it. It allows us to check quickly that nothing is missing and we know where to find each piece of information. It would help if you could add it back.

In any case, using [Harvester](https://github.com/Alhadis/Harvester), I counted only 152 repositories with Pest files from 136 users on GitHub.com. This is not enough to warrant support in Linguist yet; our rule of thumb is *hundreds of repositories*.

pchaigno:
Closing due to the low usage on GitHub.com. Let's revisit once usage is higher!

dragostis:
@pchaigno, I've run Harvester today and I'm now counting 304 repos from 259 users with continuously increasing usage. Linguist support would be useful especially because there are a lot of snippets riddled in issues that would benefit greatly from highlighting.

PR #4281, opened 2018-10-01
Reclassify Protobuf as programming language

ZJvandeWeg:
Protobuf was first a markup language, but changed to a data type in 5633fd36683241a052461538f0c98657d2325c5d. However, it's a programming language
just as Cap'n Proto

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):
  
### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3661,7 +3661,7 @@ Propeller Spin:
   ace_mode: text
   language_id: 296
 Protocol Buffer:
-  type: data
+  type: programming
   aliases:
   - protobuf
   - Protocol Buffers


pchaigno:
Isn't it used only to serialized structured data? I'm not sure most GitHub users would want to see Protocol Buffer files counted in their language statistics...

ZJvandeWeg:
@pchaigno No, for example: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/deviceplugin/v1alpha/api.proto

That file is created by a user, which that by compiling a file will generate code for the languages of your choosing. The resulting, and thus generated code, is not included but the proto files itself are.

Alhadis:
@ZJvandeWeg If it's a programming language, how does one perform loops or branching? Can you demonstrate a basic `if { … } else { … }` construct in Protocol Buffer? Does the language have any programmatic features?

A "data" language can also cover configuration files, which are written by hand (instead of being machine-generated).

Alhadis:
Yacc *is* because it contains entire code chunks which can be passed through to the compiled output.

`sed` is an odd one, because it actually *is* a Turing-complete programming language, and people have even [programmed games](https://github.com/uuner/sedtris) using nothing but `sed` syntax.

pchaigno:
@Alhadis I agree it's not a programming language, but why the change from markup to data?

Alhadis:
How/why on earth would you consider this markup? 😕

pchaigno:
I agree that's weird too, but it's not data, it's a description of how to format data, kind of like XSD.

Alhadis:
Correct, like a configuration file. Those too, get lumped under the `data` umbrella.

pchaigno:
> Correct, like a configuration file.

Fair enough.

Alhadis:
@ZJvandeWeg I'd like to mention that if you want a non-detectable language type to appear in your repository's statistics, you can do so [using an override](https://github.com/github/linguist#detectable). Add this to your project's `.gitattributes` file (create one if it doesn't exist):

~~~shell
*.proto linguist-detectable=true
~~~

PR #4285, opened 2018-10-03
Language alias for the Windows file system

### LINKED ISSUE #4234: Cannot clone repo on Windows
I suspect that Windows files cannot contain '*' character.

PS C:\projects> git clone https://github.com/github/linguist.git
Cloning into 'linguist'...
remote: Counting objects: 32811, done.
remote: Compressing objects: 100% (13/13), done.
remote: Total 32811 (delta 4), reused 0 (delta 0), pack-reused 32798
Receiving objects: 100% (32811/32811), 33.56 MiB | 1.04 MiB/s, done.
Resolving deltas: 100% (17916/17916), done.

**fatal: cannot create directory at 'samples/F*': Invalid argument**

warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'

pchaigno:
With this pull request, languages with a non-Windows compatible name must include an `fs_name` alias. That alias is used to create the `samples` and `fixtures` directories. This change enables repository cloning under Windows.

The F* entry now uses Fstar as the Windows compatible alias.

Fixes #4234.

@lildude:
> Please also remind me that I'll need to perform the DB nudge when I make the next release once that PR has been merged so the language rename is picked up in the DB.


lib/linguist/languages.yml
```diff
@@ -1292,6 +1292,7 @@ F#:
   codemirror_mime_type: text/x-fsharp
   language_id: 105
 F*:
+  fs_name: Fstar
```
Alhadis: We should describe this new field in `language.yml`'s header, along with the other fields. Its purpose or meaning isn't immediately obvious unless the reader knows that `fs_` stands for `filesystem_` / `file_safe_`).

pchaigno: Oh, right! We'll do as soon as I get home.
`filesystem` here, I guess I'm used to seeing it in the system community.

pchaigno: Done.

RobQuistNL: Or even fstar_name in this case!

PR #4289, opened 2018-10-05
Heuristic rule for Puppet and Pascal .pp files

### LINKED ISSUE #3605: Puppet is often misclassified as many different languages
We've noticed internally that Puppet is often misclassified as Pascal.  After [a little further digging](https://github.com/search?l=Pascal&q=extension%3App+include&type=Code&utf8=%E2%9C%93), it looks like Puppet is misidentified as a host of different languages from Pascal to Ruby. 

Prior art:
https://github.com/github/linguist/pull/2333
https://github.com/github/linguist/pull/2031
https://github.com/github/linguist/pull/1267

pchaigno:
This pull request adds a heuristic rule for Puppet and Pascal `.pp` files. Fixes #3605.

/cc @Caged

## Checklist:
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -268,6 +268,12 @@ disambiguations:
     named_pattern: perl6
   - language: XPM
     pattern: '^\s*\/\* XPM \*\/'
+- extensions: ['.pp']
+  rules:
+  - language: Pascal
+    pattern: '^\s*end[.;]'
+  - language: Puppet
+    pattern: '^\s+\w+\s+=>\s'
 - extensions: ['.pro']
   rules:
   - language: Prolog
--- a/test/test_heuristics.rb
+++ b/test/test_heuristics.rb
@@ -322,6 +322,14 @@ class TestHeuristics < Minitest::Test
     })
   end
 
+  # Candidate languages = ["Pascal", "Puppet"]
+  def test_pp_by_heuristics
+    assert_heuristics({
+      "Pascal" => all_fixtures("Pascal", "*.pp"),
+      "Puppet" => all_fixtures("Puppet", "*.pp") - ["#{samples_path}/Puppet/stages-example.pp", "#{samples_path}/Puppet/hiera_include.pp"]
+    })
+  end
+
   # Candidate languages = ["IDL", "Prolog", "QMake", "INI"]
   def test_pro_by_heuristics
     assert_heuristics({


lib/linguist/heuristics.yml
```diff
@@ -268,6 +268,13 @@ disambiguations:
     named_pattern: perl6
   - language: XPM
     pattern: '^\s*\/\* XPM \*\/'
+- extensions: ['.pp']
+  rules:
+  - language: Pascal
+    pattern: '^\s*end[.;]'
+  - language: Puppet
+    # Curly brackets mark comments in Pascal.
```
Alhadis: Pascal comments can span multiple lines, according to [Free Pascal's wiki](http://wiki.freepascal.org/Comments) (going by the distinction the page makes between block and line comments).

So a Pascal file with a multiline header might get flagged as Puppet.

pchaigno: I was [under the impression](https://www.tutorialspoint.com/pascal/pascal_basic_syntax.htm) that multiline comments had to start and end with `{*` and `*}`. Is that incorrect?

pchaigno: Looks like I was wrong: https://github.com/github/linguist/blob/b70a976e12d52184d91780db2607d8152f524e88/samples/Pascal/cwindirs.pp#L77-L79

Alhadis: It might be safer to assert `Puppet` by matching a keyword preceding the opening brace. To use the [examples here](https://puppet.com/docs/puppet/5.3/lang_summary.html), the heuristic might look like:

~~~regexp
/^[ \t]+(package|service|file)[ \t]+}$/
~~~

Alhadis: BTW, it appears the author of [that page you linked](https://www.tutorialspoint.com/pascal/pascal_basic_syntax.htm) is misinformed about comment syntax — the [Free Pascal Wiki](http://wiki.freepascal.org/Comments) explains that `(* … *)` is a relic from the old days when keyboards didn't always have curly brackets (!). The single-line comment form is `// `, just like C++. =)

So the asterisks in that page's example are redundant...

Alhadis: Ah yes, sorry. 😅 That's what I meant. Fixed.

pchaigno: > BTW, it appears the author of [that page you linked](https://www.tutorialspoint.com/pascal/pascal_basic_syntax.htm) is misinformed about comment syntax — the [Free Pascal Wiki](http://wiki.freepascal.org/Comments) explains that `(* … *)` is a relic from the old days when keyboards didn't always have curly brackets (!). The single-line comment form is `// `, just like C++. =)

To be fair, there are many Pascal dialects and even more implementations, so both statements could be true. We need to support both cases though, so I'll change the heuristic rule.

pchaigno: That heuristic matches only 1 of our 5 sample files. They're not very big, but still.
Maybe
```regexp
/^\s+\w+\s+=>\s/
```
would be a better match?

Alhadis: Assuming `=>` isn't legal syntax in most Pascal dialects (you seem to know more here than I do), I'd say that's an acceptable heuristic. :+1:

pchaigno: I don't know much on Pascal either; only what I remember from classes a decade ago...

To check, I downloaded 4,957 `.pascal` and `.pas` Pascal files and searched for the above regular expression. None of the files contain it so I'm fairly confident that, if  `=>` is legal syntax, it's very uncommon.

pchaigno:
I've updated the pull request with the new heuristic rule for Puppet.

PR #4297, opened 2018-10-16
Make sure a GMS2 generated file is classified as such

RobQuistNL:
I'm making sure the generated .yy file is being matches as a generated GMS2 file, instead of Yacc.

## Checklist:
    - Sample source(s):
      - https://github.com/RobQuistNL/DSNet/blob/ad8bc9ffc37b920610f6524a86324f1b6503d3c0/options/main/inherited/options_main.inherited.yy
    - Sample license(s): (MIT)[https://github.com/RobQuistNL/DSNet/blob/ad8bc9ffc37b920610f6524a86324f1b6503d3c0/LICENSE]


lib/linguist/generated.rb
```diff
@@ -588,7 +588,8 @@ def generated_graphql_relay?
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines[2].match(/\"modelName\"\:\s*\"GM/)
+      return lines[2].match(/\"modelName\"\:\s*\"GM/) ||
+             lines[0].start_with?("1.0.0")
```
RobQuistNL: Would there be a way for me to include the weird character back in there? Its this one: ←(``U+2190`` - ``0xE2 0x86 0x90`` in utf-8).

It might make the checking a lot stricter, and then could possible replace starts with ``1.0.0`` with ``\d\.\d\.\d←`` which would then support future versions :)

RobQuistNL: PS: Just throwing the character in there breaks the testsuite;
```
 1) Error:
TestBlob#test_language:
Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8
    /home/travis/build/github/linguist/lib/linguist/generated.rb:592:in `start_with?'
    /home/travis/build/github/linguist/lib/linguist/generated.rb:592:in `generated_gamemakerstudio?'
    /home/travis/build/github/linguist/lib/linguist/generated.rb:93:in `generated?'
    /home/travis/build/github/linguist/lib/linguist/generated.rb:12:in `generated?'
    /home/travis/build/github/linguist/lib/linguist/blob_helper.rb:362:in `generated?'
    /home/travis/build/github/linguist/test/test_blob.rb:281:in `block (2 levels) in test_language'
    /home/travis/build/github/linguist/test/test_blob.rb:271:in `each'
    /home/travis/build/github/linguist/test/test_blob.rb:271:in `block in test_language'
    /home/travis/build/github/linguist/test/test_blob.rb:265:in `each'
    /home/travis/build/github/linguist/test/test_blob.rb:265:in `test_language'
  2) Error:
TestFileBlob#test_language:
Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8
    /home/travis/build/github/linguist/lib/linguist/generated.rb:592:in `start_with?'
    /home/travis/build/github/linguist/lib/linguist/generated.rb:592:in `generated_gamemakerstudio?'
    /home/travis/build/github/linguist/lib/linguist/generated.rb:93:in `generated?'
    /home/travis/build/github/linguist/lib/linguist/generated.rb:12:in `generated?'
    /home/travis/build/github/linguist/lib/linguist/blob_helper.rb:362:in `generated?'
    /home/travis/build/github/linguist/test/test_file_blob.rb:676:in `block (2 levels) in test_language'
    /home/travis/build/github/linguist/test/test_file_blob.rb:666:in `each'
    /home/travis/build/github/linguist/test/test_file_blob.rb:666:in `block in test_language'
    /home/travis/build/github/linguist/test/test_file_blob.rb:660:in `each'
    /home/travis/build/github/linguist/test/test_file_blob.rb:660:in `test_language'
```

Alhadis: @RobQuistNL Is `LANG` set in your environment? On OpenBSD, every test belched spurious encoding errors until I executed tests using `LANG=en_AU.UTF-8 bundle exec rake` (aliased in my shell as simply `brake`).

It fixed errors which looked extremely similar to the ones you've posted.

RobQuistNL: And it was fixed by removing the ``←`` character from the matching line

RobQuistNL: I must note that I have not ran the suite locally - I tried it a while back but I feel like I'm too unexperienced with ruby / whatnot and am missing a lot of tooling ^_^ Full-blown PHP developer here

Alhadis: Not an expert on how encodings work in Ruby (especially in CI), but you can express Unicode characters as 7-bit ASCII using:

~~~rb
/\u2190/ # RegExp escape
"\u2190" # Analogous string escape

"\u2190" =~ /\u2190/ # => Returns 0, the index of the first match
"foobar" =~ /\u2190/ # => nil No match
~~~

Alhadis: The `=~` operator is ~~shameless ripped off from Perl~~ Ruby's way of performing a match between a regular expression and a value.

In PHP, the equivalent syntax is `pcre_match("/\\u2190/", "input");` (if memory serves, it's been years since I bothered working with the language. #NODE4LYF

RobQuistNL: Ah cool, will try :)

lib/linguist/generated.rb
```diff
@@ -588,7 +588,8 @@ def generated_graphql_relay?
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines[2].match(/\"modelName\"\:\s*\"GM/)
+      return lines[2].match(/\"modelName\"\:\s*\"GM/) ||
+             lines[0] =~ /^\d\.\d\.\d/\u2190/
```
Alhadis: Syntax error. That's what's causing the build to fail:

~~~
/home/travis/build/github/linguist/lib/linguist/generated.rb:592: syntax error, unexpected ')'
t.match(/\d\.\d\.\d/\u2190.*/)
~~~


Escape or delete this forward-slash:

~~~
t.match(/\d\.\d\.\d/\u2190.*/)
                   ^~~~~~~~~~~
~~~

RobQuistNL: Yep, I was working on it and letting travis run the tests ^_^

RobQuistNL:
Hmm, that explodes again..
```
TestFileBlob#test_language:
Encoding::CompatibilityError: incompatible encoding regexp match (UTF-8 regexp with ASCII-8BIT string)
```

Alhadis:
@lildude Could you check this out, please? You're likely more familiar with this pipeline (and Ruby) than I am.

RobQuistNL:
@Alhadis can you confirm that the tests run on your own machine? I can't since I don't have the project running :x

RobQuistNL:
IMHO we could remove that one character - allthough it would be a pretty solid detection method :)

lildude:
I've not had the chance to look into this in-depth but this smells very similar to https://github.com/github/linguist/issues/4028

Alhadis:
Is there any chance of it being related to `charlock_holmes`? I keep seeing this warning emitted whenever I run `bundle exec rake …`:

> /home/travis/build/github/linguist/vendor/bundle/ruby/2.5.0/gems/charlock_holmes-0.7.6/lib/charlock_holmes/encoding_detector.rb:66: warning: instance variable encoding_list not initialized

... and [this doesn't look very good](https://github.com/brianmario/charlock_holmes/blob/fee4cad36c4481b5bba98e060e5af78f1cf0e11c/lib/charlock_holmes/encoding_detector.rb#L45-L66):

~~~rb
    # A mapping table of supported encoding names from EncodingDetector
    # which point to the corresponding supported encoding name in Ruby.
    # Like: {"UTF-8" => "UTF-8", "IBM420_rtl" => "ASCII-8BIT"}
    #
    # Note that encodings that can't be mapped between Charlock and Ruby will resolve
    # to "ASCII-8BIT".
    @encoding_table = {}
~~~

Since `charlock_holmes` appears to be consulting a lower-level system dependency (ICU) I'm willing to bet this is undefined behaviour subject to environmental factors. It would explain why OpenBSD gets flooded with errors (to an extend considered unacceptable by *anybody*), whereas nobody else seems to have reported this on other systems.

RobQuistNL:
I really wish I could help, but I'm a real big newbie when it comes to Ruby - I will at least try to setup the project on my Xenial machine soon.

lib/linguist/generated.rb
lildude: I've looked into this today and attempting to match the `←` is definitely a bit of a struggle. I can get it working locally on macOS by using that char directly, but it doesn't play nice on Travis.

An alternate solution may be to more closely match the format of the line ignoring the content, for example:

```
lines[0] =~ /^\d\.\d\.\d.+\|\{/
```

... matches the `1.0.0` at the beginning of the line and the `|{` at the end, ignoring everything in between and does the trick locally and on [Travis here](https://travis-ci.org/github/linguist/builds/449346384). We could probably get more fancy with the regex, but I'm not sure it's worth the effort.

RobQuistNL: Ah thanks, I didn't see your comment there for a while. Sorry! Yeah, I think we're going to have to walk this way since the actual char is not going to work.

Alhadis:
Seems like a decent compromise. 👍 Thanks for your patience!

lib/linguist/generated.rb
```diff
@@ -588,7 +588,8 @@ def generated_graphql_relay?
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines[2].match(/\"modelName\"\:\s*\"GM/)
+      return lines[2].match(/\"modelName\"\:\s*\"GM/) ||
+             lines[0] =~ /^\d\.\d\.\d.+\|\{/
```
pchaigno: Is this (version?) number always at the very beginning of the file?

RobQuistNL: Yes, currently i've only seen `1.0.0`, but I think this is the most important part for the IDE - look at it as a header, specifying the version, so the IDE knows how to read the file - in that way it makes sense to be the very first piece of text.

You can find a couple this way; https://www.google.com/search?q=site:github.com%20options/main/inherited/options_main.inherited.yy

pchaigno: Hm, I was going to suggest that we make that a `\A`, but I missed that you are already matching on the first line only. So all good for me :+1:

PR #4299, opened 2018-10-17
Add the .workflow extension

lildude:
## Description
GitHub [announced](https://blog.github.com/2018-10-16-future-of-software/#github-actions-limited-public-beta) [GitHub Actions](https://github.com/features/actions) yesterday and the raw files for the workflows are written in HashiCorp Configuration Language aka HCL, and have a `.workflow` extension.

Support for this extension was rolled out to GitHub.com nearly a month ago, in preparation for yesterday's announcement so this PR is making these changes public and also adding in the necessary sample files.

## Checklist:

     🤫 This isn't the case yet, but it will be and as this is by GitHub, for GitHub, and already used on GitHub.com, an exception has already been made 😉.

    The actual workflow files we're interested in: https://github.com/search?q=extension%3Aworkflow+NOT+xml+language%3Ahcl&type=Code
    
    There's also an XML `.workflow` file used for Salesforce which is incredibly common too: https://github.com/search?q=extension%3Aworkflow+Workflow+language%3Axml&type=Code

    Unfiltered also shows a few text files are are clearly READMEs and already picked up as documentation and a few shell scripts detected by the shebang: https://github.com/search?q=extension%3Aworkflow+NOT+hack&type=Code

    - Sample source(s):
      - [`main.workflow`](https://github.com/actions/heroku/blob/master/.github/main.workflow)
      - [`Case.workflow`](https://github.com/cabinetoffice/Government-Grants-Information-System-Metadata/blob/master/src/workflows/Case.workflow)
    - Sample license(s): MIT for both

    Not applicable as all other strategies filter out other matches found to date.

/cc @bkeepers


samples/XML/Asset.workflow
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pchaigno: Could we replace this with a bit of a longer sample file?

lildude: Sure. I didn't as the rest of the content is never considered and this sample is incredibly popular already 😄

lildude: Done.

pchaigno: Sure, it's not critical. The content of `.workflow` XML files might be considered if the file doesn't have a root tag though, so it's important that we train the classifier in case that happens. It's not a very frequent case so we should be fine. Tiny files are also a pain when/if we later add a heuristic rule and have to manually exclude all tiny files from the tests.

timruffles:
@lildude I'm seeing some workflow files not being highlighted - e.g [this file](https://github.com/fried-oreos/repo-test/blob/e01a787894d7569c590afc63da067837d28d4268/.github/main.workflow?short_path=bbd9158) is not highlighted (but parses fine), whereas [this one works](https://github.com/fried-oreos/repo-test/blob/master/.github/test.workflow?short_path=34a2452)

pchaigno:
@timruffles We (non-GitHub maintainers of Linguist) might be able to help if you can reproduce in a public repository.

timruffles:
@pchaigno thanks. I've created a minimal reproduction - removing the [line comment](https://github.com/timruffles/orange-soda/blob/c6d8b2ba4afd1f752ad3de0ff2c6ebfdd2ec6665/not-highlighting.workflow) causes it to highlight correctly.

lildude:
Hmmm, this definitely appears to be a detection thing and I somehow think this is related to the "compiled" samples in `samples.json`.

I tested locally directly from my clone and sure enough, I managed to reproduce the issue and the debug view shows us why:

```
$ LINGUIST_DEBUG=2 bundle exec bin/github-linguist ~/tmp/trash/orange-soda/not-highlighting.workflow --breakdown
not-highlighting.workflow: 8 lines (7 sloc)
  type:      Text
  mime type: text/plain
            #       HCL       XML
       [    1     5.795         -
       ]    1     4.146         -
     and    1         -     8.568
      it    1         -     7.685
      on    1         -     8.014
    this    1         -     8.656
       {    1     4.030         -
       }    1     4.030         -
       HCL =   -138.191 +  -6.235 =   -144.427
       XML =   -123.269 +  -3.479 =   -126.748
  language:  XML
$
```

One thing you'll notice above is the lack of the "workflow" word. This would indicate my `samples.json` file doesn't include the necessary processed workflow samples, and sure enough it doesn't.

I updated the `samples.json` locally and 🎉 , correct analysis:

```
$ LINGUIST_DEBUG=2 bundle exec bin/github-linguist ~/tmp/trash/orange-soda/not-highlighting.workflow
not-highlighting.workflow: 8 lines (7 sloc)
  type:      Text
  mime type: text/plain
            #       HCL       XML
       #    1     8.425         -
       [    1     5.931         -
       ]    1     4.282         -
     and    1         -     0.144
 comment    1     8.425         -
      it    1     0.738         -
      on    1     1.103         -
  remove    1     8.425         -
resolves    1     9.118         -
    this    1         -     0.232
workflow    1     9.118         -
       {    1     4.052         -
       }    1     4.052         -
       HCL =    -60.011 +  -5.837 =    -65.848
       XML =   -123.305 +  -3.470 =   -126.774
  language:  HCL
$
```

Now the crazy thing... the `samples.json` file in the v7.0.0 gem _does_ contain all the workflow samples and running linguist from the extracted gem also works. As does copying the `samples.json` file from the v7.0.0 gem to my local clone.

I suspect we may have an old `samples.json` or version of the gem cached somewhere on the GitHub servers. I test using SHA-based versions specifically to avoid caching problems, but I wonder if something has changed on that side resulting in one of my earlier test gems being used instead of the latest.

I plan to make a new release today and will pay special attention to the `samples.json` and will test this issue specifically to see if an updated gem will sort things up.

As an aside, I've just checked @pchaigno's test repo "behind the scenes" and it recognises both files as HCL too:

screenshot 2018-11-08 at 10 29 59

So the override is working, though the search indexing/caching appears to still be holding onto the old info and reindexing doesn't clear this up.

pchaigno:
> I was thinking that maybe this is happening because the Bayesian classifier wasn't retrained on the production server where Linguist runs

Wouhou! My guess was right! :stuck_out_tongue: 

> So the override is working, though the search indexing/caching appears to still be holding onto the old info and reindexing doesn't clear this up.

First time we get a caching issue with the detection, right? (Usually, they're only on the search side.)

lildude:
> First time we get a caching issue with the detection, right? (Usually, they're only on the search side.)

Yeah, though if my suspicion is correct, this isn't unique to Linguist. We very aggressively cache gems, hence I introduced the changes to step 6 in test using the SHA rather than a real version. I have a sneaking suspicion the caching may have become more aggressive since 😞  🤞 a new gem solves this.

lildude:
@timruffles your file is now rendering correctly.

timruffles:
@lildude @pchaigno thanks! 🎉

AndreaGriffiths11:
Hi friends, passing along this feedback: 

I found a syntax highlighting bug in the GitHub Actions “.workflow” file format. https://github.com/closeio/elasticsales.com/blob/master/.github/main.workflow?short_path=f9100ed

lildude:
@AndreaGriffiths11 any syntax highlighting issues should be taken up with the grammar maintainer and not with Linguist, as the grammar isn't controlled by Linguist. In this case it would be the HCL grammar, so an issue needs to be raised in https://github.com/alexlouden/Terraform.tmLanguage

PR #4303, opened 2018-10-23
Added HAProxy configuration language

NickMRamirez:
Adding a new language for HAProxy configuration files, e.g. haproxy.cfg. The ".cfg" file extension is already used by the INI language, but I have added samples to both sample folders to differentiate them.

## Checklist:
    - Sample source(s):
    - Sample license(s):

- [X ] **I am adding a new language.**
  - [X ] The extension of the new language is used in hundreds of repositories on GitHub.com.

    Search for haproxy.cfg filename:
       https://github.com/search?q=filename%3Ahaproxy.cfg&type=Code
    Search for .cfg that is not haproxy.cfg:
       https://github.com/search?q=roundrobin+extension%3Acfg+-filename%3Ahaproxy.cfg

  - [X ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://github.com/NickMRamirez/linguist/tree/haproxy/samples/HAProxy
      - https://github.com/NickMRamirez/linguist/blob/haproxy/samples/INI/ms.cfg
    - Sample license(s):
  - [X ] I have included a syntax highlighting grammar: https://github.com/abulimov/atom-language-haproxy

    - Sample source(s):
    - Sample license(s):


lib/linguist/languages.yml
```diff
@@ -1758,6 +1758,16 @@ Groovy Server Pages:
   codemirror_mode: htmlembedded
   codemirror_mime_type: application/x-jsp
   language_id: 143
+HAProxy:
+  type: data
+  extensions:
+  - ".cfg"
```
pchaigno: If the filename of these files is known, do we need to add the extension as well?

lib/linguist/languages.yml
```diff
@@ -1758,6 +1758,16 @@ Groovy Server Pages:
...
+  filenames:
+  - hapee-lb.cfg
```
pchaigno: I found [only 2 such files](https://github.com/search?utf8=%E2%9C%93&q=filename%3Ahapee-lb.cfg&type=Code) on GitHub.com. That won't be enough to warrant support in Linguist.

samples/HAProxy/haproxy.cfg
```diff
@@ -0,0 +1,46 @@
+global
```
pchaigno: Where did you get this sample file? What license was it under?

lib/linguist/languages.yml
NickMRamirez: The file will almost always be called haproxy.cfg, but there's nothing stopping someone from naming it something else. A person could split their configuration into multiple files line haproxy-1.cfg or sample.cfg.

lib/linguist/languages.yml
NickMRamirez: That makes sense. The Enterprise version calls it hapee-lb.cfg, but id expect way more people to be hosting the Community Edition haproxy.cfg. Should I remove the former?

samples/HAProxy/haproxy.cfg
NickMRamirez: It's my own creation, using some common, but demonstration only, settings.

lib/linguist/languages.yml
pchaigno: If we keep the `.cfg` extension will need a search query to estimate the usage of that extension for HAProxy configuration files on github.com.

lib/linguist/languages.yml
pchaigno: > Should I remove the former?

Yes :)

samples/HAProxy/haproxy.cfg
pchaigno: The sample files are used in tests and to train the naive Bayesian classifier. We therefore try to add representative samples from "real" applications only. If you think those files are representative of common usage in `haproxy.cfg` files, that's fine by me. If not, you can replace them with any files under permissive license found on github.com.

lib/linguist/languages.yml
NickMRamirez: I updated the PR comment with a second search query for files that contain "roundrobin" that have the .cfg extension and are _not_ haproxy.cfg.

samples/HAProxy/haproxy.cfg
NickMRamirez: It's representative in that it uses many of the configuration settings that are commonly used. However, whitespace varies a lot between instances (between words). Will that cause it to not be a match? Also, the values that different projects use for each setting can vary. I've tried to use common values though. Does a line need to match in its entirety? Or only on a keyword at the beginning of the line?

pchaigno: Great, thanks! Yes the Bayesian classifier acts on tokens, not on the raw files.

pchaigno:
@NickMRamirez Could you add one or two other sample files? I'm still seeing a lot of misclassifications by the Bayesian classifier between HAProxy and INI files. I'm expecting that it should be able to do a better job with more training data since HAProxy files contain a set of easily identifiable keywords. Files of the same size as the one you created, or a bit larger should be fine; try to avoid huge files, as the Bayesian then keeps all tokens in memory.

lib/linguist/languages.yml
NickMRamirez: Removed it

NickMRamirez:
Added more sample files. Let me know if I should add more.

pchaigno:
I've downloaded 2347 files from your search query and ran Linguist on it. As far as I can see, it detects all HAProxy files perfectly. I'd say we're good to go!

lildude:
@pchaigno regarding:

> I'm still seeing a lot of misclassifications by the Bayesian classifier between HAProxy and INI files. I'm expecting that it should be able to do a better job with more training data since HAProxy files contain a set of easily identifiable keywords.

Shouldn't we be using a heuristic first? This PR doesn't have a heuristic but introduces a duplicate extension.

pchaigno:
@lildude Adding more samples was enough to improve the Bayesian classifier's accuracy significantly. We went from 1 to 4 samples and now don't have miss-classifications anymore (in the 2347 files corpus I downloaded). In the past, we've kept heuristic rules for cases where the Bayesian classifier is unable to identify recurring keywords. Here, we have a set of recurring keywords (`frontend`, `backend`, `server`, `listen`, etc.) which the classifier can easily identify as long as we have enough samples.

lildude:
@pchaigno 👍 Thanks for the explanation and validation work. 🙇

PR #4309, opened 2018-11-01
Add several more Roff extensions

jordemort:
To go along with https://github.com/github/markup/pull/1196 this PR adds several more extensions to the Roff language.

## Description

I surveyed the manpages in the Ubuntu Cosmic, Debian Stretch, and CentOS 7, and in the OpenIndiana Hipster Live DVD. Here are the extensions used, by count, aggregated by unique filename: https://gist.github.com/c69fa7739d6e4731e90d5749c887f29b

I then went and searched for any of the extensions that had more than 100 uses on GitHub - this PR adds the extensions that, as a result of that search:

  - Appear to be in wide use on GitHub
  - Appear to primarily be used for manpages, as opposed to some other format
  - Appear in more than one or two repositories

They went through a further round of winnowing as I was writing up this PR and deciding if I was prepared to defend the addition of each extension.

This ended up excluding a lot of the nonstandard ones from OpenIndiana, which only appear in one or two mirrors on GitHub, along with a lot of assorted exotica. Those will have to wait until @Alhadis adds a shiny new smart Roff strategy. 

## Checklist:
    - `0p`: https://github.com/search?p=1&q=extension%3A0p+NOT+nothack&type=Code
    - `1p`: https://github.com/search?p=1&q=extension%3A1p+NOT+nothack&type=Code
    - `1pm`: https://github.com/search?q=extension%3A1pm+NOT+nothack&type=Code
    - `3G`/`3g`: https://github.com/search?p=1&q=extension%3A3G+NOT+nothack&type=Code (used by Apple SDK docs, apparently)
    - `3o`: https://github.com/search?p=1&q=extension%3A3o+NOT+nothack&type=Code&utf8=%E2%9C%93 (Common for Ocaml, it seems)
    - `3p`: https://github.com/search?q=extension%3A3p+NOT+nothack&type=Code
    - `3perl`: https://github.com/search?q=extension%3A3perl+NOT+nothack&type=Code
    - `3pm`: https://github.com/search?q=extension%3A3pm+NOT+nothack&type=Code

    - Licenses inline unless otherwise noted
    - `0p`: https://github.com/kitserver/kitserver5/blob/995dfb80cb2a5ca2cd1506ca772f629cc1fe92b6/src/soft/libarchive-2.4.12-1/man/man0p/cpio.h.0p
    - `1p`: https://github.com/mirrete/perlblog/blob/5696f2f67ce6e28600988df2d0e1aed74dd92d0f/blib/man1/blog_cgi.pl.1p
    - `1pm`: https://github.com/germanleonz/scm_perl/blob/cbc22b52369ed7f0521b700da6778d60961e91d1/lib/share/man/man1p/instmodsh.1pm
    - `3G`: https://github.com/hjc4869/osxcross-bin/blob/62c9b301d472f81187dd2611299bfe106ad484a1/SDK/MacOSX10.11.sdk/usr/share/man/man3/glBeginConditionalRender.3G
    - `3o`: https://github.com/EnjoyLifeFund/macHighSierra-cellars/blob/49a477d42f081e52f4c5bdd39535156a2df52d09/ocaml/4.05.0/share/man/man3/Ast_helper.Const.3o (Ocaml License https://ocaml.org/docs/license.html LGPL 2.1)
    - `3p`: https://github.com/dank101/4.3BSD-Reno/blob/00328b5a67ffe35e67baeba8f7ab75af79f7ae64/share/doc/ucs/spms/pgrep.3p (BSD License)
    - `3perl`: https://github.com/jiaojie1989/graphviz-2.38.0/blob/65568018aee8a060c0acbd219d90b9ca3ddcb411/tclpkg/gv/gv.3perl (GraphViz License https://graphviz.org/license/ CPL 1.0)
    - `3pm`: https://github.com/part1zano/scripts/blob/2a4392dbce8503ebdd9192dcbe9c8ab78253a468/perl/rfc/usr/Vlad/Crypt-DH-0.06/blib/man3/Crypt::DH.3pm

    - There don't appear to be any other uses of this exension


Alhadis:
> […] maybe working on a dedicated strategy as @Alhadis suggested would be a better approach.

[I got this](https://github.com/github/linguist/blob/24d3e96621f0bbb3eb786be77db94454e47816b6/lib/linguist/strategy/manpage.rb).

jordemort:
@pchaigno 
  - `.Np` extensions - I added these because they're used in https://mirrors.edge.kernel.org/pub/linux/docs/man-pages/man-pages-posix/ and there seemed to be a bit of concern about supporting them over in the discussion on https://github.com/github/markup/pull/1196 but I'm fine with leaving them out in favor of waiting for a better Roff strategy
  - `.1pm` / `.3pm` - The `.3pm` extension is really the one I want added most out of all of these, since it seems to be in wide use by various Perl projects. I waffled on `.1pm` but it seems weird to add `.3pm` but not `.1pm`
  - `.3perl` - added because it seemed weird to do `.3pm` but not `.3perl` - I think `.3perl` might mostly be a Solarisism, though; it's probably fine to wait
  - `.3o` - not sure why I added this, it can also probably wait

pchaigno:
I'll always be leaning for the less-accurate but clearer solution for heuristics. Ugly code is harder to maintain, and we have to consider the performance impact.

Alhadis:
> and we have to consider the performance impact.

Actually, that's a large part of why this looks like a mess. 😀 Everything is hard-wired to short-circuit as soon as possible, leaving to a very inside-out looking flow of logic.

I think I'll open a W.I.P pull-request so we can discuss this a bit clearer. Just wait until I've finished rewriting the first mess. :D

Nweman (commented):
Nwes Thailand

PR #4314, opened 2018-11-06
Add Razor (cshtml) highlighting

worldbeater:
## Description

Currently, Github won't highlight `.cshtml` files properly. See https://github.com/github/linguist/issues/3760. Let's use highlighting from https://github.com/austincummings/razor-plus

Razor is like Django - a templating engine based on HTML and C#. See [MSDN documentation](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-2.1&tabs=visual-studio) for more info. Now, GitHub treats `.cshtml` files as regular C# files, but this is not quite right. If we use the grammar @austincummings made, we'll get a much better syntax highlighting.

## Checklist:

- [x] **I am changing the source of a syntax highlighting grammar**
  - [Old grammar](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fatom%2Flanguage-csharp%2Fblob%2Fmaster%2Fgrammars%2Fcsharp.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Faspnet%2FMusicStore%2Fblob%2Fmaster%2Fsamples%2FMusicStore%2FViews%2FStore%2FBrowse.cshtml&code=)
  - [New grammar](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Faustincummings%2Frazor-plus%2Fmaster%2Fsyntaxes%2Fcshtml.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Faspnet%2FMusicStore%2Fblob%2Fmaster%2Fsamples%2FMusicStore%2FViews%2FStore%2FBrowse.cshtml&code=)
  - [x] The extension of the language is used in hundreds of repositories on GitHub.com.
    - [Search results for `.cshtml`](https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Acshtml+NOT+nothack)
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [MusicStore](https://github.com/aspnet/MusicStore/blob/master/samples/MusicStore/Views/Store/Browse.cshtml)

PR #3933, opened 2017-12-02
Remove Arduino as a language

oldmud0:
See #3929 for the original issue.

PR #4315, opened 2018-11-07
Extend Language ID tests to prevent re-use and manual generation

lildude:
## Description

As discovered in https://github.com/github/linguist/pull/4314 our current tests don't detect if someone has manually incremented the language ID instead of using `script/update-ids`.  This PR attempts to implement changes to catch this scenario.

## Checklist:

Template doesn't really apply, but gonna keep this as the closest.

### DIFF
--- a/test/test_language.rb
+++ b/test/test_language.rb
@@ -385,7 +385,8 @@ class TestLanguage < Minitest::Test
   end
 
   def test_all_languages_have_a_valid_id
-    invalid = Language.all.select { |language| language.language_id < 0 || language.language_id >= (2**31 - 1) }
+    deleted_language_ids = [21]  # Prevent re-use of deleted language IDs
+    invalid = Language.all.select { |language| language.language_id < 0 || deleted_language_ids.include?(language.language_id) || (language.language_id > 431 && language.language_id < 1024) || language.language_id >= (2**31 - 1) }
 
     message = "The following languages do not have a valid language_id. Please run `script/update-ids` as per the contribution guidelines.\n"
     invalid.each { |language| message << "#{language.name}\n" }


test/test_language.rb
```diff
@@ -385,7 +385,8 @@ def test_all_languages_have_a_language_id_set
   end
 
   def test_all_languages_have_a_valid_id
-    invalid = Language.all.select { |language| language.language_id < 0 || language.language_id >= (2**31 - 1) }
+    deleted_language_ids = [21]  # Prevent re-use of deleted language IDs
```
lildude: 21 was removed in https://github.com/github/linguist/pull/3933. This prevents re-use "to be safe".

test/test_language.rb
```diff
@@ -385,7 +385,8 @@ def test_all_languages_have_a_language_id_set
...
+    invalid = Language.all.select { |language| language.language_id < 0 || deleted_language_ids.include?(language.language_id) || (language.language_id > 431 && language.language_id < 1024) || language.language_id >= (2**31 - 1) }
```
lildude: 431 is the last manually generated ID. 1024 is picked at random to catch attempts to manually set an ID. I also contemplated using `(2**20 - 1)` as a lower limit but that's also just picked at random with no qualification.

I was going to implement this as:

```ruby
language.language_id != Digest::SHA256.hexdigest(language.name).to_i(16) % (2**30 - 1)
```

... however this won't work when a language name changes as the ID determined in this test would no longer be based off the original SHA generated off the original name and adding a list of "exemptions" feels a little 🤢 to me.

I'm open to suggestions on how best to catch _most_ attempts at setting an ID manually.

lildude: > Is there no way to run tests on diffs?

There is, but that could get quite hairy and time consuming and would mean shelling out or getting fancy with something like Rugged.

If the gen'd ID has a predictable format or guaranteed minimum size (in value of number of chars), that could be an alternate and more robust solution too.

pchaigno: We could write a Ruby script to validate the `git diff HEAD^`, but I'm not sure it's worth it here. I'll investigate later, when I'll have more time.

lildude: > We could write a Ruby script to validate the `git diff HEAD^`, but I'm not sure it's worth it here. I'll investigate later, when I'll have more time.

👍 I looked into this but it was quite slow, especially if we want to catch staged and unstaged files for testing locally before committing.

I'll leave that in your capable hands - when you have time - and merge this for now. We can iterate if we need to.

PR #4317, opened 2018-11-08
Add strategy to identify Roff man pages

Alhadis:
This PR adds a strategy to identify man-pages with unrecognised file extensions. Added in support of github/markup#1196.

## Description
Conventionally, manpage filenames are of the form:

	title "." section [group]

Where `section` is a digit between 1-9, and `group` is an arbitrary alphanumeric identifier. Currently, we have no way to accommodate the latter, [except](https://github.com/github/linguist/blob/58ef7278579a6dca0bc16729836833e19d6d28cd/lib/linguist/languages.yml#L4118-L4133) for group suffixes which are the most common:

~~~
.1in
.1m
.1x
.3in
.3m
.3qt
.3x
~~~

This leaves out a lot, especially in SunOS-based distributions [like illumos](https://github.com/illumos/illumos-gate/tree/master/usr/src/man). This strategy fixes that with a solution similar to @pchaigno's XML header. It might be too strict with what it expects of manpage headers; let me know if you think the strategy should be more lenient.

I've tested this with two corpora of man-pages:
* [`1-9`](https://github.com/Alhadis/Linguist-Silo/tree/1-9): Harvested from in-the-wild search results
* [`manpages`](https://github.com/Alhadis/Linguist-Silo/tree/manpages): Taken from the ,manual directories of four different systems

They all pass, save for a handful of `.n` files which get misclassified as Tcl. We can fix that later with a heuristic, if needed.

/cc @pchaigno, @lildude, @jordemort, @pali

## Checklist:


Alhadis:
Closing this because I've a hunch we're going about things the wrong way...

Given that `github/markup` has no control over the "renderability" of individual files, there's a *huge* risk of github/markup#1196 screwing up the readability of both non-Roff and Roff files alike. See [my question here](https://github.com/github/markup/pull/1196#issuecomment-442025473).

I'll discuss my intended solution in a new PR. We can still add a manpage strategy, but it can be a hell of a lot simpler than the mess I've dumped upon you all here...

Issue #3924, opened 2017-11-30
Outstanding Grammar Issues


vmg:
The following is a detailed list of all the outstanding issues in the grammars that GitHub.com uses for syntax highlighting the code in our website.

These issues are detected by our grammars compiler (https://github.com/github/linguist/pull/3915) and are probably causing minor rendering bugs in the website. 

**Help is very much welcome!** If you're seeing bugs or rendering issues in your source code in GitHub, please start by taking a look at this list to make sure we're not detecting any issues in your language's grammar.

Feel free to ask any questions about any given issue and what would be the appropriate way to fix it. I'll keep the issue up-to-date as I work through grammar fixes myself.

cc @github/linguist @pchaigno @Alhadis 

----

- [ ] repository `https://bitbucket.org/Clams/sublimesystemverilog/get/default.tar.gz` (1 errors)
    - [ ] Unknown keys in grammar: `source.systemverilog` (in `Clams-sublimesystemverilog-8c37a53ac265/SystemVerilog.tmLanguage`) contains invalid keys (`hidden`)

- [ ] repository `vendor/grammars/Lean.tmbundle` (from https://github.com/leanprover/Lean.tmbundle) (1 errors)
    - [ ] Unknown keys in grammar: `source.lean` (in `Syntaxes/Lean.tmLanguage`) contains invalid keys (`Patterns[3].pattern`)

- [ ] repository `vendor/grammars/MagicPython` (from https://github.com/MagicStack/MagicPython) (1 errors)
    - [ ] Unknown keys in grammar: `source.python` (in `grammars/MagicPython.cson`) contains invalid keys (`first_line_match`)

- [ ] repository `vendor/grammars/NimLime` (from https://github.com/Varriount/NimLime) (2 errors)
    - [ ] Missing include in grammar: `source.nim` (in `Syntaxes/Nim.YAML-tmLanguage`) attempts to include `source.asm` but the scope cannot be found
    - [ ] Missing include in grammar: `source.nim` (in `Syntaxes/Nim.YAML-tmLanguage`) attempts to include `text.html.markdown.multimarkdown` but the scope cannot be found

- [ ] repository `vendor/grammars/Scalate.tmbundle` (from https://github.com/scalate/Scalate.tmbundle) (1 errors)
    - [ ] Missing include in grammar: `source.scaml` (in `Syntaxes/Scaml.tmLanguage`) attempts to include `source.js.jquery` but the scope cannot be found

[...]

PR #4322, opened 2018-11-12
Use correct scope for Razor-plus grammar

lildude:
## Description

Whilst generating and testing the upcoming v7.1.0 release I noticed that the cshtml/razor change in https://github.com/github/linguist/pull/4314 introduced the wrong scope for the grammar and this is picked up by the grammar compiler:

```
- [ ] repository `vendor/grammars/razor-plus` (from https://github.com/austincummings/razor-plus) (2 errors)
    - [ ] Unexpected scope in repository: `text.html.cshtml` declared in `text.html.cshtml` (in `syntaxes/cshtml.tmLanguage.json`) was not listed in grammars.yml
    - [ ] Missing scope in repository: `text.html.razor` is listed in grammars.yml but cannot be found
```

The correct scope is `text.html.cshtml` as the error reports and can be confirmed by [checking the grammar itself](https://github.com/austincummings/razor-plus/blob/d1d3a9b5178f1afb963034c788eb143a090df5b9/syntaxes/cshtml.tmLanguage.json#L3).  This PR corrects the scope as using the wrong scope results in _no_ syntax highlighting at all.

This would have been picked up by the `script/add-grammar` script that should have been used to replace the grammar. I suspect it wasn't, hence this slipped through.

## Checklist:

A quick review on this would be ✨ so I can get the release back into testing today.

/cc @worldbeater FYI as the author of https://github.com/github/linguist/pull/4314

### DIFF
--- a/grammars.yml
+++ b/grammars.yml
@@ -637,7 +637,7 @@ vendor/grammars/r.tmbundle:
 vendor/grammars/rascal-syntax-highlighting:
 - source.rascal
 vendor/grammars/razor-plus:
-- text.html.razor
+- text.html.cshtml
 vendor/grammars/ruby-slim.tmbundle:
 - text.slim
 vendor/grammars/sas.tmbundle:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1876,7 +1876,7 @@ HTML+PHP:
   language_id: 151
 HTML+Razor:
   type: markup
-  tm_scope: text.html.razor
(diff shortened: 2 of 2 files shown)


Alhadis:
We should really have a test which catches this...

lildude:
> We should really have a test which catches this...

Yeah, though we really need to address all of the outstanding issues in https://github.com/github/linguist/issues/3924 first, or maybe whitelist those outstanding grammars in the mean time.

Issue #4219, opened 2018-08-04
Extension popularity monitoring


pchaigno:
I'm opening this issue to track the popularity of submitted file extensions (pull requests with the label `Pending Popularity`). I've used [Harvester](https://github.com/Alhadis/Harvester) to count the number of repositories for all extensions where it makes sense (i.e., if the number of files is below 200 we already know there won't be hundreds of repositories as required). All numbers were updated today, in the last few hours.

I'll update the numbers in a few months and will probably stop counting for extensions whose number of repositories stagnated or reduced.

Extension | PR | Files | Repo. | Date | +/- at Feb. 21
---|---|---|---|---|---
[~~`.mxl`~~](https://github.com/search?p=3&q=extension%3Amxl+www.musicxml.org&type=Code&utf8=%E2%9C%93) | #3651 | 106 | - | Aug. 4 18 | +0 files
[~~`.exw`~~](https://github.com/search?q=extension%3Aexw+NOT+djfhdjfjd) | #3754 | 618 | 41 | Aug. 4 18 | +8 repo.
[~~`.exu`~~](https://github.com/search?q=extension%3Aexu+NOT+djfhdjfjd) | #3754 | 54 | - | Aug. 4 18 | +4 files
[~~`.sarl`~~](https://github.com/search?utf8=%E2%9C%93&q=extension%3Asarl+io.sarl.lang.core&type=Code) | #3772 | 252 | 49 | Aug. 4 18 | +5 repo.
[~~`.hsig`~~](https://github.com/search?q=extension%3Ahsig+NOT+dkfjdfh) | #3855 | 459 | 64 | Aug. 4 18 | +21 repo.
[~~`.imba`~~](https://github.com/search?utf8=%E2%9C%93&q=extension%3Aimba+NOT+djfhdjfdj&type=) | #3869 | 611 | 58 | Aug. 4 18 | +57 repo.
[~~`.coco`~~](https://github.com/search?utf8=%E2%9C%93&q=extension%3Acoco+NOT+nothack&type=Code) | #3872 | 641 | 79 | Aug. 4 18 | +11 repo.
[~~`.pbf`~~](https://github.com/search?utf8=%E2%9C%93&q=extension%3Apbf+Event+NOT+nothack&type=Code) | #3926 | 62 | - | Aug. 4 18 | +0 files
[...]

PR #4323, opened 2018-11-12
Add support for Svelte.

umanghome:
Add support for [Svelte](https://svelte.technology).

## Checklist:

    - Sample source(s):
    - Sample license(s):

      -  https://github.com/search?q=extension%3Asvelte+export+default+NOT+nothack&type=Code
    - Sample source(s):
      - [`Form.svelte`](https://github.com/umanghome/linguist/blob/svelte/samples/Svelte/Form.svelte)
    - Sample license(s):
  
  - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fumanghome%2Fsvelte-atom%2Fmaster%2Fgrammars%2Fsvelte.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fumanghome%2Flinguist%2Fsvelte%2Fsamples%2FSvelte%2FCatList.svelte&code=

    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4353,6 +4353,7 @@ Svelte:
   color: "#e34c26"
   extensions:
   - ".svelte"
+  - ".htmlx"
 SVG:
   type: data
   extensions:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4343,6 +4343,16 @@ STON:
   tm_scope: source.smalltalk
   ace_mode: text
   language_id: 336
+Svelte:
+  type: markup
+  tm_scope: text.html.svelte
+  group: HTML
+  ace_mode: html
+  codemirror_mode: htmlmixed
+  codemirror_mime_type: text/html
+  color: "#e34c26"
+  extensions:
+  - ".svelte"
 SVG:
   type: data
   extensions:
--- a/.gitmodules
+++ b/.gitmodules
@@ -901,6 +901,9 @@
 [submodule "vendor/grammars/sublimetext-cuda-cpp"]
 	path = vendor/grammars/sublimetext-cuda-cpp
 	url = https://github.com/harrism/sublimetext-cuda-cpp
+[submodule "vendor/grammars/svelte-atom"]
+	path = vendor/grammars/svelte-atom
+	url = https://github.com/umanghome/svelte-atom
 [submodule "vendor/grammars/swift.tmbundle"]
 	path = vendor/grammars/swift.tmbundle
 	url = https://github.com/textmate/swift.tmbundle
--- a/grammars.yml
+++ b/grammars.yml
@@ -738,6 +738,8 @@ vendor/grammars/sublimeprolog:
 - source.prolog.eclipse
 vendor/grammars/sublimetext-cuda-cpp:
 - source.cuda-c++
+vendor/grammars/svelte-atom:
+- source.svelte
 vendor/grammars/swift.tmbundle:
 - source.swift
 vendor/grammars/tcl.tmbundle:
--- a/samples/Svelte/Form.htmlx
+++ b/samples/Svelte/Form.htmlx
@@ -1,0 +1,37 @@
+<form on:submit="createHandler(event)">
+  <div>
+    <label for="slinky-title">Title</label>
+    <input
+      bind:value='title'
+      id="slinky-title"
+      name="title"
+      type="text"
+    />
+  </div>
+  <button type="submit">Add Now</button>
+  <button type="button" on:click="logState()">Log</button>
+</form>
+
+<script>
+  const defaultData = {
+    title: '',
+  }
+
+  export default {
+    data: () => defaultData,
+
+    methods: {
+      logState() {
+        console.log(this.get())
+      },
+
+      createHandler(event) {
+        event.preventDefault()
+        // parse, validate, fire, reset
+        const slinky = this.get()
+        this.fire('create', slinky)
+        this.set(defaultData)
+      }
+    },
+  }
+</script>
--- a/samples/Svelte/Form.svelte
+++ b/samples/Svelte/Form.svelte
@@ -1,0 +1,37 @@
+<form on:submit="createHandler(event)">
+  <div>
+    <label for="slinky-title">Title</label>
+    <input
+      bind:value='title'
+      id="slinky-title"
+      name="title"
+      type="text"
+    />
+  </div>
+  <button type="submit">Add Now</button>
+  <button type="button" on:click="logState()">Log</button>
+</form>
+
+<script>
+  const defaultData = {
+    title: '',
+  }
+
+  export default {
+    data: () => defaultData,
+
+    methods: {
+      logState() {
+        console.log(this.get())
+      },
+
+      createHandler(event) {
+        event.preventDefault()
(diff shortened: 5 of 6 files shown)


umanghome:
@pchaigno Done.

Rich-Harris:
Svelte creator here, via [Twitter](https://twitter.com/umanghome/status/1061945549161750528). Amazing work Umang, thank you!

I noticed that the [sample file](https://raw.githubusercontent.com/umanghome/linguist/svelte/samples/Svelte/CatList.svelte) is using the old Svelte 1 template syntax (e.g. double `{{curlies}}`) — is that intentional? Relatedly, we created a separate project to maintain the Svelte 2 syntax called [HTMLx](https://github.com/htmlx-org/HTMLx) (the intention being that other frameworks are free to adopt it without having to invent their own syntax). Eventually our recommendation will be that people use the `.htmlx` file extension. Is it possible to support `.htmlx` here alongside `.svelte`?

umanghome:
@Rich-Harris Didn't notice the example was written in an older version. I picked up a file from an example on GitHub because the Svelte code that I write for work is confidential. I will pick up an updated sample and update the repository.

I can add `.htmlx` too, but the [usage doesn't seem to be very much](https://github.com/search?utf8=%E2%9C%93&q=extension%3Ahtmlx+export+default+NOT+nothack&type=Code) yet. I took a look at the README and it seems like it uses the same Grammar as Svelte does. In which case, it would make more sense to add support for `HTMLx` syntax rather than `Svelte` syntax as the base, and then let `Svelte` inherit this base.

Rich-Harris:
Yeah, usage will be effectively zero for now since we haven't promoted the extension at all (because of the lack of linguist support, which this PR solves!) — we generally just go with good old `.html`, which gets you 90% of the way there. I spot-checked a few existing uses of `.htmlx` (without restricting it to files that also include `export default`) and it seems that in most cases it's temporary renaming of files (i.e. people adding an 'x' as an alternative to e.g. 'file.OLD.html')

umanghome:
@Rich-Harris Would you agree on keeping `HTMLx` as the base syntax and then inheriting `Svelte` syntax support from it for the sole purpose of syntax highlighting?

Rich-Harris:
That makes sense to me, yes

umanghome:
I'm keeping `.htmlx` another extension for Svelte as of now. We can move it later to its own language once it gains enough traction.

lib/linguist/languages.yml
```diff
@@ -4343,6 +4343,17 @@ STON:
...
+Svelte:
+  type: markup
+  tm_scope: text.html.svelte
+  group: HTML
+  ace_mode: html
+  codemirror_mode: htmlmixed
+  codemirror_mime_type: text/html
+  color: "#e34c26"
+  extensions:
+  - ".svelte"
```
pchaigno: I counted 117 repositories by 102 users from the search query you provided. That a bit low considering our usual requirement of *hundreds* of repositories, but it's unlikely we'll get any conflicts in the future given the extension. @lildude What do you think?

lib/linguist/languages.yml
```diff
@@ -4343,6 +4343,17 @@ STON:
   tm_scope: source.smalltalk
   ace_mode: text
   language_id: 336
+Svelte:
+  type: markup
+  tm_scope: text.html.svelte
```
pchaigno: This needs to be `source.svelte`.

lib/linguist/languages.yml
```diff
@@ -4343,6 +4343,17 @@ STON:
...
+  group: HTML
+  ace_mode: html
+  codemirror_mode: htmlmixed
+  codemirror_mime_type: text/html
+  color: "#e34c26"
```
pchaigno: You don't need to set a color since it has a group; it will automatically take the parent's color.

lib/linguist/languages.yml
```diff
@@ -4343,6 +4343,17 @@ STON:
   tm_scope: source.smalltalk
   ace_mode: text
   language_id: 336
+Svelte:
```
pchaigno: Items in `languages.yml` should be alphabetically sorted, with uppercase letters first.

Rich-Harris:
They're not Svelte files — my hypothesis is:

> in most cases it's temporary renaming of files (i.e. people adding an 'x' as an alternative to e.g. 'file.OLD.html'

Happily, HTML files highlighted as HTMLx will be identical except in some unusual edge cases, so the worst that will result from the false positive is that those files will now be highlighted as HTML instead of not at all.

At present, the vast majority of Svelte files are written with an `.html` extension. There are enough `.svelte` files to show up in search, but ideally we'd want to use the `.htmlx` extension instead — we've just held off doing that someone got around to sending a PR to linguist (and that someone turned out to be @umanghome!).

pchaigno:
> They're not Svelte files — my hypothesis is:
> 
> > in most cases it's temporary renaming of files (i.e. people adding an 'x' as an alternative to e.g. 'file.OLD.html'

Sorry, I missed that before.

> Happily, HTML files highlighted as HTMLx will be identical except in some unusual edge cases, so the worst that will result from the false positive is that those files will now be highlighted as HTML instead of not at all.

Usage is way too low to add them now. Maybe we could add `.htmlx` as a HTML extension though (in a separate pull request)? I counted 351 repositories by 327 users so usage is fine.

If you think usage of `.htmlx` for Svelte files will grow rapidly, I can add the extension to #4219.

> we've just held off doing that someone got around to sending a PR to linguist.

With this pull request, you'll be able to switch on syntax highlighting of `.htmlx` Svelte files on a per-repository basis, using [Linguist overrides](https://github.com/github/linguist#overrides).

Rich-Harris:
Ah, nice, I didn't know about overrides.

> Maybe we could add `.htmlx` as a HTML extension though (in a separate pull request)?

That would make it practical to start using that extension, which could hopefully solve the chicken-and-egg problem we have at the moment, so yes — that sounds like an excellent starting point. Could you point us in the right direction please?

I expect usage *will* grow in time, yes — if you could add it to #4219 that'd be great. Thank you!

umanghome:
If we are going to add support for `.htmlx`, I would recommend making `.svelte` one of the file extensions for the HTMLx language. I would be happy to work on adding HTMLx in a different PR if we decide to close this one and go with HTMLx instead.

lib/linguist/languages.yml
lildude: Unfortunately, that usage is not high enough for inclusion right now, it's barely over a _hundred_ let alone _hundred**s**_.

pchaigno:
> Could you point us in the right direction please?

You can follow [the contribution guidelines](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language) for that.

> If we are going to add support for .htmlx, I would recommend making .svelte one of the file extensions for the HTMLx language. I would be happy to work on adding HTMLx in a different PR if we decide to close this one and go with HTMLx instead.

I wasn't suggesting to add HTMLx as a language, just `.htmlx` as a new HTML extension.

> if you could add it to #4219 that'd be great.

Done.

PR #4319, opened 2018-11-09
Improve Rust color

pmaddams:
The Rust color on GitHub doesn't look much like rust, and I don't see where it could have come from (not on the website or part of the logo). In fact, it looks rather pale and sickly, rather than vibrant and robust.

#a62c00 is much nicer. It resembles the red paints that have traditionally been made with iron oxide.

PR #4326, opened 2018-11-13
Revert rust color change

0xpr03:
The previous change #4319  made it hard to distinguish the language and made it hard to recognize for all users that are used to the old color. Practically speaking the change introduced more problems than the esthetics it wanted to solve gained us. Also esthetics are personal opinions, so I'd leave a new color up to the rust team to decide.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4232,7 +4232,7 @@ Ruby:
   language_id: 326
 Rust:
   type: programming
-  color: "#a62c00"
+  color: "#dea584"
   extensions:
   - ".rs"
   - ".rs.in"


atouchet:
I am in favour of this. As has been mentioned in the comments of #4319 the deeper red colour is harder to distinguish from other languages (D, Ruby, Scala, etc.) while the original colour worked just fine. Also I do not believe that the #a62c00 shade has any real association with the Rust community.

coder543:
FWIW, I started a humorous thread [over here](https://www.reddit.com/r/rust/comments/9wsas2/github_changed_the_language_color_for_rust/) discussing the change. It seems like most people favor the old color, although I don't have as strong of an opinion on this subject. I slightly favor the old color.

xtian:
I am in favor of this as well. The color that people would most associate with Rust would be that of its unofficial mascot, which is orange: http://rustacean.net

Additionally, the rationale that the old color doesn't look like iron oxide is misguided since the language was named after the rust fungus, which has an orange color: https://en.wikipedia.org/wiki/Rust_(programming_language)#History

meme:
I, too, prefer the older colour. Please bring it back. Annoying that this change was done without the O.K. of the Rust team.

quininer:
I think [Rust color](https://en.wikipedia.org/wiki/Rust_(color)) is a better choice.

xieyuheng:
I, too, prefer the older colour. Please bring it back!!!!!!!!!!!

amatuni:
I think the new color is 10000x better than the old one. The color suggested by @quininer could be a nice compromise though.

xieyuheng:
**OLD COLOR BACK !!!!!!!!!!!**

xieyuheng:
Otherwise, I will delete all my repos and move to **gitlab**.

CharlotteMoore927:
the new colour is great and definitely better than the old one!

elpiel:
I am in favor of the old one as well. It's too dark and distinguishable from other languages.

lildude:
I'll try get this out today.

lildude:

Issue #3394, opened 2016-12-19
Repository lang stats graph: Poor color contrast


jonsuh:
I’m color blind (deutan) and found the CoffeeScript blue nearly impossible to tell apart from the CSS purple with the naked eye:

![image](https://cloud.githubusercontent.com/assets/359871/21320838/f2d2fb9a-c5d7-11e6-8b1a-5e34e91f278b.png)
From repo https://github.com/postcss/autoprefixer

Nothing critical, but I thought it was worth mentioning from an accessibility standpoint.

I was directed to open an issue here by @jasonlong on Twitter: https://twitter.com/jasonlong/status/810887402260692992

Thanks!

gabru-md:
The same is the case with `Go` and `Python`.
I am not color blind, but even I find it difficult to differentiate between the two.
![cb](https://user-images.githubusercontent.com/25414121/39965239-d33acc0a-56b2-11e8-978d-a0b605f57113.png)

From the repository [https://github.com/taskcluster/json-e](https://github.com/taskcluster/json-e)

Alhadis:
Colour-blind as well (deutan), and just chiming in to say some colours are impossible for me to distinguish... I often find myself clicking the language bar out-of-habit just to bring up the language stats, because I'm so used to seeing CSS/CoffeeScript repos... 😞

tiehuis:
Wrote a small program that compares the colors for languages as specified by linguist. Would probably require a bit of context in what languages are commonly found together but may prove helpful.

https://github.com/tiehuis/linguist-color-checker

jonsuh:
@Alhadis Dividers definitely help define the breakdown! I think it’s a good solution—consider this solved :) Thank you for the follow up!

PR #4331, opened 2018-11-15
Improve Golang color

alassanecoly:
The actual color of Go doesn't looks like the color of the official Gopher mascot.

`#5FC9D8` fits more to the mascot which is a bright and joyful blue turquoise.

Also the actual deep blue color (`#375EAB`) looks a lot like PHP (`#4F5D95`) & Python (`#3572A5`) ones.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1649,7 +1649,7 @@ Gnuplot:
   language_id: 131
 Go:
   type: programming
-  color: "#375eab"
+  color: "#5FC9D8"
   aliases:
   - golang
   extensions:


Alhadis:
Going by [Golang's style-guide](https://storage.googleapis.com/golang-assets/Go-brand-book-v1.9.5.pdf#page=16), I'd say **Gopher Blue** `#00ADD8` is the more appropriate and "official" choice of colour to represent Go.

Since this is endorsed by the Go project's official branding, I don't see too much ire being raised over this change... nonetheless, @alassanecoly should be aware of #4326 (which eventuated rather recently),. Changing the colours of well-known programming languages should be done with caution and only when project styleguides warrant it...

alassanecoly:
You're right. I directly took the color from the Gopher mascot. So does the actual color should be changed to `#00ADD8` or it would better to leave it like that ?

Alhadis:
Stick to their official blue `#00ADD8`. The one which looks the most "official" of the bunch. It's harder to justify a wanton colour change that doesn't start out like *"...well, *I* thought the new colour looked more vibrant and energetic!"*

alassanecoly:
😹 I got you

pchaigno:
The color conflicts with Dockerfile's. @alassanecoly Could you check whether changing Dockerfile's color to [the second color associated with the language](https://brandcolors.net/b/docker) fixes the issue?

alassanecoly:
### Before
capture d ecran 2018-11-16 a 14 18 30


### After
capture d ecran 2018-11-16 a 14 20 07

alassanecoly:
Should I open a new PR to fix the issue ?

alassanecoly:
@pchaigno @lildude Any updates ?

pchaigno:
Did you check that changing Dockerfile to its second color fixes build failures?

alassanecoly:
@pchaigno Build fixed 🙌

pchaigno:
@lildude @Alhadis Do you think it's okay if we change the Dockerfile color to be able to change the Golang one? The new color for Dockerfile is selected with the [same rationale](https://github.com/github/linguist/pull/4169#issuecomment-399171517) as the old one.

lildude:
Yeah, I'm also a little hesitant changing the colours for a prominent language like Go after the Rust furore, especially as this PR, at a quick glimpse, almost looks like we're switching the two colours around.

The current Go colour has also been in use for nearly 5 years - this change _will_ be noticed.

The question is how much of a stink is this likely to cause? 

Maybe not much, as in both cases we're using official brand colours, and I think the proximity to Perl won't be too much of a problem either as the current colour is already close to several others and I don't think many peeps will be mixing the two, though you never know.

Alhadis:
> The question is how much of a stink is this likely to cause?

Probably not nearly as much as the Rust colour change caused. We have Go's branding as a sound justification for the colour change, and googling "Go Lang" in Google Images reveals a swarth of cyan-coloured mascots. It's actually *harder* to justify the current colour, really.

pchaigno:
> The question is how much of a stink is this likely to cause?

Should we raise the issue at https://groups.google.com/forum/#!forum/golang-nuts for this as well?
<br>

@lildude - I think @Alhadis' improvement makes a lot of sense and might also help color-blind peeps (cf #3394). Should we move it to a separate issue and ping the appropriate Githubbers?

lildude:
> Should we raise the issue at https://groups.google.com/forum/#!forum/golang-nuts for this as well?

That might be best. We've now got the Crystal peeps upset too 😢 

> @lildude - I think @Alhadis' improvement makes a lot of sense and might also help color-blind peeps (cf #3394). Should we move it to a separate issue and ping the appropriate Githubbers?

No need. We've got an issue open on the main GitHub repo already, and it includes a suggestion like @Alhadis's. I'll give it another nudge as it hasn't been touched in a while.

alassanecoly:
> > Should we raise the issue at https://groups.google.com/forum/#!forum/golang-nuts for this as well?
> 
> @lildude: That might be best. We've now got the Crystal peeps upset too 😢
> 

Posted an issue at golang-nuts on december 5th: https://groups.google.com/forum/#!topic/golang-nuts/Xr_x9REPC-c.

98 views but nobody replied 😓

Alhadis:
> 98 views but nobody replied 😓

22 days later, still no response. I guess we're either right to change this or they simply don't care. 😅

Alhadis:
Heh. That's true, in a way. 😉 But out of the 111 views the topic's received, none of the readers felt strongly enough to comment. Hopefully that's indicative of the Go community at large. 🤞 If not, well, we can use Go's official styleguide as a defence.

It's possible we're being overly cautious in light of the Rust backlash. Rust's colour was changed on purely arbitrary grounds, Go's colour change is not.

Alhadis:
Nudging @lildude.

leighmcculloch:
This change is great, thanks for updating Go to display with the correct color. Can't wait for it go live. 🎉

Lomanic:
Don't want to be a trouble-maker but I was used for years to this color and I would have preferred it to stay unchanged.

welkie:
I came here from Reddit discussions about the color change. It's a shame that the community at large doesn't know about this. I had no idea that the color changing and style changes like that new white border between the colors were done publicly through PRs like this. I'm a new Go dev, and I love the new color. It makes much more sense. I also like the new white border. It was so subtle a change that I never actually noticed it, but it works great.

If I would have known about this PR I would have chimed in with my opinion, as I'm sure many other devs would have, so that it wouldn't have stagnated like that with weeks of inactivity.

pchaigno:
We did try to reach out to the community (see above) but it's not always obvious how to do that.

pmaddams:
I'm the perpetrator of #4319...and I approve this change.

welkie:
I think this is a good example of how to do this right. @pmaddams reverted the change when the community stated they didn't like it. Here, the Go community seems to like this change. If the communities are happy, we're probably doing something right. All's well that ends well?

PR #4334, opened 2018-11-20
add vvvv as new language

colorchestra:
## Description

I'm adding vvvv as a new language. vvvv is a visual programming language that has been around since 1998 and stores its projects (also called _Patches_) in `v4p` files, which internally use XML. Since vvvv is programmed completely through a graphical interface, there is no 'code' as such; for that reason, no grammar is included.

~~The chosen gray (#e6e6e6) is vvvv's characteristic background color.~~
The color, #C0C0C0, is that of a vvvv node. The originally chosen color conflicted with another one already in use.

## Checklist:
      
      -  `v4p`: https://github.com/search?q=extension%3Av4p+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/colorchestra/vvvv-add-sample
    - Sample license(s): MIT

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5581,7 +5581,7 @@ sed:
   tm_scope: source.sed
   language_id: 847830017
 vvvv:
-  type: data
+  type: programming
   color: "#C0C0C0"
   extensions:
   - ".v4p"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5087,6 +5087,13 @@ Vue:
   tm_scope: text.html.vue
   ace_mode: html
   language_id: 391
+vvvv:
+  type: markup
+  color: "#e6e6e6"
+  extensions:
+  - ".v4p"
+  - ".vl"
+  language_id: 315707027
 Wavefront Material:
   type: data
   extensions:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5581,12 +5581,11 @@ sed:
   tm_scope: source.sed
   language_id: 847830017
 vvvv:
-  type: markup
+  type: data
   color: "#C0C0C0"
   extensions:
   - ".v4p"
-  - ".vl"
-  tm_scope: none
+  tm_scope: text.xml
   ace_mode: xml
   language_id: 315707027
 wdl:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5087,13 +5087,6 @@ Vue:
   tm_scope: text.html.vue
   ace_mode: html
   language_id: 391
-vvvv:
-  type: markup
-  color: "#e6e6e6"
-  extensions:
-  - ".v4p"
-  - ".vl"
-  language_id: 315707027
 Wavefront Material:
   type: data
   extensions:
@@ -5587,6 +5580,15 @@ sed:
   ace_mode: text
   tm_scope: source.sed
   language_id: 847830017
+vvvv:
+  type: markup
+  color: "#C0C0C0"
+  extensions:
+  - ".v4p"
+  - ".vl"
+  tm_scope: none
+  ace_mode: xml
+  language_id: 315707027
 wdl:
   type: programming
   color: "#42f1f4"
--- a/samples/vvvv/add.v4p
+++ b/samples/vvvv/add.v4p
@@ -1,0 +1,101 @@
+<!DOCTYPE PATCH  SYSTEM "http://vvvv.org/versions/vvvv50beta37.dtd" >
+
+   <PATCH nodename="C:\Users\case-12345\Desktop\example.v4p" scrollx="30" scrolly="0" systemname="example" filename="C:\Users\case-12345\Desktop\example.v4p">
+
+   <BOUNDS type="Window" left="3975" top="6540" width="9000" height="6000">
+
+   </BOUNDS>
+
+   <NODE systemname="IOBox (Value Advanced)" nodename="IOBox (Value Advanced)" componentmode="InABox" id="0">
+
+   <BOUNDS type="Node" left="1410" top="795" width="100" height="100">
+
+   </BOUNDS>
+
+   <BOUNDS type="Box" left="1410" top="795" width="795" height="240">
+
+   </BOUNDS>
+
+   <PIN pinname="Y Input Value" slicecount="1" values="9">
+
+   </PIN>
+
+   <PIN pinname="Y Output Value" visible="1">
+
+   </PIN>
+
+   </NODE>
+
+   <NODE systemname="IOBox (Value Advanced)" nodename="IOBox (Value Advanced)" componentmode="InABox" id="1">
+
+   <BOUNDS type="Node" left="3180" top="960" width="100" height="100">
+
+   </BOUNDS>
+
+   <BOUNDS type="Box" left="3180" top="960" width="795" height="240">
+
+   </BOUNDS>
+
+   <PIN pinname="Y Input Value" slicecount="1" values="12">
+
+   </PIN>
+
+   <PIN pinname="Y Output Value" visible="1">
+
+   </PIN>
+
+   </NODE>
+
+   <NODE systemname="Add (Value)" nodename="Add (Value)" componentmode="Hidden" id="2">
+
+   <BOUNDS type="Node" left="2490" top="1665" width="100" height="100">
+
+   </BOUNDS>
+
+   <PIN pinname="Input 1" visible="1">
+
+   </PIN>
+
+   <PIN pinname="Input 2" visible="1">
+
+   </PIN>
+
+   <PIN pinname="Output" visible="1">
+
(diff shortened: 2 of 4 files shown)


samples/vvvv/add.v4p
```diff
@@ -0,0 +1,101 @@
+<!DOCTYPE PATCH  SYSTEM "http://vvvv.org/versions/vvvv50beta37.dtd" >
```
pchaigno: Any chance you could find a sample under a permissive license?

lib/linguist/languages.yml
```diff
@@ -5580,6 +5580,15 @@ sed:
   ace_mode: text
   tm_scope: source.sed
   language_id: 847830017
+vvvv:
```
pchaigno: [It looks like](https://vvvv.org/documentation/vl) vvvv is the name of the software and VL the name of the language, no?

lib/linguist/languages.yml
```diff
@@ -5580,6 +5580,15 @@ sed:
...
+  type: markup
```
pchaigno: Why the markup type?

lib/linguist/languages.yml
```diff
@@ -5580,6 +5580,15 @@ sed:
...
+  color: "#C0C0C0"
+  extensions:
+  - ".v4p"
+  - ".vl"
+  tm_scope: none
```
pchaigno: `source.xml` would be better than nothing here.

samples/vvvv/add.v4p
colorchestra: Yes, I changed it to the MIT license.

lib/linguist/languages.yml
colorchestra: I figured that would be the right one because vvvv patches are stored as XML. Which type would you suggest?

lib/linguist/languages.yml
colorchestra: Ok, I'll change that

samples/vvvv/add.v4p
pchaigno: Oh, I missed that you were the author. How did you write this sample file? Is it representative of in-the-wild usage? Sample files are used to train the Bayesian classifier so it's important that they be from "real" applications.

lib/linguist/languages.yml
```diff
@@ -5087,6 +5087,13 @@ Vue:
   tm_scope: text.html.vue
   ace_mode: html
   language_id: 391
+vvvv:
+  type: markup
+  color: "#e6e6e6"
+  extensions:
+  - ".v4p"
+  - ".vl"
```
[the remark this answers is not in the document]
colorchestra: Yes, that first search link returns a lot of unrelated files, because other languages seem to use the `.vl` extension as well. Is that a problem?

From that second link, I've skimmed the first few pages of the search results - all of them are regular, valid vvvv / vl patches.

lib/linguist/languages.yml
colorchestra: No, vvvv is a language unto itself. vl is a relatively new, additional language that can be used within vvvv to solve certain problems more easily. I'd view vvvv and vl as a unit, but if you find that cleaner, I can remove the `.vl` extension from `languages.yml`, treat vl as a completely separate language and use this pull request to only add vvvv and its .v4p files to Linguist.

lib/linguist/languages.yml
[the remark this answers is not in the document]
pchaigno: > Yes, that first search link returns a lot of unrelated files, because other languages seem to use the `.vl` extension as well. Is that a problem?

Yes, with the present pull request, all `.vl` files will be recognized as vvvv. Could you update the search query provided in the original post for `.vl` to only return vvvv files (it doesn't need to be perfect, but should give us a good estimation of the number of `.vl` vvvv files)?

> From that second link, I've skimmed the first few pages of the search results - all of them are regular, valid vvvv / vl patches.

Oh, I thought there were not because they don't look much like your sample file. Maybe we need a more representative sample file? Or a few (3-4) instead of a single one?

lib/linguist/languages.yml
colorchestra: Actually, thinking about it - I think treating them as separate languages would be the best way to go, because they also look a bit different, which is what caused the confusion about the sample file, which is pure vvvv. I'll remove the .vl extension in the next commit.

lib/linguist/languages.yml
Alhadis: XML is actually classified by Linguist as `data`, as the language is used more for general purpose data serialisation than marking up readable documents.

Since you said that these files are generated via a graphical interface, it makes sense to treat them as data too, rather than human-written source code.

colorchestra: Ok, I've changed it to `data` for the next commit.

lib/linguist/languages.yml
colorchestra: `source.xml` doesn't seem to exist in `grammars.yml`. What should I use instead? `text.xml`?

pchaigno: Yes, my bad.

lib/linguist/languages.yml
pchaigno: Maybe we should change this pull request to add `.v4p` as a XML extension instead of a new language. It would achieve the same effect.

colorchestra: Well, the goal is to have "vvvv" show up as a language in repositories that use it. That wouldn't work that way, or would it?

samples/vvvv/add.v4p
colorchestra: It is a normal vvvv patch, albeit a very simple one. I added another one that is more the way vvvv is typically used, so that should be enough for the classifier.

lib/linguist/languages.yml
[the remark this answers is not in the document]
colorchestra: As I said below, I completely removed vl from this pull request to treat it as a completely separate language in the future. Which it technically isn't, but for this purpose (and training the classifier) this is the best option.

colorchestra:
Whoops, forgot to run `list-grammars`, just did it and the error went away. I'm still getting this one though and I don't understand it:

```
  1) Failure:
TestLanguage#test_no_unused_colours [/home/work/src/linguist/test/test_language.rb:469]:
Unused colour assigned to vvvv
```
How can I fix it?

lib/linguist/languages.yml
pchaigno: No but `data` languages don't show up in statistics either, at least not by default.

colorchestra:
Hey, is there anything still needed from my side? All issues / comments should have been resolved with the last commit.

pchaigno:
I'm a bit wary of adding this as a *programming* language. Many other XML files could be considered programming because they correspond to the backup files of visual programming editors.
@colorchestra Is there somewhere we could address the wider vvvv community?

/cc @Alhadis @lildude

Alhadis:
@colorchestra If you're interested in making your project detectable as a "VVVV repository", then you can use a [`gitattributes` override](https://github.com/github/linguist/blob/master/README.md#detectable). But marking this as "programming" is neither correct nor consistent with the current semantics of Linguist.

lib/linguist/languages.yml
```diff
@@ -5580,6 +5580,14 @@ sed:
   ace_mode: text
   tm_scope: source.sed
   language_id: 847830017
+vvvv:
```
Alhadis: Change the type to `data`, please.

pchaigno: I'd rather we simply add it as an XML extension. It'll achieve the same result.

colorchestra: Just so I understand it, what effect would it have to add it as an XML extension?

Using gitattributes overrides is of course pointless, because I want all vvvv projects to be visible as such and not just mine.

colorchestra:
Thinking about it, I'll close this pull request instead, adding vvvv as a data language / XML extension seems rather pointless to me. Although I'm a bit disappointed about these outdated views on what makes something a programming language.

lib/linguist/languages.yml
pchaigno: > Just so I understand it, what effect would it have to add it as an XML extension?

It would be correctly syntax highlighted and would be "searchable" as XML.

pchaigno:
> Although I'm a bit disappointed about these outdated views on what makes something a programming language.

We do not claim to define what makes a programming languages in general. We are only careful of what the consequence will be to classify a language as **"programming" inside Linguist**. In particular, we don't want to associate vvvv to repositories where users don't expect it. If you know a way to address the wider vvvv community, we could raise this issue there.

colorchestra:
But the search query I initially posted seems to be working just fine, looking through the results for a while I can’t find a single false positive. Or is that not what you mean by „where users don’t expect it“? 

vvvv has a [forum](https://discourse.vvvv.org/?q=forums) and a [Matrix chat room](https://vvvv.org/blog/chat-vvvv-on-matrix.org) that can be used without having to sign up.

PR #4336, opened 2018-11-22
Release v7.1.2

lildude:
A small quick patch release to update the Kotlin grammar to address the request in https://github.com/github/linguist/issues/4329 and fix Kotlin syntax highlighting.

Once deployed on GitHub.com, this will render correctly:

```kotlin
class PostalAddress(
  val streetAddress : String,
  val city : String,
  val zip : String,
  val state : USState?,
  val country : Country
) {
   assert {(state == null) xor (country == Countries["US"]) }
}

class PhoneNumber(
  val country : Country,
  val areaCode : Int,
  val number : Long
)
```

Fixes https://github.com/github/linguist/issues/4329

/cc @nishtahir @DanielJette


lildude:
This release is now live on GitHub.com and the sample in the OP is now rendering correctly.

PR #4339, opened 2018-12-01
Add chez scheme as an interpreter

amirouche:
Add chez `scheme` as scheme an interpreter.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4419,6 +4419,7 @@ Scheme:
   - ".sps"
   - ".ss"
   interpreters:
+  - scheme
   - guile
   - bigloo
   - chicken


Alhadis:
Hey there, 👋

What's "chez"? Could you please provide a link which details the existence of a Scheme interpreter named `scheme`?

Note that the `interpreters` list in `languages.yml` is only for programs which execute the file directly. E.g., `bash` for shell-scripts, `perl` for `*.pl` programs, etc.

amirouche:
The first paragraph at https://cisco.github.io/ChezScheme/csug9.5/use.html#./use:h1:

> When Chez Scheme is installed with default options, entering the command `scheme` at the shell's prompt starts an interactive Scheme session.

Alhadis:
> Chez Scheme Version 9.5.1 . 
> Copyright 1984-2017 Cisco Systems, Inc.

Been around for over 30 years. Strange that nobody thought to add it until now... :\

PR #4349, opened 2018-12-10
Added Jinja language

mwadman:
Adding Jinja2 as a separate language.

## Description
Previously, support for Jinja2 (https://github.com/pallets/jinja) was provided by the "HTML+Django" language in linguist.
Unfortunately, while Django and Jinja2 are similar (Jinja is "inspired by" Django), Jinja2 uses slightly different syntax in some situations. Jinja2 filters are a good example of this.
From Jinja2's docs (http://jinja.pocoo.org/docs/2.10/faq/#how-compatible-is-jinja2-with-django):
"For example filter arguments use a function call syntax rather than a colon to separate filter name and arguments."

I am using ".j2" as the main extension as this has greater representation than the ".jinja" and ".jinja2" file extensions.

For one example, I've taken the .jinja example from the Django language.
Let me know your thoughts on whether this is appropriate or not.

I apologize in advance as I'm not sure how to get Lightshow highlighting the language using the sourced tmbundle.

## Checklist:
      
    - Sample source(s):
    - Sample license(s):

      -  https://github.com/search?q=extension%3Aj2+NOT+nothack&type=Code
      -  https://github.com/search?q=extension%3Ajinja+NOT+nothack&type=Code
      -  https://github.com/search?q=extension%3Ajinja2+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/openstack/openstack-ansible/blob/677d0bdc08a93634ad1b09bae8859a5bc8a14288/tests/roles/bootstrap-host/templates/user_variables.aio.yml.j2
    - Sample license(s): Apache License, Version 2.0
  
  - [X] I have included a syntax highlighting grammar: https://github.com/kudago/jinja2-tmbundle

    - Sample source(s):
    - Sample license(s):


mwadman:
I'd like some assistance with the failing builds if possible.

"TestLanguage#test_codemirror_modes_present" is failing due to the mime type of Jinja2 being "null" (https://github.com/codemirror/CodeMirror/blob/master/mode/meta.js#L78).


I believe "TestLanguage#test_find_by_extension" is failing due to the presence of line 182 in test_language.rb (https://github.com/github/linguist/blob/master/test/test_language.rb#L182).

Should I be including a change to these tests in my pull request?

lildude:
> "TestLanguage#test_codemirror_modes_present" is failing due to the mime type of Jinja2 being "null" (https://github.com/codemirror/CodeMirror/blob/master/mode/meta.js#L78).

Yup. In such cases I turn to `file` and in this case it tells me (on macOS):

```
$ file --mime-type samples/Jinja2/user_variables.aio.yml.j2
samples/Jinja2/user_variables.aio.yml.j2: text/plain
$
```

So try `text/plain` instead of `null`.

> I believe "TestLanguage#test_find_by_extension" is failing due to the presence of line 182 in test_language.rb (https://github.com/github/linguist/blob/master/test/test_language.rb#L182).

Yup. You'll need to adjust that test... BUT... this last point brings up an important question... your PR appears to be lumping the old Django "Jinja1" into the new Jinja2 and calling them all "Jinja2". This doesn't sound right to me as it will cause all current uses of "Jinja1" to now be called "Jinja2" and you've sort of confirmed this isn't right with:

> Unfortunately, while Django and Jinja2 are similar (Jinja is "inspired by" Django), Jinja2 uses slightly different syntax in some situations. Jinja2 filters are a good example of this.

If Jinja2 really is its own language, this PR should _not_ change the behaviour of the older and now completely distinct language.

As it's quite similar to "Jinja1", it's probably also a good idea to include samples that distinctly show this difference so the classifier can identify the differences, if it needs to.

As for the final failing test:

```
/home/travis/build/github/linguist/vendor/licenses/grammar/jinja2-tmbundle.txt:
  - missing license text
  - license needs reviewed: none.
```

This is because the grammar you've sourced doesn't have a license so it wasn't added when you ran `script/add-grammar` to add the grammer. I've checked the grammar repo itself and there is no clear indication as to what license the grammar is licensed under - it talks about Jinja2's license, but not its own. We can't accept grammars that are not clearly licensed.

lildude:
> Would your preference be that I submit a pull request to CodeMirror to have the MIME changed, or make additional changes to linguist to allow the "null" MIME to pass tests?

I think a PR to the upstream repo is probably best as we won't be the only peeps hitting this problem.

Regarding Jinja1 vs Jinja2... as there's very little difference between the two and no easy way to differentiate the two, how about simply calling the language in this PR "Jinja", like we do for Python 2/3, and have it cater for both?

> I can't seem to find a suitable grammar that also has the appropriate license associated with it.
> I might take a look at writing my own and posting this shortly. Would there be any objections to this?

Not at all. You might also want to try opening an issue in the grammar you've found to see if they're prepared to explicitly license it under one of [these licenses](https://github.com/github/linguist/blob/257425141d4e2a5232786bf0b13c901ada075f93/vendor/licenses/config.yml#L2-L11).

mwadman:
CodeMirror PR submitted and referenced.
I'll take a look at either getting a license changed or writing my own grammar.

I'll close this PR for now and will re-open this at a later stage once I've determined how to proceed.

Thanks again for your time, @lildude.

mgeier:
For the record, https://github.com/codemirror/CodeMirror/pull/5745 (replacing https://github.com/codemirror/CodeMirror/pull/5709) has been merged some time ago.

PR #4354, opened 2018-12-14
Add new Hack file extensions.

azjezz:
## Description

Add support for hack `hhi` extension.

## Checklist:
      - `.hhi` : https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Ahhi+hh
    - Sample source(s):
      - [Map.hhi](https://github.com/facebook/hhvm/blob/3df035714e281ea56d2d99592757277fb8e44401/hphp/hack/src/utils/findUtils.ml#L16)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2051,6 +2051,7 @@ Hack:
   extensions:
   - ".hack"
   - ".hh"
+  - ".hhi"
   - ".php"
   tm_scope: source.hack
   color: "#878787"
--- a/samples/Hack/Map.hhi
+++ b/samples/Hack/Map.hhi
@@ -1,0 +1,769 @@
+<?hh
+/**
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the "hack" directory of this source tree.
+ *
+ */
+
+/**
+ * This file provides type information for some of HHVM's builtin classes.
+ *
+ * YOU SHOULD NEVER INCLUDE THIS FILE ANYWHERE!!!
+ */
+
+namespace HH {
+
+  /**
+   * `Map` is an ordered dictionary-style collection. HHVM provides a native
+   * implementation for this class. The PHP class definition below is not
+   * actually used at run time; it is simply provided for the typechecker and for
+   * developer reference.
+   *
+   * Like all objects in PHP, `Map`s have reference-like semantics. When a caller
+   * passes a `Map` to a callee, the callee can modify the `Map` and the caller
+   * will see the changes. `Map`s do not have "copy-on-write" semantics.
+   *
+   * `Map`s preserve insertion order of key/value pairs. When iterating over a
+   * `Map`, the key/value pairs appear in the order they were inserted. Also,
+   * `Map`s do not automagically convert integer-like `string` keys (ex. `"123"`)
+   * into integer keys.
+   *
+   * `Map`s only support `int` keys and `string` keys. If a key of a different
+   * type is used, an exception will be thrown.
+   *
+   * `Map`s support `$m[$k]` style syntax for getting and setting values by key.
+   * `Map`s also support `isset($m[$k])` and `empty($m[$k])` syntax, and they
+   * provide similar semantics as arrays. Adding an element with square bracket
+   * syntax `[]` is supported either by providing a key between the brackets or
+   * a `Pair` on the right-hand side. e.g.,
+   * `$m[$k] = $v` is supported
+   * `$m[] = Pair {$k, $v}` is supported
+   * `$m[] = $v` is not supported.
+   *
+   * `Map`s do not support iterating while new keys are being added or elements
+   * are being removed. When a new key is added or an element is removed, all
+   * iterators that point to the `Map` shall be considered invalid.
+   *
+   * `Map`s do not support taking elements by reference. If binding assignment
+   * (`=&`) is used with an element of a `Map`, or if an element of a `Map` is
+   * passed by reference, of if a `Map` is used with `foreach` by reference, an
+   * exception will be thrown.
+   *
+   * @guide /hack/collections/introduction
+   * @guide /hack/collections/classes
+   */
+
+  final class Map<Tk as arraykey, Tv> implements \MutableMap<Tk, Tv> {
+    /**
+     * Creates a `Map` from the given `KeyedTraversable`, or an empty `Map` if
+     * `null` is passed.
+     *
+     * @param $it - any `Traversable` object from which to create a `Map`
+     *              (e.g., `array`). If `null`, then an empty `Map` is created.
+     */
+    <<__Rx, __AtMostRxAsArgs>>
+    public function __construct(
+      <<__MaybeMutable, __OnlyRxIfImpl(\HH\Rx\KeyedTraversable::class)>>
+      ?KeyedTraversable<Tk, Tv> $it,
+    );
+
+    /**
+     * Returns an `array` containing the key/value pairs from the current `Map`.
+     *
+     * @return - an `array` containing the key and value pairs from the current
+     *           `Map`.
+     */
+    <<__Rx, __MaybeMutable, __PHPStdLib>>
+    public function toArray(): array<Tk, Tv>;
+
+    /**
+     * Returns an `array` containing the values from the current `Map`.
+     *
+     * @return - an integer-indexed `array` containing the values from the
+     *           current `Map`.
+     */
+    <<__Rx, __MaybeMutable>>
+    public function toValuesArray(): varray<Tv>;
+
+    /**
+     * Returns an `array` whose values are the keys of the current `Map`.
+     *
+     * @return - an integer-indexed `array` containing the keys from the current
+     *           `Map`.
+     */
+    <<__Rx, __MaybeMutable>>
+    public function toKeysArray(): varray<Tk>;
+
+    /**
+     * Returns a `Vector` with the values of the current `Map`.
+     *
+     * @return - a `Vector` that contains the values of the current `Map`.
+     */
+    <<__Rx, __MutableReturn, __MaybeMutable>>
+    public function toVector(): Vector<Tv>;
+
+    /**
+     * Returns an immutable vector (`ImmVector`) with the values of the current
+     * `Map`.
+     *
+     * @return - an `ImmVector` that is an immutable copy of the current `Map`.
+     */
+    <<__Rx, __MaybeMutable>>
+    public function toImmVector(): ImmVector<Tv>;
+
+    /**
+     * Returns a deep copy of the current `Map`.
+     *
+     * @return - a `Map` that is a deep copy of the current `Map`.
+     */
+    <<__Rx, __MutableReturn, __MaybeMutable>>
+    public function toMap(): Map<Tk, Tv>;
+
+    /**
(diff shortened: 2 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -1904,7 +1904,10 @@ Hack:
   codemirror_mode: php
   codemirror_mime_type: application/x-httpd-php
   extensions:
+  - ".hack"
+  - ".hck"
```
pchaigno: Usage for that extension is [still too low](https://github.com/search?q=extension%3Ahck+NOT+sdhdfjdhfjd&type=Code). Could you remove it?

azjezz:
@pchaigno i have removed the `.hck` extension, also i didn't add heuristic changes as the `.hack` and `.hhi` extensions are not used by any other language

azjezz:
cc @pchaigno

pchaigno:
Usage for the `.hhi` extension seems fine. I've counted 830 repositories by 746 users using [Harvester](https://github.com/Alhadis/Harvester) and [the `extension:hhi hh` search query](https://github.com/search?o=desc&q=extension%3Ahhi+hh&s=&type=Code).

We'll have an issue with the `.hack` extension however, if we merge this pull request as-is. There are [plenty of `.hack` files that contain just zeros and ones](https://github.com/search?o=desc&q=extension%3Ahack+NOT+hh&s=indexed&type=Code) which will be classified as Hack with this pull request. I don't know what these files are. Either we can identify them and classify them as their own (data?) language, or we'll need to remove the `.hack` extension from Hack for the moment.

azjezz:
@pchaigno 

#### Problems with `.php` extension :

currently its hard to detect hack files that use `.php` extension, for example i noticed that today my repository was reporting `11%` hack, when its a pure php app. the issue was that some `.php` files contained only HTML, since there's no `<?php`, linguist detects them as `hack`.

see : https://github.com/azjezz/web-php-mock-ups

#### Problems with `.hh` extension :
hh files used to be detected as `c++` header files, but since we added a heuristic rule last time to check for `<?hh` it was fixed.

the issue here is that hack is not planning to keep the `<?hh` tag after `hhvm 4.0` 
#### Solution : 
`.hack` seems to be the only reasonable solution.
there's no other language using this extension, makes it easy to detect ( note: there's no `<?hh` tag in `.hack` files )

#### Solution for 01 files
the solution would be to  identify them and classify them as their own (data or binary) language.

---

cc @fredemmott

fredemmott:
We’ve not yet announced support for .hack files - an while the support was committed, it is about to change, and these pull request examples will no longer be considered valid - as of next monday’s release, having any ‘<?’ token in a .hack file is a syntax error. They’re always strict mode.

azjezz:
@fredemmott i suggested these examples before we discussed the `<?hh` tag in `.hack` files :stuck_out_tongue: i have removed the `<?hh` tag.

jeffomatic:
Just for context: the existing `.hack` files are almost overwhelmingly going to be from solutions to the [Nand2Tetris](https://www.nand2tetris.org/) course, a popular MOOC for computer science (I recommend it!). `.hack` is a "compiled" machine code format for the fictitious CPU that students implement. As such, it's not a language that can be syntax-highlighted in any meaningful way. The content of those files are ASCII `1`'s and `0`'s, plus whitespace.

pchaigno:
Thank you all for the information!

I've checked 2817 `.hack` files from [the `extension:hack 0000000000010000 OR 0000000000000000` search query](https://github.com/search?q=extension%3Ahack+0000000000010000+OR+0000000000000000) using Harvester and found 931 repositories among 918 users. It's enough to add support for that second language. Once we've added that language, it shouldn't be hard to distinguish between the two with 100% accuracy :-)
@jeffomatic Any idea what we should name it? Looks like the official name is Hack, so maybe "Nand2Tetris Hack"?

> @fredemmott i suggested these examples before we discussed the `<?hh` tag in `.hack` files  i have removed the `<?hh` tag.

We try to follow usage and not specifications in Linguist. If one language says some syntax is illegal but a large number of users still rely on that syntax, we'll try to support it. So I'm not sure we should remove the tags here. The best thing would be to preserve the sample files in their original form.

fredemmott:
> We try to follow usage and not specifications in Linguist. If one language says some syntax is illegal but a large number of users still rely on that syntax, we'll try to support it. So I'm not sure we should remove the tags here. The best thing would be to preserve the sample files in their original form.

There's definitely not a large number of users; Hack uses a of variant PHP-style autoloading, and no release of the autoloader supports the `.hack` extension *at all* - and it's only been in master [since Jan 15th](https://github.com/hhvm/hhvm-autoload/commits/master). Any usage of `<?hh` is extremely early adopters running a locally patched version or master from the last two weeks - as far as I'm aware, it's just @azjezz.

fredemmott:
Though, `<?hh` should remain supported in `.php`, `.hh`, `.hhi` files etc.

azjezz:
@fredemmott the `.hack` files in @nuxed were generated using `hh-to-hack` so they don't use `<?hh` already

jeffomatic:
> @jeffomatic Any idea what we should name it? Looks like the official name is Hack, so maybe "Nand2Tetris Hack"?

I think that's fine. It is pretty unlikely that anyone would intentionally configure a language context for these files since they are far closer to binaries than they are human-consumable text.

Caveat that I don't speak for the Nand2Tetris project or community in any real way other than having gone through the course as a hobbyist.

pchaigno:
Thanks @jeffomatic!

@azjezz - Could you add the Nand2Tetris Hack language to Linguist in this pull request as well, with a [heuristic](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml) to disambiguate between the two `.hack` language? I think we can probably add Nand2Tetris as a data language.

Something along the lines of:
```yaml
- extensions: ['.hack']
  rules:
  - language: Nand2Tetris Hack
    pattern: ^[01]+$
  - language: Hack
```
should be enough.

azjezz:
@pchaigno done :+1:

lib/linguist/languages.yml
```diff
@@ -1904,7 +1904,9 @@ Hack:
   codemirror_mode: php
   codemirror_mime_type: application/x-httpd-php
   extensions:
+  - ".hack"
```
pchaigno: Is `.hack` considered the primary extension for Hack now?

azjezz: yes

lib/linguist/languages.yml
```diff
@@ -2986,6 +2988,12 @@ Myghty:
   tm_scope: none
   ace_mode: text
   language_id: 239
+Nand2Tetris Hack:
+  type: data
+  ace_mode: text
```
pchaigno: You'll need to add `tm_scope: none` to tell Linguist we don't have a syntax highlighting grammar for that language.

lib/linguist/languages.yml
```diff
@@ -2986,6 +2988,12 @@ Myghty:
   tm_scope: none
   ace_mode: text
   language_id: 239
+Nand2Tetris Hack:
```
pchaigno: Languages are sorted with uppercase letters first.

test/test_language.rb
```diff
@@ -343,6 +345,8 @@ def test_extensions
     assert Language['LSL'].extensions.include?('.lsl')
     assert Language['Perl'].extensions.include?('.pl')
     assert Language['Python'].extensions.include?('.py')
+    assert Language['Hack'].extensions.include?('.hack')
+    assert Language['Hack'].extensions.include?('.hhi')
```
pchaigno: You can remove these tests. We already have tests in place to check this. I'm not even sure what these hardcoded tests are used for.

azjezz: :+1:

lib/linguist/languages.yml
azjezz: :+1:

lib/linguist/languages.yml
azjezz: :+1:

samples/Hack/EmptyPipelineException.hack
```diff
@@ -1,7 +1,7 @@
-namespace Waffle\Http\Server\Exception;
+namespace Nuxed\Http\Server\Exception;
```
pchaigno: Why the changes for the two Hack sample files?

test/test_language.rb
```diff
@@ -177,6 +177,8 @@ def test_find_by_extension
     assert_equal [Language['Ruby']], Language.find_by_extension('foo/bar.rb')
     assert_equal [Language['Ruby']], Language.find_by_extension('PKGBUILD.rb')
     assert_equal ['C', 'C++', 'Objective-C'], Language.find_by_extension('foo.h').map(&:name).sort
+    assert_equal [Language['Hack']], Language.find_by_extension('HomeController.hack')
+    assert_equal [Language['Hack']], Language.find_by_extension('string_functions.hhi')
```
pchaigno: Same thing for these tests.

samples/Hack/EmptyPipelineException.hack
azjezz: the waffle project is deprecated and nuxed uses more hack features, might help in the future when updating the syntax highlight for hack :stuck_out_tongue: ( PHP Syntax highlight is not working well - but its not a priority now )

test/test_language.rb
azjezz: :+1:

samples/Hack/EmptyPipelineException.hack
pchaigno: I'd rather we did not change the existing sample files. They are not meant to be execute, but simply to represent in-the-wild usage of Hack.

pchaigno:
My guess would be that the tests are failing because the tokenizer is unable to extract tokens from the Nand2Tetris hack sample file. If that's the case, we'll probably need to adapt the tokenizer or add an exception mechanism. I have found time to debug this yet.

Lexidor:
@pchaigno if the Nand2Tetris files are only ones and zeroes (and unlimited whitespace), then there are no real tokens. This _language_ is not read as source code. Would it be okay if I did something dirty and just specified the token definition as "Each byte is the start of a new token"? Would this somehow cause something unrelated to break? I would like to have syntax highlighting in hack files, so sorry for not prioritising the Nand2Tetris hack format more.

Lexidor:
@pchaigno Are we following the route of not requiring tokenizers. Tokenizing such files is rather pointless, since there are no tokens. Hack code is being used in dothack files in many more places than it was at the release of HHVM4. This leaves entire repositories as black text on GitHub.

azjezz:
cc @pchaigno, any updates here ? after #4544 , the `.hack` files syntax is "highlightable" when using a gitattribute override to classify `.hack` as `Hack`.

example : [`.gitattribute`](https://github.com/nuxed/framework/blob/rx/.gitattributes) / [`Translator.hack`](https://github.com/nuxed/framework/blob/rx/src/Nuxed/Translation/Translator.hack)

lildude:
Wow! How did I miss this one for so long. Sorry folks. 

Support for the `.hack` extension has now been added in https://github.com/github/linguist/pull/4588 and we completely forgot about the Nand2Tetris files. I don't think anyone has noticed... or really cares about those files being identified as Hack (famous last words 😆) so lets update this branch and change it to only add support for the `.hhi` extension.

azjezz:
@lildude i myself forgot about this PR :) 

i have update this PR to only address `.hhi` extension.

PR #4357, opened 2018-12-17
Don't treat go.mod as a "generated" file

borgstrom:
While it's true that the `go` tool does modify `go.mod` I would like to propose that `go.mod` be excluded from being treated as a "generated" file since understanding changes to it are a crucial part of the review process.

For example, here's the default files view from a recent PR to a project that groomed some dependencies:

screen shot 2018-12-17 at 8 12 16 am

A reviewer must "load diff" on `go.mod`.  And for PRs that have lots of changes across many files it is easy to gloss over the two collapsed files without actually expanding the contents of `go.mod`.

From a precedence standpoint `dep` (`Gopkg.toml` & `Gopkg.lock`) has similar properties in that `dep` modifies `Gopkg.toml`, but only `Gopkg.lock` is considered "generated".

Thanks!


borgstrom:
I totally missed #4340... Closing in favor of that.

PR #4363, opened 2018-12-27
fix unnecessary CRLF and indentation in Pawn examples

Southclaws:
Just a quick QoL thing for the Pawn examples. Nothing crazy so I felt the checkboxes didn't really apply!


pchaigno:
Please do not change the sample files. These represent usage in-the-wild and should be preserved with any error or imperfection. They are use to train the Bayesian classifier and test the overall classification.

PR #4376, opened 2019-01-10
Disambiguate Slice and JSON languages for .ice

bzz:
This a fix for Slice vs JSON language disambiguation heuristic. It was preliminary discussed at https://github.com/github/linguist/pull/4243#discussion_r246500293

Before:
 - all `.ice` files detected as Slice language.
  But as noted in https://github.com/github/linguist/pull/4243#issuecomment-414971169, many are [just JSON](https://github.com/Obijuan/Entregas-Tutorial-Electronica-Digital-FPGAs/tree/master/Bloques/Notas%20musicales) as `*.ice` is used by [IceStudio](https://github.com/FPGAwars/icestudio)

After:
 - some .ice are Slice
 - others, IceStudio ones, are JSON

## Description
Fix includes multiple moving parts:
 - A fix Regexp syntax in `heuristics.yml` disambiguating "Slice" and "JSON".
 - Add `.ice` extension to JSON in `languages.yml`.
   Without this, the above heuristics will never kick in, as Extension strategy will always have only 1 match.
 - That requires to have at least one `.ice` sample in JSON, so a 8kb [block-sync-counter8.ice](https://github.com/scanlime/icebreaker-icestudio-ledmatrix/blob/27815aa2406eedaff34a86cfee3d9ea1cade30a4/block-sync-counter8.ice#L1) was added, [The Unlicense](https://github.com/scanlime/icebreaker-icestudio-ledmatrix/blob/27815aa2406eedaff34a86cfee3d9ea1cade30a4/LICENSE) (took quite some time to find a non-GPL IceStudio example).
 - That requires to have more samples of actual Slice, as otherwise Classifier would not be able to distinguish samples, so a 40kb [Murmur.ice](https://github.com/mumble-voip/mumble/blob/e31d267a11b4ed0597ad41309a7f6b715837141f/src/murmur/Murmur.ice), [BSD 3-Clause](https://github.com/mumble-voip/mumble/blob/e31d267a11b4ed0597ad41309a7f6b715837141f/LICENSE) was added.

**Why this was not noticed in original #4243?**
Moved to #4391 

## Checklist:

- [x] **I am fixing a misclassified language** Slice VS JSON for `.ico`
    - Sample 
      - 40 kb Slice [Murmur.ice](https://github.com/mumble-voip/mumble/blob/e31d267a11b4ed0597ad41309a7f6b715837141f/src/murmur/Murmur.ice), [BSD 3-Clause](https://github.com/paulkoerbitz/language-slice/blob/113b2e2c0f4b8413f72bc9567976f87ae797460e/LICENSE)
      - 8kb JSON [block-sync-counter8.ice](https://github.com/scanlime/icebreaker-icestudio-ledmatrix/blob/27815aa2406eedaff34a86cfee3d9ea1cade30a4/block-sync-counter8.ice#L1), [The Unlicense](https://github.com/scanlime/icebreaker-icestudio-ledmatrix/blob/27815aa2406eedaff34a86cfee3d9ea1cade30a4/LICENSE)

### DIFF
--- a/samples/Slice/Murmur.ice
+++ b/samples/Slice/Murmur.ice
@@ -1,0 +1,885 @@
+// Copyright 2005-2018 The Mumble Developers. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file at the root of the
+// Mumble source tree or at <https://www.mumble.info/LICENSE>.
+
+/**
+ *
+ * Information and control of the murmur server. Each server has
+ * one {@link Meta} interface that controls global information, and
+ * each virtual server has a {@link Server} interface.
+ *
+ **/
+
+#include <Ice/SliceChecksumDict.ice>
+
+module Murmur
+{
+
+	/** A network address in IPv6 format.
+	 **/
+	["python:seq:tuple"] sequence<byte> NetAddress;
+
+	/** A connected user.
+	 **/
+	struct User {
+		/** Session ID. This identifies the connection to the server. */
+		int session;
+		/** User ID. -1 if the user is anonymous. */
+		int userid;
+		/** Is user muted by the server? */
+		bool mute;
+		/** Is user deafened by the server? If true, this implies mute. */
+		bool deaf;
+		/** Is the user suppressed by the server? This means the user is not muted, but does not have speech privileges in the current channel. */
+		bool suppress;
+		/** Is the user a priority speaker? */
+		bool prioritySpeaker;
+		/** Is the user self-muted? */
+		bool selfMute;
+		/** Is the user self-deafened? If true, this implies mute. */
+		bool selfDeaf;
+		/** Is the User recording? (This flag is read-only and cannot be changed using setState().) **/
+		bool recording;
+		/** Channel ID the user is in. Matches {@link Channel.id}. */
+		int channel;
+		/** The name of the user. */
+		string name;
+		/** Seconds user has been online. */
+		int onlinesecs;
+		/** Average transmission rate in bytes per second over the last few seconds. */
+		int bytespersec;
+		/** Client version. Major version in upper 16 bits, followed by 8 bits of minor version and 8 bits of patchlevel. Version 1.2.3 = 0x010203. */
+		int version;
+		/** Client release. For official releases, this equals the version. For snapshots and git compiles, this will be something else. */
+		string release;
+		/** Client OS. */
+		string os;
+		/** Client OS Version. */
+		string osversion;
+		/** Plugin Identity. This will be the user's unique ID inside the current game. */
+		string identity;
+		/**
+		   Base64-encoded Plugin context. This is a binary blob identifying the game and team the user is on.
+
+		   The used Base64 alphabet is the one specified in RFC 2045.
+
+		   Before Mumble 1.3.0, this string was not Base64-encoded. This could cause problems for some Ice
+		   implementations, such as the .NET implementation.
+
+		   If you need the exact string that is used by Mumble, you can get it by Base64-decoding this string.
+
+		   If you simply need to detect whether two users are in the same game world, string comparisons will
+		   continue to work as before.
+		 */
+		string context;
+		/** User comment. Shown as tooltip for this user. */
+		string comment;
+		/** Client address. */
+		NetAddress address;
+		/** TCP only. True until UDP connectivity is established. */
+		bool tcponly;
+		/** Idle time. This is how many seconds it is since the user last spoke. Other activity is not counted. */
+		int idlesecs;
+		/** UDP Ping Average. This is the average ping for the user via UDP over the duration of the connection. */
+		float udpPing;
+		/** TCP Ping Average. This is the average ping for the user via TCP over the duration of the connection. */
+		float tcpPing;
+	};
+
+	sequence<int> IntList;
+
+	/** A text message between users.
+	 **/
+	struct TextMessage {
+		/** Sessions (connected users) who were sent this message. */
+		IntList sessions;
+		/** Channels who were sent this message. */
+		IntList channels;
+		/** Trees of channels who were sent this message. */
+		IntList trees;
+		/** The contents of the message. */
+		string text;
+	};
+
+	/** A channel.
+	 **/
+	struct Channel {
+		/** Channel ID. This is unique per channel, and the root channel is always id 0. */
+		int id;
+		/** Name of the channel. There can not be two channels with the same parent that has the same name. */
+		string name;
+		/** ID of parent channel, or -1 if this is the root channel. */
+		int parent;
+		/** List of id of linked channels. */
+		IntList links;
+		/** Description of channel. Shown as tooltip for this channel. */
+		string description;
+		/** Channel is temporary, and will be removed when the last user leaves it. */
+		bool temporary;
+		/** Position of the channel which is used in Client for sorting. */
+		int position;
+	};
+
+	/** A group. Groups are defined per channel, and can inherit members from parent channels.
+	 **/
+	struct Group {
+		/** Group name */
+		string name;
+		/** Is this group inherited from a parent channel? Read-only. */
+		bool inherited;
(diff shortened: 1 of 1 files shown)


Alhadis:
Ouch. That JSON sample is *huge*. Any chance it could be shortened?

While we appreciate large, detailed samples, stuff like this is normally overkill for Linguist, and adds undue weight/processing time to the classifier.

Other than that, this LGTM. :+1:

bzz:
@Alhadis I have updated the comment above.

The smallest JSON sample could find that passes the tests is this (3104 sloc) 93.9 KB  [sample](https://github.com/scanlime/icebreaker-icestudio-ledmatrix/blob/27815aa2406eedaff34a86cfee3d9ea1cade30a4/block-ledmatrix-generic.ice)

Please advice on how to proceed.

lib/linguist/heuristics.yml
```diff
@@ -158,7 +158,7 @@ disambiguations:
   - language: Slice
     pattern: '^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)'
```
Alhadis: Would it be invalid syntax for a Slice file to start with `{`?

If so, we can replace this heuristic with:

~~~diff
   - language: Slice
-    pattern: '^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)'
+    pattern: '\A(?!\s*[{\[])'
~~~

I suspect the reason why the tests are failing is because Slice files which *don't* contain lines like `#include…` are being left to the classifier.

externl: A "proper" Slice file should never start with a `{`. I think it might be possible to contrive an example, but no one should write it like that.

Alhadis: That should be enough then. 👍

bzz: Changed in b213793

lib/linguist/heuristics.yml
```diff
@@ -156,9 +156,9 @@ disambiguations:
 - extensions: ['.ice']
   rules:
   - language: Slice
-    pattern: '^\s*(#\s*(include|if[n]def|pragma)|module\s+[A-Za-z][_A-Za-z0-9]*)'
+    pattern: '\A(?!\s*[{\[])'
```
pchaigno: Isn't this the exact opposite of the JSON regex? Do we need both?

bzz:
Have replaced 120kb JSON sample for .ice with 8kb one in 361ce64 - Classifier test on [CI fails](https://travis-ci.org/github/linguist/jobs/478857700#L1589) the same.

Added the smallest sample I could find to be passing Classifier tests, 96kb [block-ledmatrix-generic.ice](https://github.com/scanlime/icebreaker-icestudio-ledmatrix/blob/27815aa2406eedaff34a86cfee3d9ea1cade30a4/block-ledmatrix-generic.ice) (from same repo under the same license) in d8de0cb

lib/linguist/heuristics.yml
Alhadis: We *should* be able to get by with only `'\A(?!\s*[{\[])'` for Slice. That, or remove the pattern altogether and use an open-ended fallback to Slice with JSON checked first. Which is probably clearer in the long-run. I'm still puzzled as to why the tests are failing. 😕

Also, I can't even see the diff I'm reviewing. 😀 Scrollbar isn't going away:
a

Alhadis:
@pchaigno Think we should add a test to guard against heuristics with patterns that look like RegExp literals? Stuff like this is easy to overlook, since we're used to seeing RegExp patterns as Ruby/Perl/JavaScript literals:

~~~yaml
- '/foo/'
- '/bar/i'
~~~

Patterns which actually need to match stuff enclosed by forward slashes can always use character classes:

~~~yaml
- '[/]foo/'
- '/foo/[i]'
~~~

The fact I completely overlooked this mistake in review suggests it might happen again, so it wouldn't hurt to add a test for it.

(Have I mentioned that YAML is the worst format in the world, and that Ruby is more suited to being a configuration language than anything programming-related? 😁 `</snark>`).

pchaigno:
> Think we should add a test to guard against heuristics with patterns that look like RegExp literals

Shouldn't this be covered by the tests for the new heuristic rules though? Maybe we should check that there are no empty fixture in `test_heuristics.rb` instead?

> I belive that happened because despite the [advised test](https://github.com/github/linguist/pull/4243#issuecomment-423562889) that supposed to be checking disambiguation by heuristics was added, it was failing silently due to empty fixture in [test_heuristics.rb#L194](https://github.com/github/linguist/blob/4d8d65eb5d342f6df5f7dc65b9f7bb4fea31e434/test/test_heuristics.rb#L194). It was empty because there were no samples of `.ice` in JSON.

bzz:
> Could you open a new issue to discuss the issue you found with the tests? (Copy and paste is fine.)

@pchaigno done in https://github.com/github/linguist/issues/4391

Alhadis:
[What the hell](https://travis-ci.org/github/linguist/jobs/483941794#L1591-L1596):

~~~
  1) Failure:
TestClassifier#test_classify_ambiguous_languages [/home/travis/build/github/linguist/test/test_classifier.rb:56]:
/home/travis/build/github/linguist/samples/JSON/block-sync-counter8.ice
[["Slice", -572.4646846236486], ["JSON", -604.6475163156096]].
Expected: "JSON"
  Actual: "Slice"
~~~

That's it. I'm cloning your branch locally and investigating this myself. This isn't right.

lib/linguist/heuristics.yml
bzz: > remove the pattern altogether and use an open-ended fallback to Slice with JSON checked first

addressed in 9c2302a

> I'm still puzzled as to why the tests are failing. 

I've put back smallest JSON sample for .ice in bd3b268 and one can see that it's not heuristics but only a Classifier test that is failing `TestClassifier#test_classify_ambiguous_languages`.

This [Classifier test](https://github.com/github/linguist/blob/58193319b3a1daa8fae0fec897029137a958568e/test/test_classifier.rb#L46) seems to be designed to look for all the languages that are ambiguous by extension (JSON and Slice in `.ice` case) and then assume `Classifier.classify` would be able to distinguish those without using heuristics strategy that we are adjusting in here.

This seems to be not possible, only based on samples that we have for this pair of languages 

8kb for Slice (that also uses `{} and []`) VS 216kb total for JSON:
```
      JSON =   -600.095 +  -4.551 =   -604.646
     Slice =   -565.507 +  -6.949 =   -572.456
```

instead of adding bigger JSON sample, I tried adding +16kb of Slice one and it seems to _almost_ work

24kb Slice VS  216kb for JSON:
```
      JSON =   -600.095 +  -4.552 =   -604.647
     Slice =   -596.006 +  -6.544 =   -602.551
```

bzz: +12kb of Slice samples (total +32kb VS +96kb in previous approach) seems to do the trick

```
      JSON =   -600.095 +  -4.552 =   -604.647
     Slice =   -624.850 +  -6.257 =   -631.108
```

🎉

Alhadis: Good work! Wait, so, the heuristics strategy *doesn't* always run??

pchaigno:
> That's it. I'm cloning your branch locally and investigating this myself. This isn't right.

I suspect a bug in the Bayesian classifier... Good luck! :stuck_out_tongue_winking_eye:

lib/linguist/heuristics.yml
bzz: It does not in `TestClassifier`, as this test directly uses only a low-level Baesian classifier API `Classifier.classify()`

I belive it may be the assumptions of this test that is too strong - for some reason it _assumes_ that all colliding extensions can be disambiguated by Baesian classifier directly. 
When in Linguist \w multiple strategies this classifier should be able to disambiguate only the cases that are _not_ covered by heuristics already.

PR #4378, opened 2019-01-12
Heuristic to disambiguate GAP and GDScript

anantn:
GDScript files (.gd) are often misclassified as GAP code on Github.

## Description
This rule updates classification of .gd files as either GAP or GDScript. Currently, several GDScript files are misclassified as GAP on Github (example search: https://github.com/search?q=language%3AGAP+game&type=Repositories)

## Checklist:
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -137,6 +137,7 @@ disambiguations:
   - language: GAP
     pattern: '\s*Declare'
   - language: GDScript
+    pattern: '\s*(extends|var|const|enum|func|class|signal|tool|yield|assert|onready)'
 - extensions: ['.gml']
   rules:
   - language: XML
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -132,6 +132,11 @@ disambiguations:
     pattern: '^\s*(#version|precision|uniform|varying|vec[234])'
   - language: Filterscript
     pattern: '#include|#pragma\s+(rs|version)|__attribute__'
+- extensions: ['.gd']
+  rules:
+  - language: GAP
+    pattern: 'gap> '
+  - language: GDScript
 - extensions: ['.gml']
   rules:
   - language: XML
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -135,7 +135,7 @@ disambiguations:
 - extensions: ['.gd']
   rules:
   - language: GAP
-    pattern: 'gap> '
+    pattern: '\s*Declare'
   - language: GDScript
 - extensions: ['.gml']
   rules:
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -135,7 +135,7 @@ disambiguations:
 - extensions: ['.gd']
   rules:
   - language: GAP
-    pattern: '\s*Declare'
+    pattern: '\s*(Declare|BindGlobal|KeyDependentOperation)'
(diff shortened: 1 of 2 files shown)


lib/linguist/heuristics.yml
```diff
@@ -132,6 +132,11 @@ disambiguations:
     pattern: '^\s*(#version|precision|uniform|varying|vec[234])'
   - language: Filterscript
     pattern: '#include|#pragma\s+(rs|version)|__attribute__'
+- extensions: ['.gd']
+  rules:
+  - language: GAP
+    pattern: 'gap> '
```
pchaigno: Should all GAP files have this pattern? Our GAP `.gd` sample file doesn't, which is why the Travis build tests are failing.

anantn: My mistake, the 'gap> ' heuristic is for GAP files with '.tst' extension. Updated to include the correct heuristic for GAP declaration files (.gd extension).

pchaigno: Okay. Though my question remains: should all `.gd` GAP files have this `\s*Declare` pattern? Is a `.gd` GAP file without that pattern invalid?

pchaigno: So without that statement the program is invalid?

anantn: That's not technically true, I think it is possible to have a .gd file with no Declare identifiers but it wouldn't do anything interesting (like a source file with only comments). I'm no GAP expert (my interest is actually in GDScript) but this is what I could surmise through code search for .gd files on Github.

Also worth noting: the current state of affairs are flipped, there are ~3x more repositories using [GDScript](https://github.com/search?q=language%3AGDScript) than [GAP](https://github.com/search?q=language%3AGAP), so .gd files misclassifying as GDScript is the more prevalent problem than the other way around.

It is possible to make this heuristic more robust by flipping the pattern and enumerating all valid GDScript identifiers first, but that would be a more inefficient regex. Let me know if you'd like me to explore that option.

pchaigno:
@anantn Could you add a test for these changes in [`test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb)?

lib/linguist/heuristics.yml
```diff
@@ -132,6 +132,12 @@ disambiguations:
     pattern: '^\s*(#version|precision|uniform|varying|vec[234])'
   - language: Filterscript
     pattern: '#include|#pragma\s+(rs|version)|__attribute__'
+- extensions: ['.gd']
+  rules:
+  - language: GAP
+    pattern: '\s*Declare'
```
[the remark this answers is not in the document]
pchaigno: Thanks for the review @fingolfin!

> I have no idea whether this "heuristic" here is just a hint to break a tie

If a file matches one of the regexp for its file extension, the associated language is returned. If it doesn't match any regexp for its file extension, then [we continue with the next and last strategy](https://github.com/github/linguist/blob/58193319b3a1daa8fae0fec897029137a958568e/lib/linguist.rb#L61-L69), the Bayesian classifier.

@anantn - If you agree with @fingolfin's suggestion, please merge it.

anantn:
If the heuristic doesn't match I think we should still classify the script as GAP since we would try other strategies as @pchaigno mentioned. But no harm making the heuristic more robust!

Thanks to @fingolfin and @pchaigno for your help with this patch 👍

anantn:
@pchaigno Could we land this now? 😄

PR #4383, opened 2019-01-15
Lock Linguist to Bundler 1.x

lildude:
As seen in https://github.com/github/linguist/pull/4340, tests are starting to fail because Travis has decided to switch to Bundler 2.x by default as of 3 Jan 2019 - https://docs.travis-ci.com/user/languages/ruby/#bundler-20 - and this is clashing with Licensed which expressly requires `~>1.10` due to it using some internal bundler functionality. 

This PR locks us to bundler `~>1.10` too.

It took a few attempts for me to get Travis to actually do what I want it to do as the documented method wasn't working 😞 but I got there in the end.

_Template removed as it's not relevant_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - libicu52
 
 before_install:
-  - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
+  - gem uninstall -v '>= 2' -ax bundler || true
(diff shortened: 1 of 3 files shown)


lildude:
> Do we know when Licensed will move to Bundler 2.x? Or even if they plan to?

Not sure when, but my colleague has it on his list of things to test and see if anything breaks. Once it updates to 2.x, we can revert this PR.

PR #4397, opened 2019-01-31
Language recognition with file path pattern (regex)

Varun-garg:
Fixes vscode config json files on github. Github shows errors for comments in vscode files. This can be fixed by classifying them as jsonc. broken vscode config files are included.

## Description
Added a feature for language recognition with file path pattern (regex). example pattern: `'\.vscode\/.*\.json$'`

## Checklist:
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist.rb
+++ b/lib/linguist.rb
@@ -61,6 +61,7 @@ class << Linguist
   STRATEGIES = [
     Linguist::Strategy::Modeline,
     Linguist::Strategy::Filename,
+    Linguist::Strategy::FilePathPattern,
     Linguist::Shebang,
     Linguist::Strategy::Extension,
     Linguist::Strategy::XML,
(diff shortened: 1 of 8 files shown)


Varun-garg:
Heu @lildude , thanks for follow up, won't extension take precedence before heuristics?

lildude:
> won't extension take precedence before heuristics?

Yup, that's why we have heuristics 😄 - they're used to differentiate one or more uses of the same extension. So in this case, the `.json` extension would need to be added to "JSON with Comments" (it's already associated with JSON) and then a heuristic is needed to differentiate the two.

Varun-garg:
Ah, regex to detect comments in json is hard to implement for me.

Varun-garg:
@lildude Sorry for replying late. On second thoughts I was wondering if we are going to scan if the json file has comments and switch to jsonc, we are basically allowing all json files to have comments (unlike hardcoded files / patterns). Can't we just enable comments in json and ditch jsonc?

Issue #4388, opened 2019-01-21
Gnuplot files with .p extension classified as OpenEdge ABL


smola:
## Problem Description
Some repositories use the `.p` extension for Gnuplot. Linguist classifies these files as OpenEdge ABL.

### URL of the affected repository:
Test search: https://github.com/search?p=90&q=extension%3Ap+plot+set+output&type=Code

### Last modified on:
2019-01-21

### Expected language:
Gnuplot

### Detected language:
OpenEdge ABL

lildude:
I think this is because the `.p` extension is only associated with the OpenEdge ABL language though I'm not sure why there are a few files that are successfully identified as GNUplot - I've not dug into it yet.

Adding the `.p` extension to GNUplot with a heuristic to differentiate it from OpenEdge ABL should "fix" this.

smola:
@lildude Could you reopen the issue / remove the `Stale` label? I filed an updated PR at https://github.com/github/linguist/issues/4388 Thanks!

PR #4402, opened 2019-02-04
improve GLSL classification

smola:
improve GLSL classification

## Description
* added extensions .gs, .comp and .glslf.
  See: https://stackoverflow.com/q/6432838

* add disambiguation rule for .gs files:
  if it starts with `#version <number>`, it's GLSL.
  This seems to work for most GLSL files with few exceptions
  that are catched by the classifier.

* added sample extrude_normal.gs from kyle-piddington/ShaderTool
  (licensed under MIT):
  https://github.com/kyle-piddington/ShaderTool/blob/c753a53bde6eab942da617adab9483c945f27f51/assets/shaders/extrude_normals.gs

* added sample blend_120.glslf from gfx-rs/gfx (licensed under
  Apache 2.0):
  https://github.com/gfx-rs/gfx/blob/7b084cfbb95e3e4d1ea51e91746fc75b2efb3be3/examples/blend/shader/blend_120.glslf

* test_heuristics.rb now supports alt_name to ake a different file name
for a fixture.

## Checklist:
- [x] **I am associating a language with a new file extension.**
  - [x] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      - https://github.com/search?l=JavaScript&q=extension%3Ags+version+void+layout&type=Code
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://github.com/kyle-piddington/ShaderTool/blob/c753a53bde6eab942da617adab9483c945f27f51/assets/shaders/extrude_normals.gs
      - https://github.com/gfx-rs/gfx/blob/7b084cfbb95e3e4d1ea51e91746fc75b2efb3be3/examples/blend/shader/blend_120.glslf
    - Sample license(s): MIT, Apache 2.0
  - [x] I have included a change to the heuristics to distinguish my language from others using the same extension.

PR #4403, opened 2019-02-04
add .p extension to Gnuplot

smola:
add .p extension to Gnuplot

Fixes https://github.com/github/linguist/issues/4388

## Description
- add .p extension to Gnuplot
- add disambiguation rule for Gnuplot and OpenEdge ABL

## Checklist:
      - https://github.com/search?p=90&q=extension%3Ap+plot+set+output&type=Code
    - No. Simulated with fake extension.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -284,6 +284,7 @@ disambiguations:
     - '^s?plot\b'
     - '^set\s+(term|terminal|out|output|[xy]tics|[xy]label|[xy]range|style)\b'
   - language: OpenEdge ABL
+    pattern: '^\s*(?i:(DEFINE|FOR|PROCEDURE|MESSAGE|RUN|ASSIGN|REPEAT|CLASS|METHOD|&SCOPED-DEFINE|&MESSAGE|&GLOBAL-DEFINE))'
 - extensions: ['.php']
   rules:
   - language: Hack
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -277,6 +277,13 @@ disambiguations:
   - language: NL
     pattern: '^(b|g)[0-9]+ '
   - language: NewLisp
+- extensions: ['.p']
+  rules:
+  - language: Gnuplot
+    pattern:
+    - '^s?plot\b'
+    - '^set\s+(term|terminal|out|output|[xy]tics|[xy]label|[xy]range|style)\b'
+  - language: OpenEdge ABL
 - extensions: ['.php']
   rules:
   - language: Hack
--- a/test/test_heuristics.rb
+++ b/test/test_heuristics.rb
@@ -7,9 +7,13 @@ class TestHeuristics < Minitest::Test
     File.read(File.join(samples_path, name))
   end
 
-  def file_blob(name)
+  def file_blob(name, alt_name=nil)
     path = File.exist?(name) ? name : File.join(samples_path, name)
-    FileBlob.new(path)
+    blob = FileBlob.new(path)
+    if !alt_name.nil?
+      blob.instance_variable_set("@path", alt_name)
+    end
+    blob
   end
 
   def all_fixtures(language_name, file="*")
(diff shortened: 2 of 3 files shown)


smola:
Added here the same `alt_name` functionality in `test_heuristics.rb` as in https://github.com/github/linguist/pull/4402

I'll update any of these PRs if the other is merged.

lib/linguist/heuristics.yml
```diff
@@ -277,6 +277,11 @@ disambiguations:
   - language: NL
     pattern: '^(b|g)[0-9]+ '
   - language: NewLisp
+- extensions: ['.p']
+  rules:
+  - language: Gnuplot
+    pattern: '^s?plot\b'
```
pchaigno: Could you provide a bit of context on this heuristic rule? Does it mean that all Gnuplot files should contain `splot` or `plot` and if they don't, they're invalid?

test/test_heuristics.rb
```diff
@@ -317,6 +321,13 @@ def test_nl_by_heuristics
     })
   end
 
+  def test_p_by_heuristics
+    assert_heuristics({
+      "Gnuplot" => all_fixtures("Gnuplot", "*"),
+      "OpenEdge ABL" => all_fixtures("OpenEdge ABL", "*")
+    }, alt_name="test.p")
```
pchaigno: As I understand, the `alt_name` is meant to ensure all files actually go through the heuristic for the `.p` extension, regardless of their extension. Could you add a comment in your code here, just to clarify why we're doing this?

lib/linguist/heuristics.yml
smola: Most of them have, at least, one plot command (`plot` for 2D or `splot` for 3D). I think are still valid if there is no plot command, but useless. I found no example for that: https://github.com/search?q=language%3AGnuplot+set+-plot+-splot&type=Code

My guess is that a Gnuplot file with `.p` extension (rare) and no plot command (very rare) is very unlikely, so it's better than current situation, where all Gnuplot with `.p` extensions are misidentified.

test/test_heuristics.rb
smola: Done. In the `assert_heuristics` function.

lib/linguist/heuristics.yml
pchaigno: I think there's an issue with your search query: https://github.com/search?q=language%3AGnuplot+set+NOT+plot+NOT+splot&type=Code.

smola: Ouch, yes. I will do another iteration to refine the rule.

smola:
Moved the `alt_name` functionality to https://github.com/github/linguist/pull/4410

lib/linguist/heuristics.yml
```diff
@@ -277,6 +277,13 @@ disambiguations:
   - language: NL
     pattern: '^(b|g)[0-9]+ '
   - language: NewLisp
+- extensions: ['.p']
+  rules:
+  - language: Gnuplot
+    pattern:
+    - '^s?plot\b'
+    - '^set\s+(term|terminal|out|output|[xy]tics|[xy]label|[xy]range|style)\b'
+  - language: OpenEdge ABL
```
pchaigno: It would be better if we could find a pattern for OpenEdge ABL as well.

/cc @abevoelker, who added support for OpenEdge ABL in Linguist (#115), in case they have an idea.

abevoelker: It's been quite a few years but I think a pattern like this should catch most Progress / OpenEdge ABL programs:

```suggestion
  - language: OpenEdge ABL
    pattern: '^\s*(?i:(DEFINE|FOR|PROCEDURE|MESSAGE|RUN|ASSIGN|REPEAT|CLASS|METHOD|&SCOPED-DEFINE|&MESSAGE|&GLOBAL-DEFINE))'
```

siawash96:
HI I WANT THE ENABLE THE GITHUB IN MY WINDOWS 10
PLEASE SET THE GITHUB AND HELP ME IN THE GO ON YHIS JAVA & HTML MY GIT HUB
THANKS ALOT

PR #4412, opened 2019-02-09
detect Vimball (.vba, .vmb) files as Vim script

### LINKED ISSUE #501: .vba is not always Visual Basic
If you look at the percentage of language of the project chrisbra/Recover.vim...

https://github.com/chrisbra/Recover.vim

...you will see that it is reported as about 47% "Visual Basic".  I presume this is because the project has files with the .vba extension.  FYI, that extension here is being used to mean VimBall Archiver, which has nothing whatsoever to do with Visual Basic.

smola:
Fixes #501

## Description

- added .vba and .vmb extensions to Vim script
- added .vba disambiguation
- added Vimball sample with .vba extension (Vim license) from https://github.com/nelstrom/vim-textobj-rubyblock/blob/master/textobj-rubyblock.vba
- added Vimball sample with .vmb extension (WTFPL) from https://github.com/CDSoft/todo/blob/master/todo.vmb
- added Visual Basic sample with .vba extension (MIT license) from https://github.com/Akaizoku/concatenate/blob/master/procedures.vba

## Checklist:
      - https://github.com/search?q=extension%3Avba+usevimball+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Avmb+usevimball+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/nelstrom/vim-textobj-rubyblock/blob/master/textobj-rubyblock.vba
    - Sample license(s): Vim License

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -426,6 +426,11 @@ disambiguations:
     pattern: '^\s*(import.+(from\s+|require\()[''"]react|\/\/\/\s*<reference\s)'
   - language: XML
     pattern: '(?i:^\s*<\?xml\s+version)'
+- extensions: ['.vba']
+  rules:
+  - language: Vim script
+    pattern: '^UseVimball'
+  - language: Visual Basic
 - extensions: ['.w']
   rules:
   - language: OpenEdge ABL
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5170,6 +5170,8 @@ Vim script:
   - nvim
   extensions:
   - ".vim"
+  - ".vba"
+  - ".vmb"
   filenames:
   - ".gvimrc"
   - ".nvimrc"
--- a/samples/Vim script/textobj-rubyblock.vba
+++ b/samples/Vim script/textobj-rubyblock.vba
@@ -1,0 +1,216 @@
+" Vimball Archiver by Charles E. Campbell, Jr., Ph.D.
+UseVimball
+finish
+plugin/textobj/rubyblock.vim	[[[1
+58
+if exists('g:loaded_textobj_rubyblock')  "{{{1
+  finish
+endif
+
+" Interface  "{{{1
+call textobj#user#plugin('rubyblock', {
+\      '-': {
+\        '*sfile*': expand('<sfile>:p'),
+\        'select-a': 'ar',  '*select-a-function*': 's:select_a',
+\        'select-i': 'ir',  '*select-i-function*': 's:select_i'
+\      }
+\    })
+
+" Misc.  "{{{1
+let s:comment_escape = '\v^[^#]*'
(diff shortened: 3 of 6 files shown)


samples/Vim script/textobj-rubyblock.vba
```diff
@@ -0,0 +1,216 @@
+" Vimball Archiver by Charles E. Campbell, Jr., Ph.D.
```
pchaigno: Could you also add an `.vmb` sample file?

lib/linguist/languages.yml
```diff
@@ -5169,7 +5169,9 @@ Vim script:
   - viml
   - nvim
   extensions:
+  - ".vba"
```
pchaigno: The first extension should be the "primary extension", only subsequent extensions need to be sorted alphabetically. `.vim` is the primary extension for Vim scripts, no?

lib/linguist/heuristics.yml
```diff
@@ -426,6 +426,11 @@ disambiguations:
     pattern: '^\s*(import.+(from\s+|require\()[''"]react|\/\/\/\s*<reference\s)'
   - language: XML
     pattern: '(?i:^\s*<\?xml\s+version)'
+- extensions: ['.vba']
+  rules:
+  - language: Vim script
+    pattern: '^UseVimball'
```
pchaigno: Do you know if Vim scripts are invalid if they don't match this pattern?

test/test_heuristics.rb
```diff
@@ -475,6 +475,13 @@ def test_tsx_by_heuristics
     })
   end
 
+  def test_vba_by_heuristics
+    assert_heuristics({
+      "Visual Basic" => all_fixtures("Visual Basic", "*.vba"),
+      "Vim script" => all_fixtures("XML", "*.vba")
```
pchaigno: Typo?
```suggestion
      "Vim script" => all_fixtures("Vim script", "*.vba")
```

smola: Yep. Thanks!

lib/linguist/languages.yml
smola: Indeed, `.vim` is still the primary extension. `.vba` and `.vmb` are Vimball archives (Vim plugins archived by [Vimball](https://www.vim.org/scripts/script.php?script_id=1502)). They are Vim scripts, but they will work only if the Vimball script is installed. They invoke it with `UseVimball` to extract the contents, which are in the rest of the archive.

AFAIK all Vimball archives have this `UseVimball` call. And if something has `.vba` and `.vmb` and it's not a Vimball archive, it's very unlikely that it's a different kind of Vim script.

lib/linguist/heuristics.yml
smola: Se above. Vim scripts will be valid if they don't match this, but then they'll have `.vim` extension.

pchaigno: Good. Let's keep the heuristic as it is then.

PR #4415, opened 2019-02-11
Revisiting #3815 (Update Racket color)

pmaddams:
In #3815, I wanted to update the Racket color to match the blue color used in the logo. Unfortunately, Go was very close in the colorspace. In #4331, Go was changed to a lighter blue. Now, with a minor tweak, the Racket color can be successfully updated from `#22228f` to `#3c5caa`.

This won't turn out like the #4319 (Rust) debacle for the following reasons:
- The blue color is official, not just a random opinion.
- We are going from a dark and not easily visible shade of blue to a lighter shade of the same color.
- Racket is not as popular as Rust, so an army of Reddit users will not show up armed with negative emojis.

Of course, if a Racket developer shows up and explains why this is a disaster, it will be easy enough to revert. But I think a nicer color will help Racket, an underappreciated Lisp dialect, look more attractive.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4079,7 +4079,7 @@ RUNOFF:
   language_id: 315
 Racket:
   type: programming
-  color: "#22228f"
+  color: "#3c5caa"
   extensions:
   - ".rkt"
   - ".rktd"


pchaigno:
Anywhere we could address the Racket community at large to get their opinion (e.g., mailing list)?

pmaddams:
Yes, that's a good idea. I've mailed the users list: https://groups.google.com/forum/#!topic/racket-users/suv7Kv7qUBw

lildude:
From an email sent directly to me 😢 :

> Is there a reason the changes have not been merged?

Because I've not had the time to do it :-D - Linguist is very low on my list of priorities. And when I approve and merge a PR isn't really much of an issue as changes won't appear on GitHub until such time as I make a release, which in itself needs waaay more time than you'd think.

As an aside, it's not particularly nice emailing maintainers directly asking about an issue when you could have simply commented on the issue.

pmaddams:
Sorry about that. Thanks Colin.

Issue #4291, opened 2018-10-09
Revisiting language groups


Alhadis:
This issue is a continuation of what @pchaigno started with github/linguist#3093:

> This pull request makes SASS a language of its own, distinct from CSS.
> This change was requested several times in #2933, #3084, #2585 and #2650.

There are several languages on GitHub which presently fall under the usage statistics of another, "parent" language, which certainly deserve reconsideration — or at the very least, some public discussion for highlighting the reasons *why* they fall under another language's umbrella.

To start, here are the languages which I believe are valid candidates for degrouping. I'll extend this list over time as discussion from other users confirms other candidates:

Languages which should be degrouped
---------------------------------------

| Candidate      | Currently grouped under | Reason(s) for decoupling
| :------------- | :---------------------- | :-------------------------------------
| Svelte | HTML | https://github.com/github/linguist/issues/4291#issuecomment-569400535
| Sass/SCSS      | CSS                     | Extremely different syntax and semantics. Sass has programmatic features and some "object-oriented" features; CSS is strictly declarative.
| Less           | CSS                     | See above. Less's syntax is much closer to "pure" CSS than Sass/SCSS, but it's still programmatic in nature and considerably different enough to warrant separation.
| ~~JSON~~ Fixed in #4345           | JavaScript              | JSON is a general-purpose data serialisation language, and virtually every modern programming language has support for reading and parsing JSON syntax to some extent (either natively or via a library). It's the closest thing we have to a universally interoperable data exchange format.<br><br>Moreover, there's little point in retaining a connection with JavaScript. JSON is classed as a data language, so it won't appear in usage statistics anyway.

I've refrained from bringing up any languages I've never worked with or lack familiarity with (such as PostCSS and Stylus), each of which might be candidates as well. Comments are welcome.

Good examples of language groups
-----------------------------------
Here are some languages which *are* justifiable in having a parent language:

* __HTML__          ← The various `HTML+…` languages (HTML+Django, HTML+ECR, HTML+ERB, etc)
* __Python__        ← Python console, Python traceback
* __JSON__          ← JSON with comments
* __Haskell__       ← Literate Haskell
* __CoffeeScript__  ← Literate CoffeeScript
* __Pic__           ← Roff <sup><a href="#1">[1]</a></sup>

/cc @pchaigno, @lildude, @controversial, @nazar-pc, @EmmaRamirez, @plibither8

Footnotes
----------
[...]

PR #4417, opened 2019-02-12
Add Salesforce Blue colour for Salesforce Apex

davidsbond:
Currently, Apex doesn't have a colour. It should use the Salesforce blue

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -254,6 +254,7 @@ ApacheConf:
   language_id: 16
 Apex:
   type: programming
+  color: #1797c0
   extensions:
   - ".cls"
   tm_scope: source.java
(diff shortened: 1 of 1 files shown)


Alhadis:
@davidsbond The colour is (currently) too close to those of other languages, which is causing the [colour proximity test](https://travis-ci.org/github/linguist/jobs/492192567#L1597-L1601) to fail.

Hold off on attempting a fix, however. It's *possible* (🤞) that the GitHub team [will drop the colour difference requirement](https://github.com/github/linguist/issues/4291#issuecomment-453052275) now that language bars have dividers to separate colours. @lildude is currently waiting on feedback from the rest of the team, so if all goes well, we can use the official Salesforce blue instead of a compromise.

Alhadis:
@davidsbond I'll make a note to update Salesforce's colour should GitHub's colour-proximity restriction be lifted in future. 👍

PR #4418, opened 2019-02-13
 A more specific heuristic for Mathematica

### LINKED ISSUE #4385: MATLAB file misidentified as Mathematica
## Preliminary Steps

Please confirm you have...

## Problem Description

One of my many MATLAB files is being incorrectly identified as Mathematica. (Yes, I will create a `.gitattributes` file to fix this, this issue is to hopefully improve your heuristics.)

The current heuristic looks for the pattern `\*\)$`, which identifies Mathematica. If it fails, it continues with the MATLAB pattern. This one file contains this comment, which matches the pattern:

    %       LCH (or L*C*H*)

Previously, Mathematica was identified by the opening comment tag: `^\s*\(\*` (see #1652). That would be an illegal sequence in a MATLAB file, as far as I know, and therefore seems to be a much better test. Is there a reason this was changed?

Is matching both the opening and the closing tag too expensive?

Otherwise, I could suggest a different strategy: if a directory contains hundreds of files with the same extension, and most of them are MATLAB, it is very likely that they are all MATLAB. To me it seems highly unlikely that a repository would have, in one directory, multiple files with the same extension that are not all the same language. Such a test could help you skip the regular expression tests on many files.

Repository: https://github.com/DIPlib/diplib/

The affected file is: https://github.com/DIPlib/diplib/blob/master/dipimage/@dip_image/dip_image.m

### Last modified on:
2018-11-20

### Expected language:
MATLAB

### Detected language:
Mathematica

crisluengo:
This commit extends the heuristic test for Mathematica, to prevent some false positives. Fixes #4385.

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):
  
    - Sample source(s): Modified from DIPlib
      - https://github.com/DIPlib/diplib
    - Sample license(s): Apache 2.0

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -224,7 +224,9 @@ disambiguations:
   - language: M
     pattern: '^\s*;'
   - language: Mathematica
-    pattern: '\*\)$'
+    and:
+      - pattern: '\(\*'
+      - pattern: '\*\)$'
   - language: MATLAB
     pattern: '^\s*%'
   - language: Limbo
--- a/test/fixtures/MATLAB/colorspace.m
+++ b/test/fixtures/MATLAB/colorspace.m
@@ -1,0 +1,39 @@
+%COLORSPACE   Gets/sets/changes the color space.
+%    IN = COLORSPACE(IN,COL), with IN a color image, changes the color
+%    space of image IN to COL, converting the pixel values as required.
+%    If COL is 'grey', a scalar image is returned with the luminosity (Y).
+%    If COL is '', no transformation is made, the color space information
+%    is simply removed.
+%
+%    A color space is any string recognized by the system. It is possible
+%    to specify any other string as color space, but no conversion of pixel
+%    values can be made, since the system doesn't know about that color
+%    space. Color space names are not case sensitive. Recognized color
+%    spaces are:
+%
+%       grey (or gray)
+%       RGB
+%       sRGB
+%       CMY
+%       CMYK
+%       HSI
+%       ICH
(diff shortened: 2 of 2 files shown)


samples/MATLAB/colorspace.m
```diff
@@ -0,0 +1,36 @@
+function obj = colorspace(obj,col)
```
pchaigno: You mentioned you modified the file. Could you clarify how and why you modified it?
Was it [this file](https://github.com/DIPlib/diplib/blob/cf71858a91991fc2c8796e4f9a12f57643a24f0b/dipimage/%40dip_image/dip_image.m) initially? If it was, I think we should rather put it in `tests/fixtures/MATLAB/`.

crisluengo: Yes, that's the file. It's kind of large, I thought you would not want such a large example just for the sake of one particular comment line. But I'd be happy changing that commit to include the whole file.

crisluengo: Is this what you had in mind?

pchaigno: Nop, you were right to change it. We avoid including large files because of memory issues (see #2117 for details).

We also try to only use original files to train the Bayesian classifier, so I think we should put this file in `tests/fixtures/MATLAB/`, where the Bayesian classifier won't use it.

crisluengo: OK, I've put the file I originally submitted in the `tests/fixtures/MATLAB` directory.

PR #4423, opened 2019-02-15
Add Go testdata to vendor.yml

fgblomqvist:
This adds Go `testdata` folders to vendors.yml as per #4420.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -49,7 +49,7 @@
 - Godeps/_workspace/
 
 # Go fixtures
-- **/testdata/**
+- (^|/)testdata/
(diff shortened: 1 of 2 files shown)


fgblomqvist:
I'm a little lost regarding how you want this tested. There doesn't seem to be any other tests for vendor.yml patterns (only one for README in test_blob#test_vendored?) but I put a test in there s since that's what the instructions say in vendor.yml. Lmk if the test isn't of the correct format (wasn't sure which function you want to be used, but sample_blob seemed to do the job, not requiring any empty dummy-files etc.).

fgblomqvist:
Alright, sounds good!

PR #4427, opened 2019-02-18
Correct misclassification of .js.erb files

CHTJonas:
Change the language classification of `*.js.erb` files from HTML+ERB to JavaScript+ERB. I think I've covered all the relevant bases but let me know if there's something I can correct - thanks for taking the time to review! 😄 

## Description
Previously JavaScript files containing embedded Ruby syntax were misclassified as HTML+ERB. This PR would correct this by classifying such source code files as JavaScript+ERB and grouping under the parent JavaScript language.

Motivation for this is that JavaScript files meant for client-side execution in response to an AJAX request in a Rails application are listed under HTML when running `github-linguist --breakdown`.

## Checklist:
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Aerb+escape_javascript+NOT+nothack
    - Sample source(s):
      - https://edgeguides.rubyonrails.org/working_with_javascript_in_rails.html#a-simple-example
    - Sample license(s):
      - MIT


lib/linguist/languages.yml
```diff
@@ -2371,6 +2371,16 @@ JavaScript:
   interpreters:
   - node
   language_id: 183
+JavaScript+ERB:
+  type: programming
+  tm_scope: none
```
pchaigno: Couldn't we use the JavaScript grammar for highlighting here?

lib/linguist/languages.yml
```diff
@@ -2371,6 +2371,16 @@ JavaScript:
...
+  group: JavaScript
+  extensions:
+  - ".js.erb"
+  ace_mode: text
+  codemirror_mode: htmlembedded
```
pchaigno: The [HTML Embedded](https://codemirror.net/mode/htmlembedded/index.html) page mentions that it supports the `x-erb` MIME type, but did you check how the actual highlighting looks like? Wouldn't JavaScript be a better fit here as well?

CHTJonas:
Thanks for the comments @pchaigno! Syntax highlighting updated now to use JavaScript now.

I wasn't sure whether `tm_scope` should be `source.js` (same as JavaScript language proper) or whether I should add `text.js.erb` to the `grammars.yml` file to match what HTML+ERB does?

pchaigno:
> I wasn't sure whether `tm_scope` should be `source.js` (same as JavaScript language proper) or whether I should add something like `text.js.erb` to the `grammars.yml` file to match what HTML+ERB and NetLinx+ERB do?

The `grammar.yml` should not be manually updated so `source.js` is fine. If we're using JavaScript for both the TextMate scope and the CodeMirror grammar, we might as well add the extension to the list of JavaScript extensions. It would achieve the exact same effect.

CHTJonas:
I was considering that, but I think(?) listing separately does allow some benefits eg. when searching you can restrict the scope to JavaScript+ERB as opposed to all JavaScript, such as [this example](https://github.com/RailsApps/rails-composer/search?l=HTML%2BERB) with HTML+ERB.

CHTJonas:
Did you have any thoughts on the above RE: whether to keep the `.js.erb` extension separate or just add it to the list of JS extensions @pchaigno?

PR #4433, opened 2019-02-24
Add strategy to identify Roff man pages: Take 2

Alhadis:
This PR adds a strategy to identify man-pages with unrecognised file extensions. Added in support of github/markup#1196.

## Description
This was previously attempted in #4317, but ditched after the complexity of the PR convinced me a simpler solution was needed to safeguard against incorrectly rendered man-pages. That concern has since been addressed by #4393, leaving us free to adopt a much simpler implementation.

Note that the `Manpage` strategy deliberately returns two language entries, deferring the exact classification to the heuristics (which determine whether a Roff file should be rendered as a man-page or not).

**References:** #4258, #4309, #4317

/cc @pchaigno, @lildude, @jordemort, @pali

## Checklist:


jordemort:
pls no close stalebot, I'll try and find a reviewer

pali:
any update? :-)

Alhadis:
Nudging @lildude for input.

lildude:
From a quick look and based on your explanation, it makes sense to me, but I'll let @pchaigno take a peek and comment before making a final decision.

As an aside, you'll need to merge in master to get the tests passing as I fixed the current failure in https://github.com/github/linguist/pull/4478.

Alhadis:
Whoops, completely missed that. 😀 Fixed!

Alhadis:
Nudging @pchaigno.

Alhadis:
If somebody — *anybody* — could review this, shut stalebot up, and make many man-page lovers happy, that'd be *seriously* copacetic.

Seriously.

/cc @pchaigno again

lib/linguist.rb
```diff
@@ -53,7 +54,7 @@ def detect(blob, allow_empty: false)
   #
   #   blob - An object that quacks like a blob.
   #   languages - An Array of candidate Language objects that were returned by the
-  #               previous strategy.
+  #               previous strategy.
```
pchaigno: Uh? What happened here?

lib/linguist/strategy/manpage.rb
```diff
@@ -0,0 +1,42 @@
+module Linguist
+  module Strategy
+    # Detects man pages based on numeric file extensions with group suffixes.
+    class Manpage
+
+      # Public: RegExp for matching conventional manpage extensions
+      #
+      # This is the same expression as that used by `github/markup`
+      MANPAGE_EXTS = /\.(?:[1-9](?![0-9])[a-z_0-9]*|0p|n|man|mdoc)(?:\.in)?$/i
```
pchaigno: That regular expression is also going to match the following extensions from other languages:
- Forth: `.4th`.
- Perl6: `.6pl` and `.6pm`.
- TI Program: `.8xp` and `.8xk`.

lib/linguist.rb
Alhadis: Looks like a pair of non-breaking spaces (U+00A0) were inserted at some point (probably be a macOS user who accidentally held the <kbd>Option</kbd> key while hitting the spacebar. Doing that inserts `\xA0` instead of `\x20`).

My editors are setup to convert those nasty little bastards to ordinary spaces when saving source code. So I must've unknowingly fixed a minor formatting error. 😜

lib/linguist/strategy/manpage.rb
Alhadis: Will it? `Linguist::Strategy::Manpage` comes after `Linguist::Strategy::Extension`, which is deliberately to address the problem you pointed out.

Alhadis:
> It's probably better if we avoid hardcoding the problematic extensions...

Well, that can always be solved by generating a `RegExp` from at runtime from every extension which happens to start with a number.

But in any case, this shouldn't be a problem since the manpage-strategy has a lower priority than the filename one does, so "problematic" extensions shouldn't be a problem. 😉

lib/linguist/strategy/manpage.rb
pchaigno: Ah, right. The tests would also fail if someone changed the order of strategies, so :+1: for me!

Alhadis:
*Merci!* 😀

@lildude Alright to merge? ^^

Alhadis:
/cc'ing @lildude in case he's missed this.

jordemort:
Yessssssssssss

Alhadis:
@jordemort Any update on github/markup#1196?

Issue #4449, opened 2019-03-06
Installing dependencies (and other tasks) not working on macOS Mojave


nwhetsell:
I can’t seem to get the steps for [contributing](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) to Linguist to work on macOS Mojave 10.14.3.

First, I have CMake, ICU4C, and pkg-config installed through Homebrew.

Running [script/bootstrap](https://github.com/github/linguist/blob/master/script/bootstrap) results in

```
script/bootstrap: line 7: bundle: command not found
```

I then ran `sudo gem install bundler` successfully, but running script/bootstrap then resulted in:

```
You are replacing the current local value of path, which is currently nil
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 1.10)

  Current Bundler version:
    bundler (2.0.1)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 1.10)' in any of the relevant sources:
  the local ruby installation
```

I then ran:

```
sudo gem uninstall bundler
sudo gem install bundler -v 1.10
```

But running script/bootstrap then resulted in:

```
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies...
Installing rake 12.3.2
Installing public_suffix 3.0.3
Installing addressable 2.5.2
Using bundler 1.10.0
with native extensions Installing byebug 11.0.0
with native extensions 
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /Users/nathanwhetsell/Documents/linguist/vendor/gems/ruby/2.3.0/gems/charlock_holmes-0.7.6/ext/charlock_holmes
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby -r ./siteconf20190306-87297-m6l1ib.rb extconf.rb
checking for main() in -licui18n... no
checking for main() in -licui18n... yes
checking for unicode/ucnv.h... yes
checking for main() in -lz... yes
checking for main() in -licuuc... yes
checking for main() in -licudata... yes
creating Makefile

To see why this extension failed to compile, please check the mkmf.log which can be found here:

[...]

PR #4451, opened 2019-03-07
Add rake task to check grammars

lildude:
## Description

In my ongoing quest to prevent losing a grammar again without noticing, this PR adds a new rake task that is run whenever the grammars are updated. It should confirm that no grammars have suddenly gone missing and also flag if there are any new errors.

Sadly, we have to rely on a hardcoded known error account until such time as we've got 100% error free grammars, which isn't likely to happen any time soon.  This is however sufficient enough to prompt anyone making a new release to check the output before actually releasing.

While I was updating the release section on the CONTRIBUTING.md, I also tweaked the dependencies and getting started sections following the feedback in https://github.com/github/linguist/issues/4449

## Checklist:


CONTRIBUTING.md
```diff
@@ -8,6 +8,14 @@ This project adheres to the [Contributor Covenant Code of Conduct](http://contri
...
 The majority of contributions won't need to touch any Ruby code at all.
 
+### Dependencies
+
+Linguist uses the [`charlock_holmes`](https://github.com/brianmario/charlock_holmes) character encoding detection library which in turn uses [ICU](http://site.icu-project.org/), and the libgit2 bindings for Ruby provided by [`rugged`](https://github.com/libgit2/rugged). [Bundler](https://bundler.io/) v1.10.0 or newer is required for installing the Ruby gem dependencies. [Docker](https://www.docker.com/) is also required when adding or updating grammars. These components have their own dependencies - `icu4c`, and `cmake` and `pkg-config` respectively - which you may need to install before you can install Linguist.
+
+For example, on macOS with [Homebrew](http://brew.sh/): `brew install cmake pkg-config icu4c && brew cask install docker` and on Ubuntu: `apt-get install cmake pkg-config libicu-dev docker-ce`.
+
+The latest version of Bundler v1 can be installed with `gem install bundler -v "~>1.10"`.
+
```
lildude: I've moved the dependencies above "Getting started" as it's a lot easier to get started once you've met all the dependencies 😁

davidhq:
Perfect :) Since then I sent you a pull request for a preliminary version of Syntax Highlighter for Vyper (second Ethereum smart contract language), used for example in these contracts: https://github.com/Uniswap/contracts-vyper/blob/master/contracts/uniswap_exchange.vy

I think it's a low risk to include this in current release, so if you manage... if not, would be great if next time. Highlighter follows the same agreement as the one for Solidity (.sublime-syntax on master and tmLanguage on `linguist` branch)... both will further improve in next weeks to nail down the smaller details.

Rakefile
```diff
@@ -54,6 +55,18 @@ task :flex do
...
 
+# The error count will need to be adjusted here until such time as all grammars are 100% error free.
+desc "Check that compiling the grammars doesn't introduce any new unexpected errors"
+task :check_grammars do
+  expected_error_count = 35  # This count should only ever go down. If it goes up, there's a new issue that needs to be addressed before updating the grammar.
+  rm_rf "linguist-grammars"
+  output, status = Open3.capture2e("script/grammar-compiler", "compile", "-o", "linguist-grammars")
+  errors_found = output[/the grammar library contains ([0-9]+) errors/, 1].to_i
+  missing_grammars = output.scan(/Missing scope in repository: `([^`].+)` is listed in grammars.yml but cannot be found/)
+  fail "#{output}\n\nERROR: The following grammars appear to have gone missing:\n\n#{missing_grammars.join("\n")}\n\nPlease review the output above.\n\n" unless missing_grammars.empty?
```
Alhadis: Squiggle here-docs might be clearer here:

~~~ruby
fail <<~EOS
    #{output}

    ERROR: …
  EOS
~~~

I'd also get rid of the _"appear to have gone missing"_ bit and keep the error message punchy and assertive:

> ERROR: Could not find the following grammars:
> …

Rakefile
```diff
@@ -54,6 +55,18 @@ task :flex do
...
+  fail "#{output}\n\nERROR: An unexpected number of errors have been found. Expected: #{expected_error_count}, Found: #{errors_found}.\nPlease review the output and adjust the rake task expected error count if the number has gone down.\n\n" unless errors_found == expected_error_count
```
Alhadis: Going from my own experience in Perl, I can tell you that putting a condition like `unless` at the end of a very lengthy statement is a maintainability accident waiting to happen. 😜 

I'd replace this with a here-doc or an `unless … end` block, personally.

Alhadis:
Eh... that review was supposed to have been a comment, not a change request. Bit gauche of me, sorry. 😞

lildude:
> Eh... that review was supposed to have been a comment, not a change request. Bit gauche of me, sorry. 😞

NP. All the same to me 😄

PR #4455, opened 2019-03-11
Add Jsonnet language.

sparkprime:
This is an update of the following PR which has now been accepted.  

https://github.com/github/linguist/pull/2653

@pchaigno 

Things that I updated:
- Update submodule to latest language-jsonnet
- Add libsonnet and libjsonnet extensions
- Change colour to the one used by the new website and logo

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2428,10 +2428,11 @@ Jsonnet:
   type: programming
   ace_mode: text
   extensions:
-  - .jsonnet
-  - .libsonnet
-  - .libjsonnet
+  - ".jsonnet"
+  - ".libjsonnet"
+  - ".libsonnet"
   tm_scope: source.jsonnet
+  language_id: 664885656
 Julia:
   type: programming
   extensions:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2423,6 +2423,15 @@ Jolie:
   ace_mode: text
   tm_scope: source.jolie
   language_id: 998078858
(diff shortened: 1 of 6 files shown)


sparkprime:
I ran harvester on .libsonnet and it returned 244 repos with
`xsel -b | cut -f 5 -d '/' | sort -u | wc`

The .libjsonnet extension is much less used, we can drop it if it's a problem

lildude:
> The .libjsonnet extension is much less used, we can drop it if it's a problem

Yes, please. A simple GitHub search shows only 10 uses - https://github.com/search?q=extension%3Alibjsonnet+NOT+nothack&type=Code

`.libsonnet` appears sufficiently widespread to include support.

lib/linguist/languages.yml
```diff
@@ -2423,6 +2423,16 @@ Jolie:
   ace_mode: text
   tm_scope: source.jolie
   language_id: 998078858
+Jsonnet:
+  color: "#0064bd"
+  type: programming
+  ace_mode: text
+  extensions:
+  - ".jsonnet"
+  - ".libjsonnet"
```
lildude: 🔥 this extension for now - it's no where near prevalent enough at only 10 files.

sparkprime: Done.

lildude:
@pchaigno or @Alhadis wanna take a quick peek at this in case I'm missing anything. I'm in the process of testing a new release and I think we can squeeze this one in.

Alhadis:
LGTM. :+1:

PR #4459, opened 2019-03-13
Remove yarn.lock from auto-generated files list (follow-up)

### LINKED ISSUE #4348: Reconsidering auto-detecting Yarn files as auto-generated
## Preliminary Steps

Please confirm you have...

## Problem Description

#3431 added support for generated file detection for Yarn. I'm not sure this is a good idea.

While generated, the lockfile contains important information regarding the dependencies used by a project; maybe it would make sense to display them so that they can be properly reviewed? For example, it would be important for a reviewer to understand that a PR adds N packages to the lockfile (which might be much higher than the number of lines added to the `package.json` file).

This might also have security implication. A malicious author could easily change some resolved versions inside the `yarn.lock` file while also upgrading a `package.json` dependency to a innocent release. Since Github won't show the lockfile content by default the reviewer might forget to check it, accept the apparently harmless upgrade, and let the malicious override make its way to the lockfile.

Note that the same concerns might apply to the `package-lock.json` file from npm (except that they are much harder to review due to their nested design).

edmorley:
This is a follow-up to #4445, which intended to make `yarn.lock` no longer be treated as an auto-generated file (to fix #4348), however only removed the tests rather than also the handling in `generated.rb`.

As of this PR, b36ea7a is now fully reverted.

Fixes #4348.


Alhadis:
Thanks! This was my blunder... :sweat

edmorley:
Many thanks :-)

lildude:
@Alhadis Hmmm, merging this whilst the release PR was pending has caused a bit of a complication WRT the versions and contents 😞  I'm going to need to get a bit creative to get this into v7.3.1

Alhadis:
Sorry about that. :/ If I'd known that it wasn't possible to squeeze in last minute changes, I wouldn't have been so quick to merge.

PR #4464, opened 2019-03-14
Add unicode filename fixture

lildude:
## Description
This PR adds a file with a unicode char in the filename to the fixtures used during testing.

This is required by the test that will be modified in a PR for master to ensure filenames are being forcefully UTF-8 encoded.

As this branch is only used for testing gitattributes from the master branch, and is waaaaay out of date, the tests here don't need to pass.


lildude:
This is needed by https://github.com/github/linguist/pull/4465

Issue #2988, opened 2016-05-06
Old-style CR terminators can affect classification


Alhadis:
This was discovered in [#2971](https://github.com/github/linguist/pull/2971#issuecomment-217193513), where one file was still being misclassified because of its unusual line-endings.

With support for WoW-ToC files added, `ChatKeys.toc` should've been identified as addon metadata. However, Linguist still interpreted it as TeX until the line endings were converted from old-style CR to proper modern LF. Suddenly, it identified as WoW-ToC again.

Could choice of line-ending be potentially skewing the classification?

Converting the file to LF endings wasn't possible in Atom due to a bug (which was [recently addressed](https://github.com/atom/line-ending-selector/pull/27)), so I had to use Perl. I've [attached the file](https://github.com/github/linguist/files/251768/ChatKeys.zip) with both line-ending types to spare you the trouble.

Alhadis:
.

lildude:
> .

Removing the Stale label is also enough to indicate activity and stalebot will ignore the issue/PR for another 30 days.

Alhadis:
Oh. :sad Sorry.

Alhadis:
Alright, so it's not the CR line-terminators that's screwing up classification. If I rerun `github-linguist` with an edited copy of `ChatKeys-CR.toc` with all carriage-returns replaced with nothing, it gives me the same result (`TeX`).  It might be line-length that's the factor:  because CR hasn't been supported as a line-terminator since the "classic Macintosh" died,

In text, the `\r` sequence becomes an invisible, zero-width indicator of absolutely nothing... that is, programs act as though they simply aren't *there* leaving CR-delimited runs of text to flop into one huge ass paragraph, which is probably what's gotten our bot confused with the likeness.

In any case, a few new samples and a carefully-written heuristic should put this matter to rest. 👍

Alhadis:
Also, it's kind of strange that this offending fixture has been sitting in Linguist's `samples` directory for quite some time now, and the analysis is blatantly incorrect:

~~~~diff
index a71f827..ec9ba7a 100644
--- /samples/World of Warcraft Addon Data/ChatKeys-CR.toc
+++ /samples/World of Warcraft Addon Data/ChatKeys-LF.toc
@@ -1,16 +1,3 @@
-  ChatKeys-CR.toc                                      
-     10 lines (9 sloc)                                 
-     type:      Text                                   
-     mime type: text/plain                             
[...]

PR #4465, opened 2019-03-14
Force UTF-8 for filenames in breakdown analysis

lildude:
As reported in https://github.com/github/linguist/issues/4028 the JSON output for repos which contain filenames that contain unicode chars fails with and encoding error like this:

```console
$ git init foo
Initialized empty Git repository in /Users/lildude/Downloads/trash/foo/.git/
$ cd foo
$ echo "foo" > foo/fileã.rb
$ git add fileã.rb
$ git commit -m 'Foo'
[master (root-commit) c1e9aa2] Foo
 1 file changed, 1 insertion(+)
 create mode 100644 "file\303\243.rb"
$ github-linguist --json
/Users/lildude/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/json-2.1.0/lib/json/common.rb:224:in `encode': "\xC3" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
	from /Users/lildude/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/json-2.1.0/lib/json/common.rb:224:in `generate'
	from /Users/lildude/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/json-2.1.0/lib/json/common.rb:224:in `generate'
	from /Users/lildude/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/json-2.1.0/lib/json/common.rb:394:in `dump'
	from /Users/lildude/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/github-linguist-7.0.0/bin/github-linguist:47:in `<top (required)>'
	from /Users/lildude/.rbenv/versions/2.4.3/bin/github-linguist:23:in `load'
	from /Users/lildude/.rbenv/versions/2.4.3/bin/github-linguist:23:in `<main>'
$
```

Performing a breakdown analysis behind the scenes on GitHub.com will result in a 500 error there too.

As pointed out by @tenderlove:

>  It looks like we're getting this error because we're trying to dump out filesystem information (specifically file paths) as JSON. Unfortunately:
> 
> 1. JSON can only represent UTF-8 strings
> 2. Filesystems can use any encoding for file names
> 3. Git must support storing these paths
> 4. Git doesn't store encoding information about the paths, they are just a list of bytes

... and he proposed two possible solutions. I've gone for the latter in this PR and forcing UTF-8 on all filenames when performing a breakdown.

@tenderlove do you see any issues with this approach? Seems too easy :wink:.

Now analysis of my example produces:

```console
$ bundle exec bin/github-linguist ~/tmp/trash/foo --json | jq
{
  "Ruby": [
    "fileã.rb"
  ]
}
$
```

I've added a new fixture file to the `test/attributes` branch in https://github.com/github/linguist/pull/4464 (will need to change the SHA in this PR when that PR has been merged) in order to test this encoding enforcement. 

I've gone with this as we don't care about the content of the file and can't add it to the samples as the content is deliberately descriptive about the purpose of the file. As `test/fixtures` is vendored by default the normal testing wouldn't pick up the file, so I've piggy-backed onto the repo tests when we un-vendor `test/fixtures`.

If the enforcement doesn't happen, the test fails as such:

```
  1) Failure:
TestRepository#test_repo_git_attributes [/Users/lildude/github/linguist/test/test_repository.rb:69]:
Expected false to be truthy.
```

Fixes https://github.com/github/linguist/issues/4028


Alhadis:
I'm not sure if this issue is connected with the encoding errors I get on OpenBSD. Here's what `bundle exec rake` looks like when I haven't overridden its environment with `LANG`:

~~~console
$ type brake
brake is aliased to `LANG=en_AU.UTF-8 bundle exec rake'
~~~

This seems to fix the encoding errors. I'm still not sure what's going on, or *why* this issue only happens on OpenBSD...

lildude:
@tenderlove If you've got a free moment, I'd like your thoughts on this fix based on your comment at https://github.com/github/linguist/issues/4028#issuecomment-405307413 hence I've requested a review from you.

lib/linguist/repository.rb
```diff
@@ -96,7 +96,7 @@ def breakdown_by_file
       @file_breakdown ||= begin
         breakdown = Hash.new { |h,k| h[k] = Array.new }
         cache.each do |filename, (language, _)|
-          breakdown[language] << filename
+          breakdown[language] << filename.dup.force_encoding("UTF-8")
```
tenderlove: We might want to call `scrub` here.  If the filename contains bytes that aren't valid UTF-8, we'll end up with strings that aren't "valid":

```
irb(main):004:0> path = "/foo/ba\xe9r/baz"
=> "/foo/ba\xE9r/baz"
irb(main):005:0> path.force_encoding "UTF-8"
=> "/foo/ba\xE9r/baz"
irb(main):006:0> path.encoding
=> #<Encoding:UTF-8>
irb(main):007:0> path.valid_encoding?
=> false
irb(main):008:0> path.scrub
=> "/foo/ba�r/baz"
irb(main):009:0> scrubbed = path.scrub
=> "/foo/ba�r/baz"
irb(main):010:0> scrubbed.encoding
=> #<Encoding:UTF-8>
irb(main):011:0> scrubbed.valid_encoding?
=> true
```

tenderlove:
@Alhadis the `LANG` environment variable impacts the encoding of strings read in by the program.  It may have been previously set to `C` which would cause strings read in (including the names of files) to be tagged as `US-ASCII` even though they contained unicode characters.  Changing `LANG` to include `UTF-8` would cause the strings read to be tagged as UTF-8 so it makes sense that your environment variable fixed this issue.

I hope that helps!

lib/linguist/repository.rb
lildude: Oooo. I've added the scrub and also updated the tests to confirm this is working as expected. Thanks @tenderlove.

Alhadis:
Ah, that would make sense. Still, isn't there any way we could detect problematic `locale` settings up-front, so as to not confuse the next user who's faced with a waterfall of unhelpful error messages...

lildude:
> Still, isn't there any way we could detect problematic `locale` settings up-front, so as to not confuse the next user who's faced with a waterfall of unhelpful error messages...

🤔 this is more of a user env issue than a Linguist issue. We could put some sort of locale check at the beginning of `github-linguist` like:

```ruby
abort "A UTF-8 compatible locale is required to run github-linguist" unless Encoding.default_external.to_s =~ /UTF-8/
```

But we may need to think carefully about this. Def a topic for another issue, though you're the only person I know who's hit this issue 😉.

Alhadis:
> this is more of a user env issue than a Linguist issue.

Is it? Every time I see that warning emitters by `charlock_holmes` about the default encoding variable not being set, I can't help but wonder if that's got something to do with it.

In any case, I've grown so used to running my `brake` alias that I no longer think anything of it. 😝 Sometimes I use Docker to remind me that I've forgotten to override the `LANG=` variable again (where "docker" is really a shell-script in my `~/.files/bin` directory on OpenBSD whose sole purpose is to cheat the grammar registration script.

... but don't blame me though, there's no Docker port for OpenBSD, and I barely know a single damn thing about Docker. 😢

lildude:
> Is it? Every time I see that warning emitters by `charlock_holmes` about the default encoding variable not being set, I can't help but wonder if that's got something to do with it.

Oh, I get those too and everything I do is in UTF-8 😀

Alhadis:
I know, I see the warnings emitted on TravisCI as well. When I did some digging, I remember reading something about it deferring to the first locale it could find in the site's `iconv` headers or... something along those lines. I came away with the impression that the locale handling was nondeterministic, and subject to variation between platforms.

ANYWAYYYYY, don't mind me. :D I'm happy with my Perl and JavaScript, where locale is more sane than, erm, Ruby's. 😀

(I also realised I've been side-tracked the whole day and haven't finished tending to #2988 yet. Time to get another oldie off our lingering issues list. 👍

Issue #3949, opened 2017-12-13
Addition: Reasonml


Haroenv:
There seems to be a file with it already in the vendors folder, but it's not picked up on GitHub.com. Which files need to be edited for it to be possible?

lildude:
Reason should be detected. If you're not seeing it being detected, there's likely another explanation.

Can you provide a link to a repository where you are seeing that it isn't being detected.

Haroenv:
* in https://github.com/reasonml-community/reason-react-hacker-news/ it's detected as OCaml or C++
* https://github.com/BuckleScript/bucklescript/blob/master/jscomp/build_tests/jsxv/src/demo.re doesn't seem to be detected
* https://github.com/reasonml-community/bs-react-native has detected as Reason in the search, but not in the overview

I've never seen Reason in the language overview

Haroenv:
* https://github.com/reasonml-community/bs-director
* https://github.com/reasonml-community/reductive

and a bunch of other projects on https://github.com/reasonml-community as well

pchaigno:
Reason is [part of the OCaml group](https://github.com/github/linguist/blob/d4c2d83af9f576aacdd97d58796d2a22fa90def4/lib/linguist/languages.yml#L3813) so Reason files are counted as OCaml files (only the syntax highlighting differs).

Haroenv:
Interesting, I knew it's a different syntax for the same language, but I'm wondering if it wouldn't be better if it was split out in its own language colour

mlms13:
I agree that it would be nice for Reason to be shown separately from OCaml, but independent of that, I can also confirm that some Reason files are getting misidentified as C++.

lpil:
Given ReasonML is just an alternative parser for the OCaml compiler I don't think it should be a different language.

Haroenv:
* https://github.com/reasonml-community/reason-react-hacker-news/search?l=c%2B%2B detected as c++ still

Not sure if you want to fix this still

lildude:
Thanks for the nudge @Haroenv. Looks like this is caused by a lack of a heuristic for the `.re` extension. It is associated with both C++ and Reason but without an heuristic so is falling through to the classifier.

I have no experience with either language, but if you've got the bandwidth and some experience, we'd appreciate a PR adding the necessary heuristic to distinguish the two languages.

Haroenv:
Sorry, not my area of expertise :)

Haroenv:
[...]

PR #4473, opened 2019-03-28
Fix Reason detection

dawee:
* I removed `.re` from C++ extensions.
* I removed the group OCaml of Reason

## Description
This change is linked to the comment I made here: https://github.com/github/linguist/issues/3949#issuecomment-475834977

On the ungroup of OCaml: Reason might be an extension of OCaml, it has its own identity. I'm a Reason developer I'm not an OCaml developer at all. I don't even know how to read OCaml properly and I work with Reason everyday. I really want the ability to give a ReasonML its "Reason" identity so I can find them in github trendings.

On the removal of .re in C++: I've been C++ developer and I never saw this extension. I just checked on cppreference and I don't see it either. It's colliding with the official Reason extension.

The overall idea is to have our ReasonML projects having a "Reason" identity, so we can share as we do with all the other languages.

## Checklist:


pchaigno:
> On the ungroup of OCaml: Reason might be an extension of OCaml, it has its own identity. I'm a Reason developer I'm not an OCaml developer at all. I don't even know how to read OCaml properly and I work with Reason everyday. I really want the ability to give a ReasonML its "Reason" identity so I can find them in github trendings.

We might want to add this to #4291. @Alhadis What do you think?

> On the removal of .re in C++: I've been C++ developer and I never saw this extension. I just checked on cppreference and I don't see it either. It's colliding with the official Reason extension.

`.re` was initially added as a C++ extension because of [those files](https://github.com/search?p=1&q=extension%3Are+include+define&ref=searchresults&type=Code&utf8=%E2%9C%93).

Alhadis:
> @Alhadis What do you think?

I'm not familiar with either OCaml *or* Reason, but it certainly sounds like grouping them was a mistake:

>  I don't even know how to read OCaml properly and I work with Reason everyday.

dawee:
> We might want to add this to #4291. @Alhadis What do you think?

I think the issue is a very good initiative, maybe it would still be easier to do the job language by language?

> `.re` was initially added as a C++ extension because of [those files]

The weirdest thing is that the extension has been added to the C++ ones in the same commit Reason was added in Linguist.
The examples you're referencing seem to be specific to this code generator: [re2c](http://re2c.org/manual/syntax/syntax.html). I don't know if it helps but at least removing it wouldn't mean removing an official C++ extension.

Is there a way through linguist to detect with both source content and extension?

Alhadis:
> an official C++ extension.

Just an FYI: if a file extension has enough usage in-the-wild, it doesn't matter if it's "official" or not. If hundreds of repositories on GitHub associate an extension with a language, that's important enough for Linguist to consider, irrespective of whether a language deems it official.

anmonteiro:
> On the ungroup of OCaml: Reason might be an extension of OCaml, it has its own identity.

I don't agree with this. Reason _is_ and has been OCaml from its inception. It's just a frontend to the OCaml compiler, insofar as after the first compiler pass you can't know whether your code was Reason or OCaml to begin with

Additionally, Reason works with OCaml dependencies, and I'll go as far as saying it _needs_ OCaml dependencies, given how more widespread the OCaml ecosystem of libraries is in comparison to Reason.

>  I'm a Reason developer I'm not an OCaml developer at all. 

You may see it that way, but that's not true. You're using the OCaml compiler therefore I consider you an OCaml developer :)

> I don't even know how to read OCaml properly and I work with Reason everyday.

This is a personal shortcoming which IMO is orthogonal to this argument.

dawee:
The community is disagreeing with this so I'm closing the pull request.

dawee:
Sorry I closed it very fast because it created a big debate in the community, we may come back here but for now we need a consensus.

And @Alhadis for what I said about removing `.re` extension on C++, saying it's not official was pushy of me, I'm sorry. It make total sense that a lot of people using it is a good reason to keep it working.

imbsky:
Reason is definitely just a syntax extension of OCaml. but if your opinion is reasonable for this project, TypeScript should be recognized as JavaScript.

alavkx:
Same deal for Elixir/Erlang—there's a precedent for this already. Not having detection for the language hurts the growth of the community. I can't see the downside.

imbsky:
Solved by https://github.com/github/linguist/pull/4713!

PR #4477, opened 2019-03-30
Add the 'constexpr' keyword to C++ named patterns

M4444:
## Description
A header file in my repo was misclassified as C instead of C++:
https://github.com/M4444/TMatrix/search?l=c
https://github.com/M4444/TMatrix/blob/43cd9ce1830bb7e78b7746972296cd3fc368b1e9/include/tmatrix.h

The file contains the `constexpr` keyword which only exists in C++.
This change is the simplest one that fixes the problem.

## Checklist:

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -456,11 +456,10 @@ named_patterns:
   cpp:
   - '^\s*#\s*include <(cstdint|string|vector|map|list|array|bitset|queue|stack|forward_list|unordered_map|unordered_set|(i|o|io)stream)>'
   - '^\s*template\s*<'
-  - '^[ \t]*try'
+  - '^[ \t]*(try|constexpr)'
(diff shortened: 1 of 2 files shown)


lib/linguist/heuristics.yml
```diff
@@ -460,6 +460,7 @@ named_patterns:
   - '^[ \t]*catch\s*\('
   - '^[ \t]*(class|(using[ \t]+)?namespace)\s+\w+'
   - '^[ \t]*(private|public|protected):$'
+  - '^[ \t]*constexpr'
```
pchaigno: Couldn't we merge this pattern with the `try` one above?

M4444:
Addressed the comments. The file in the mentioned repo is under GPL-2.0-only so I created a header file with a similar structure for the sample.
I've pushed a new commit for this because I'm now sure what effect squashing them would have on the pull request but you can squash them before they get merged.

PR #4487, opened 2019-04-05
Document requirements for a language color change

lildude:
## Description

Changing a language's colour has become a bit of a hot topic recently so I'm adding a section to the CONTRIBUTING.md, PR template and issue template advising on what is now required for a colour change.

## Checklist:

Doesn't apply as this is a documentation change.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-<!--- Briefly describe what you're changing in the pull request title. -->
+<!--- Briefly describe your changes in the field above. -->
 
 ## Description
 <!--- If necessary, go into depth of what this pull request is doing. -->
@@ -50,7 +50,7 @@
   <!-- This includes modifying the vendor, documentation, and generated lists. -->
   - [ ] I have added or updated the tests for the new or changed functionality.
 
-- [ ] **I am changing the color of a language**
+- [ ] **I am changing the color associated with a language**
   <!-- Please ensure you have gathered agreement from the wider language community _before_ opening this PR -->
   - [ ] I have obtained agreement from the wider language community on this color change.
     - [URL to public discussion]
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,10 +1,11 @@
-<!--- Briefly describe what you're changing. -->
+<!--- Briefly describe what you're changing in the pull request title. -->
 
 ## Description
 <!--- If necessary, go into depth of what this pull request is doing. -->
 
 ## Checklist:
 <!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- Feel free to remove whole sections, not points within the sections, that do not apply -->
 <!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
 - [ ] **I am associating a language with a new file extension.**
   - [ ] The new extension is used in hundreds of repositories on GitHub.com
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -157,6 +157,18 @@ If problems are found, please report these problems to the grammar maintainer as
...
+The colors were often chosen based on the colors used by the language at the time and since then users will have become familiar with those colors as they appear on GitHub.com.
+If you would like to change the color of a language, we ask that you propose your suggested color change to the wider community for your language to gain consensus before submitting a pull request.
+Please do this in a community forum or repository used and known by the wider community of that language, not the Linguist repository.
+
+Once you've received consensus that the community is happy with your proposed color change, please feel free to open a PR making the change and link to the public discussion where this was agreed by the community.
+If there are official branding guidelines to support the colour choice, please link to those too.
+
+Please note that Linguist currently implements a [color proximity test][color-proximity-test] to ensure colors are sufficiently different from one another so you may not be able to use the precise color you want - reds and blues are really popular.
+As such, we recommend you test the color change locally before making your plea to the wider language community.
```
Alhadis: Before we promulgate this as a restriction, we should *really* wait for the final consensus as to whether or not the colour-proximity requirements will ever be axed in future.

We have no idea how long it might take to reach a consensus. Members of a language's community could spend weeks bikeshedding as to whether a compromise for their project's official branding should be taken, simply their shade of red was too close to 5-6 other shades of red. How bad would we (GitHub/Linguist) look if the colour restriction was suddenly abolished the day after they reached an agreement?

Obviously we can't expect it to take *that* long to decide on a hex colour value, but you know what people are like...

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,10 +1,11 @@
-<!--- Briefly describe what you're changing. -->
+<!--- Briefly describe what you're changing in the pull request title. -->
```
Alhadis: Nitpick: replace with

> Briefly describe your changes in the field above.

It's clear from context that it's a pull-request, and "title" is  less obvious than pointing to a form-field.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -41,10 +42,16 @@
...
-  
+
 - [ ] **I am updating a grammar submodule**
   <!-- That's not necessary, grammar submodules are updated automatically with each new release. -->
 
 - [ ] **I am adding new or changing current functionality**
   <!-- This includes modifying the vendor, documentation, and generated lists. -->
   - [ ] I have added or updated the tests for the new or changed functionality.
+
+- [ ] **I am changing the color of a language**
```
Alhadis: I'd use

> I am changing the colour **_associated with a language_**

Since language is a human concept with no baryonic form capable of absorbing or reflecting bands of the visible electromagnetic spectrum. =)

CONTRIBUTING.md
```diff
@@ -157,6 +157,18 @@ If problems are found, please report these problems to the grammar maintainer as
 
 Please then open a pull request for the updated grammar.
 
+## Changing the color of a language
```
Alhadis: See above, discarding smart-alecky remark.

CONTRIBUTING.md
lildude: > Before we promulgate this as a restriction, we should _really_ wait for the final consensus as to whether or not the colour-proximity requirements will ever be axed in future.

We may be waiting a while. I've given my issue a nudge to see if I can get a quick response, but it's in a long list of other higher priority issues. The best thing to do is assume the status quo for now, we can easily remove the note when we remove the test.

lildude:
> It took all my effort not to review this with only `s/color/colour/`, but I managed.

😆  It really hurt to write "color" so many times but it's GitHub's preferred version of English so we gotta use it.

CONTRIBUTING.md
Alhadis: Fair enough, then.

I might be getting anxious for the outcome because there's a lot hanging on the removal of proximity requirements. We can *finally* degroup the languages touched on in #4291, add official colours for data/prose formats (which matters now that we have `linguist-detectable`), and a whole heap of other nags...

PR #4529, opened 2019-05-15
Add ".graphqls" as a GraphQL file extension

steebchen:
## Description

Often, developers use .graphqls specifically for GraphQL schema definitions to separate them from GraphQL queries, which are put into .graphql files.

It's easy to know what's in a file by just looking at its extension. This is especially useful in projects where both schema and queries are used. For example, one could have .graphql files for queries which tests their schema defined in .graphqls files. Or in a frontend project where an apollo client schema is defined in a .graphqls file while using .graphql files to define the queries.

## Checklist:

      - https://github.com/search?l=&p=100&q=type+extension%3Agraphqls&type=Code
    - Sample source(s):
      - https://github.com/magento/magento2/blob/779f39c71e3b801c9a9e8650f2acd05889643daf/app/code/Magento/CatalogGraphQl/etc/schema.graphqls
      - https://github.com/digiaonline/graphql-php/blob/f9b538f074f961c74bfe941106b7a1da9630e404/tests/Functional/Language/schema-kitchen-sink.graphqls
    - Sample license(s): 
      - Open Software License ("OSL") (Magento)
      - MIT (graphql-php)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1772,6 +1772,7 @@ GraphQL:
   type: data
   extensions:
   - ".graphql"
+  - ".graphqls"
   - ".gql"
   tm_scope: source.graphql
   ace_mode: text
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1772,8 +1772,8 @@ GraphQL:
   type: data
   extensions:
   - ".graphql"
-  - ".graphqls"
   - ".gql"
+  - ".graphqls"
   tm_scope: source.graphql
   ace_mode: text
   language_id: 139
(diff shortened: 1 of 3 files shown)


lib/linguist/languages.yml
```diff
@@ -1772,6 +1772,7 @@ GraphQL:
   type: data
   extensions:
   - ".graphql"
+  - ".graphqls"
```
pchaigno: Extensions should be sorted, except for the first one, so `.graphqls` should come last.

steebchen:
About the sample file, do I just put the schema-kitchen-sink.graphqls from https://github.com/digiaonline/graphql-php in the sample folder (MIT license is probably best)? Should I add a comment in the file for illustration?

pchaigno:
You need to add at least one, you can add both. Both licenses look okay. It's better if you don't modify the file.

steebchen:
I added both schema files and the check passed.

PR #4531, opened 2019-05-16
Update dependencies: use mini mime instead of mime type

ksolo:
## Description
Part of dependency clean up for gh/gh; working moving to a single mime library over having 2 installed. Linguist is the only dependency requiring `mime-types`, everything else is using `mini_mime`.


lib/linguist/VERSION
```diff
@@ -1 +1 @@
-7.4.0
+7.4.1
```
lildude: @ksolo if you're going to make the next release yourself, the next version _will_ include new languages (they've already been merged) so you'll need to use 7.5.0. 

Whilst you're at it, it would be good to update all the grammars too.

https://github.com/github/linguist/blob/master/CONTRIBUTING.md#releasing has all the details on the procedure I normally follow when making a release.

If you'd prefer I make the release, skip changing the version in this PR and I'll do it as part of the release I plan to make after Satellite (when the freeze on gh/gh has been lifted).

lildude:
I'm planning a new Linguist release this week @ksolo so if you're ok with it, I'm happy to remove the version bump in this PR and include this in my release.

ksolo:
@lildude Sorry I didn't see this earlier. Yes, please include it in the new release.

Thank you!

lildude:
Great! Thanks @ksolo. Can you please remove the version bump from this PR and I think we're all set.

ksolo:
@lildude done and done! Just waiting on CI now

Issue #2965, opened 2016-04-22
Syntax highlighting bug


henrywright:
In one of my .scss files I have stored on GitHub, the @ symbol is being highlighted red, but I see no reason for it to be an error. See here for an example:

https://github.com/6ty8/6ty8.github.io/blob/master/_sass/_footer.scss

Is this a Linguist issue?

pchaigno:
> Is this a Linguist issue?

No, it's not. Linguist only selects the grammar to use for highlighting. For SCSS, GitHub uses [MarioRicalde/SCSS.tmbundle](https://github.com/MarioRicalde/SCSS.tmbundle/tree/49a74571e7346b082016168fd702d41d1d319727). You should open an issue there if there isn't already one ;)

henrywright:
Thanks @pchaigno, I've opened an issue upstream. See [#207](https://github.com/MarioRicalde/SCSS.tmbundle/issues/207).

pchaigno:
Perfect! When the bug is fixed upstream, the dependency will be automatically updated with the next release of Linguist ;)

arfon:
Closing as this is an upstream issue.

aaronlademann-wf:
@arfon is there any concern that the upstream dependency being used... hasn't been updated in over 2 years?  What are the odds of the upstream dependency actually fixing it?

pchaigno:
@aaronlademann-wf I just pinged the actual maintainer of the project (that sometimes works). If he doesn't answer in the next week, please open an issue and we'll try to find a replacement grammar.

PR #4535, opened 2019-05-22
Use atom's language-sass grammar for SASS and SCSS

jonmast:
Switch to using https://github.com/atom/language-sass grammars for both SASS and SCSS languages.

## Description
SCCS.tmbundle is badly out of date and [no longer maintained](https://github.com/MarioRicalde/SCSS.tmbundle/issues/217). https://github.com/github/linguist/issues/2965 discussed switching to a different grammar, but it was never followed through.

This PR switches to [language-sass](https://github.com/atom/language-sass) instead since it's used by both atom and vscode, and should be better maintained in the future.

### Errors
There were a couple of errors associated with this change:
- The new sassdoc grammar references `source.html` which doesn't exist. Obviously we do have a grammar for HTML, I'm not sure if there's a way to have this use that?
- The scss grammar is now `source.css.scss` instead of `source.css`, which is included in slim. I've reversed the alias in `compiler/data.go` which I think should address this, but it seems like the test scripts use a docker image locally so I'm not sure how to test it.

## Checklist:

  - Old: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2FMarioRicalde%2FSCSS.tmbundle%2Fmaster%2FSyntaxes%2FSCSS.tmLanguage&grammar_text=&code_source=from-text&code_url=&code=.foo+%7B%0D%0A++%5Bbar%3D%22baz%22%5D+%7B%0D%0A++%7D%0D%0A%7D
  - New: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fatom%2Flanguage-sass%2Fmaster%2Fgrammars%2Fscss.cson&grammar_text=&code_source=from-text&code_url=&code=.foo+%7B%0D%0A++%5Bbar%3D%22baz%22%5D+%7B%0D%0A++%7D%0D%0A%7D


jonmast:
Looks like CI is failing due to the license. It looks like it's just two MIT licenses concatenated together, but it gets detected as "other".

vendor/licenses/grammar/language-sass.txt
```diff
@@ -0,0 +1,53 @@
+---
+type: grammar
+name: language-sass
+version: d07f9e7bebac73a2c6a0dceae4c099ea6d60fd4c
+license: other
```
lildude: > Looks like CI is failing due to the license. It looks like it's just two MIT licenses concatenated together, but it gets detected as "other".

Yeah, these Atom/Textmate licenses always trip things up and need manual reviewing. Consider this my review for the license and feel free to change "other" to "mit".

jonmast: Done, thanks for the review @lildude.

PR #4538, opened 2019-05-28
Support for Altium Designer Files

sciguy14:
This pull request adds support for Altium Designer files. Altium Software is used to design printed rigid and flexible circuit boards. More info about Altium: https://www.altium.com/altium-designer/

## Description

Altium uses many file types, but I've included the ones that can be represented in an ASCII format in this pull request. Many others are binary files only.

## Checklist:
      
      -  [.OutJob Files](https://github.com/search?q=extension%3AOutJob+NOT+nothack&type=Code)
      -  [.PcbDoc Files](https://github.com/search?q=extension%3APcbDoc+NOT+nothack&type=Code)
      -  [.PrjPCB Files](https://github.com/search?q=extension%3APrjPCB+NOT+nothack&type=Code)
      -  [.SchDoc Files](https://github.com/search?q=extension%3ASchDoc+NOT+nothack&type=Code)
    - Sample source(s): I made the simplest possible PCB design with Altium Designer 19 and am providing the resulting sample files in this PR.
    - Sample license(s): I hereby release these sample files under a permissive license.
  
_Question:_ It wasn't clear to me from the Contributing Guidelines - How does color selection work? I added a color to the languages.yml file, but running `bundle exec rake test` complained when a color was specified. The color used on all of Altium's branding is: #a5915f


lib/linguist/languages.yml
```diff
@@ -216,6 +216,20 @@ Alpine Abuild:
...
+  type: data
+  aliases:
+  - altium
+  extensions:
+  - ".OutJob"
+  - ".PcbDoc"
+  - ".PrjPCB"
+  - ".SchDoc"
+  interpreters:
+  - Altium Designer
```
pchaigno: Can these files really have [shebangs](https://en.wikipedia.org/wiki/Shebang_(Unix))?

sciguy14: Good catch - they can't. I've removed the "interpreters" parameter for Altium and updated the PR. It wasn't clear to me which parameters were required and which were optional, so I took a stab at filling in a reasonable value there without realizing I could just omit it.

PR #4547, opened 2019-06-05
Release v7.5.0

lildude:
Time for a new release.  Standard procedure of adding new languages etc.


lildude:

olafurpg:
It seems like syntax highlighting for Scala on GitHub.com has regressed. I'm unable to reproduce the same highlighting in our VS Code extension, which uses the same `Scala.tmLanguage.json` file. Is there any documentation on how debug this situation?

Context: we recently changed from a `Scala.tmLanguage` file to `Scala.tmLanguage.json`, which is autogenerated by a script written in TypeScript (PR: https://github.com/scala/vscode-scala-syntax/pull/20). We validated the fidelity of the translation by re-generating the original `Scala.tmLanguage` file from the new `Scala.tmLanguage.json` file, they were identical excluding a few reordered fields and comments (diff: https://www.diffnow.com/report/0z8ht).

lildude:
> Is there any documentation on how debug this situation?

Unfortunately not, but you might be able to work things out by looking at the grammar compiler we use at https://github.com/github/linguist/tree/master/tools/grammars and the resulting JSON file (it's included in the `linguist-grammars.tar.gz` added to each release - [v7.5.0](https://github.com/github/linguist/releases/download/v7.5.0/linguist-grammars.tar.gz) vs [v7.4.0](https://github.com/github/linguist/releases/download/v7.4.0/linguist-grammars.tar.gz)) to see what we do and use.

I did notice the compiler has recently started reporting an error with the grammar:

- [ ] repository `vendor/grammars/vscode-scala-syntax` (from https://github.com/scala/vscode-scala-syntax) (1 errors)
    - [ ] Unknown keys in grammar: `source.scala` (in `syntaxes/Scala.tmLanguage.json`) contains invalid keys (`$schema`)

... but wasn't too concerned with it as it shouldn't affect the syntax highlighting as the key isn't used. 

As an aside, this is the first time I've ever seen a `$schema` key in a grammar too. Is this something VS Code specific or new to TextMate compatible grammars? If so, is there some supporting documentation for it?

olafurpg:
Thank you for your help! I ran a diff on `source.scala.json` between v7.4 and v7.5 and got the following output https://gist.github.com/olafurpg/125fc86cfb4d6a3728285411a5bdbcc2 Could it be that the regexp engine used by VS Code is different from the regexp engine used by GitHub? cc/ @PanAeon

While we investigate, would it be possible to revert the vscode-scala-syntax submodule upgrade and cut a new linguist release? It may take a while to track down the root cause of this regression.

> As an aside, this is the first time I've ever seen a $schema key in a grammar too. Is this something VS Code specific or new to TextMate compatible grammars? If so, is there some supporting documentation for it?

I don't know much about the `$schema` field, it was introduced in https://github.com/scala/vscode-scala-syntax/pull/20. I believe it's used by editors to automatically provide completions and tooltip documentation when you open a JSON file.

lildude:
> Could it be that the regexp engine used by VS Code is different from the regexp engine used by GitHub?

They are different. GitHub uses PCRE whilst VS Code uses Oniguruma and I know there are a few expressions in oniguruma that aren't supported by PCRE, however the grammar compiler should pick those up and report them.

> While we investigate, would it be possible to revert the vscode-scala-syntax submodule upgrade and cut a new linguist release? It may take a while to track down the root cause of this regression.

Yes, sure.

olafurpg:
Thank you! I'll try to see what's going on here. Maybe we can add linguist to our test suite somehow to ensure the highlighting remains identical for VS Code and GitHub.

lildude:
You can also use [Lightshow](https://github-lightshow.herokuapp.com/) to check how the grammar will behave on GitHub.com.

PanAeon:
@lildude thank you for reverting the changes. The `$schema` field is json specific. It enables automatic validation in vscode, I'm not sure about the other editors.

The issue with scala grammar has boiled down to the difference between pcre and oniguruma regex engines: while it's valid in oniguruma to use class union like `[...[..]..]`, it is not valid in pcre.

Do you mind if I add a mention about the PCRE in the `CONTRIBUTING.md`? It references textmate's documentation and this threw me off a little bit since textmate is using oniguruma regex.

lildude:
> Do you mind if I add a mention about the PCRE in the `CONTRIBUTING.md`? It references textmate's documentation and this threw me off a little bit since textmate is using oniguruma regex.

Good idea. Go for it.

olafurpg:
The regressions for Scala highlighting should be fixed now thanks to https://github.com/scala/vscode-scala-syntax/pull/34 by @PanAeon. It should be safe to upgrade the vscode-scala-syntax submodule again.

lildude:
>  It should be safe to upgrade the vscode-scala-syntax submodule again.

Thanks. This will happen the next time I make a release.

PR #4550, opened 2019-06-06
Release v7.5.1

lildude:
Small point release to go back to the previous version of the Scala grammar as the newer one doesn't highlight correctly as advised in https://github.com/github/linguist/pull/4547#issuecomment-499467939


lildude:

PR #4584, opened 2019-07-19
Fix BibTeX highlighting and use LaTeX grammar for TeX files

Alhadis:
## Description
This PR adds a separate language entry for BibTeX to fix the syntax highlighting of `.bib` files. In addition, it enhances the highlighting of TeX files by switching to the [LaTeX grammar](https://github.com/textmate/latex.tmbundle/blob/master/Syntaxes/LaTeX.plist) instead of the [plain TeX one](https://github.com/textmate/latex.tmbundle/blob/master/Syntaxes/TeX.plist).

References: #3679
/cc @ruffsl, @blegat

## Checklist:
- [x] **I am adding a new language.**
	- [x] **The extension of the new language is used in hundreds of repositories.**
	- [x] **I have included a real-world usage sample.**  
	Assembled from BibTeX samples posted by @blegat in #3679, as well from [Wikipedia's BibTeX page](https://en.wikipedia.org/wiki/BibTeX#Bibliographic_information_file). Given that BibTeX files are just citations, I'd say this qualifies as "real-world" usage.

- [x] **I am changing the source of a syntax highlighting grammar**
[...]

PR #4585, opened 2019-07-19
Require all grammars to have a `tm_scope` defined

Alhadis:
The current behaviour of Linguist is to infer a missing `tm_scope` from a language's lowercased name instead. This is utterly unintuitive and offers very little benefit.

@lildude and @pchaigno [both agreed](https://github.com/github/linguist/pull/4584#issuecomment-513199027) that a `tm_scope` should be explicitly declared for each `languages.yml` entry.

## Checklist:

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -273,17 +273,7 @@ module Linguist
       # Set aliases
       @aliases = [default_alias] + (attributes[:aliases] || [])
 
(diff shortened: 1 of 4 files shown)


Alhadis:
@lildude I also added a test to help pick up on unrecognised fields in `languages.yml`. I made a few typos while editing that which I didn't pick up, and something like this is likely to slip through review when many similar edits are being submitted at once (like now).

PR #4588, opened 2019-07-23
trivial: add .hack extension for Hack language

jjergus:
`.hack` was agreed on as the main extension for Hack files in the future, and is already used by standard Hack libraries ([example](https://github.com/hhvm/hacktest/tree/master/src)) and official documentation ([example](https://docs.hhvm.com/hack/getting-started/getting-started#your-first-hack-program__3.-write-your-first-hack-program)).

This should fix syntax highlighting for pull requests such as https://github.com/hhvm/hacktest/pull/84/files

For context, `.php` is deprecated for Hack files since the languages are no longer compatible, and `.hh` is discouraged because it can cause confusion with C++ header files.

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Ahack+function+NOT+nothack
    - Sample source(s):
      - https://github.com/hhvm/hacktest/tree/master/src
      - https://docs.hhvm.com/hack/getting-started/getting-started#your-first-hack-program__3.-write-your-first-hack-program

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2003,6 +2003,7 @@ Hack:
   codemirror_mode: php
   codemirror_mime_type: application/x-httpd-php
   extensions:
+  - ".hack"
   - ".hh"
   - ".php"
   tm_scope: source.hack
--- a/samples/Hack/first.hack
+++ b/samples/Hack/first.hack
@@ -1,0 +1,17 @@
+// from https://docs.hhvm.com/hack/getting-started/getting-started
+// note: no <?hh header needed for .hack files
+
+namespace Hack\GettingStarted\MyFirstProgram;
+
+<<__EntryPoint>>
+function main(): noreturn{
+  echo "Welcome to Hack!\n\n";
+
+  \printf("Table of Squares\n" .
+          "----------------\n");
+  for ($i = -5; $i <= 5; ++$i) {
+    \printf("  %2d        %2d  \n", $i, $i * $i);
+  }
+  \printf("----------------\n");
+  exit(0);
+}


lib/linguist/languages.yml
```diff
@@ -2003,6 +2003,7 @@ Hack:
   codemirror_mode: php
   codemirror_mime_type: application/x-httpd-php
   extensions:
+  - ".hack"
```
ukos-git: would be great to have at least one `*.hack` file in the [samples section](https://github.com/github/linguist/tree/master/samples/Hack) if this is the standard extension. Instead of linking to external sites, you could then add directly a lightshow example for these sample files.

lib/linguist/languages.yml
```diff
@@ -2003,6 +2003,7 @@ Hack:
...
   - ".hh"
   - ".php"
```
ukos-git: if this is a deprecated extension and not working anymore anyway with hack, then remove it because we are better of having the extension assigned to plain php.

jjergus: It's still very commonly used: https://github.com/search?q=extension%3Aphp+async+await&type=Code

So we should probably not remove it in the near future.

lib/linguist/languages.yml
jjergus: Good idea, I didn't know there was a samples section in the repository. I'll add some.

jjergus:
Sample file added.

lib/linguist/languages.yml
Alhadis: @jjergus We don't retroactively remove extensions due to declining popularity. If an extension was proven to have enough real-world distribution at the time it was added, that's enough for us.

jjergus:
Thanks for the review and suggestions!

PR #4575, opened 2019-07-07
Support for Coccinelle's SmPL language

pchaigno:
This pull request adds support for [the SmPL language](http://coccinelle.lip6.fr/docs/index.html) used by [the Coccinelle tool](http://coccinelle.lip6.fr) to write semantic patches.

## Description

I didn't find a grammar to highlight SmPL files.

## Checklist:
- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      - [`extension:cocci NOT fhdifhdfdggd`](https://github.com/search?q=extension%3Acocci+NOT+fhdifhdfdggd)
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source: [torvalds/linux](https://github.com/torvalds/linux/blob/fa1b5d09d0771247d407df89228b3902de8e2ce6/scripts/coccinelle/api/atomic_as_refcounter.cocci)
    - Sample license: GPLv2
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

PR #4601, opened 2019-08-12
Fix failing tests

Alhadis:
In #4575, the `aliases` field was erroneously entered as `alias`, a mistake which wasn't noticed until it was caught by a [test](https://github.com/github/linguist/pull/4585#issuecomment-513302506) I added in #4585 (for reasons exactly like this). Both PRs were merged earlier, revealing the error.


pchaigno:
Thanks @Alhadis!

PR #4604, opened 2019-08-13
Add pull_request to on events list for CI

lildude:
## Description

When switching from Travis to GitHub Actions, I missed out that we need to trigger builds for the `pull_request` event else CI won't run for PRs submitted from forks.

This isn't very clear from the docs or the available examples.

I'm also taking this opportunity to use more descriptive names so the checks look better.

Before:

Screenshot 2019-08-13 at 10 57 13

After:

Screenshot 2019-08-13 at 10 59 35

I'll update the list of required status checks once this PR has been merged.

_[Template removed as it is not relevant]_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,10 @@
-name: CI
+name: Run Tests
 
 on: [push, pull_request]
 
 jobs:
-  build:
+  run:
+    name: Ruby ${{ matrix.ruby }}
     runs-on: ubuntu-latest
     strategy:
       matrix:


lildude:
> I'll update the list of required status checks once this PR has been merged.

No I won't. I can do it now they've passed 😄

lildude:
And yes, it is known that certain events that happen when modifying a pull request, eg labelling, requesting reviewers etc, cause CI to run again even if the actual code hasn't changed. The ability to filter those is coming 🔜

Alhadis:
Would it be possible to use GitHub Actions to automate some of our scripts? Stuff like `script/add‑grammar` and `script/list‑grammars` still need to be run manually, despite requiring little or no user intervention.

I know absolutely nothing about GitHub Actions, so forgive me if this is a dumb question.

lildude:
> Would it be possible to use GitHub Actions to automate some of our scripts? Stuff like `script/add‑grammar` and `script/list‑grammars` still needs to be run manually (and in a local checkout, at that), despite requiring little or no intervention from the user.

Most definitely, though I'm not sure how much success we'd have with automating `script/add‑grammar` right now as we've still got quite a few outstanding issues (https://github.com/github/linguist/issues/3924). Once we've got those all sorted, I don't see why we couldn't do this automatically.

lildude:
Oooof. Looks like something has changed around the UTF-8 encoding on the underlying CI env since the tests on this PR last passed and merging. Digging into it now.

Alhadis:
Try setting `LANG=en.UTF-8` in the environment if it's missing. That's what solves a similar encoding issue I always get when running tests on OpenBSD.

lildude:
> Try setting `LANG=en.UTF-8` in the environment if it's missing. That's what solves a similar encoding issue I always get when running tests on OpenBSD.

Read my mind. It was the the first thing I was going to try as Docker has the same issue and really should have remembered this as I stumble upon it almost every time I touch docker 😄

Alhadis:
Heh. Since a Docker port doesn't exist on OpenBSD, I actually solved two problems at once by having a shell-script named `docker` in my $PATH:

~~~shell
#!/bin/sh
export LANG=en_AU.UTF-8
~~~

Not sure if this is the best or worst hack I've ever thought of (okay, definitely *not* the worst...). I can't export `LANG` in my dotfiles because it does screwy things with other locale-related variables (`LC_COLLATE`, etc)... 😞

melvisglez:
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index e8252244b6..5b11077efa 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,10 @@
-name: CI
+name: Run Tests
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
-  build:
+  run:
+    name: Ruby ${{ matrix.ruby }}
     runs-on: ubuntu-latest
     strategy:
       matrix:

PR #4610, opened 2019-08-15
Add MLIR grammar

jpienaar:
Add support for MLIR

## Description
Add support for [MLIR](https://github.com/tensorflow/mlir).

## Checklist:

      -  https://github.com/search?q=extension%3Amlir+NOT+nothack&type=Code
    - Sample source(s):
      - [const-fold.mlir](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/tests/const-fold.mlir)
      - [core-ops.mlir](https://github.com/tensorflow/mlir/blob/master/test/IR/core-ops.mlir)
      - [executor_to_control_dialect.mlir](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/tensorflow/tests/executor_to_control_dialect.mlir)
      - [sample.mlir](https://github.com/jpienaar/mlir-grammar/blob/master/sample.mlir)
  
    [core-ops](https://github-lightshow.herokuapp.com/?utf8=✓&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fjpienaar%2Fmlir-grammar%2Fmaster%2Fgrammars%2Fmlir.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Ftensorflow%2Fmlir%2Fmaster%2Ftest%2FIR%2Fcore-ops.mlir&code=)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -725,6 +725,9 @@
 [submodule "vendor/grammars/mercury-tmlanguage"]
 	path = vendor/grammars/mercury-tmlanguage
 	url = https://github.com/sebgod/mercury-tmlanguage
+[submodule "vendor/grammars/mlir-grammar"]
+	path = vendor/grammars/mlir-grammar
+	url = https://github.com/jpienaar/mlir-grammar
 [submodule "vendor/grammars/monkey"]
 	path = vendor/grammars/monkey
 	url = https://github.com/gingerbeardman/monkey.tmbundle
--- a/grammars.yml
+++ b/grammars.yml
@@ -632,6 +632,8 @@ vendor/grammars/mediawiki.tmbundle:
 - text.html.mediawiki
 vendor/grammars/mercury-tmlanguage:
 - source.mercury
+vendor/grammars/mlir-grammar:
+- source.mlir
 vendor/grammars/monkey:
 - source.monkey
 vendor/grammars/moonscript-tmbundle:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2926,6 +2926,13 @@ MAXScript:
   tm_scope: source.maxscript
   ace_mode: text
   language_id: 217
+MLIR:
+  type: programming
+  extensions:
+  - ".mlir"
+  tm_scope: source.mlir
+  ace_mode: text
+  language_id: 448253929
 MQL4:
   type: programming
   color: "#62A8D6"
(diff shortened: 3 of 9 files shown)


jpienaar:
@pchaigno Thanks! I combined it from multiple of the unit tests to get something a bit more interesting mix of types (all from the same repo and with same license). Would you prefer if I use one/more exact file instead?

jpienaar:
Updated template to point to exact file instead.

pchaigno:
> I combined it from multiple of the unit tests to get something a bit more interesting mix of types (all from the same repo and with same license).

Thanks for the explanation. Including a few files instead of a single one would have been fine too.

jpienaar:
> > I combined it from multiple of the unit tests to get something a bit more interesting mix of types (all from the same repo and with same license).
> 
> Thanks for the explanation. Including a few files instead of a single one would have been fine too.

Sounds good to me, added a few more (exact samples from real usage). Let me know if anything else is needed. Thanks

jpienaar:
> The test failure here is legit:
> 
> ```
> /home/runner/work/linguist/linguist/vendor/licenses/grammar/mlir-grammar.txt:
>   - cached license data out of date
> ```
> 
> You'll need to run `script/licensed` (which should update the `version` in the license) and commit the updated file. You might need to manually add back the license text though, so feel free to.

Updated, that did it thanks

jpienaar:
This seems to have stopped working (previously highlighted file no longer works), I don't see anything in the commit logs that would indicate a change though.

lildude:
I've taken a look into this as this appears to only affect languages or grammars added in v7.6.0.  Work is underway to improved the syntax highlighting service used by GitHub and I suspect it has missed the v7.6.0 update. I've opened an issue to bring this to the attention of the team responsible for the improvements.

I'll update when I know more or when things are working as expected again.

lildude:
> Work is underway to improved the syntax highlighting service used by GitHub and I suspect it has missed the v7.6.0 update.

All sorted. This was indeed the case.

jpienaar:
> > Work is underway to improved the syntax highlighting service used by GitHub and I suspect it has missed the v7.6.0 update.
> 
> All sorted. This was indeed the case.

Great, thanks!

PR #4625, opened 2019-08-26
Add some rules for TSQL and PLSQL

ktaranov:
Add some new rules for TSQL and PLSQL code.

## Checklist:

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -400,12 +400,12 @@ disambiguations:
     pattern: "(?i:(alter module)|(language sql)|(begin( NOT)+ atomic)|signal SQLSTATE '[0-9]+')"
   # Oracle
   - language: PLSQL
-    pattern: '(?i:\$\$PLSQL_|XMLTYPE|sysdate|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function)'
+    pattern: '(?i:\$\$PLSQL_|XMLTYPE|sysdate|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function|FROM DUAL)'
   # T-SQL
   - language: TSQL
     and:
     - negative_pattern: '(?i:IDENTIFIED|NUMBER|VARCHAR2|REPEAT|UNTIL|IMMEDIATE)'
-    - pattern: '(?i:(GO)|(@@)|(CREATE PROCEDURE)|BEGIN( TRY| CATCH)|OUTPUT( INSERTED)|IF|ELSE|IIF|CHOOSE|CURSOR|FETCH|DEALLOCATE|DECLARE)'
+    - pattern: '(?i:(GO)|(@@)|(CREATE PROCEDURE)|BEGIN( TRY| CATCH)|OUTPUT( INSERTED)|IF|ELSE|IIF|CHOOSE|CURSOR|FETCH|DEALLOCATE|DECLARE|ISNULL|GETDATE|SET NOCOUNT|datetime2|bit|datetimeoffset|sql_variant|uniqueidentifier|USE )'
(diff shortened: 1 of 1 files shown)


lib/linguist/heuristics.yml
```diff
@@ -400,12 +400,12 @@ disambiguations:
     pattern: "(?i:(alter module)|(language sql)|(begin( NOT)+ atomic)|signal SQLSTATE '[0-9]+')"
   # Oracle
   - language: PLSQL
-    pattern: '(?i:\$\$PLSQL_|XMLTYPE|sysdate|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function)'
+    pattern: '(?i:\$\$PLSQL_|XMLTYPE|sysdate|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function|FROM DUAL)'
```
lildude: The test failures are legitimate and suggest `FROM DUAL` is _not_ unique to PLSQL and can be used for "normal" SQL too. From a quick search, this appears correct as MySQL allows the use of it too: https://dev.mysql.com/doc/refman/8.0/en/select.html

lib/linguist/heuristics.yml
```diff
@@ -400,12 +400,12 @@ disambiguations:
...
   # T-SQL
   - language: TSQL
     and:
     - negative_pattern: '(?i:IDENTIFIED|NUMBER|VARCHAR2|REPEAT|UNTIL|IMMEDIATE)'
-    - pattern: '(?i:(GO)|(@@)|(CREATE PROCEDURE)|BEGIN( TRY| CATCH)|OUTPUT( INSERTED)|IF|ELSE|IIF|CHOOSE|CURSOR|FETCH|DEALLOCATE|DECLARE)'
+    - pattern: '(?i:(GO)|(@@)|(CREATE PROCEDURE)|BEGIN( TRY| CATCH)|OUTPUT( INSERTED)|IF|ELSE|IIF|CHOOSE|CURSOR|FETCH|DEALLOCATE|DECLARE|ISNULL|GETDATE|SET NOCOUNT|datetime2|bit|datetimeoffset|sql_variant|uniqueidentifier|USE )'
```
lildude: I don't think all of these are unique to TSQL. I've not tried very hard and I've already found `ISNULL` is [usable in MySQL](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#function_isnull).

ktaranov: `ISNULL` in SQL Server has 2 parameters, maybe change regex expression to handle this metric?
https://docs.microsoft.com/ru-RU/sql/t-sql/functions/isnull-transact-sql?view=aps-pdw-2016

lib/linguist/heuristics.yml
ktaranov: Agree with this.

lib/linguist/heuristics.yml
pchaigno: Detecting the number of parameters is not an easy task and the regular expression is already long. I'm not sure it would be worth it.

ktaranov:
Close this. But problem with correct determining tsql scripts is still exists, too many tsql repo defined as PLpgSQL language. See, for example, [this](https://github.com/olahallengren/sql-server-maintenance-solution) or [this](https://github.com/LowlyDBA/ExpressSQL).

I can create new pull request only for tsql regular expression:
`- pattern: '(?i:(GO)|(@@)|(CREATE PROCEDURE)|BEGIN( TRY| CATCH)|OUTPUT( INSERTED)|IF|ELSE|IIF|CHOOSE|CURSOR|FETCH|DEALLOCATE|DECLARE|ISNULL|GETDATE|SET NOCOUNT|datetime2|bit|datetimeoffset|sql_variant|uniqueidentifier|USE )'`

PR #4629, opened 2019-08-29
Modula 2 definition

idrougge:
Add ".def" file suffix for Modula-2 definition files (similar to C header files).
## Description

As with C, typical Modula-2 repositories contain almost as many DEF files as MOD files.
## Checklist:

      - https://github.com/search?q=%22definition+module%22+NOT+nothack+extension%3Adef&type=Code
    - Sample source(s):
      - https://github.com/axtens/res/blob/4e749b53aace81e5f08a9b7706b816e56857d90f/v2/RESERROR.DEF
    - Sample license(s): MIT
  - [-] I have included a change to the heuristics to distinguish my language from others using the same extension.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3173,6 +3173,7 @@ Modula-2:
   type: programming
   extensions:
   - ".mod"
+  - ".def"
   tm_scope: source.modula2
   ace_mode: text
   language_id: 234
--- a/samples/Modula-2/RESERROR.DEF
+++ b/samples/Modula-2/RESERROR.DEF
@@ -1,0 +1,17 @@
+DEFINITION MODULE RESerrors;
+
+TYPE
+        RESerrorcodes = (USERERROR, NOENVAR, NOMEMAVAIL,
+                         NOFILEFOUND, FILEEMPTY,
+                         PICKLIMITERROR
+                         );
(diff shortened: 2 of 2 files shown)


idrougge:
As far as I can tell, a definition file must include the words `DEFINITION MODULE ModuleName`, that should be a rather strict heuristic, shouldn't it?

lildude:
> As far as I can tell, a definition file must include the words `DEFINITION MODULE ModuleName`, that should be a rather strict heuristic, shouldn't it?

Yes, for the Modula 2 classification. However There are still loads of other languages using this extension that Linguist doesn't know about so the fallthrough to the classifier isn't going to do a very job of identifying the languages of those other files.

This is the same problem Solidity has with the `.sol` extension as you can see from all the PRs we've closed attempting to add this extension: https://github.com/github/linguist/pulls?q=is%3Apr+solidity+is%3Aclosed

idrougge:
So Linguist is greedy rather than lazy when classifying files, not giving up until it has found a fitting language?

lildude:
Yup... it essentially works like a funnel. All languages go in the top and we whittle it down using the various methods detailed in [How Linguist Works](https://github.com/github/linguist#how-linguist-works) until we get one language.

PR #4628, opened 2019-08-29
Fix generated_html failure when the meta line doesn't have a "content" attribute

lildude:
## Description
It has come to light that the changes in `generated_html()` added in https://github.com/github/linguist/pull/4574 don't correctly handle meta lines that contain `name="generated"` attribute, but not a `content="?"` attribute, for example:

```
  <meta name='generator' value='Ronn/v0.7.3 (http://github.com/rtomayko/ronn/tree/0.7.3)'>
``` 

... from https://github.com/tj/git-extras/blob/a56e14488e5cd713ad27370ec432fdd510353f53/man/git-alias.html#L5

This results in a 500 error (try clicking the first commit in this PR 😉) on GitHub.com with the following failure stack:

```
NoMethodError: undefined method `match' for nil:NilClass
    /home/runner/work/linguist/linguist/lib/linguist/generated.rb:624:in `block in generated_html?'
    /home/runner/work/linguist/linguist/lib/linguist/generated.rb:622:in `any?'
    /home/runner/work/linguist/linguist/lib/linguist/generated.rb:622:in `generated_html?'
    /home/runner/work/linguist/linguist/lib/linguist/generated.rb:89:in `generated?'
    /home/runner/work/linguist/linguist/lib/linguist/generated.rb:12:in `generated?'
    /home/runner/work/linguist/linguist/test/test_generated.rb:31:in `generated_loading_data'
    /home/runner/work/linguist/linguist/test/test_generated.rb:40:in `generated_fixture_loading_data'
    /home/runner/work/linguist/linguist/test/test_generated.rb:161:in `test_check_generated'
```

... as can be seen in the test failure I deliberately forced. See https://github.com/github/linguist/runs/206726836#step:5:68

This PR updates `generated_html()` to better handle the case where `content` doesn't exist, and also adds support for the `value` attribute and additionally marks those generated by [Ronn](http://github.com/rtomayko/ronn/) as generated too.

Now the catch: I can't include the two sample files I want to include in this PR as doing so causes a 500 error when attempting to create this pull request.

I can add them back now the PR has been created, but then the files can't be viewed in the "Files Changed" tab as we hit the same 500 error. You can't view the individual commits containing those files either. All of these are caused by the issue I'm trying to fix here.

So for the moment, I'm going to leave out the tests and fixture files and add them once this PR has been merged and a new release made and pushed to GitHub.com.

The fix can be verified locally as follows:

[...]

PR #4632, opened 2019-08-30
Add tests for generated_html changes in #4628

lildude:
As promised and prevented by the bug fixed by https://github.com/github/linguist/pull/4628, here are the tests for the changes introduced in that PR.


test/fixtures/HTML/no-content.html
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE html>
+<html>
+<meta name="generator" value="Some sick tool nobody's using yet" />
```
Alhadis: > Some sick tool

You can tell an Australian wrote this. 😁 Didn't even need `<html lang="en-AU">`.

PR #3112, opened 2016-07-12
Add support for Apollo Guidance Computer extension

chrislgarry:
I accidentallied the previous PR by trying to use the web interface to make changes...

PR #4637, opened 2019-09-01
Add common extensions to Motorola 68k Assembly

idrougge:
## Description
Common extensions for m68k assembly are `.asm`, `.s`, `.i`, `.inc`. The only currently registered extension `x68` is only used within one specific development environment AFAIK.

## Checklist:
	- [X] **The new extension is used in hundreds of repositories on GitHub.com**
		- [`.asm`](https://github.com/search?q=moveq+extension%3Aasm+NOT+nothack&type=Code)
		- [`.i`](https://github.com/search?q=moveq+extension%3Ai+NOT+nothack&type=Code)
		- [`.inc`](https://github.com/search?q=movem+extension%3Ainc+NOT+nothack&type=Code)
		- [`.s`](https://github.com/search?q=movem+extension%3As+NOT+nothack&type=Code)

	- [X] **I have included real-world usage samples for all extensions added in this PR:**
		- Motorola 68K Assembly
			- [`bls_routines.inc`](https://github.com/retro16/blastsdk/blob/572bd8489e6a2c1cc638120aa62241c99badc2e5/include/bls_routines.inc): [MIT](https://github.com/retro16/blastsdk/blob/572bd8489e6a2c1cc638120aa62241c99badc2e5/LICENSE)
			- [`cpu.s`](https://github.com/flynn-nrg/tornado-amiga/blob/master/src/amiga/cpu.s): [Public Domain](https://github.com/flynn-nrg/tornado-amiga/blob/ac261e2aad6f15c26d871a24823e7f42230ba2f9/src/amiga/cpu.s#L2) / [BSD 2-Clause](https://github.com/flynn-nrg/tornado-amiga/blob/477b5d754537e114e5aa115ea93e5723961404e3/LICENSE)
			- [`iff_ilbm.i`](https://github.com/smart-fun/Amiga/blob/2a6e28c2e765571ad907449d6a6d7c8eca578b9f/AsmLib/iff_ilbm.i): [Apache 2.0](https://github.com/smart-fun/Amiga/blob/2a6e28c2e765571ad907449d6a6d7c8eca578b9f/LICENSE)
			- [`rom_testbench.asm`](https://github.com/fredrequin/j68_cpu/blob/538badf8def66dc900947be665d0eac6f9ff833e/roms/rom_testbench.asm): [ISC](https://github.com/fredrequin/j68_cpu/blob/538badf8def66dc900947be665d0eac6f9ff833e/README.md)
			- [`system.s`](https://github.com/flynn-nrg/tornado-amiga/blob/master/src/amiga/system.s): [BSD 2-Clause](https://github.com/flynn-nrg/tornado-amiga/blob/477b5d754537e114e5aa115ea93e5723961404e3/LICENSE)
		
		- “General” Assembly
			- [`3D_PRG.I`](https://github.com/malbanGit/Vide/blob//0e0a35a5dc06fc9a0ab2c9b0f6f7793fbc4b13fd/codelib/CompleteReleases/Malban/VRelease/3d/3D_PRG.I): [Public Domain](https://github.com/malbanGit/Vide/blob/0e0a35a5dc06fc9a0ab2c9b0f6f7793fbc4b13fd/codelib/CompleteReleases/Malban/VRelease/README.txt#L5)
			- [`A8514.I`](https://github.com/AnimatorPro/Animator-Pro/blob//b63d1d862ec6ab45db776e69d58c970e7000e030/Pristine-Pro/FLIC386P/LIBSRC/8514/A8514.I): [BSD 3-Clause](https://github.com/AnimatorPro/Animator-Pro/blob/master/LICENSE)
			- [`audio.i`](https://github.com/kervinck/gigatron-rom/blob//b0471f594ac33017c3c5af023bfc1bcca1a3c9f8/Contrib/at67/gasm/tetronis/audio.i): [BSD 2-Clause](https://github.com/kervinck/gigatron-rom/blob/b0471f594ac33017c3c5af023bfc1bcca1a3c9f8/Contrib/at67/LICENSE)

	- [X] **I have included a change to the heuristics to distinguish my language from others using the same extension.**

	- [X] **The extension of the new language is used in hundreds of repositories on GitHub.com.**
		- [`.i`](https://github.com/Alhadis/Silos/tree/60496ae7220110c59817c8fedf3f8868b0b67bdd/files/SWIG)
	- [X] **I have included a real-world usage sample for all extensions added in this PR:**
		- [`CGAL_AABB_tree.i`](https://github.com/jon-daeh/cgal-swig-bindings-maven/blob//197aa7da78aa8e2a30931b666e1bc6e45e7c1a42/SWIG_CGAL/AABB_tree/CGAL_AABB_tree.i): [Boost License](https://en.wikipedia.org/wiki/Boost_(C%2B%2B_libraries)#License)
		- [`dictionary.i`](https://github.com/InnerPortal/mds-utils/blob//2ede32e0590ac9469cdd976a33cd408961f138db/include/mds_utils/python/dictionary.i): [MIT](https://github.com/InnerPortal/mds-utils/blob/2ede32e0590ac9469cdd976a33cd408961f138db/include/mds_utils/python/dictionary.i#L5)
		- [`gauss.i`](https://github.com/aptech/gsoop/blob//c79dae019b86274ffa202b29e790bf5a97972a7c/gauss.i): [MIT](https://github.com/aptech/gsoop/blob/c79dae019b86274ffa202b29e790bf5a97972a7c/LICENSE)]
	- [X] **I have updated the heuristics to distinguish my language from others using the same extension.**


lib/linguist/languages.yml
```diff
@@ -3224,11 +3224,16 @@ MoonScript:
...
   type: programming
-  group: Assembly
+  aliases:
+  - m68k
   extensions:
-  - ".X68"
+  - ".asm"
+  - ".i"
+  - ".inc"
+  - ".s"
```
Alhadis: Are there any registers or opcodes unique to Motorola we can use to disambiguate assembly files with?

We're *definitely* going to need some heuristics for `.asm` and `.inc`. The latter of which is particularly important because it sees very general use across a range of unrelated) languages…

lib/linguist/languages.yml
```diff
@@ -3224,11 +3224,16 @@ MoonScript:
...
+  - ".x68"
   tm_scope: source.m68k
-  ace_mode: assembly_x86
+  ace_mode: text
```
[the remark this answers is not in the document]
idrougge: Even when the languages have almost nothing in common?

lib/linguist/languages.yml
idrougge: M68k assembly is easily distinguished from other assembly languages, both by registers and opcodes. 

How would such a heuristic look?

lib/linguist/languages.yml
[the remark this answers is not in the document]
Alhadis: Well, [I'll let you see for yourself](https://ace.c9.io/build/kitchen-sink.html). In the sidebar, select `Assembly x86` from the `Mode` menu, and compare its sample's highlighting with that of Motorola assembly. Not perfect, but far from terrible.

A lot of languages have imperfect grammars or editor modes assigned, simply because it's a passable approximation. Motorola has comments and labels highlighted, and that's really a step up from absolutely no highlighting whatsoever...

lib/linguist/languages.yml
Alhadis: A regular expression; I'm happy to write it for you, provided you give me the names of substrings guaranteed (or highly unlikely) to appear in the source code of any other assembler language.

lib/linguist/languages.yml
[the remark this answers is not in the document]
Alhadis: It's worth pointing out that Intel's is the only assembly dialect supported by ACE, meaning it may as well have the `x86` part removed and just serve as an umbrella for the editor's assembly support.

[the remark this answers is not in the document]
idrougge: Yes, an ACE syntax only supporting the most generic assembly directives like includes, sections, equations and comments would be most useful.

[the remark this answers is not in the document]
Alhadis: Yes, a sarcastic quip about keyword recognition from a browser-based code editor will certainly get this PR approved and merged sooner.  

Are we done here?

lib/linguist/languages.yml
idrougge: `(?im:moveq\b.*?d\d|move\.[bwl]\s+.*\b[ad]\d|movem\.[bwl]\b|btst\b|dbra\b)`

lib/linguist/languages.yml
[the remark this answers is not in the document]
Alhadis: Heh. Well, I'm used to seeing assembly incorrectly highlighted ([except when I'm not](https://github.com/github/linguist/pull/3112#issuecomment-232148398)), so I naturally can't see basic code-editors grokking full instruction sets for every known dialect of assembler. 😜

Anywho... if you could revert this line back to `ace_mode: assembly_x86` the way it was, that'd be copacetic. 👍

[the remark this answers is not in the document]
idrougge: Done.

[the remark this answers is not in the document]
pchaigno: > Yes, an ACE syntax only supporting the most generic assembly directives like includes, sections, equations and comments would be most useful.

If the `x86` part is breaking syntax highlighting for other assembly dialects, that may make sense. In any case, it would require a pull request to Ace, not here.

In the meantime, unless there are cases for which the `assembly_x86` Ace mode is making it worse, I agree with @Alhadis and think it would be a good fallback compared to no highlighting at all.

lib/linguist/languages.yml
Alhadis: Would it be reasonable to limit the `moveq` heuristic to match two registers (one address, one data)? From what I hear, 68k is unique for differentiating between the two.

If so, we could try this:

~~~regexp
(?xi)
	# Mnemonic
	\b moveq (\.l)? \s+
	
	# Address
	\#( \$ -? [0-9a-f]{1,3}
	  |    %  [0-1]{1,8}
	  |    -? [0-9]{1,3}
	  )
	, \s*
	
	# Register
	d[0-7] \b
~~~

When writing heuristics, it's best to be as specific as possible; anything which doesn't match is passed down to the (less accurate) classification techniques.

idrougge: If I read that regex correctly, it looks perfect. I'm just not as good at writing them.

Alhadis: Credit for this expression belongs to @zerkman, since it's taken from the [`language-m68k`](https://github.com/zerkman/language-m68k/blob/f7f7bf11c322976bd9fd00a0d94a1cfc3d1c6c8f/grammars/m68k.cson#L144) grammar we're using to highlight 68k on GitHub (I did clean it up and remove some redundant syntax for clarity).

I've amended the other parts of that expression to use other bits of that grammar, bringing us down to:

~~~regexp
(?xim:
	# Mnemonic
	\b moveq (\.l)? \s+
	
	# Address
	\#( \$ -? [0-9a-f]{1,3}
	  |    %  [0-1]{1,8}
	  |    -? [0-9]{1,3}
	  )
	, \s*
	
	# Register
	d[0-7] \b
	
	| ^ \s* move     (\.[bwl])? \s+ (sr|usp), \s* [^\s]+
	| ^ \s* movem     \.[bwl]  \b
	| ^ \s* move[mp] (\.[wl])? \b
	| ^ \s* btst  \b
	| ^ \s* dbra  \b
)
~~~

Notice that I've anchored the remaining parts to match at the beginning of a line (with or without indentation). This reduces the risk of incorrectly matching part of a comment in an unrelated file. For the same reason, you'll notice I avoid using wildcards when possible (`.*`).

Alhadis: @idrougge If the above revisions look good to you, then the changes to make to `heuristics.yml` are below.

I'll still need to test them thoroughly on my end, as well as investigate any possible formats using the `.i` extension that we've not registered yet.

<details><summary><b>Click to show diff</b></summary>

~~~diff
--- heuristics.yml	2019-10-03 21:45:48.000000000 +1000
+++ heuristics.yml	2019-10-03 22:08:15.000000000 +1000
@@ -49,8 +49,12 @@
   rules:
   - language: ActionScript
     pattern: '^\s*(package\s+[a-z0-9_\.]+|import\s+[a-zA-Z0-9_\.]+;|class\s+[A-Za-z0-9_]+\s+extends\s+[A-Za-z0-9_]+)'
   - language: AngelScript
+- extensions: ['.asm']
+  rules:
+  - language: Motorola 68K Assembly
+    named_pattern: m68k
 - extensions: ['.asc']
   rules:
   - language: Public Key
     pattern: '^(----[- ]BEGIN|ssh-(rsa|dss)) '
@@ -191,8 +195,10 @@
     pattern: '\A\s*[{\[]'
   - language: Slice
 - extensions: ['.inc']
   rules:
+  - language: Motorola 68K Assembly
+    named_pattern: m68k
   - language: PHP
     pattern: '^<\?(?:php)?'
   - language: SourcePawn
     pattern: '^public\s+(?:SharedPlugin(?:\s+|:)__pl_\w+\s*=(?:\s*{)?|(?:void\s+)?__pl_\w+_SetNTVOptional\(\)(?:\s*{)?)'
@@ -383,8 +389,12 @@
   - language: Rust
     pattern: '^(use |fn |mod |pub |macro_rules|impl|#!?\[)'
   - language: RenderScript
+- extensions: ['.s']
+  rules:
+  - language: Motorola 68K Assembly
+    named_pattern: m68k
 - extensions: ['.sc']
   rules:
   - language: SuperCollider
     pattern: '(?i:\^(this|super)\.|^\s*~\w+\s*=\.)'
@@ -486,7 +496,14 @@
   - 'std::\w+'
   fortran: '^(?i:[c*][^abd-z]|      (subroutine|program|end|data)\s|\s*!)'
   key_equals_value: '^[^#!;][^=]*='
+  m68k:
+  - '(?im)\bmoveq(?:\.l)?\s+#(?:\$-?[0-9a-f]{1,3}|%[0-1]{1,8}|-?[0-9]{1,3}),\s*d[0-7]\b'
+  - '(?im)^\s*move(?:\.[bwl])?\s+(?:sr|usp),\s*[^\s]+'
+  - '(?im)^\s*movem\.[bwl]\b'
+  - '(?im)^\s*move[mp](?:\.[wl])?\b'
+  - '(?im)^\s*btst\b'
+  - '(?im)^\s*dbra\b'
   objectivec: '^\s*(@(interface|class|protocol|property|end|synchronised|selector|implementation)\b|#import\s+.+\.h[">])'
   perl5: '\buse\s+(?:strict\b|v?5\.)'
   perl6: '^\s*(?:use\s+v6\b|\bmodule\b|\b(?:my\s+)?class\b)'
~~~

</details>

idrougge: I tested the `^ \s* move     (\.[bwl])? \s+ (sr|usp), \s* [^\s]+` line, and it only catches moves to `sr` or `usp`, not moves to any given register. I feel that the Motorola syntax of `move.size with source or destination as a register named Dn or An is sufficiently dissimilar to other assembly syntaxes to avoid confusion with other assembly languages while also catching even the shortest snippet.
Testing may prove the totality of heuristics to still be sufficient to catch all m68k assembly sources. 

`.i `, like `.inc`, `.asm` or `.s` is used by assemblers on most platforms AFAIK.

Alhadis: In that case, the third line of the `m68k` heuristic becomes:

~~~yaml
- '(?im)^\s*move\.[bwl]\s+.*\b[ad]\d'
~~~

I'll update the diff I just posted.

> `.i` , like `.inc`, `.asm` or `.s` is used by assemblers on most platforms AFAIK.

There are currently [11,273,157 `.i` files](https://github.com/search?q=extension%3Ai+NOT+nothack&type=Code) publicly indexed on GitHub. Surely there must be other formats hidden out there...

idrougge: That regex looks fine for heuristics.

A quick glance at those results indicate that a lot of `.i` files are SWIG files, which may need a language definition of their own.

Alhadis:
Christ. Going through the [search results for `.i`](https://github.com/Alhadis/Silos/tree/i/files) files reveals not one, but *two* potential language additions: [SWIG interfaces](http://www.swig.org/) and something called [Yorick](https://software.llnl.gov/yorick-doc/), which has apparently been around since 1996, [according to Wikipedia.](https://en.wikipedia.org/wiki/Yorick_(programming_language)) Making matters worse is the large number of assembly files that weren't matched by our proposed heuristic, suggesting we either need to improve our heuristic or register `.i` as an ordinary assembly extension.

@idrougge [Are](https://github.com/Alhadis/Silos/blob/601e7c614c7afa0f64a3858e7a3947566e891ff0/files/Other/rawsoundroutines.i) [these](https://github.com/Alhadis/Silos/blob/601e7c614c7afa0f64a3858e7a3947566e891ff0/files/Other/functions.i) [files](https://github.com/Alhadis/Silos/blob/601e7c614c7afa0f64a3858e7a3947566e891ff0/files/Other/data.1.i) Motorola 68k assembly? If so, we can add `asm6809` as an alias, which will identify them to Linguist through their Vim modelines (Vim also has a [`asm68k`](https://github.com/vim/vim/blob/f37506f60f87d52a9e8850e30067645e2b13783c/runtime/syntax/asm68k.vim) mode, so it's worth adding that as an alias as well…).

What makes me unsure is that when I googled `asm6809`, it led me to [a cross-assembler](https://www.mankier.com/1/asm6809) targeting Motorola 6809 and Hitachi 6309 processors:

> `asm6809` is a portable macro cross assembler targeting the Motorola 6809 and Hitachi 6309 processors. These processors are most commonly encountered in the Dragon and Tandy Colour Computer.

idrougge:
@Alhadis The files you linked are all Motorola 6809 assembly, which is an 8-bit processor totally distinct from 68000.

I agree that `.i` should be added as an ordinary assembly extension — it is analogous to `.s`.

Alhadis:
Ah, my bad. I took `Motorola 68k` to mean anything in the range of _"Motorola 68,000—68,999"_. Wasn't aware that it was specific to 68000 only. Guess that answers my question about Vim modelines, then. 👍

We'll need to add `.i` to the `Assembly` entry then. In addition, SWIG will need to be added as a new entry. Regarding Yorick, I concluded that it fails to meet our usage requirements (there were hundreds of files, but only a handful of repositories).

I can push these changes to your branch later, as I'll need to run Linguist on the files I harvested (I'm not currently on a computer where I'm able to do so). Thanks for your input and patience, I realise this PR's been left hanging open fo some time. 👍

idrougge:
Indeed 68k is 68000…68999, but 6809 < 68000. ;)

Alhadis:
Erp. So it is.

This just keeps getting better. 😅 I should stop talking now.

Alhadis:
Alright. I've pushed the aforementioned changes to your branch; running Linguist locally on the harvested `.i` files produced satisfactory results. Have updated the original post to include source/license links for the new samples.

All that's left now is for Lildude to give a final 👍 before we can merge.

idrougge:
Thanks!

PR #4639, opened 2019-09-04
Add 4DForm as a JSON language

Ganbin:
## Description
4DForm extension contains JSON data representing 4D Form description. It is used in the context of 4D software application which start to provide source control compatibility in beta within the new v17 R6 version and will be released officially in v18 (between end 19 - begin 20). More and more repositories contains that kind of files.

## Checklist:

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3A4DForm+pages+rightMargin+NOT+nothack
    - Sample source(s):
      - Example not on GitHub. It is a sample form that I have used during a training of 4D software development platform.
    - Sample license(s):
      - MIT

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2338,6 +2338,7 @@ JSON:
   - ".tern-project"
   - ".watchmanconfig"
   - composer.lock
+  - form.4DForm
   - mcmod.info
   language_id: 174
 JSON with Comments:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2315,6 +2315,7 @@ JSON:
   searchable: false
   extensions:
   - ".json"
+  - ".4DForm"
   - ".avsc"
   - ".geojson"
   - ".gltf"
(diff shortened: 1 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -2337,6 +2338,7 @@ JSON:
   - ".tern-project"
   - ".watchmanconfig"
   - composer.lock
+  - form.4DForm
```
pchaigno: This shouldn't be necessary as you already added `.4DForm` as a JSON extension.

Ganbin: Ok I was not sure If I had to add it in the filename too.

pchaigno:
@Alhadis Did you already check usage? Each repository seems to hold a lot of files.

Ganbin:
Hi, thanks for the interest and the check. Effectively, the non-beta (v18) should be released in early 2020. We could expect an increasing number of directories in the mid-2020. I also think that a lot of repositories may be private and are therefore not measurable.

pchaigno:
@Ganbin I usually check usage every ~6 months and stop monitoring file extensions for which usage hasn't grown. So it's maybe best if I start checking around the time the non-beta is released... Couldn't you post here when that happens?

Ganbin:
@pchaigno Ok I will try to remember to post here in about 6 month when the non-beta will be released.

PR #4642, opened 2019-09-10
Add esy.sh dependencies

Et7f3:
## Description

This adds the [esy](https://esy.sh) lockfile and build dir to the vendor list.

## Checklist:

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -62,6 +62,7 @@ module Linguist
       node_modules? ||
       go_vendor? ||
       go_lock? ||
+      esy_lock? ||
       npm_shrinkwrap_or_package_lock? ||
       godeps? ||
       generated_by_zephir? ||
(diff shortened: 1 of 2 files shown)


Alhadis:
Linguist doesn't acknowledge the `.lock` extension, so adding `.esy.lock` as a vendored file is redundant. Instead, we mark such files as *generated*, which is handled in [`lib/linguist/generated.rb`](https://github.com/github/linguist/blob/50005d6e607d956479f03e378c9c54aed03e7a6e/lib/linguist/generated.rb).

Could you give a link that explains more about `_esy` build directories?

Et7f3:
I only see `composer.lock`, `Cargo.lock`, `Pipfile.lock` and `(Gopkg|glide)\.lock` also side note.
https://github.com/github/linguist/blob/50005d6e607d956479f03e378c9c54aed03e7a6e/lib/linguist/generated.rb#L364 we have the two `!!` we didn't have it here https://github.com/github/linguist/blob/50005d6e607d956479f03e378c9c54aed03e7a6e/lib/linguist/generated.rb#L430

for `_esy` The better link I have found is https://esy.sh/docs/en/how-it-works.html#local-build-store
but it deal only with some specific folder. Maybe @andreypopp can give you more information/link because it is the main developper (2084 commits vs <113 commits for others)

ulrikstrid:
`_esy` should not be committed avd is ignored in all "official" templates.

`esy.lock` is a directory with a bunch of files that describes the current snapshot including overrides of some packages. When something is updated there is often a huge amount of changed files here.

Et7f3:
If we should add test I want to add theese one https://gist.github.com/Et7f3/c36059408ffb2f648618e798b54d1d6f

PR #4561, opened 2019-06-18
Add support for RBI files

peterzhu2118:
## Description

`.rbi` files will be for Sorbet type definition. See [here](https://sorbet.org/docs/rbi) for more details. TLDR: Sorbet is a static typechecker for Ruby developed by Stripe that will open sourced soon, it uses `.rbi` (Ruby interface) files to define the structure of gems and Ruby code. These `.rbi` files are valid Ruby code that contain no method body.

## Checklist:

- [x] **I am associating a language with a new file extension.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?q=extension%3Arbi+sig&type=Code
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      -  https://github.com/sorbet/sorbet-typed/blob/master/lib/ruby/all/open3.rbi
    - Sample license(s): [Apache 2](https://github.com/sorbet/sorbet-typed/blob/master/LICENCE)
  - [x] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

- [ ] **I am fixing a misclassified language**
  - [ ] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am changing the source of a syntax highlighting grammar**
  
  - Old: https://github-lightshow.herokuapp.com/
  - New: https://github-lightshow.herokuapp.com/

- [ ] **I am updating a grammar submodule**
  

- [ ] **I am adding new or changing current functionality**
  
  - [ ] I have added or updated the tests for the new or changed functionality.

- [ ] **I am changing the color associated with a language**
  
[...]

PR #4650, opened 2019-09-25
Add .rbi for Sorbet files

ghiculescu:
Adds https://sorbet.org/docs/rbi. They are valid Ruby files, with a `.rbi` extension.

## Checklist:

      - https://github.com/search?q=extension%3Arbi+typed&type=Code
    - Sample source(s):
      - https://github.com/sorbet/sorbet/blob/master/gems/sorbet/lib/gem_loader.rbi
    - No, as this extension isn't used elsewhere.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4522,6 +4522,7 @@ Ruby:
   - ".podspec"
   - ".rabl"
   - ".rake"
+  - ".rbi"
   - ".rbuild"
   - ".rbw"
   - ".rbx"
--- a/samples/Ruby/gem_loader.rbi
+++ b/samples/Ruby/gem_loader.rbi
@@ -1,0 +1,15 @@
+# typed: true
+
+# Write this as an RBI because the gem_loader.rb file itself cannot be typed.
+# By nature, it references constants in other gems in an attempt to load them.
+# They will not load in this project. It's marked as `typed: ignore`, and this
+# file exists so that we can call into gem_loader from typed code.
+
+class Sorbet::Private::GemLoader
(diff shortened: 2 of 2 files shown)


ghiculescu:
I realise this was previously rejected in https://github.com/github/linguist/pull/4561 but since Sorbet is now public, usage appears to have gone up a lot.

Alhadis:
> since Sorbet is now public, usage appears to have gone up a lot.

It would appear so. [~3,538 `.rbi` results](https://github.com/search?q=extension%3Arbi+NOT+nothack&type=Code) currently indexed, with the overwhelming majority of them being Ruby source:

- [2,396 results harvested](https://github.com/Alhadis/Silos/blob/rbi/urls.log)
- [117 unique repositories](https://github.com/Alhadis/Silos/blob/rbi/repos.log)
- [80 users](https://github.com/Alhadis/Silos/blob/rbi/users.log)

Not bad for something relatively-recent. 👍 LGTM. @lildude?

lildude:
> Not bad for something relatively-recent. 👍 LGTM. @lildude?

To be consistent with prior decisions, the usage doesn't quite meet the "hundreds of repos" requirement yet. Given it's jumped over 400 files in just over a week, I'm gonna give it a little longer before 👍  as it probably won't be long now.

pchaigno:
There are now [about 5k `.rbi` files](https://github.com/search?q=extension%3Arbi+NOT+dsddfdfkdj&type=Code). I downloaded 2555 of them using [Harvester]() and counted 120 repositories by 84 users. If we extrapolate to the total 5k files, that would be approximately 240 repositories. I think we're now okay to merge this!

PR #4656, opened 2019-09-30
Add Vim snippet language

ObserverOfTime:
## Description
This PR adds detection for Vim snippets (#4537). I couldn't find any compatible grammars.

## Checklist:
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com
      - https://github.com/search?q=extension%3Asnip+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Asnippets+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/Shougo/neosnippet-snippets/blob/master/neosnippets/vim.snip
      - https://github.com/honza/vim-snippets/blob/master/UltiSnips/vim.snippets
    - Sample license(s):
      - MIT
      - MIT

---

Edit: force-pushed to remove the conflicting color.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5451,6 +5451,20 @@ Verilog:
   codemirror_mode: verilog
   codemirror_mime_type: text/x-verilog
   language_id: 387
+Vim Snippet:
+  type: markup
+  aliases:
+  - SnipMate
+  - UltiSnip
+  - UltiSnips
+  - NeoSnippet
+  extensions:
+  - ".snip"
+  - ".snippet"
+  - ".snippets"
+  tm_scope: none
+  ace_mode: text
+  language_id: 81265970
 Vim script:
   type: programming
   color: "#199f4b"
--- a/samples/Vim Snippet/vim.snip
+++ b/samples/Vim Snippet/vim.snip
@@ -1,0 +1,236 @@
+snippet     if
+abbr        if endif
+options     head
+    if ${1:#:condition}
+      ${0:TARGET}
+    endif
+
+snippet     elseif
+options     head
+    elseif ${1:#:condition}
+        ${0:TARGET}
+
+snippet     ifelse
+abbr        if else endif
+options     head
+    if ${1:#:condition}
+      ${2:TARGET}
+    else
+      ${3}
+    endif
+
+snippet     for
+abbr        for in endfor
+options     head
+    for ${1:#:var} in ${2:#:list}
+      ${0:TARGET}
+    endfor
+
+snippet     while
+abbr        while endwhile
+options     head
+    while ${1:#:condition}
+      ${0:TARGET}
+    endwhile
+
+snippet     function
+abbr        func endfunc
+alias       func
+options     head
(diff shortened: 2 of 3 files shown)


lib/linguist/languages.yml
```diff
@@ -5473,6 +5473,18 @@ Vim script:
   - vimrc
   ace_mode: text
   language_id: 388
+Vim snippet:
```
Alhadis: Could you capitalise `Snippet`, please?

~~~diff yml
-Vim snippet
+Vim Snippet
~~~

lib/linguist/languages.yml
```diff
@@ -5473,6 +5473,18 @@ Vim script:
...
+  type: markup
+  aliases:
+  - SnipMate
+  - UltiSnip
```
Alhadis: According to [its repository](https://github.com/SirVer/ultisnips), this engine's name should be `UltiSnips` (pluralised).

Keep the singular spelling too, though, because it's easily confused.

lib/linguist/languages.yml
```diff
@@ -5473,6 +5473,18 @@ Vim script:
...
+  - NeoSnippet
+  extensions:
+  - ".snip"
+  - ".snippets"
```
Alhadis: You can also add [`.snippet`](https://github.com/genghuiluo/vimrc/tree/65badb75815e4b21b941e89879d68ad4dbedb17c/sources_non_forked/snipmate-snippets/ruby-rspec) as an extension, too (no plural).

I got halfway through researching each extension's usage earlier this year before getting interrupted, but I'd already confirmed their distribution is adequate. 👍

lib/linguist/languages.yml
ObserverOfTime: I didn't capitalise it because I saw that `Vim script` is lowercase.

Alhadis: Normally, we prefer to keep things capitalised, unless non-standard capitalisation is canon.

I chose to name `Vim script` with a lowercase `s` because that's apparently how Bram Moolenaar always referred to the language. The entry was previously named `VimL`, which wasn't just incorrect, but the name appeared to have originated on GitHub and began creeping into casual usage.

I wanted to make up for that, so I was cautious with the precise spelling I used.

By no means a common reason for an exception, but there you go. 😉

Alhadis:
For posterity and administrative correctness, [here's what Harvester collected](https://github.com/Alhadis/Silos/tree/snippet) from the `.snippet` extension search.

- [516,129 indexed search results](https://github.com/search?q=extension%3Asnippet+-language%3Axml+-filename%3Aaspx+NOT+nothack&type=Code)
- [2,745 harvested samples](https://github.com/Alhadis/Silos/blob/snippet/urls.log)
- [80 unique repos](https://github.com/Alhadis/Silos/blob/snippet/repos.log)
- [79 unique users](https://github.com/Alhadis/Silos/blob/snippet/users.log)

The repo count looks low because most of these snippet repos number in the hundreds, and Harvester has a hard limit of <= 3,000 results, so the uniquity is skewed.

PR #4660, opened 2019-10-03
Consider go.sum to be generated

nathany:
Go 1.11 and newer include built in support for Go modules, with a user-editable go.mod file and a generated go.sum file.

This flags go.sum as a generated file so that it will be suppressed in diffs.

## Checklist:

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -62,6 +62,7 @@ module Linguist
       node_modules? ||
       go_vendor? ||
       go_lock? ||
+      go_sum? ||
       npm_shrinkwrap_or_package_lock? ||
       godeps? ||
       generated_by_zephir? ||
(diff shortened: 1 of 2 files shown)


Alhadis:
See #4340. This was discussed in the past, and the eventual consensus was `go.sum` files should remain visible to be reviewed in diffs.

You can still exclude them in your projects by marking them as generated in a `.gitattributes` file.

nathany:
Apologies, I only searched against open issues and pull requests.

I respectfully disagree with the previous conclusion. When I have a pull request that updates dependencies, the actual code changes get lost after the giant go.sum diff.

While the go.mod file should be reviewed, the go.sum is generated and need not be reviewed. Also, a generated file can still be expanded as needed.

It doesn't make sense to consider Gopkg.lock and glide.lock as generated but not do the same for go.sum.

Alhadis:
Sigh... yes, this was precisely the catalyst of a long, *long* debate.

In any case, you still have the option of hiding these files yourself.

nathany:
Thanks for the tip. Works just as desired, so now I just need to go add it to every one of my Go repos. 🤦‍♂️

I guess I can see how some people might want to review the transitive dependency changes in the diff, but I really don't see reading through `go.sum` as a good way to do that. A much cleaner interface would be nice. In any case, thanks for the rapid response, and providing a workaround.

PR #4662, opened 2019-10-03
make the change requested by the review

Et7f3:
lock file go in generated section so it disapear from vendor.yml


Et7f3:
Sorry I missclicked. The right PR is here https://github.com/github/linguist/pull/4642

Issue #3562, opened 2017-04-11
JSON output not working when running on a single file


armijnhemel:
I tried the command line interface to linguist (updated to master) and JSON output does not seem to be working when trying to process a single file:

$ ./bin/linguist ./grammars.yml --json
grammars.yml: 693 lines (692 sloc)
  type:      Text
  mime type: text/x-yaml
  language:  YAML

The help text seems to suggest that JSON output should be available:

$ ./bin/linguist -h
  Linguist v5.0.8
  Detect language type for a file, or, given a directory, determine language breakdown.

  Usage: linguist <path>
         linguist <path> [--breakdown] [--json]
         linguist [--breakdown] [--json]

armijnhemel:
Hm, it appears that some output went missing. Anyway, once you run linguist -h it should be clear that the --json option is also advertised for individual paths, but that's not working.

pchaigno:
Thanks for reporting the issue.
We'd welcome a pull request to fix this!

armijnhemel:
I'd like to but unfortunately Ruby is not my forte :(

PR #4664, opened 2019-10-05
Add JSON output for single file classification

lbonanomi:
This PR attempts to resolve https://github.com/github/linguist/issues/3562 by outputting JSON-format results for a single file.

## Description

Example based on issue:

```
$ ./bin/github-linguist grammars.yml --json
{"grammars.yml":{"lines":857,"sloc":856,"type":"Text","mime_type":"text/x-yaml","language":"YAML"}}
```

### DIFF
--- a/bin/github-linguist
+++ b/bin/github-linguist
@@ -58,20 +58,20 @@ elsif File.file?(path)
     'Binary'
   end
 
-  if !json_breakdown
+  if json_breakdown
+    puts JSON.generate( { blob.name =>  {
+                                          :lines => blob.loc,
+                                          :sloc => blob.sloc,
+                                          :type => type,
+                                          :mime_type => blob.mime_type,
+                                          :language => blob.language
+                                        }
+                        } )
+  else
     puts "#{blob.name}: #{blob.loc} lines (#{blob.sloc} sloc)"
     puts "  type:      #{type}"
     puts "  mime type: #{blob.mime_type}"
     puts "  language:  #{blob.language}"
-  else
-    puts JSON.generate( { blob.name =>	{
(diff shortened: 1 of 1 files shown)


bin/github-linguist
```diff
@@ -56,10 +57,21 @@ elsif File.file?(path)
...
+  else
+    puts JSON.generate( { blob.name =>	{
+	    					:lines => blob.loc,
+						:sloc => blob.sloc,
+						:type => type,
+						:mime_type => blob.mime_type,
+						:language => blob.language
+    					}
+    			} )
+  end
```
lildude: Negation in the first part of an `if` statement can lead to confusion. I know we have this smelly behaviour elsewhere in this file and we can address that at a later date. For the moment, please flip this around, ie:

```ruby
if json_breakdown
  puts JSON.generate( { blob.name =>	{
  [...]
else
  puts "#{blob.name}: #{blob.loc} lines (#{blob.sloc} sloc)"
  [...]
end

File: bin/github-linguist [hunk 2]
```diff
@@ -22,10 +22,10 @@ end
 ARGV.shift
 breakdown = true if ARGV[0] == "--breakdown"
 json_breakdown = true if ARGV[0] == "--json"
-
```
lildude: Please put this blank line back. It makes things clearer.

bin/github-linguist
lbonanomi: I'll refrain from making a joke about perl, here. Done.

bin/github-linguist
lbonanomi: You're right, it is re-added.

Issue #4506, opened 2019-04-25
List of languages pending colour correction or addition


Alhadis:
#4417 was closed yesterday, so I felt it might be beneficial to keep a record of languages whose colours are incorrect (according to their branding or community's "preferred" colour). That way they won't fade into obscurity after being blocked for so long.

Now, these fixes are blocked because of an ugly tangle of colour-proximity violations, which was previously discussed in [`#4291`](https://github.com/github/linguist/issues/4291#issuecomment-452278428). Since it's unclear when (or even *if*) a decision will be made on GitHub's part, I'm compiling them here so nobody forgets. :+1: The list is incomplete and I've probably missed a lot, so feel free to make amendments as you see fit.

* [ ] **Asymptote:** `#ff0000` (source: [logo](http://asymptote.sourceforge.net/logo.svg))
* [ ] **Apex:** `#1797c0` (source: [`#4417`](https://github.com/github/linguist/issues/4417))
* [ ] **Blade:** `#f7523f`
* [ ] **CSON:** `#244776` (source: "inherited" from CoffeeScript)
* [ ] **Eiffel:** `#4d6977` (source: [logo](https://upload.wikimedia.org/wikipedia/commons/6/6a/Eiffel_logo.svg))
* [ ] **Grammatical Framework**: `#ff0000` (source: [logo](http://www.grammaticalframework.org/))
* [ ] **Haml:** `#ece2a9`
* [ ] **Handlebars:** `#f7931e`
* [ ] **Latte:** `#f2a542`
* [ ] **Less:** `#1d365d`
* [ ] **Marko:** `#42bff2`
* [ ] **Markdown:** `#083fa1` (source: [original colour value](https://github.com/github/linguist/pull/3112#issuecomment-232171926))
* [ ]  **mIRC Script Language (mSL)**: `#3D57C3` (source: "m" in [logo](https://www.mirc.com/images/indexlogorb.gif))
* [ ] **PostScript:** `#da291c` (source: [`#3141`](https://github.com/github/linguist/pull/3141))
* [ ] **Prisma**: `#0C344B`
* [ ] **Pug:** `#a86454`
* [ ] **SCSS:** `#c6538c`
* [ ] **Sass:** `#a53b70`
* [ ] **Slim:** `#2b2b2b`
* [ ] **SourcePawn:** `#f69e1d` (source: [logo](https://www.sourcemod.net/))
* [ ] **Stylus:** `#ff6347`
[...]

PR #4667, opened 2019-10-08
added scss color

Tusko:


Alhadis:
This PR won't do anything. Certain languages are considered "sub-languages" of others… in this case, SCSS is a "sublanguage" is CSS. That means their parent language's colour is shown in the stats-bar, making a `color:` field completely redundant.

See #4291 and #4506. Until GitHub staff address those issues, there's nothing we can do. 😞

PR #4671, opened 2019-10-10
Add "bzl" and "build" (Starlark) as Python aliases

sfreilich:
Starlark files (*.bzl, BUILD, BUILD.bazel) have been included under Python for almost four years, since 864a6c0a2 (by @Dominator008). Would be good for "bzl" and "build" fenced code blocks to be able to get that highlighting, too.

## Checklist:

Where are tests relevant to this change?

Would it be better to split this into its own language entry?

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4111,6 +4111,8 @@ Python:
   - python2
   - python3
   aliases:
+  - bzl
+  - build
   - rusthon
   - python3
   language_id: 303


lib/linguist/languages.yml
```diff
@@ -4111,6 +4111,8 @@ Python:
   - python2
   - python3
   aliases:
+  - bzl
+  - build
```
Alhadis: Wouldn't it make more sense to add `starlark` as an alias instead?

`bzl` is fine, but I don't think we should be associating a word as generic as `build` with a specific language (unless that language is *literally* named "Build"...)

sfreilich: Sounds good.

PR #4680, opened 2019-10-20
add language DirectX 3D File

malytomas:
## Description

See https://github.com/github/linguist/issues/4679

## Checklist:

      -  https://github.com/search?l=&q=extension%3Ax+size%3A%3E10000+%22xof+0303txt+0032%22&type=Code
    - Sample source(s):
    - Sample license(s):
       The file added to samples is model of a cube, created and exported from blender. CC0.
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-etc%2Fblob%2Fmaster%2Fgrammars%2Fdirect-x.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fdinghim%2Fmagic_cube%2Fblob%2Fd4aa1b0d3438ce39c3d90c6a77a161fe69cfe0b8%2FmagicCube%2Fcube.X&code=
     - the command scripts/add-grammar is failing: cannot connect to docker daemon.
    - help needed :D


lib/linguist/languages.yml
```diff
@@ -1132,6 +1132,11 @@ Diff:
   codemirror_mode: diff
   codemirror_mime_type: text/x-diff
   language_id: 88
+DirectX 3D File:
+  type: data
+  extensions:
+  - ".x"
+  language_id: 201049282
```
Alhadis: Alright, first, let's add the syntax highlighting fields:

~~~yaml
- ace_mode: text
- tm_scope: source.direct-x
~~~

(The [ACE editpr](https://ace.c9.io) doesn't have that many editing modes, so we usually leave this field as `text`. It's a mandatory field, however).

Alhadis:
> the command `scripts/add-grammar` is failing: cannot connect to docker daemon

In this instance, you actually don't need to run that script; I've already added the grammar to a repository that's already added as a submodule. What you *do* need to do is update it manually; I'll handle this by pushing changes to your branch, since it's a little complicated.

lib/linguist/languages.yml
malytomas: what about the codemirror_* fields?

Alhadis: Nope, those are optional. CodeMirror [doesn't have that many language-modes either](https://codemirror.net/mode/index.html), so unless it has one that produces similar-enough highlighting, we generally just leave these fields out (it defaults to `text/plain` in any case).

Alhadis:
@malytomas I can't push to your branch. Could you check that you have this checkbox ticked on the right-hand side of this page, please? :)

Figure 1

Koszi!

malytomas:
> 
> 
> @malytomas I can't push to your branch. Could you check that you have this checkbox ticked on the right-hand side of this page, please? :)
> 
> Figure 1
> 
> Koszi!

It is. I have also added you as a collaborator.

Alhadis:
Okay, that shouldn't be happening. GitHub also indicated I could push to your branch, but my push was rejected:

Figure 2

So, uh, apologies for making this PR more complicated than it had to be. 😁

Alhadis:
BTW, when forking for a pull-request, it always helps to create a separate branch for your changes (instead of working off your forked `master` branch). It makes stuff like this a bit messier to deal with.

malytomas:
> BTW, when forking for a pull-request, it always helps to create a separate branch for your changes (instead of working off your forked `master` branch). It makes stuff like this a bit messier to deal with.

Apologies. I never had issues like this before.
Anyway, glad you figured it out. :D

Is there anything else I can help with? :D
(Not that I helped much so far... :( )

Alhadis:
Don't worry, this is my fault. I should have rolled the grammar into [my other pull-request](https://github.com/github/linguist/pull/), which is the reason for all our troubles (I think)...

To make things worse, I mistakenly merged the wrong file locally, causing the [license-hash](https://github.com/github/linguist/commit/a1925c150d81de7a16188247e6003b2cf1f681ef/checks?check_suite_id=272805228) to change erroneously... 😥 (This isn't a normal part of the procedure for adding grammars to Linguist, so don't let this dissuade you from future contributions. 😀)

Alhadis:
Ugh, the frigging tests are passing locally, I've no idea what's going on with the license cache. 😞

Tell you what, let's try again, but *without* the grammar this time. Just put `none` as the `tm_scope` field's value; I can add it in a follow-up PR if need be (probably in #4678).

Once again, sorry for this. 😢

Issue #1247, opened 2014-06-03
Add Faust


magnetophon:
http://faust.grame.fr/

FAUST (Functional Audio Stream) is a functional programming language specifically designed for real-time signal processing and synthesis. FAUST targets high-performance signal processing applications and audio plug-ins for a variety of platforms and standards.

It's quite popular in the linux audio world:
http://lac.linuxaudio.org/2014/?page=program

and on github:
https://github.com/search?q=extension%3Adsp+faust&type=Code&ref=searchresults

PR #4688, opened 2019-10-28
Add the Faust programming language

jpcima:
Hi, this adds a grammar for the [Faust programming language](https://github.com/grame-cncm/faust).

These are related issues: #1247 and grame-cncm/faust#354

## Description

Doing the steps as indicated in the documentation.

## Checklist:

      -  https://github.com/search?q=extension%3Adsp+stdfaust.lib&type=Code
    - Sample source(s):
      - https://github.com/grame-cncm/faust
      - https://github.com/grame-cncm/faustlibraries
    - Sample license(s): MIT, LGPL2.1, GPL3.0, MIT-style STK-4.3
  
  - [x] I have included a syntax highlighting grammar: [Link](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fjpcima%2Flanguage-faust%2Fmaster%2Fgrammars%2Ffaust.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fjpcima%2Fstone-phaser%2Fmaster%2Fdsp%2Fstone_phaser.dsp&code=)


pchaigno:
Thanks for the contribution!

Are [these files](https://github.com/search?p=1&q=extension%3Adsp+NOT+stdfaust.lib&type=Code) related to Faust as well?

jpcima:
You're welcome. No, these files by Visual Studio are strictly unrelated.

pchaigno:
Ok. We'll need to identify their language and add it to Linguist as well. Otherwise they will be recognized as Faust :-/

jpcima:
Hmm, I don't think these Microsoft files have a public specification, do they?
Does it need a full grammar to detect it, or there is a simpler way?

pchaigno:
Grammars are only needed for syntax highlighting and not mandatory. To add the "language", we'll need its name and a way to differentiate it from Faust files (see [`lib/heuristics.yml`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml)).

jpcima:
To my knowledge, these files are IDE-generated, and I added a pattern to identify the format according to some comments in the heading.
It will need to add a few samples..

lib/linguist/heuristics.yml
```diff
@@ -113,6 +113,12 @@ disambiguations:
     #  : dependency
     # path/file.ext1 : some/path/../file.ext2
     pattern: '([\/\\].*:\s+.*\s\\$|: \\$|^[ %]:|^[\w\s\/\\.]+\w+\.\w+\s*:\s+[\w\s\/\\.]+\w+\.\w+)'
+- extensions: ['.dsp']
+  rules:
+  - language: Microsoft Developer Studio Project
+    pattern: '#.*Microsoft Developer Studio.*'
```
pchaigno: I think we can replace this with only `# Microsoft Developer Studio Generated Build File` to avoid the any backtracking.

Could you also add support for these files in [`lib/linguist/generated.rb`](https://github.com/github/linguist/blob/master/lib/linguist/generated.rb), since they appear to be generated?

lib/linguist/heuristics.yml
```diff
@@ -113,6 +113,12 @@ disambiguations:
...
+  - language: Faust
+    pattern: '='
```
pchaigno: This pattern is too general. Almost all languages use `=` for assignment, including `Microsoft Developer Studio Project`. Any other pattern, specific to Faust, we could use? It doesn't need to match *all* Faust files, we have other strategies to distinguish the remaining files.

lib/linguist/heuristics.yml
```diff
@@ -113,6 +113,12 @@ disambiguations:
     #  : dependency
     # path/file.ext1 : some/path/../file.ext2
     pattern: '([\/\\].*:\s+.*\s\\$|: \\$|^[ %]:|^[\w\s\/\\.]+\w+\.\w+\s*:\s+[\w\s\/\\.]+\w+\.\w+)'
+- extensions: ['.dsp']
```
pchaigno: Could you also add a test in [`test/test_heuristics.rb`](https://github.com/github/linguist/blob/master/test/test_heuristics.rb) for these heuristic rules?

vendor/licenses/grammar/language-faust.txt
```diff
@@ -0,0 +1,52 @@
+---
+type: grammar
+name: language-faust
+version: 93a620e7db713a9858c123e5be5e880ca53856c9
+license: other
```
pchaigno: @lildude I'm not sure I remember how we're supposed to whitelist licenses now? This one looks okay, it's just duplicated with the original repository's license. Would just changing the above line to `license: MIT` be okay?

lib/linguist/heuristics.yml
jpcima: I have collected a set of regular expressions which I think relevant.
All are optional but some will match most times.

1. the main function known as `process`
2. importing statements : `import` or `library`, followed by paren and double-quote
3. some common top-level declarations

Is it good?

```
\bprocess\b
\b(library|import)\s*\(\s*"
\bdeclare\s+(name|version|author|copyright|license)\s+"
```

pchaigno: The last two look good (although I don't think the `\s*` at the end of the second is necessary and the `\s+` could be `\b` for the last), but the first one is too common a keyword. Maybe we could tie that keyword to part of the construct of a function (kinda of like you did for the last two)?

lib/linguist/generated.rb
```diff
@@ -618,6 +619,14 @@ def generated_gimp?
              lines[0].match(/\/\*  GIMP header image file format \([a-zA-Z0-9\- ]+\)\: .+?\.h  \*\//)
     end
 
+    # Internal: Is this a generated Microsoft Visual Studio 6.0 build file?
+    #
+    # Return true or false
+    def generated_visualstudio6?
+      return false unless extname.downcase == '.dsp'
+      lines.any? { |l| l.include? '# Microsoft Developer Studio Generated Build File' }
```
pchaigno: Do we really need to look further than the first three lines?

lib/linguist/heuristics.yml
jpcima: >The last two look good (although I don't think the \s* at the end of the second is necessary and the \s+ could be \b for the last)

These spacings precede a `"` character following. I intended this double-quote as part of the regular expression.

Regarding `process`, I'm able to check for the next token, which is necessarily "(" or "=".
examples are `process = _;` and `process(a,b) = a,b;`

lib/linguist/generated.rb
jpcima: Probably not.

lib/linguist/heuristics.yml
jpcima: This is done, and tried on my side, looking fine.

lib/linguist/languages.yml
```diff
@@ -1455,6 +1455,15 @@ Fantom:
   tm_scope: source.fan
   ace_mode: text
   language_id: 110
+Faust:
+  type: programming
+  color: "#c37240"
+  extensions:
+  - ".dsp"
+  - ".lib"
```
pchaigno: Hm, I missed that file extension. You'll need to provide a search query for that one too. We'll have to be particularly careful for that one, it's likely to conflict with other languages.

jpcima: Oh yes, for the ".lib" file suffix:
https://github.com/search?p=2&q=extension%3Alib+%22declare+name%22&type=Code

It's nothing special compared to ".dsp", it's just an optional naming which hints at the file being a library of functions.

What steps does it take to disambiguate this one?

pchaigno: Piouf, that one is huge: [`extension:lib NOT "declare name"`](https://github.com/search?o=desc&q=extension%3Alib+NOT+%22declare+name%22&s=indexed&type=Code). Are you sure you don't want to add it in a second pull request? There seem to be a lot of different languages there and it's going to take us a while to sort them out. It might be better to get `.dsp` in first, and then address `.lib`.

jpcima: It is perfectly fine with me, I removed `.lib` in the commit f16e596f.

vendor/licenses/grammar/language-faust.txt
lildude: Yes, manually changing this to mit should do the trick.

jpcima: Hi, I have edited the license as suggested.

jpcima:
It's no problem, I've removed a few.
Now 2 for Faust, and 1 for Microsoft DSP.

jpcima:
Hi, I think I have done everything as asked, but if it needs something else, please let me know.

lib/linguist/heuristics.yml
```diff
@@ -113,6 +113,12 @@ disambiguations:
...
+  rules:
+  - language: Microsoft Developer Studio Project
+    pattern: '# Microsoft Developer Studio Generated Build File'
```
Alhadis: Tabs are awesome and everybody should use them, but I'm afraid they're invalid in YAML; these lines need to use spaces for indentation (much as I hate to say it…)

lib/linguist/heuristics.yml
```diff
@@ -113,6 +113,12 @@ disambiguations:
...
+  - language: Faust
+    pattern: '\bprocess\s*[(=]|\b(library|import)\s*\(\s*"|\bdeclare\s+(name|version|author|copyright|license)\s+"'
```
Alhadis: See above.

lib/linguist/languages.yml
```diff
@@ -1455,6 +1455,14 @@ Fantom:
...
   ace_mode: text
   language_id: 110
+Faust:
+  type: programming
+  color: "#c37240"
+  extensions:
+  - ".dsp"
+  tm_scope: source.faust
+  ace_mode: text
+  language_id: 622529198
```
Alhadis: These lines are also using tabs instead of spaces.

lib/linguist/languages.yml
```diff
@@ -3148,6 +3156,13 @@ Metal:
   codemirror_mode: clike
   codemirror_mime_type: text/x-c++src
   language_id: 230
+Microsoft Developer Studio Project:
+  type: data
+  extensions:
+  - ".dsp"
+  tm_scope: none
+  ace_mode: text
+  language_id: 800983837
```
Alhadis: See above.

Alhadis:
@jpcima I'm not sure what text editor you use, but I recommend installing a plugin for [EditorConfig](https://editorconfig.org/) if one is available. This will solve issues like the one I mentioned above.

Also, just to reiterate: tabs > spaces.

lib/linguist/heuristics.yml
jpcima: Hi @Alhadis,
I do not understand the remark; I have checked the file, and it does not contain any tab character. I don't use them usually and I try my best to adapt, and in this case I cannot see where exactly is the problem here.

Alhadis: Crap, I forgot to disable a browser extension. Nevermind. 😂

PR #4089, opened 2018-04-04
improve GLSL classification

smola:
* added extensions .gs, .comp and .glslf.
  See: https://stackoverflow.com/q/6432838

* add disambiguation rule for .gs files:
  if it starts with `#version <number>`, it's GLSL.
  This seems to work for most GLSL files with few exceptions
  that are catched by the classifier.

* added sample extrude_normal.gs from kyle-piddington/ShaderTool
  (licensed under MIT):
  https://github.com/kyle-piddington/ShaderTool/blob/c753a53bde6eab942da617adab9483c945f27f51/assets/shaders/extrude_normals.gs

* added sample blend_120.glslf from gfx-rs/gfx (licensed under
  Apache 2.0):
  https://github.com/gfx-rs/gfx/blob/7b084cfbb95e3e4d1ea51e91746fc75b2efb3be3/examples/blend/shader/blend_120.glslf

## Checklist:

- [x] **I am associating a language with a new file extension.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a syntax highlighting grammar.
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am fixing a misclassified language**
  - [ ] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am changing the source of a syntax highlighting grammar**
  
[...]

PR #3747, opened 2017-07-31
improve GLSL classification

smola:
* added extensions .gs, .comp and .glslf.
  See: https://stackoverflow.com/q/6432838

* add disambiguation rule for .gs files:
  if it starts with `#version <number>`, it's GLSL.
  This seems to work for most GLSL files with few exceptions
  that are catched by the classifier.

* added a sample extrude_normal.gs from kyle-piddington/ShaderTool
  (licensed under MIT):
  https://github.com/kyle-piddington/ShaderTool/blob/c753a53bde6eab942da617adab9483c945f27f51/assets/shaders/extrude_normals.gs

PR #4692, opened 2019-11-01
Improve GLSL classification

smola:
## Description
* added extensions .gs, .comp and .glslf.
  See: https://stackoverflow.com/q/6432838

* add disambiguation rule for .gs files:
  if it starts with `#version <number>`, it's GLSL.
  This seems to work for most GLSL files with few exceptions
  that are catched by the classifier.

* added sample extrude_normal.gs from kyle-piddington/ShaderTool
  (licensed under MIT):
  https://github.com/kyle-piddington/ShaderTool/blob/c753a53bde6eab942da617adab9483c945f27f51/assets/shaders/extrude_normals.gs

* added sample blend_120.glslf from gfx-rs/gfx (licensed under
  Apache 2.0):
  https://github.com/gfx-rs/gfx/blob/7b084cfbb95e3e4d1ea51e91746fc75b2efb3be3/examples/blend/shader/blend_120.glslf

* test_heuristics.rb now supports alt_name to ake a different file name
for a fixture.

## Checklist:
* Continuation of #3747 #4089 #4402.
* Last pending review at #4402. Now solved.
* After merging this, an issue should be opened to find proper disambiguation for `.comp`. See https://github.com/github/linguist/pull/4402#discussion_r265324224


lib/linguist/languages.yml
```diff
@@ -1604,7 +1604,9 @@ GLSL:
   - ".fshader"
   - ".geo"
   - ".geom"
+  - ".glslf"
   - ".glslv"
+  - ".gs"
```
nihui: hi, please add .comp extension

PR #4695, opened 2019-11-03
Support YAML's .yaml.sed extension

pchaigno:
This pull request adds support for YAML's `.yaml.sed` file extension. It seems to indicate YAML files in which a couple constants are expected to be "initialized" with sed.

## Checklist:

      - https://github.com/search?o=desc&q=extension%3Ased+apiVersion&s=&type=Code (288 repositories by 238 users according to Harvester)
    - Sample source(s): [`coredns.yaml.sed`](https://github.com/bastlelight/googlek8s/blob/9e52d14eb9db297523ec5541a1131856a498f518/cluster/addons/dns/coredns.yaml.sed)
    - Sample license(s): [Apache v2.0](https://github.com/bastlelight/googlek8s/blob/9e52d14eb9db297523ec5541a1131856a498f518/LICENSE)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5973,6 +5973,7 @@ YAML:
   - ".syntax"
   - ".yaml"
   - ".yaml-tmlanguage"
+  - ".yaml.sed"
   - ".yml.mysql"
   filenames:
   - ".clang-format"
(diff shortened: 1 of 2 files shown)


pchaigno:
Uh, not again... Thanks @Alhadis! It's fixed.

Issue #4681, opened 2019-10-20
Whitespace in .gitattributes linguist-language (again)


vinriviere:
Hi, I wonder how to specify a language name with spaces in `.gitattributes` for `linguist-language`. This question had been asked in issue #2751, but as the issue seemed to be solved differently, the question was not answered.

More specifically, I want to colorize [this file](https://github.com/Vretrocomputing/atari-devpac2-afficher_image/blob/master/IMAGE.S) with the language `Motorola 68K Assembly`. I found that name in [languages.yml](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3226), but I'm stuck as spaces don't seem to be allowed in `.gitattributes` values.

BTW, to avoid confusion, it would be nice if GitHub could display the name of the currently detected language used for colorization.

Alhadis:
Unfortunately, this is a limitation of the `gitattributes(5)` format itself; the best you can do is use an alias that doesn't contain whitespace.

Motorola assembly doesn't have any aliases defined yet, but `m68k` would be a good one. You're more than welcome to submit a pull-request to add it. 👍 Bear in mind the changes won't take place until the next release of Linguist (which will be soon, I believe — late this month).

vinriviere:
@Alhadis Thanks for your feedback! I have just created the pull request #4683, as requested.

Alhadis:
Aaaaand it's already been approved. 😉 Will get merged once GitHub staff (i.e., @lildude) gives it the final review.

Thanks!

vinriviere:
Many thanks, it works fine now to colorize my assembler files on GitHub 😀
https://github.com/Vretrocomputing/atari-devpac2-afficher_image/blob/master/IMAGE.S

PR #4699, opened 2019-11-04
Update docs to explain whitespace in gitattributes

### LINKED ISSUE #4697: Update docs to mention hyphenated aliases
**UPDATE:** Our docs suck, let's make them not suck. Original issue below:

------------

An [acknowledged](https://github.com/github/linguist/issues/2751) [limitation](https://github.com/github/linguist/issues/4681) of ~~Linguist~~ Git is that it's currently impossible to override a language if its name contains whitespace (and lacks an alias that doesn't). For example, none of the following lines work:

~~~gitattributes
*.glyphs linguist-language=OpenStep Property List
*.glyphs linguist-language=OpenStep\ Property\ List
*.glyphs linguist-language="OpenStep Property List"
~~~

I propose a second override to enable users to specify a language using its [`language_id`](https://github.com/github/linguist/blob/770b6e636683d7eebf295adaec9ab7b175563710/lib/linguist/languages.yml#L23-L25):

~~~gitattributes
*.glyphs linguist-language-id=598917541
~~~

This is neither readable nor convenient, but it's better than nothing. Thoughts?

Alhadis:
Fixes #4697 and, by proxy, #4681.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,19 @@ File and folder paths inside `.gitattributes` are calculated relative to the pos
 *.rb linguist-language=Java
 ```
 
+Be aware that `gitattribute` values cannot contain whitespace.
+If a language's name contains spaces, replace them with hyphens instead:
+
+```gitattributes
+# These lines won't work:
+*.glyphs linguist-language=OpenStep Property List
+*.glyphs linguist-language=OpenStep\ Property\ List
+*.glyphs linguist-language="OpenStep Property List"
+
+# But this line will:
+*.glyphs linguist-language=OpenStep-Property-List
+```
+
 #### Vendored code
 
 Checking code you didn't write, such as JavaScript libraries, into your git repo is a common practice, but this often inflates your project's language stats and may even cause your project to be labeled as another language.


README.md
```diff
@@ -181,6 +181,19 @@ File and folder paths inside `.gitattributes` are calculated relative to the pos
...
 ```
 
+Be aware that `gitattribute` values cannot contain whitespace.
+If a language's name contains spaces, replace them with hyphens instead:
+
+```gitattributes
+# These lines won't work:
+*.glyphs linguist-language=OpenStep Property List
+*.glyphs linguist-language=OpenStep\ Property\ List
+*.glyphs linguist-language="OpenStep Property List"
```
pchaigno: I would remove these. The README is already longer than it should, and these examples are likely to be confusing for anyone who's skimming through.

To be honest, I would even prefer simply adding a new line to the above `=Java` example, with a comment that says `Replace whitespaces with hyphens`.

lildude: Ooops. Eager beaver over here merged to soon 😊

Alhadis: Ah yeah, true.

This is what happens when I proof-read my own writing too many times: I develop tunnel-vision. 😁

pchaigno: I'm too slow these days...

Alhadis: @pchaigno So, like this?

> ```gitattributes
> # Example of a `.gitattributes` file which reclassifies `.rb` files as Java:
> *.rb linguist-language=Java
>
> # Replace any whitespace in the language name with hyphens.
> ```

Or this?

> ```gitattributes
> # Example of a `.gitattributes` file which reclassifies `.rb` files as Java:
> *.rb linguist-language=Java
> ```
>
> Replace any whitespace in the language name with hyphens.

pchaigno: Hm, too slow and unclear :)

I was thinking of something like this:
```gitattributes
# Example of a `.gitattributes` file which reclassifies `.rb` files as Java:
*.rb linguist-language=Java

# Replace any whitespace in the language name with hyphens:
*.glyphs linguist-language=OpenStep-Property-List 
```
When I'm skimming through documentation, I usually look only at the code/config. examples, barely reading any text. It might not be how other people read it though.

PR #4713, opened 2019-11-12
remove reason from ocaml group

### LINKED ISSUE #4684: ReasonML detected as Ocaml
I think it happens because reason is in the `ocaml` group.

I know and agree that reason is just a syntax for ocaml but seprate detection will give some benfits such as searching through trending repos, know that how much of a monorepo, or a library which uses both reason and ocaml is written in reason / ocaml.

amiralies:
Closes #4684

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4373,7 +4373,6 @@ Readline Config:
   language_id: 538732839
 Reason:
   type: programming
-  group: OCaml
   ace_mode: rust
   codemirror_mode: rust
   codemirror_mime_type: text/x-rustsrc
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4373,6 +4373,7 @@ Readline Config:
   language_id: 538732839
 Reason:
   type: programming
+  color: "#dd4b39"
   ace_mode: rust
   codemirror_mode: rust
   codemirror_mime_type: text/x-rustsrc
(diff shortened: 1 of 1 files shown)


amiralies:
@jordwalke do you agree with this?

jordwalke:
I think your argument about knowing how much of a repo is written in either Reason or OCaml is a nice feature. It would be cool if eventually there were some higher level Github/linguist feature available to be able to group based on language semantics and not just syntax, but until then I support this change.

ryyppy:
Just to be sure to comply to the CONTRIBUTING guidelines:

## Checklist

Goal was to display Reason as its own language instead of showing it as OCaml.
This is ready for review, the Reason community would be super happy to see this merged 😄

\cc @lildude

ryyppy:
@pchaigno awesome, thanks for chiming in! I don't have push rights to this branch, but the Reason logo / most recognizable color in the Reason ecosystem is `#dd4b39` (orange-red) .

@amiralies could you add this color to the Reason config?

rafaelfesi:
@ryyppy Why was it added as **"Reason"** instead of **"ReasonML"**?

ryyppy:
Reason is the official language name, ReasonML is the "googleable name" afaik. kinda similarly on how go / rust call themselves rustlang / golang

rafaelfesi:
Thanks for the explanation, @ryyppy.

PR #4720, opened 2019-11-20
Add ".postcss" to PostCSS language

Akryum:
## Checklist:

      - https://github.com/search?q=extension%3Apostcss+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/Akryum/awesomejs.dev/blob/e59a65f6d318cfad9f340681e843d2de20d032fb/packages/frontend/src/assets/styles/components/markdown.postcss#L1
    - Sample license(s):
      - MIT

### DIFF
--- a/samples/PostCSS/sample.postcss
+++ b/samples/PostCSS/sample.postcss
@@ -1,0 +1,13 @@
+@define-mixin size $size {
+  width: $size;
+}
+
+$big: 100px;
+
+/* Main block */
+.block {
+  &_logo {
+    background: inline("./logo.png");
+    @mixin size $big;
+  }
+}
--- a/samples/PostCSS/sample.postcss
+++ b/samples/PostCSS/sample.postcss
@@ -1,13 +1,165 @@
-@define-mixin size $size {
-  width: $size;
-}
+.markdown {
+  @apply overflow-hidden;
 
-$big: 100px;
+  img {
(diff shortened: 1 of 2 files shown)


samples/PostCSS/sample.postcss
```diff
@@ -0,0 +1,13 @@
...
+
+$big: 100px;
+
+/* Main block */
+.block {
+  &_logo {
+    background: inline("./logo.png");
+    @mixin size $big;
+  }
+}
```
pchaigno: Where is this sample from? It looks very simple. Is it something you wrote specifically for this pull request?

We generally prefer more realistic samples as they are used to train the naive Bayes classifier.

Akryum: It's the existing postcss sample with the new extension. 😸

pchaigno: Ah. Any chance you could find another sample from a "real" application under a permissive license?

Akryum: https://github.com/Akryum/awesomejs.dev/blob/e59a65f6d318cfad9f340681e843d2de20d032fb/packages/frontend/src/assets/styles/components/markdown.postcss#L1 (MIT)

pchaigno: That looks fine :+1:

Issue #2418, opened 2015-05-23
VBA vs. Visual Basic


peterennis:
VBA != vb
I can only search for Visual Basic. It returns mostly .net code and that is just noise.
VBA should be separate.

pchaigno:
Do they use different sets of file extensions?

peterennis:
VBA is embedded in ms office apps.

If files are exported then default is .bas

Sometimes .mod for modules or

You can make up your own.


Distribution is often inside an office app

So extensions could be

Xls, xlsx, mdb, accdb, doc, docx etc. Ad nauseum...


Someone searching for VBA is typically

Not interested in .net


Thanks


Peter


E.g.

https://www.bing.com/search?q=peterennis+github&PC=U316&FORM=CHROMN


These are VBA repos, not visual basic, so the default github

Description is false


Sent from Yahoo Mail on Android

pchaigno:
[Linguist currently associates the following extensions with Visual Basic](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3390-L3406):
- `.vb`
- `.bas`
- `.cls`
- `.frm`
- `.frx`
- `.vba`
- `.vbhtml`
- `.vbs`

Which of these are VBA? Visual Basic?

peterennis:
Not .vb

Not .frx

Not .vbhtml

Not .vbs


However, I could create all them if I wanted :)


Removing those would at least reduce the search spam.


In the spirit of cross platform development, ms 

Honest embracement of os(?), and github expansion,

Perhaps you could have "Microsoft Office VBA"

As the 'language' selection?


Thanks,


Peter




Sent from Yahoo Mail on Android

NattyNarwhal:
VBA is actually just embedded classic VB. Modern VB.NET uses .vb, but classic VB will use the older extensions like .bas and .frm.

.vbs is VBScript, which is its own language altogether.

peterennis:
Yes. The issue us that search for VBA returns too much github spam as it brings back all results for VB. Ironically I find better results using google while searching the universe in comparison to the "targeted" subset of github.

chadebeauche:
I confirm :+1: 
.vb => VB.NET
.bas, .cls, .frm => VBA (or to be more precise, it's originally VB6 extensions)
.frx => binary file associated with .frm
.vbs => VB Script, sort of sub-langage based on VB6
.vba and .vbhtml => no idea

[...]

PR #4725, opened 2019-11-25
Separate Visual Basic into VB.NET, VBA/VB6 and VBScript

zspitz:
## Description
See #2418 .

## Checklist:
- [x] **I am adding a new language.** -- VBA/VB6
      -  https://github.com/search?q=extension%3Abas+%28%22End+Function%22+OR+%22End+Sub%22%29+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Acls+%28%22End+Function%22+OR+%22End+Sub%22%29+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Afrm+%28%22End+Function%22+OR+%22End+Sub%22%29+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Afrx+%28%22End+Function%22+OR+%22End+Sub%22%29+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Avba+%28%22End+Function%22+OR+%22End+Sub%22%29+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/VBA-tools/VBA-Dictionary/blob/master/dictionary.cls
      - https://github.com/VBA-tools/VBA-Dictionary/blob/master/specs/specs.bas
      - https://github.com/github/linguist/blob/master/samples/Visual%20Basic/cApplication.cls
      - https://github.com/github/linguist/blob/master/samples/Visual%20Basic/procedures.vba
    - Sample license(s): mit

- [x] **I am adding a new language.** -- VBScript
      -  https://github.com/search?p=3&q=extension%3Avbs+%28%22End+Function%22+OR+%22End+Sub%22%29+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/smjburton/vbase/blob/master/v_Data_ArrayList.vbs
    - Sample license(s): mit


zspitz:
Per [this suggestion](https://github.com/github/linguist/issues/2418#issuecomment-465794235) I am attempting to create a PR. However, I am running into the following issues:

1. I can't get a working Ruby install on my Windows box. This means I can't update the language ids, and update the grammar.yml file. It also means I have no way of confirming that tests actually succeed, short of committing and waiting for the GitHub actions to complete. (NB. I have no experience with Ruby, but if I had a working install I would manage somehow.) My next step is a Linux VM on a different Windows machine, but it would be helpful if there is some way to bypass this.

2. The `languages.yml` file is supposed to allow an entry as follows:

> ```
> # fs_name               - Optional field. Only necessary as a replacement for the sample directory name if the
> #                         language name is not a valid filename under the Windows filesystem (e.g., if it
> #                         contains an asterisk).
> ```

In this case, it would be useful for the `VBA/VB6` language entry, containing a `/`, which is an invalid Windows path character. But the test run still expects the samples to be in a folder called `VBA/VB6`, and fails on this.

lib/linguist/languages.yml
```diff
@@ -5488,6 +5488,49 @@ V:
   codemirror_mode: go
   codemirror_mime_type: text/x-go
   language_id: 603371597
+VB.NET:
```
lildude: I think it would be better to just append `.NET` to the original name, ie  "Visual Basic .NET".

```suggestion
Visual Basic .NET:
```

lib/linguist/languages.yml
```diff
@@ -5488,6 +5488,49 @@ V:
...
+  - visual basic
+  - visual basic .net
+  - vbnet
+  - vb .net
+  tm_scope: source.vbnet
+  ace_mode: text
+  codemirror_mode: vb
+  codemirror_mime_type: text/x-vb
+  language_id: 389
+VBA/VB6:
```
lildude: It's best not to have `/` in language names as this is very rare and can cause problems due to `/` being valid in *nix paths.

This is better named using the formal name: Visual Basic For Applications:

```suggestion
Visual Basic For Applications:
```

You'll need to adjust the samples' path too.

lib/linguist/languages.yml
```diff
@@ -5488,6 +5488,49 @@ V:
...
+  fs_name: VBA_VB6
```
lildude: You don't need this if you change the language name as suggested above.

lib/linguist/languages.yml
zspitz: Thanks for the suggestions and review. I'll write this in a separate commit.

lib/linguist/languages.yml
zspitz: > This is better named using the formal name: Visual Basic For Applications.

Out of curiosity, is this the Linguist convention? Or something specific about VBA?

Also, it's a little hard to say which should be the primary name of the language -- it's the same language, but with different global objects, and different environments (VBA programs are stored within other file types, and run in the context of a host application; VB6 usually runs standalone). Nevertheless, if I had to choose, I would go with Visual Basic for Applications, as it hasn't really been replaced by anything else, and is still widely used; unlike VB6, for which there are a number of better alternatives.

cbeauchesne:
@zspitz actually, VBA **is** VB6, it's the VB6 runtime used in dev mode (binary include debugger symbols, that allow the famous yellow line). It's exactly the mode used in the VB6 editor when you develop.

MS simply included the VB6 IDE inside Office Apps. For instance, you can see in VBA a legacy `compile` method on `VBProject` object, but calling it will raise an exception : "not supported for this project".

So, you will never find something in VBA that does not exists in VB6, and IMO, you should use VB6 for `fs_name`.

zspitz:
@cbeauchesne 

> @zspitz actually, VBA **is** VB6, it's the VB6 runtime used in dev mode (binary include debugger symbols, that allow the famous yellow line). It's exactly the mode used in the VB6 editor when you develop.
> 
> MS simply included the VB6 IDE inside Office Apps. For instance, you can see in VBA a legacy `compile` method on `VBProject` object, but calling it will raise an exception : "not supported for this project".
>

Not disagreeing, but I think that (as I said) VBA is far more commonly used than VB6, so it makes more sense that the language name should be VBA (or Visual Basic for Applications), rather than VB6.

> So, you will never find something in VBA that does not exists in VB6, and IMO, you should use VB6 for `fs_name`.

The reverse is also true to my knowledge -- in terms of the source of the language itself (not the global objects, or the runtime environment), you won't find anything in VB6 not in VBA

zspitz:
@lildude 

RE VBA vs Visual Basic for Applications -- The formal name isn't always used, or even preferred: e.g. HTML, CSS, XML, CSV; even when the acronym is not universally known to people unfamiliar with the language (such as GAMS, HLSL, VHDL). Therefore, can we keep the name as `VBA`?

I've updated the ids, but other tests are still failing.

cbeauchesne:
> The reverse is also true to my knowledge -- in terms of the source of the language itself (not the global objects, or the runtime environment), you won't find anything in VB6 not in VBA

Partially true :) It's the same runtime, but you'll find several things not possible, or hardly doable in VBA. I've in mind options for default immutable value for objects (`x = range.value` is equivalent to `x = range.value` for excel range, and object iterators (`for each item in myObject`). On VB6, it's a options in the IDE, and it's saved as a header of your .cls file. In VBA, the option is missing in the IDE, and if you export the file, you'll never find the header. But, if you import a cls file with the good header, you can add  this behavior (and if you modify your file in VBA IDE, you loose your hack :( ).

I used to build my own build tooling for VBA for this use-case.

That said, I agree that's very small differences, and as VBA usage is far more common than VB6, the less common surprise effect is probably a better argument.

lildude:
> Therefore, can we keep the name as `VBA`?

Yeah, you can keep VBA. I was more trying to draw things away from the previous attempt of VBA/VB6 whilst keeping things unambiguous. I think it's commonly known that VBA is referring to Visual Basic for Applications as opposed to any other language.

zspitz:
@lildude Is there anything I can add/change to move this forward? Thanks.

lildude:
> @lildude Is there anything I can add/change to move this forward? Thanks.

Not from what I can see. Just waiting on a review from one of the other maintainers before I make my final review and merge.

zspitz:
Out of curiosity, at what point will this be visible in existing repos? I understand from the Linguist README that GitHub only performs this analysis when code is pushed to a repo. Is there ever another time when this happens?

lildude:
> Out of curiosity, at what point will this be visible in existing repos? I understand from the Linguist README that GitHub only performs this analysis when code is pushed to a repo. Is there ever another time when this happens?

The rename should take effect in _some places_ when I make the next release, but only for "Visual Basic" -> "Visual Basic .NET" rename as the language ID is the same. Everything else will only take affect on those repos on the next push.  The reason I say "some places" is because different parts of the site are cached differently and I have no idea how they all tie in.

lildude:
These changes are now live:

Screenshot 2019-12-06 at 16 43 31

Anything not correctly reflected will need a push, and if that doesn't work, contact GitHub support as something else will be caching the old name.

PR #4730, opened 2019-12-01
fix errors on non-UTF-8 encodings

smola:
## Description
Some files failed with "invalid byte sequence in UTF-8 (ArgumentError)"
when BlobHelper#lines was called. Some problematic files include
UTF-16LE samples such as test/fixtures/Data/utf16le.

Errors were not present when computing stats from git repositories,
since git blobs are always read as ASCII-8BIT and that was working
correctly. However, when using FileBlob, encoding could be ASCII-8BIT,
UTF-8 or other, depending on the runtime value of Encoding.external_encoding.

Tests were not catching the error since they were forcing
Encoding.external_encoding to be ASCII-8BIT (introduced in #1211). So the
error would only be seen in wild usage (see issue #353).

This commit forces ASCII-8BIT on File.read calls. The error is still
present if using memory blobs with other encodings.

### DIFF
--- a/lib/linguist/file_blob.rb
+++ b/lib/linguist/file_blob.rb
@@ -35,7 +35,7 @@ module Linguist
     #
     # Returns a String.
     def data
-      @data ||= File.read(@fullpath)
+      @data ||= File.read(@fullpath, :encoding => "ASCII-8BIT")
     end
 
     # Public: Get byte size
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -17,7 +17,7 @@ end
 
 def fixture_blob_memory(name)
   filepath = (name =~ /^\//)? name : File.join(fixtures_path, name)
-  content = File.read(filepath)
+  content = File.read(filepath, :encoding => "ASCII-8BIT")
   Linguist::Blob.new(name, content)
 end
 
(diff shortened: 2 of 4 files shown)


smola:
Note that the pull request removes tests about `BlobHelper#lines` encoding. These were introduced in #1211:

> @alindeman Awkwardly enough, we've made an implicit guarantee that the data exposed through lines is encoded in its original form (in practice, ASCII-8BIT or UTF-8) even if the file is detected as a different encoding. Certain other code seems to actually depend on this.

Strictly speaking, this guarantee is still in place: `lines` result is `ASCII-8BIT` just as the files are read. However, that wouldn't be the case if an in-memory blob is created. I lack context of the original issue about this behavior, so it's hard to tell if this change is breaking any third party code.

lildude:
A quick test suggests this does the trick for the samples mentioned in https://github.com/github/linguist/issues/353. More extensive testing would be needed on GitHub.com as I don't have the context around the comment you referenced @smola and I have no idea how this change would impact GitHub.com without testing it.

I'll try schedule some time this week to test this locally and in our staging env.

lildude:
Missed the deploy window before the Christmas freeze so will try test this sometime in the next two-to-three weeks, depending on availability etc.

lildude:
I've done some testing over the past few days and haven't spotted any problems so I'm going to approve and merge this. Thanks @smola 🙇

PR #4734, opened 2019-12-05
Add syntax highlighting for Vim help files

Alhadis:
## Description
This PR adds syntax highlighting for Vim help files, which are `.txt` files identified using a modeline:

	vim:tw=78:ts=8:noet:ft=help:norl:

I was reluctant to use `help` as an alias, because it's quite generic, and Emacs has its own `help` mode. Instead, I added a heuristic that tests for a Vim-style modeline, falling back to `Text` if one wasn't found.

It's somewhat hacky, but this approach is the safest I could think of that wouldn't screw with the classifier or cause any misidentified files.

## Checklist:
	- [x] **The extension of the new language is used in hundreds of repositories:**
		- [~838,147 results](https://github.com/search?q=extension%3Atxt+vim+ft+help&type=Code)
	- [ ] ~~I have included a real-world usage sample for all extensions added in this PR~~
	- [x] **I have included a syntax highlighting grammar:**
		- [Preview](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-viml%2Fblob%2Fmaster%2Fgrammars%2Fhelp.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fvim%2Fvim%2Fblob%2Fmaster%2Fruntime%2Fdoc%2Fediting.txt&code=)
	- [x] **I have updated the heuristics to distinguish my language from others using the same extension.**


lildude:
Got some conflict that need resolving if you still want to pursue this when you're not busy @Alhadis

Alhadis:
On it. 😉  Actually forgot I still had this PR open…

samples/Vim Help File/modeline.txt
```diff
@@ -0,0 +1 @@
+ vim:tw=78:ts=8:noet:ft=help:norl:
```
pchaigno, line 1: Could you replace this with a more realistic sample files?

Alhadis: Uh, no. 😜 That'll cause random `.txt` files to be misclassified as "Vim Help"...

pchaigno: How is that?

Alhadis: Because depending on what sample I use, Linguist is going to disambiguate two `.txt` files based on whichever words they most have in common with the sample. Adding a file with a modeline in it is the least disruptive thing I could think of.

pchaigno: > Linguist is going to disambiguate two `.txt` files based on whichever words they most have in common with the sample.

Unless I'm missing something, this is only true if we're using the Bayesian classifier. Since you have a fallback to Text in your heuristic rule, we will never use the Bayesian classifier.

Alhadis: Then what's the point of having a realistic sample?

lib/linguist/heuristics.yml
```diff
@@ -498,6 +498,13 @@ disambiguations:
     pattern: '^\s*(import.+(from\s+|require\()[''"]react|\/\/\/\s*<reference\s)'
   - language: XML
     pattern: '(?i:^\s*<\?xml\s+version)'
+- extensions: ['.txt']
+  rules:
+    # The following RegExp is simply a collapsed and simplified form of the
+    # modeline-checking ptattern in `./lib/linguist/strategy/modeline.rb`.
```
lildude, line 504: Typo

```suggestion
    # modeline-checking pattern in `./lib/linguist/strategy/modeline.rb`.
```

samples/Vim Help File/modeline.txt
pchaigno: Sorry for the long delay.

Ok, I tried to think of other solutions to avoid having unrealistic files in `samples/`, but can't think of a better way :/
Could you however add a comment in the heuristic rule above the fallback to explain this (i.e., state that the `Text` fallback shouldn't be removed because we have unrealistic samples and can't rely on the Bayesian classifier)? I just want to make sure we don't end up using unrealistic samples in the future.

Alhadis: I'll, uh, just leave a permalink to that comment…

pchaigno (approved):

lildude (approved):

Alhadis:
> You can’t delete this branch because it has an open pull request

… wait, what?

Figure 2

Alhadis:
@lildude Any idea what's going on with the `vim-help` branch?  I ask only because I prune remote-tracking branches locally after a PR… and my usual OCD is itching to clean up after myself. 😢

In other words, this might be indicative of a bug.

lildude:
🤔 good question. I have no idea. Lemme see what I can find in the backend.

lildude:
No idea what was going on but I've managed to delete it manually using `git push origin --delete vim-help`

Alhadis:
Thanks! (That was driving me *bonkers*…)

Issue #4750, opened 2019-12-18
Coq / Verilog misdetection


0x00a1e9:
# Coq / Verilog misdetection

## Problem Description

Some Verilog files are misclassified as Coq files in [my repository](https://github.com/0x00a1e9/selevy).  I'm suspicious that Verilog synthesis attribute can cause this. These should be classified as Verilog, I think.

### URL of the affected repository:

https://github.com/0x00a1e9/selevy

### Last modified on:

2019-12-15

### Expected language:

Verilog

### Detected language:

Coq

pchaigno:
Any reason you removed the checkboxes from the templates? Looks like the second and third would have been useful to you...

0x00a1e9:
I just thought people didn't check items in the template wouldn't open an issue here: then I deleted.

I read the whole you mentioned.  So I know there's a way like a manual overrides feature, and linguists cannot do perfect detection. But I opened this just because I thought it's good if the problem  I mentioned is kind of one can be fixed.

lildude:
> I opened this just because I thought it's good if the problem I mentioned is kind of one can be fixed.

It certainly can, but it's going to need a PR from someone that understands Coq and Verilog as they both share the same extension and thus the heuristic at https://github.com/github/linguist/blob/643c091e8d1e8c20401d1267ca558d304ebae8ca/lib/linguist/heuristics.yml#L460-L467 ... needs to be enhanced.

If you know of a good regex that can improve the detection rate, please feel free to submit a pull request. If no one submits a PR, this issue is likely to go stale and be automatically closed.

pchaigno:
> I just thought people didn't check items in the template wouldn't open an issue here

Unfortunately, people who don't read that information open issues here all the time.

To complete @lildude's answer, you can also try to add new sample files, if you see that the current ones are not representative enough.

0x00a1e9:
I figured out that 

https://github.com/github/linguist/blob/643c091e8d1e8c20401d1267ca558d304ebae8ca/lib/linguist/heuristics.yml#L462-L463

the first pattern for Coq could also match Verilog.

PR #4751, opened 2019-12-18
Revise pattern to fix Coq/Verilog misclassification

0x00a1e9:
#4750 

Revise some patterns for avoiding Coq/Verilog misdetection.

## Description

https://github.com/github/linguist/blob/643c091e8d1e8c20401d1267ca558d304ebae8ca/lib/linguist/heuristics.yml#L462-L463

The first pattern for Coq could also match Verilog synthesis attribute; then I revised patters for Coq and Verilog.  

## Checklist:
    - Sample source(s): samples/Verilog/ram.v
    - Sample license(s): MIT

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -460,9 +460,9 @@ disambiguations:
 - extensions: ['.v']
   rules:
   - language: Coq
-    pattern: '\(\*.*?\*\)|(?:^|\s)(?:Proof|Qed)\.(?:$|\s)|(?:^|\s)Require[ \t]+Import\s'
+    pattern: '(?:^|\s)(?:Proof|Qed)\.(?:$|\s)|(?:^|\s)Require[ \t]+(Import|Export)\s'
   - language: Verilog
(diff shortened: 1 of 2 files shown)


0x00a1e9:
I added [a sample Verilog file](samples/Verilog/ram.v) used to be misclassified as Coq under the MIT license.

pchaigno:
That Verilog sample files looks fairly different from those we already have (and as a result, the syntax highlighting doesn't work very well on that one). Any idea why that is? Is that a different dialect? In particular, the comments are different and they're the reason why the heuristic was failing.

0x00a1e9:
>In particular, the comments are different and they're the reason why the heuristic was failing.

I only use comment syntax defined in IEEE Std 1364-2005. What did you mention to by "comments"? Seems there's only one comment part in the sample...

pchaigno:
> What did you mention to by "comments"? Seems there's only one comment part in the sample...

That's the one I was referring to.

I'd really like to have @0x00a1e9 or @S-YOU's eyes on this.

0x00a1e9:
When there's a no comment part in the sample, the heuristic fail. This problem is caused because it has a bad pattern for Coq, which could match "Attributes Syntax" in Verilog, I thiknk.

0x00a1e9:
Plz have a look at this once again

PR #4752, opened 2019-12-18
Replace outdated Fortran tmbundle with Atom definition

ararslan:
The TextMate bundle for Fortran appears not to be maintained and does not seem to support modern, free-source Fortran. (Or if it does, its definitions are not being applied appropriately on GitHub, as all Fortran files show up as though they're fixed source.) This change swaps out the TextMate definition for David Parkinson's Atom definition, which supports both fixed- and free-source.

## Checklist:

  - Old: https://github.com/antoine-levitt/ACCQUAREL/blob/master/src/basis.f90
  - New: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fgithub.com%2Fdparkins%2Flanguage-fortran%2Fblob%2Fmaster%2Fgrammars%2Ffortran%2520-%2520free%2520form.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fantoine-levitt%2FACCQUAREL%2Fblob%2Fmaster%2Fsrc%2Fbasis.f90&code=

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -377,9 +377,6 @@
 [submodule "vendor/grammars/forth"]
 	path = vendor/grammars/forth
 	url = https://github.com/textmate/forth.tmbundle
-[submodule "vendor/grammars/fortran.tmbundle"]
-	path = vendor/grammars/fortran.tmbundle
-	url = https://github.com/textmate/fortran.tmbundle
 [submodule "vendor/grammars/gap-tmbundle"]
 	path = vendor/grammars/gap-tmbundle
 	url = https://github.com/dhowden/gap-tmbundle
@@ -1022,3 +1019,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/language-fortran"]
+	path = vendor/grammars/language-fortran
+	url = https://github.com/dparkins/language-fortran
--- a/grammars.yml
+++ b/grammars.yml
@@ -316,9 +316,6 @@ vendor/grammars/fish-tmbundle:
 - source.fish
 vendor/grammars/forth:
 - source.forth
-vendor/grammars/fortran.tmbundle:
-- source.fortran
-- source.fortran.modern
 vendor/grammars/gap-tmbundle:
 - source.gap
 vendor/grammars/gdscript:
@@ -452,6 +449,12 @@ vendor/grammars/language-fontforge:
 - source.opentype
 - source.xlfd
 - text.sfd
+vendor/grammars/language-fortran:
+- source.fortran.fixed
+- source.fortran.free
+- source.fortran.modern
+- source.fortran.preprocessor
+- source.fortran.punchcard
 vendor/grammars/language-gf:
 - source.gf
 vendor/grammars/language-gfm:


ararslan:
Actually, something is off with how I set things up, probably because the local thing failed due to a missing Gem and I tried to finish it off by hand. But really it seems the problem is not related to the grammar but rather with how the `tm_scope` is applied: it should be different based on the file extension.

grammars.yml
```diff
@@ -316,9 +316,6 @@ vendor/grammars/fish-tmbundle:
 - source.fish
 vendor/grammars/forth:
 - source.forth
-vendor/grammars/fortran.tmbundle:
-- source.fortran
```
pchaigno: It looks like the new grammar doesn't have this scope name, so you need to update `languages.yml` to point to the appropriate new scope name.

ararslan:
@pchaigno, is it possible to say that for file extensions .f, .f77, and .for, the `tm_scope` is `source.fortran`, and for .f90, .f08, ... it's `source.fortran.modern`? I think that's the root of the issue here.

.gitmodules
```diff
@@ -1022,3 +1019,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/language-fortran"]
```
pchaigno: Did you add this manually? There's a script to switch grammars, please see [*Changing the source of a syntax highlighting grammar*](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#changing-the-source-of-a-syntax-highlighting-grammar).

ararslan: I did use the script, but it failed partway through, so I tried to finish off the changes by hand. It seems they're incomplete.

ararslan:
```
$ script/add-grammar --replace fortran.tmbundle https://github.com/dparkins/language-fortran
Checking docker is installed and running
$ docker ps
Deregistering: vendor/grammars/fortran.tmbundle
$ git submodule deinit vendor/grammars/fortran.tmbundle
$ git rm -rf vendor/grammars/fortran.tmbundle
$ script/grammar-compiler update -f
Registering new submodule: vendor/grammars/language-fortran
$ git submodule add -f https://github.com/dparkins/language-fortran vendor/grammars/language-fortran
$ script/grammar-compiler add vendor/grammars/language-fortran
Confirming license
$ script/licensed
  > /Library/Ruby/Site/2.3.0/bundler/resolver.rb:287:in `block in verify_gemfile_dependencies_are_found!': Could not find gem 'rake-compiler (~> 0.9)' in any of the gem sources listed in your Gemfile. (Bundler::GemNotFound)
  > 	from /Library/Ruby/Site/2.3.0/bundler/resolver.rb:255:in `each'
  > 	from /Library/Ruby/Site/2.3.0/bundler/resolver.rb:255:in `verify_gemfile_dependencies_are_found!'
  > 	from /Library/Ruby/Site/2.3.0/bundler/resolver.rb:49:in `start'
  > 	from /Library/Ruby/Site/2.3.0/bundler/resolver.rb:22:in `resolve'
  > 	from /Library/Ruby/Site/2.3.0/bundler/definition.rb:258:in `resolve'
  > 	from /Library/Ruby/Site/2.3.0/bundler/definition.rb:170:in `specs'
  > 	from /Library/Ruby/Site/2.3.0/bundler/definition.rb:237:in `specs_for'
  > 	from /Library/Ruby/Site/2.3.0/bundler/definition.rb:226:in `requested_specs'
  > 	from /Library/Ruby/Site/2.3.0/bundler/runtime.rb:108:in `block in definition_method'
  > 	from /Library/Ruby/Site/2.3.0/bundler/runtime.rb:20:in `setup'
  > 	from /Library/Ruby/Site/2.3.0/bundler.rb:107:in `setup'
  > 	from /Library/Ruby/Site/2.3.0/bundler/setup.rb:20:in `<top (required)>'
  > 	from /Library/Ruby/Site/2.3.0/rubygems/core_ext/kernel_require.rb:54:in `require'
  > 	from /Library/Ruby/Site/2.3.0/rubygems/core_ext/kernel_require.rb:54:in `require'
  > 	from script/licensed:3:in `<main>'
Command failed. Aborting.
```
Notably the license was not updated.

ararslan:
I'm going to close this, since it's incomplete and the issue is actually elsewhere.

pchaigno:
> @pchaigno, is it possible to say that for file extensions .f, .f77, and .for, the `tm_scope` is `source.fortran`, and for .f90, .f08, ... it's `source.fortran.modern`? I think that's the root of the issue here.

Yes, you can break the language entry in two in `languages.yml`. One will be Fortran and the other a dialect for instance. Then you set `group: Fortran` for the dialect. This way, the dialect can have its own grammar but it will appear as Fortran in statistics, searches, etc.

This is not the only issue though. It looks like you switched the grammar manually, which is not a good idea when it comes to Linguist. We have scripts to do that because there's a bunch of things to update.

ararslan:
I've done as you suggest in #4753 but I'll need some help with it if you're willing.

PR #4760, opened 2020-01-06
Add interpreters 'csh' and 'tcsh' for language 'Tcsh'

friedc:
Make GitHub automatically recognize files with `#!/bin/csh` and `#!/bin/tcsh` shebangs as Tcsh.
## Checklist:

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5265,6 +5265,9 @@ Tcsh:
   extensions:
   - ".tcsh"
   - ".csh"
+  interpreters:
+  - tcsh
+  - csh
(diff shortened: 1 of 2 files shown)


friedc:
There is for example
- https://github.com/barlage/WRF-kill/blob/master/tools/regtest_hwrf.csh
- https://github.com/barlage/WRF-kill/blob/master/tools/regtest_nmmnest.csh

Are they complicated enough?

pchaigno:
Yes, I think one of those would suffice! They're pretty big, but since we don't have any samples for Tcsh so far, it's probably okay :+1:

PR #4761, opened 2020-01-06
Add syntax highlighting to Yarn lockfiles

arcanis:
## Description

Yarn 1 uses a Yaml-like format which is [good enough](https://gist.github.com/arcanis/899cdb2ed30eeba45a0730d9af0f36bc) with the YAML highlighting, and Yarn 2 will switch to the good old regular YAML. Adding syntax highlighting to this file will help JS developers to review the changes, potentially preventing security issues.

## Checklist:

      - https://github.com/search?q=filename%3Ayarn.lock&type=Code
    - Sample source(s):
      - https://gist.github.com/arcanis/899cdb2ed30eeba45a0730d9af0f36bc
      - https://github.com/yarnpkg/berry/blob/master/yarn.lock (truncated sample included).


Alhadis:
FYI: you can use [Lightshow](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=source.yaml&grammar_format=auto&grammar_url=&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fokkyj%2Fokkysample2%2Fblob%2F8e748634eb5df63582834e73ed7e1c68e90d071b%2Fnode_modules%2Fformidable%2Fyarn.lock&code=) to show how files would be rendered with syntax highlighting. 😉

Alhadis:
> Maybe we should add them to `generated.rb`?

#4459, #4445

pchaigno:
Ah, thanks @Alhadis!

arcanis:
Samples are here, sorry for the wait 🙂

arcanis:
Hm keeping both versions seems to make the tests unhappy - how should I fix it? Add a single version?

Also, should I add `yarn.lock` to `vendor.yml`?

Alhadis:
> Hm keeping both versions seems to make the tests unhappy - how should I fix it?

Move the contents of `samples/YAML/yarn/*` to just `samples/YAML/*`. The exact error is [here](https://github.com/github/linguist/pull/4761/checks?check_run_id=472193283#step:5:60):

	Errno::EISDIR: Is a directory @ io_fread - /home/runner/work/linguist/linguist/samples/YAML/yarn

> Also, should I add `yarn.lock` to `vendor.yml`?

That shouldn't be necessary as YAML is a data-type language, so it won't skew or show up in a repository's language stats.

arcanis:
> Move the contents of samples/YAML/yarn/* to just samples/YAML/*. The exact error is here:

The problem is that the detection for both yarn formats has the same filename, so I can't have both in the same directory.

Alhadis:
Then omit whichever one has the conflicting filename; only one is needed.

Which reminds me; this should be going in [`samples/YAML/filenames`](https://github.com/github/linguist/tree/master/samples/YAML/filenames) if it's an entry being added to the `filenames` array (instead of `extensions`).

Alhadis:
Ouch. I just saw the size of the file you're adding... a 26,114-line sample is *probably* a bit overkill. 😉

Could you add a much more minimal sample, please? Something around the ~1-5 KBs range will do. Otherwise, it's going to nuke our classifier.

arcanis:
@lildude the sample is in the OP, it's the lockfile from the Yarn project, licensed BSD 2-clause.

lildude:
> @lildude the sample is in the OP, it's the lockfile from the Yarn project, licensed BSD 2-clause.

That's not the same file. https://github.com/yarnpkg/berry/blob/master/yarn.lock is 14757 lines long. The sample you've included is only 30 lines long.

arcanis:
Yes, because

> Could you add a much more minimal sample, please? Something around the ~1-5 KBs range will do. Otherwise, it's going to nuke our classifier.

I trimmed the file to a more reasonable size. Check the start of the file, it's the same content.

lildude:
> I trimmed the file to a more reasonable size. Check the start of the file, it's the same content.

Ah, that explains it and isn't clear. I've updated the OP to reflect this and the license. Thanks.

PR #4762, opened 2020-01-06
Add leex extension to HTML+EEX

lpil:
The Elixir template language sometimes uses the extension `.leex`, when used with the LiveView library.

This is documented here https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-assigns-and-liveeex-templates

I've not added an example as it is the same file contents as the regular `.eex` version. Shall I add this?

## Checklist:

      - https://github.com/search?q=extension%3Aleex+div&type=Code
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1996,6 +1996,7 @@ HTML+EEX:
   - eex
   extensions:
   - ".eex"
+  - ".leex"
   ace_mode: text
   codemirror_mode: htmlmixed
   codemirror_mime_type: text/html


lib/linguist/languages.yml
```diff
@@ -1996,6 +1996,7 @@ HTML+EEX:
   - eex
   extensions:
   - ".eex"
+  - ".leex"
```
pchaigno: Looks like these files are all named `.html.leex`. Linguist supports such extensions, so we could use it instead of just `.leex`.

pchaigno:
For other reviewers: Usage looks okay. I downloaded 1139 `.html.leex` (which is likely all such files) and counted 486 repositories by 403 users.

lib/linguist/languages.yml
lpil: `.html.leex` is a convention in the Phoenix framework to denote that the template is used to render HTML but it could be something else. You'll find that `.html.eex` is also common, even though the canonical extension is `.eex`.

pchaigno: I understand, but we try to follow usage, not standards/conventions. For example, if convention dictates that C++ users should use `.cpp`, but a large portion use `.c`, we'll keep recognizing `.c` as a C++ extension.

In `.leex`'s case, I have trouble finding even a single `.leex` file that is not `.html.leex`. Since the same seems to be true for `.eex`, we should also change it to `.html.eex`.

lpil: `.html.leex` seems OK but removing `.eex` will remove highlighting from non-Phoenix EEx projects. EEx is part of the Elixir language itself, not part of the Phoenix framework.

lpil: Here's the official documentation for EEx, which specified `.eex`, not `.html.eex`. https://hexdocs.pm/eex/EEx.html#eval_file/3-examples

In my work we use EEx to render HTML, XML, markdown, and Javascript, so adding a `.html.eex` extension would be inaccurate. The template language is not related to HTML.

pchaigno: You're editing the HTML+EEx language here though. Do you have any examples of HTML+EEx files on GitHub that do not use `.html.eex`?

lpil: To expand a little bit, in Phoenix `index.html` is the name of the template file, and `.eex` is the extension.

lpil: I've now left the company I'm afraid so I don't have access to them, but projects that use EEx without using Phoenix are the likely candidates. Plug projects for example. I'm not sure how to find any using GitHub's search

lpil:
An unrelated question around the eex extension seems to have got mixed up in this PR for the leex extension, so I'm going to close this.

lib/linguist/languages.yml
lpil: Here's the official vim editor integration that shows the supported extensions: https://github.com/elixir-editors/vim-elixir/blob/057ac39b5982a2decf52d48cffc405ff6a666ca5/ftdetect/elixir.vim#L2

PR #4776, opened 2020-01-18
Fix the regex for vscode vendored files

mgriffin:
## Description

Doing a little bit if kinda unrelated work, I found that the regexes that are defined in `lib/linguist/vendor.yml` are incorrect and caused a match against a file with `vscode` in it. This shouldn't have happened as the regex is supposed to be for a file or directory named `.vscode` only.

This PR adds in a simple test to make sure I knew what I was doing and then fixes the regex so that it matches what is expected. It tests to make sure that these strings are not matched:

- `testing-vscode-testing`
- `testing.vscode.testing`
- `.vscode-testing`

Looking further at the regexes, there are a lot of other ones that will need to be changed as well. If this PR gets accepted, I can work on the others in another PR. I decided not to change them all in one go as this has the potential to change how this library works.

## Checklist:


lildude:
@mgriffin what @Alhadis said. It would be great if you could fix 'em all in this PR. 🙇

mgriffin:
I'm working on it, I noticed that there are already some tests in place in https://github.com/github/linguist/blob/master/test/test_file_blob.rb#L292 so I'm doing a little moving around to accommodate that. 👍

mgriffin:
@lildude I think this is getting close to being ready to merge. Can you throw an 👁 over when you get a chance?

mgriffin:
@Alhadis @lildude Does this look like it's okay to merge now?

lib/linguist/vendor.yml
```diff
@@ -127,44 +127,44 @@
...
 
 # Sublime Text workspace files
-- .sublime-project
-- .sublime-workspace
+- \.sublime-project
+- \.sublime-workspace
 
 # VS Code workspace files
-- .vscode
+- ^\.vscode/
```
pchaigno: Should this also match `/somedir/.vscode`?

lib/linguist/vendor.yml
```diff
@@ -349,23 +349,23 @@
...
+- \.[Dd][Ss]_[Ss]tore$
 
 # R packages
 - ^vignettes/
 - ^inst/extdata/
 
 # Octicons
-- octicons.css
-- sprockets-octicons.scss
+- (^|/)octicons\.css
```
pchaigno: If they're all interpreted as regular expressions, shouldn't we add this `(^|/)` pattern in front of all entries (e.g., `stuff\.framework` entries above)?

lib/linguist/vendor.yml
mgriffin: Good point, yes, I suspect it should.

lib/linguist/vendor.yml
mgriffin: That's a huge maybe from me 😄 

I'm not 100% sure of all the strings that can be used, so I didn't want to do this to them all in one go.

lib/linguist/vendor.yml
pchaigno: Could you update the regex so that it does?

lib/linguist/vendor.yml
pchaigno: Ok. How did you choose where to add `(^|/)` and where not to add it then?

mgriffin: They were the ones I either knew about or was able to find out about quickly. I didn't do an exhaustive search to check the others.

pchaigno: I think it's okay to add it everywhere unless it's obvious that the expression should match the end of a filename (e.g., file extension). I'm not sure we have any such case anyway. Worst-case scenario, we break one or two of those and we'll fix it quickly afterward when users come complain.

mgriffin:
I think the test failures are not related to this PR, is anyone able to review it and merge it?

lildude:
@mgriffin Thanks for this. I've merged master into this PR and it looks like the failures now are legit.

mgriffin:
@lildude I've fixed things up now so all the tests are passing again.

PR #4777, opened 2020-01-19
Add Arend language

ice1000:
## Checklist:
      
      -  Search with `func`: https://github.com/search?q=extension%3Aard+func+NOT+nothack&type=Code
      -  Search with `import`: https://github.com/search?q=extension%3Aard+import+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/ice1k/linguist/blob/arend/samples/Arend/SeqColimit.ard
    - Sample license(s):
  
### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -212,6 +212,9 @@
 [submodule "vendor/grammars/applescript.tmbundle"]
 	path = vendor/grammars/applescript.tmbundle
 	url = https://github.com/textmate/applescript.tmbundle
+[submodule "vendor/grammars/arend-tmbundle"]
+	path = vendor/grammars/arend-tmbundle
+	url = https://github.com/arend-lang/arend-tmbundle
 [submodule "vendor/grammars/asciidoc.tmbundle"]
 	path = vendor/grammars/asciidoc.tmbundle
 	url = https://github.com/zuckschwerdt/asciidoc.tmbundle
--- a/grammars.yml
+++ b/grammars.yml
@@ -174,6 +174,8 @@ vendor/grammars/api-blueprint-sublime-plugin:
 - text.html.markdown.source.gfm.mson
 vendor/grammars/applescript.tmbundle:
 - source.applescript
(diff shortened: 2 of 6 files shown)


ice1000:
Also, running `script/bootstrap` also requires `libssl-dev` on ubuntu, which is not installed by default. It took me about half an hour to figure this out, since the error I saw is "cannot find target to make", instead of "cannot find OpenSSL".

ice1000:
The license is the MIT license

ice1000:
Yay! I made the CI passes!

ice1000:
I hope so, let's see.

ice1000:
There are 300+ files in the search result now :)

ice1000:
There's now 340 😟 not much.

ice1000:
Now 360

pchaigno:
Those 360 files make up 27 repositories by 19 users. So it looks like this is growing, albeit very slowly. I'm going to close this issue, but don't hesitate to ping me to reopen if you think we've reached the required threshold of hundreds of repositories.

ice1000:
Now 406

pchaigno:
@ice1000 It's good that you're keeping an eye on the number of files, but I don't think it's very useful to post it here unless the number of repositories significantly went up (e.g., +100) or the number of files itself went up (e.g., +1000).

ice1000:
Ok

ice1000:
How do I count the repositories? Is there an API?

pchaigno:
You can use [Harvester](https://github.com/Alhadis/Harvester).

ice1000:
> @ice1000 It's good that you're keeping an eye on the number of files, but I don't think it's very useful to post it here unless the number of repositories significantly went up (e.g., +100) or the number of files itself went up (e.g., +1000).

Hi! It looks like there's 1.8k Arend files now!

Search results: https://github.com/search?q=path%3A*.ard+AND+%28import+OR+func%29&type=code

PR #4779, opened 2020-01-23
Add minus-prefixed gitattribute overrides for tests

lildude:
## Description
This PR adds "minus-prefixed" `.gitattribute` overrides needed for testing the support for minus-prefixed gitattribute negation/unsetting/overriding added in https://github.com/github/linguist/pull/4780

CI isn't going to pass on this branch as it is soooo far behind and only really used for testing master and branches of master. We should really remove everything that aren't samples or related to the overrides testing from this branch.

_[Rest of template removed as it doesn't apply]_


.gitattributes
```diff
@@ -7,3 +7,7 @@ README.md linguist-documentation=false
 samples/Arduino/* linguist-documentation
 samples/Markdown/*.md linguist-detectable=true
 samples/HTML/*.html linguist-detectable=false
+samples/CSS/bootstrap.css -linguist-vendored
+samples/CSS/bootstrap.min.css -linguist-generated
+LICENSE -linguist-documentation
+samples/CoffeeScript/browser.coffee -linguist-detectable
```
pchaigno: No newline? :disappointed:

lildude: Arse!!

lildude: Fixed 😁

PR #4781, opened 2020-01-23
Cleanup the test/attributes branch to clearly indicate purpose

lildude:
## Description

As I noted in https://github.com/github/linguist/pull/4779, we really should clean up this `test/attributes` branch and make it very clear what its purpose is.  

This PR does just that. It also removes the very old Travis config as we don't need it and we don't need tests on this branch.


lildude:
FYI @pchaigno and @Alhadis I've just enabled "Automatically delete head branches" cos computers are much better at menial tasks than humans.

PR #4783, opened 2020-01-25
Add 2 more JavaScript interpreters

Alhadis:
This PR registers ~~4~~ 2 more JavaScript interpreters:

- [Gnome JS](https://gitlab.gnome.org/GNOME/gjs/wikis/Home): `gjs`
- ~~[MuJS](https://www.mujs.com/): `mujs`~~ <sup>[[1]](#1)</sup>
- ~~[QuickJS](https://bellard.org/quickjs/): `qjs`~~ <sup>[[1]](#1)</sup>
- [SlimerJS](https://slimerjs.org/): `slimerjs`

*Template removed as it doesn't really apply.*

-------
<a name="1"><sup>[1]</sup></a> Removed after <a href="https://github.com/github/linguist/pull/4783#pullrequestreview-348321932">discussion</a> with @pchaigno

PR #4784, opened 2020-01-25
Add Yarn releases to vendored list

SiAdcock:
## Description

Yarn 1 and 2 allow developers to check in a specific version of Yarn for their project ([`yarn policies set-version`](https://legacy.yarnpkg.com/en/docs/cli/policies) in v1 and [`yarn set version`](https://yarnpkg.com/cli/set/version) in v2). This ensures all developers use the same version of Yarn when working on the project. This adds ~147k lines of JavaScript to the project, skewing the project's language stats. 

This change ensures the generated files stored under `.yarn/releases` are vendored from Linguist's perspective.

## Checklist:
  
### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -41,6 +41,9 @@
 # Node dependencies
 - node_modules/
 
+# Yarn releases
+- .yarn/releases
+
 # esy.sh dependencies
 - (^|/)_esy$
 
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -42,7 +42,7 @@
 - node_modules/
 
 # Yarn releases
-- .yarn/releases
+- (^|/)\.yarn/releases/
 
 # esy.sh dependencies
 - (^|/)_esy$


SiAdcock:
I think those test failures are [unrelated to this change](https://github.com/github/linguist/pull/4783#pullrequestreview-348323900), but let me know if you need any more info or if there's anything I can improve 🙂

lib/linguist/vendor.yml
```diff
@@ -41,6 +41,9 @@
 # Node dependencies
 - node_modules/
 
+# Yarn releases
+- .yarn/releases
```
Alhadis: This is a regular expression, so the dot needs to be escaped to avoid matching something like `eyarn/releases` or what-have-you.

In addition, this will also target files named `releases` that simply reside in a directory whose name *ends* with `yarn`, such as:

~~~ignorelist
/node-yarn/releases
~~~

I recommend replacing this with:

~~~yaml
# Yarn releases
- (^|/)\.yarn/releases/
~~~

(Note that the other patterns in this file have similar mistakes; this was only noticed [very recently](https://github.com/github/linguist/pull/4776#pullrequestreview-344970330), hence why they're not fixed).

SiAdcock:
```suggestion
- (^|/)\.yarn/releases/
```

SiAdcock: Great, thanks for your suggestions @Alhadis! I've updated accordingly

PR #4788, opened 2020-01-29
Bump version string to v7.8.0

lildude:
My commit to bump the version string to v7.8.0 got lost in the wash somewhere, so here it is again.


lildude:
Using executive privilege here 😜

PR #4792, opened 2020-02-04
Add support for Nessus Attack Scripting Language

tchatard:
## Description
Adding support for NASL. This is a scripting language used by the suite of CyberSecurity products related to Nessus. It is developed and maintained  by Tenable Network Security, Inc.

## Checklist:
            -  .nasl: https://github.com/search?p=1&q=extension%3Anasl+NOT+nothack&type=Code
            - .inc: https://github.com/search?q=Tenable+extension%3Ainc+NOT+nothack&type=Code
    - Sample source(s): see sample directory in PR
    - Sample license(s): see in each file
    - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Ftenable%2Fsublimetext-nasl%2Fblob%2Fmaster%2FNasl%2FNASL.tmLanguage&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2FTemplar-os%2FTemplar-overlay%2Fblob%2F122e308d4ce8bc62ac5600e941a99d7c1cf054c3%2Fnet-analyzer%2Fnessus-bin%2Ffiles%2Fnmap.nasl&code=


lib/linguist/languages.yml
```diff
@@ -3332,6 +3332,14 @@ NSIS:
   codemirror_mode: nsis
   codemirror_mime_type: text/x-nsis
   language_id: 242
+Nasl:
+  type: programming
+  extensions:
+  - ".nasl"
+  - ".inc"
```
pchaigno: Could you add a search query in the original post for this extension as well?

`.inc` is already used by quite a few languages, so if we add it for Nasl as well, we're going to need a heuristic.

vendor/licenses/grammar/sublimetext-nasl.txt
```diff
@@ -0,0 +1,26 @@
+---
```
pchaigno: Where is this license coming from? I don't see it in the grammar's repository. Did you use the provided script to add the grammar, as per the contribution guidelines?

lib/linguist/languages.yml
```diff
@@ -3332,6 +3332,14 @@ NSIS:
   codemirror_mode: nsis
   codemirror_mime_type: text/x-nsis
   language_id: 242
+Nasl:
```
Alhadis: This should be changed to `NASL` (uppercase), as that appears to be the language's canonical spelling (going by its [Wikipedia page](https://en.wikipedia.org/wiki/Nessus_Attack_Scripting_Language)).

tchatard: Good point. Done with new commit!

lib/linguist/languages.yml
tchatard: Search query: Done!
Heuristic: Done! (see new commit)

vendor/licenses/grammar/sublimetext-nasl.txt
tchatard: I was working with Tenable's legal department yesterday and submitted a PR for this repo today to clearly add it. The PR also got merged today, see https://github.com/tenable/sublimetext-nasl/blob/master/LICENSE.

tchatard:
Thanks All for the quick initial review. I have made the recommended changes, please let me know if anything else is needed.

lib/linguist/languages.yml
pchaigno: Please filter the search query using some NASL-specific keywords. The current query is returning all `.inc` files on GitHub.

lib/linguist/heuristics.yml
```diff
@@ -216,6 +216,8 @@ disambiguations:
     pattern: '^<\?(?:php)?'
   - language: SourcePawn
     pattern: '^public\s+(?:SharedPlugin(?:\s+|:)__pl_\w+\s*=(?:\s*{)?|(?:void\s+)?__pl_\w+_SetNTVOptional\(\)(?:\s*{)?)'
+  - language: NASL
+    named_pattern: incNASL
```
pchaigno: We don't need a named pattern here since these heuristic rules are only used here.

samples/NASL/function.inc
```diff
@@ -0,0 +1,11 @@
+// Sample function
```
pchaigno: Most of the sample files seem to be very simple. We prefer samples from "real" applications as they will be used to train the Bayesian classifier (i.e., they should be representative of actual usage).

lib/linguist/languages.yml
tchatard: Added filtering using "Tenable" keyword, which narrows down the search a lot but at least returns no false-positives.

lib/linguist/heuristics.yml
tchatard: Updated in latest commit.

samples/NASL/function.inc
tchatard: The latest commit removes these trivial samples and adds a bunch of real-life ones.

lib/linguist/heuristics.yml
```diff
@@ -216,6 +216,13 @@ disambiguations:
     pattern: '^<\?(?:php)?'
   - language: SourcePawn
     pattern: '^public\s+(?:SharedPlugin(?:\s+|:)__pl_\w+\s*=(?:\s*{)?|(?:void\s+)?__pl_\w+_SetNTVOptional\(\)(?:\s*{)?)'
+  - language: NASL
+    pattern:
+    - '^\s*include\s*\(\s*(?:"|'')[\\/\w\-\.:\s]+(?:"|'')\s*\)\s*;'
+    - '^\s*(?:global|local)_var\s+(?:[\w\-]+(?:\s*=\s*[\w\-"'']+|)\s*)(?:,\s*[\w\-]+(?:\s*=\s*[\w\-"'']+|)\s*)*+\s*;'
```
Alhadis: This line contains capturing-groups with empty expressions: `|)`. If you intended to make these matches optional, use `)?` instead:

~~~diff
-'^\s*(?:global|local)_var\s+(?:[\w\-]+(?:\s*=\s*[\w\-"'']+|)\s*)(?:,\s*[\w\-]+(?:\s*=\s*[\w\-"'']+|)\s*)*+\s*;'
+'^\s*(?:global|local)_var\s+(?:[\w\-]+(?:\s*=\s*[\w\-"'']+)?\s*)(?:,\s*[\w\-]+(?:\s*=\s*[\w\-"'']+)?\s*)*+\s*;'
~~~

lib/linguist/heuristics.yml
```diff
@@ -216,6 +216,13 @@ disambiguations:
...
+    - '^\s*namespace\s+[\w\-]+\s*(?:$^)*\s*\{'
```
Alhadis: Change `\s*(?:$^)*\s*\{` to just `\s*{`:

~~~diff
-'^\s*namespace\s+[\w\-]+\s*(?:$^)*\s*\{'
+'^\s*namespace\s+[\w\-]+\s*{'
~~~

`$` and `^` are zero-length matches, so repeating them is meaningless (also invalid in many regex engines). If you meant to match newlines, they'll already be matched by `\s*`.

lib/linguist/heuristics.yml
```diff
@@ -216,6 +216,13 @@ disambiguations:
...
+    - '^\s*object\s+[\w\-]+\s*(?:extends\s+[\w\-:]+\s+|)\s*(?:$^)*\s*\{'
```
Alhadis: Another empty expression. See my first comment.

(Note that `{` doesn't need to be escaped unless it's followed by a numbers and a closing brace: `{1}`, `{1,2}`, `{3,}`, etc).

lib/linguist/heuristics.yml
```diff
@@ -216,6 +216,13 @@ disambiguations:
...
+    - '^\s*(?:public\s|private\s|)\s*function\s+[\w]+\s*\([\w\s,]*\)\s*(?:$^)*\s*\{'
```
Alhadis: This can be simplified with:

~~~regex
- '^\s*(?:public\s+|private\s+|\s*)function\s+\w+\s*\([\w\s,]*\)\s*{'
~~~

lib/linguist/heuristics.yml
```diff
@@ -216,6 +216,13 @@ disambiguations:
     pattern: '^<\?(?:php)?'
   - language: SourcePawn
     pattern: '^public\s+(?:SharedPlugin(?:\s+|:)__pl_\w+\s*=(?:\s*{)?|(?:void\s+)?__pl_\w+_SetNTVOptional\(\)(?:\s*{)?)'
+  - language: NASL
+    pattern:
+    - '^\s*include\s*\(\s*(?:"|'')[\\/\w\-\.:\s]+(?:"|'')\s*\)\s*;'
```
Alhadis: Is it possible to `include` files without a `.nasl` or `.inc` extension? If not, it might be a good idea to add them to the regex to avoid matching files with `include("foo.php")` or what-have-you. You can use this:

~~~regex
- '^\s*include\s*\(\s*("|'')(?:(?!\1)[^\r\n])*\.(?:inc|nasl)\1\s*\)\s*;'
~~~

lib/linguist/heuristics.yml
tchatard: Always good to learn from a regex wizard! :-) Updated in latest commit.

lib/linguist/heuristics.yml
tchatard: Technically the extension can be anything but I never saw it with anything different than nasl or inc so I think it is better to restrict this regex to avoid false positives. Updated in latest commit.

lib/linguist/heuristics.yml
tchatard: Changed in latest commit.

lib/linguist/heuristics.yml
tchatard: Thanks, I wasn't sure if the new line would be accounted for that way but I was able to test it to make sure this time. Changed in latest commit.

lib/linguist/heuristics.yml
tchatard: More elegant indeed. Changed in latest commit.

tchatard:
What is the next step to get this PR merged?

Alhadis:
@tchatard A 👍 of approval from a GitHub staff member, which I'm sure @lildude will get to sooner or later.

In any case, merged PRs won't take effect on GitHub.com until the next release of Linguist is cut, which happens roughly once a month. So there's really no rush.

tchatard:
The sample files have been extracted from a recent Tenable plugins feed.
Some previous (or modified) versions of similar files can be found on various locations on github or other code hosting services as Tenable distributes its feed in clear text for its products to download and process.
Anybody with a (trial or not) version of one of the Nessus products will be able to get the current version of the Nasl plugin feed.
Information about the plugins can be found at https://www.tenable.com/plugins.
Information on how to get access to the files: https://docs.tenable.com/nessus/Content/DownloadAndCopyPlugins.htm

Let me know if this is enough information.

lildude:
Thanks @tchatard, that makes sense and after a bit of digging, I worked out these plugins are fed off https://plugins.nessus.org/ which needs an account to access.

PR #4806, opened 2020-02-28
Make shebang detection cheaper

tenderlove:
String#lines can potentially allocate *many* strings depending on the
number of newlines in the string.  That might be OK if the file we're
looking at is actually a file with a shebang in it (because we might
eventually use every line).  But it is not OK when the file isn't a
shebang file because we end up allocating many objects for no reason.

This commit reduces the number of allocations it takes to determine
whether or not the string is interesting to the Shebang class.

Here is a benchmark:

```ruby
require "linguist/shebang"

def allocs
  x = GC.stat(:total_allocated_objects)
  yield
  GC.stat(:total_allocated_objects) - x
end

file = "foo\n" * 1000

p allocs {
  Linguist::Shebang.interpreter(file)
}
```

Before this commit:

```
$ bundle _1.17.2_ exec ruby -I lib:test thing.rb
1003
```

After this commit:

```
$ bundle _1.17.2_ exec ruby -I lib:test thing.rb
2
```

### DIFF
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -24,7 +24,7 @@ module Linguist
     # Returns a String or nil
     def self.interpreter(data)
       # First line must start with #!
-      return unless data.b.match?(/\A#!/n)
+      return unless data.b.start_with?("#!")
 
       shebang = data[/\A[^#{$/}]*#{$/}/].chomp
 
--- a/lib/linguist/shebang.rb
+++ b/lib/linguist/shebang.rb
@@ -23,10 +23,10 @@ module Linguist
     #
     # Returns a String or nil
     def self.interpreter(data)
-      shebang = data.lines.first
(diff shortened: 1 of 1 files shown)


lib/linguist/shebang.rb
```diff
@@ -23,10 +23,10 @@ def self.call(blob, candidates)
     #
     # Returns a String or nil
     def self.interpreter(data)
-      shebang = data.lines.first
-
       # First line must start with #!
-      return unless shebang && shebang.start_with?("#!")
+      return unless data.b.match?(/\A#!/n)
```
jhawthorn: Could we?

```suggestion
      return unless data.b.start_with?("#!")
```

or is there an encoding issue?

tenderlove:
It's now at 0 allocations on a "miss".

PR #4808, opened 2020-03-01
Add Haxe generated source detection

Gama11:
## Description

The [Haxe](https://haxe.org/) compiler can generate source files for a whole host of different target languages. It's fairly simple to detect this with the `Generated by Haxe <version>` comment the files start with. I've included test files for all relevant targets that Haxe 4 supports.

## Checklist:


lib/linguist/generated.rb
```diff
@@ -627,6 +628,16 @@ def generated_visualstudio6?
       lines.first(3).any? { |l| l.include? '# Microsoft Developer Studio Generated Build File' }
     end
 
+    HAXE_EXTENSIONS = ['.js', '.py', '.lua', '.cpp', '.h', '.java', '.cs', '.php']
```
Gama11: I've seen this pattern of checking the file extension in other `generated`-functions. I'm assuming this is desired for performance reasons? Otherwise, not checking the extension at all would be a bit easier to maintain (Haxe adds new targets from time to time).

pchaigno: > I'm assuming this is desired for performance reasons?

Yes, we want to avoid reading the first three lines of every single file on GitHub.

Gama11: Ok, makes sense. :+1:

Gama11:
@lildude I simply compiled a Hello World program to different targets with Haxe 4.0.5:

```haxe
class Main {
	static function main() {
		trace("Hello World");
	}
}
```

e.g. `haxe -main Main -js main.js`.

Not sure a license makes any sense for that? :)

lildude:
> Not sure a license makes any sense for that? :)

Indeed, which is why we always ask for samples from real-world applications.

Gama11:
Well, I can certainly swap them out with something else if you prefer, though for what is being tested here (just the source header comments) it doesn't really make any difference, other than increasing file size...

lildude:
🤔 thanks for the explanation @Gama11. Yeah, I think in this case, it makes sense for these samples as is. Thanks.

PR #4818, opened 2020-03-11
Improve C++ color

joelerego:
Minor change to the C++ language color from #f34b7d to #ed1d7d. The new color matches with the official colors of Google's Fuchsia OS. It's brighter and more vibrant. This change does not make it hard to recognize for all users that are used to the color since the two colors are almost indistinguishable.

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):
  
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -560,7 +560,7 @@ C++:
   ace_mode: c_cpp
   codemirror_mode: clike
   codemirror_mime_type: text/x-c++src
-  color: "#f34b7d"
+  color: "#ed1d7d"
   aliases:
   - cpp
   extensions:


Alhadis:
Google don't own C++, and Fuchsia is simply one of many, *many* projects that uses C++. By no means is it representative of the language itself.

👎 from me.

joelerego:
> Google don't own C++, and Fuchsia is simply one of many, _many_ projects that uses C++. By no means is it representative of the language itself.
> 
> In addition, the links you've given in the _"I have obtained agreement…"_ section don't include an **actual discussion** about changing the language's colour on GitHub; they're simply links to Fuchsia-related pages and images.
> 
>  from me.

On second thoughts, you are right. There is no reason for it to match Fuchsia, because Google doesn't own C++. Also, I apologize for the community links. I did have a discussion where a few people said they don't really like the current salmon/baby pink language color for CPP. I am not able to find the link right now. My argument was this color doesn't confuse people who are already used to the current color. The part of the Fuchsia match is just an additional advantage, but not a reason for the change. 

If getting used to is not a barrier, I think grey or black is the perfect color for CPP and pink doesn't accurately represent CPP at all. Do you agree with this? Please reply. Thank you.

Alhadis:
> If getting used to is not a barrier

It is. See #4319 and the [community's response](https://github.com/github/linguist/pull/4326) to see why arbitrary, wanton colour-changes are a bad idea. This incident is actually what led to us enforcing [our current guidelines](https://github.com/github/linguist/blob/b545f1c6befe87c2074e5e608f47ca0f6334d355/CONTRIBUTING.md#changing-the-color-associated-with-a-language) for proposing a colour-change. To quote the most relevant part:

> _Many of the colours associated with the languages within Linguist have been in place for a very long time. The colours were often chosen based on the colours used by the language at the time and since then users will have become familiar with those colours as they appear on GitHub.com._

**Bottom-line:** Rationale like _"I think this colour looks better"_ isn't acceptable because everybody has their own opinion of what "looks better". Unless there's a clear community consensus, *and* the community in question is representative of the language's userbase as a whole, it's best to keep colours as they are, even if they're somewhat lurid.

joelerego:
> > If getting used to is not a barrier
> 
> It is. See #4319 and the [community's response](https://github.com/github/linguist/pull/4326) to see why arbitrary, wanton colour-changes are a bad idea. This incident is actually what led to us enforcing [our current guidelines](https://github.com/github/linguist/blob/b545f1c6befe87c2074e5e608f47ca0f6334d355/CONTRIBUTING.md#changing-the-color-associated-with-a-language) for proposing a colour-change. To quote the most relevant part:
> 
> > _Many of the colours associated with the languages within Linguist have been in place for a very long time. The colours were often chosen based on the colours used by the language at the time and since then users will have become familiar with those colours as they appear on GitHub.com._
> 
> **Bottom-line:** Rationale like _"I think this colour looks better"_ isn't acceptable because everybody has their own opinion of what "looks better". Unless there's a clear community consensus, _and_ the community in question is representative of the language's userbase as a whole, it's best to keep colours as they are, even if they're somewhat lurid.

I agree with everything that you have mentioned. I am closing the pull request immediately. After the color proximity test, I have realized this #ed1d7d is difficult to distinguish from the Cold Fusion color. I did read a comment from someone who seemed like GitHub's web designer, who claimed the following:

1. He would first choose the hex code of the logo of the language if the logo was available.
2. If the logo was unavailable, he would take the color of the O'Reilly book on the language.
3. If the hex of the book cover is unavailable, he would choose an arbitrary hex code.

Standard C++ (isocpp.org) has a dark blue logo, which is available on Wikipedia. The C++ Programming Language textbook written by Bjarne Stroustrup is also of the same color. I am curious to know how the current color was chosen. Was it an arbitrary selection?

lildude:
> Was it an arbitrary selection?

Probably and we may never know as the C++ was one of the languages that Linguist knew about when colours were first assigned to the languages over 8 years ago in https://github.com/github/linguist/commit/c3d6fc5af8cf9d67afa572bba363bf0db256a900 .  This predates pull requests (and GitHub even having a formal design team) so we can't even read the substantiation within a PR, and as all the colours were added in one commit, it's impossible to tell why each colour was chosen. 

I can't remember why I made a choice for a line of code, random or otherwise, a last year, let alone 8 years ago, so asking the commitor, who may not have been the original author, is probably futile too. They don't work at GitHub anymore either.

As an aside, to my 👀, the proposed colour change is also only just perceptibly different when you compare them side-by-side: 

`#f34b7d`
`#ed1d7d`

... neither of which are close to the salmon pinks [Wikipedia](https://en.wikipedia.org/wiki/Salmon_(color)) lists to my eyes 😁:

`#FA8072`
`#FFA07A`
`#FF91A4`
`#E9967A`

So I agree with @Alhadis; there is no sound and justifiable reason to change C++ to this colour.

PR #2426, opened 2015-05-28
Detect Go files generated by go-bindata

tamird:
Modeled after https://github.com/github/linguist/commit/2ca586861025caa2887b6b7dd8ef2f0b83805343.

cc @bramgruneir

PR #4821, opened 2020-03-12
Detect go files as generated according to official rules

ernestas2k:
Use officially accepted rules to detect golang generated files https://golang.org/s/generatedcode

## Description
The rule dictates that a line matching `^// Code generated .* DO NOT EDIT\.$` can be anywhere in golang file.
I currently use github.com/deepmap/oapi-codegen that generates the following (valid according to the rules) file:
```go
// Package api provides primitives to interact the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen DO NOT EDIT.
package api

// Error defines model for Error.
type Error struct {
...
```
Sample file for test taken from: https://github.com/deepmap/oapi-codegen/blob/master/examples/petstore-expanded/echo/api/petstore-types.gen.go
The project license (apache 2.0) https://github.com/deepmap/oapi-codegen/blob/master/LICENSE

Would be good to get such files ignored automatically by github when reviewing PRs.

This issue was raised and solved multiple times but wasn't merged. 
Refs #3800 #2724 

## Checklist:


lib/linguist/generated.rb
```diff
@@ -301,7 +301,11 @@ def generated_go?
       return false unless extname == '.go'
       return false unless lines.count > 1
 
-      return lines[0].include?("Code generated by")
+      # old check for the transitional period (remove it later)
+      return true if lines[0].include?("Code generated by")
+
+      # https://golang.org/s/generatedcode
+      return lines.find {|line| line =~ /^\/\/ Code generated .* DO NOT EDIT\.$/ }
```
pchaigno: I understand the rule states this could be anywhere in the file, but in practice, isn't it almost always in the first few lines? I think we should avoid scanning entire files if possible.

lib/linguist/generated.rb
```diff
@@ -301,7 +301,11 @@ def generated_go?
       return false unless extname == '.go'
       return false unless lines.count > 1
 
-      return lines[0].include?("Code generated by")
+      # old check for the transitional period (remove it later)
+      return true if lines[0].include?("Code generated by")
```
pchaigno: Why do we need to keep this? The second rule is more general.

lib/linguist/generated.rb
ernestas2k: As far as I could find the very first (comment) lines would contain it, have seen them in 17th line too (due to long (generated) description text. I agree checking files with 10k and more lines is too much.

What would be a solution, collecting first consecutive comment lines and testing them?

lib/linguist/generated.rb
ernestas2k: The second rule would fail for lines where first would match (the first rule ignores `DO NOT EDIT.`). Didn't want to break the old behaviour. What are your thoughts on this? Should we just follow 2nd rule?

lib/linguist/generated.rb
pchaigno: > What would be a solution, collecting first consecutive comment lines and testing them?

Please check the other examples that match against several lines below (e.g., `lines.first(6).`).

> As far as I could find the very first (comment) lines would contain it, have seen them in 17th line too due to long (generated) description text. I agree checking files with 10k and more lines is too much.

If the 17 lines case happens once in a lifetime, it's probably not worth it and we can match only on say the first 10 lines. If you've seen several such cases, then let's go for 17 lines :-)

pchaigno: It looks like the original intent was to match on go-bindata generated files (cf. #2426), which won't match the second rule. So let's keep the first rule!

lib/linguist/generated.rb
pchaigno: It looks like the original intent was to match on go-bindata generated files (cf. #2426), which won't match the second rule. So let's keep the first rule!

lib/linguist/generated.rb
ernestas2k: Good points, updated the pull request to test first 10 lines

lildude:
@ernestas2k Can you please add a note to the original post linking to the source of the sample file and the license it is covered by. Thanks.

ernestas2k:
> @ernestas2k Can you please add a note to the original post linking to the source of the sample file and the license it is covered by. Thanks.

The sample file is generated code that's located in private repo of mine. Should I find public sample file instead? LIke this one https://github.com/deepmap/oapi-codegen/blob/master/examples/petstore-expanded/echo/api/petstore-server.gen.go ?

lildude:
@ernestas2k yes please. We need to be able to verify sources in case there are any licensing issues in future.

pchaigno:
@lildude Isn't posting here in a pull request the same as adding it to a public repository under a permissive license? I mean, it's literally what people are doing by posting here.

lildude:
@pchaigno sort of. If the file isn't from a source with a verifiable open license, and the author isn't deemed "authoritative" at some point in the future by the owner of the code, it potentially opens GitHub up to legal battles.  

I know this isn't the case now, but it's best and easier to be consistent across the board.

lildude:
@ernestas2k if you're quick, I might just be able to squeeze this into the linguist release I'm currently testing.

ernestas2k:
@lildude ok just a moment

ernestas2k:
@lildude updated the original post, let me know if any more changes would be valuable.

PR #4827, opened 2020-03-19
Kusto: Azure Log Analytics's query language

pjquirk:
## Description
Adds support for Azure Log Analytics's query language called [Kusto](https://docs.microsoft.com/en-us/azure/kusto/query/).

Kusto files are generally stored with `.csl` or `.kusto` file extensions.  The `.csl` extension is currently classified as XML so I've added a heuristic to distinguish the XML and Kusto files.

## Checklist:

      - https://github.com/search?q=extension%3Acsl+where&type=Code&ref=advsearch&l=&l=
      - https://github.com/search?q=extension%3Akusto+NOT+nothack&type=Code&ref=advsearch&l=&l=
      - https://github.com/search?q=extension%3Akql+NOT+nothack&type=Code&ref=advsearch&l=&l=
    - Sample source(s):
      - [example-create-sflogs-table.csl](https://github.com/microsoft/CollectServiceFabricData/blob/c848f9996154e1e86ff9f1d3889b077dfb14b541/kusto/queries/example-create-sflogs-table.csl)
      - [examples_traces.csl](https://github.com/jagilber/workflowtest/blob/db6b339d16760f1512f329035490cb79af6f221c/KustoQueries/examples_traces.csl)
      - [kusto-example-queries.csl](https://github.com/microsoft/CollectServiceFabricData/blob/c848f9996154e1e86ff9f1d3889b077dfb14b541/kusto/queries/kusto-example-queries.csl)
    - Sample license(s):
      - [MIT License](https://github.com/microsoft/CollectServiceFabricData/blob/73a1ef280654b963dcd999ac09992382320cd1fd/LICENSE)
  
    - Grammar is [here](https://github.com/josin/kusto-syntax-highlighting/blob/master/syntaxes/kusto.tmLanguage).


pjquirk:
👋 First PR to Linguist! This may not meet the bar for being "widely used", but would appreciate the consideration.

pchaigno:
:wave: Thanks for contributing to Linguist!

> Search results for each extension:
> 
>     * https://github.com/search?q=extension%3Acsl+extension%3Akusto+where+OR+project+OR+extend+OR+limit+OR+summarize&type=Code&ref=advsearch&l=&l=

We'll need a search query per file extension added.

It also looks like [the search query for `.cls`](https://github.com/search?q=extension%3Acls+where+OR+project+OR+extend+OR+limit+OR+summarize&type=Code) returns a lot of files with a syntax very different from [the `.kusto` files](https://github.com/search?q=extension%3Akusto+where+OR+project+OR+extend+OR+limit+OR+summarize&type=Code). Are these files actually Kusto?

pjquirk:
@pchaigno Thanks, I split the search into two separate queries.

The extension should be `csl`, not `cls`, so no those files are Kusto 😃  Looking at the `csl` files, I do see a few non-Kusto files in the last few pages.  I don't have any great ideas on filtering those out though.

pchaigno:
> The extension should be `csl`, not `cls`, so no those files are Kusto smiley

Ah, sorry. My bad.

The `.kusto` file extension only has about a hundred files which is not going to be enough to meet our *hundreds of repository* criteria. Do you expect that number to grow significantly in the next few months?

> Looking at the `csl` files, I do see a few non-Kusto files in the last few pages. I don't have any great ideas on filtering those out though.

We can add a `NOT xml` to filter those. There are still [a few files at the end that look different](https://github.com/search?p=57&q=extension%3Acsl+where+OR+project+OR+extend+OR+limit+OR+summarize+NOT+xml&type=Code); are they Kusto files?

pjquirk:
> Do you expect that number to grow significantly in the next few months?

I'd expect it to grow some, but not significantly.  Some of the teams that joined Actions recently use Kusto for monitoring and store their queries in a repo.  Not all have migrated their queries to a GitHub repo yet though.

> There are still a few files at the end that look different; are they Kusto files?

Kusto-related results look to stop around page 28 or so.  I do see a handful of other different types of `csl` files, so yeah there's going to be some overlap.

I understand if this doesn't meet the bar in terms of wide-spread usage.  Let me know how you want to proceed!

pchaigno:
> I'd expect it to grow some, but not significantly. The teams that joined Actions recently use Kusto for monitoring and store their queries in a repo. Not all have migrated their queries to a GitHub repo yet though.

Okay. I'll add the extension to #4219 to continue monitoring usage on GitHub.com (I'm adding the label in the meantime to remember). Once usage is there, I'll ping you here.

> Kusto-related results look to stop around page 28 or so. I do see a handful of other different types of `csl` files, so yeah there's going to be some overlap.

Unfortunately, Linguist doesn't have a `not-a-language` entry, so if it starts classifying `.csl` files, it will have to classify them all. In practice that means that, in order to add `.csl` for Kusto, we'll have to understand what are the other languages using that extension and add them to Linguist as well. Note that we only need to do that for languages that meet our usage criteria (~ hundreds of repositories).

ahayworth:
@pjquirk @pchaigno adding `.kql` as an extension turns up additional results: https://github.com/search?q=extension%3Akql+where+OR+project+OR+extend+OR+limit+OR+summarize&type=Code

pjquirk:
Thanks @ahayworth , I wasn't aware `kql` was used, and it looks to be more widely used than `.csl`, so that may push it over the bar as far as usage? I added the extension to the PR in the meantime.

pjquirk:
@lildude Pinging you on this since I think `@pchaigno` may not be here any longer and I don't want this to get lost.

[Here](https://github.com/github/linguist/pull/4827#issuecomment-609037112) they said they were going to add it to #4219 but I don't see it on there.  Can you help with that? Is there any automation that updates those numbers or is that something I or some other non-robot periodically updates?

lildude:
> Can you help with that?

Yeah, I can add to that table.

> Is there any automation that updates those numbers or is that something I or some other non-robot periodically updates?

It needs to be done manually using Harvester as mentioned in the OP of that issue.

lildude:
> Yeah, I can add to that table.

Ooof, this is going to be time consuming. I'll try remember to add the figures when I've got the time. It's worth pointing out that we only need _one_ extension to be popular enough to add a language. An override can then be used for all other extensions until popularity of the other extensions improves.

pjquirk:
> your OP on this issue needs to link to the original source of all the samples, not mention the paths in this PR/repo

Done, my mistake 😄 

I also updated some of my queries since I realized `OR` doesn't work (anymore? I thought it used to).  

> Ooof, this is going to be time consuming.

What about porting Harvest to an action that runs daily?  I could see it taking some JSON input with the searches to monitor, then dumping it to a markdown table.  A quick scan of the action shows we could probably do all that with Octokit pretty easily.  I've got some time before the holidays to try and crank that out if that'd be helpful.

lildude:
> A quick scan of the action shows we could probably do all that with Octokit pretty easily. 

Oh yes, this can most definitely be automated and switched to using Octokit instead of the scraping method used by Harvester. I've implemented a very crude method of doing this in a hacky script I use locally (https://github.com/lildude/linguist/blob/lildude/download-corpus/script/github-ext-pop), however it comes with one big caveat: it hits the API rate limits very quickly.

You'll also need to use a token that is _not_ associated with a staff account to be sure you don't hit any staff shipped changes to the search or accidentally reveal or count GitHub private repos.

> I also updated some of my queries since I realized `OR` doesn't work (anymore? I thought it used to).

I'm pretty sure it used to too, but I wouldn't be surprised if you (and I) are enjoying some staff-shipped changes that aren't quite prod-ready 😁.

> I've got some time before the holidays to try and crank that out if that'd be helpful.

Go for it. Feel free to steal code or inspiration from my script too.

lildude (changes requested):
@pjquirk it looks like `.csl where` is popular enough now for inclusion. The other two extensions aren't so we can't include them just yet. Once this PR has been merged, peeps can use an override for those extensions.

Can you please remove `.kql` and `.kusto` from the entry in the `languages.yml` file and add a test for the heuristic so the tests pass.

I've missed the boat for shipping this week (peeps don't like it when I deploy Linguist changes on Fridays 😬) but if you're quick, I can squeeze this in for my definite deploy early next week.

pjquirk:
@lildude Great to hear! Sure, I'll make those changes hopefully today.

pjquirk:
@lildude Changes made, checks are green :)

lildude (approved):
I 🔥 the old redundant cached txt license file. Hope you don't mind.

Thanks for your patience.

PR #4830, opened 2020-03-24
Add Kaitai Struct (KSY) language

dgelessus:
## Description
Adds a mapping for the .ksy extension used by [Kaitai Struct](https://kaitai.io), a YAML-based language for describing the structure of binary file formats.

## Checklist:
      - https://github.com/search?q=extension%3Aksy+NOT+nothack (according to data from Harvester, this matches 256 repos by 219 individual owners)
    - Sample source(s):
      - https://github.com/kaitai-io/kaitai_struct_formats/blob/cde1b06de6fc639df99cd62fca9cd07634c64d2a/filesystem/iso9660.ksy
      - https://github.com/kaitai-io/kaitai_struct_formats/blob/cde1b06de6fc639df99cd62fca9cd07634c64d2a/archive/zip.ksy
    - Sample license(s): CC0-1.0 (see the sample source files' `meta` section)
  - [x] ~~I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/~~ N/A: existing YAML syntax highlighting is used
  - [x] ~~I have updated the heuristics to distinguish my language from others using the same extension.~~ N/A: the .ksy extension is not used by any other languages


lib/linguist/languages.yml
```diff
@@ -2615,6 +2615,18 @@ KRL:
...
+Kaitai Struct:
+  type: programming
+  aliases:
+  - ksy
+  ace_mode: yaml
+  codemirror_mode: yaml
+  codemirror_mime_type: text/x-yaml
+  color: "#773b37"
+  extensions:
+  - ".ksy"
```
pchaigno: For other reviewers: I counted 284 repositories by 239 users with the search query [`extension:ksy meta OR seq OR types`](https://github.com/search?p=1&q=extension%3Aksy+meta+OR+seq+OR+types&type=Code).

lib/linguist/languages.yml
```diff
@@ -2615,6 +2615,18 @@ KRL:
   tm_scope: none
   ace_mode: text
   language_id: 186
+Kaitai Struct:
+  type: programming
```
pchaigno: This looks like it should be `data` instead of `programming`, no?

Also, why add this as a separate language from YAML? Why not just a new YAML extension?

dgelessus: > This looks like it should be `data` instead of `programming`, no?

Hm, I'm not really sure where Kaitai Struct falls in this categorization... It's definitely not a general-purpose programming language (you can't write standalone programs in it), but it's also not a general-purpose data language like CSV or JSON (it doesn't store/represent arbitrary data). Kaitai Struct specs are declarative specifications, so you might consider them data, but they also compile to executable code, so you might consider them programs.

Looking at some other similar languages: ANTLR and JFlex (parser generator grammars) are categorized as `programming`, but Protobuf .proto files are categorized as `data`. These languages are similar to Kaitai Struct in that they are descriptions of data formats and are compiled to source code for general-purpose languages (Java, Python, etc.).

Is there an explanation somewhere what the different `type`s mean? The comment at the top of languages.yml isn't very detailed - it just lists all valid types with no explanation of what the differences are.

> Also, why add this as a separate language from YAML? Why not just a new YAML extension?

There are already a few other YAML-based languages that have their own language entry rather than just an extra extension under the YAML language (e. g. Common Workflow Language, LookML, RAML, SaltStack), so I did the same thing here.

Also I think from a user perspective it makes more sense to have Kaitai Struct specs show up as a separate language, rather than being grouped with generic YAML files. IMHO, a .ksy file is primarily a Kaitai Struct spec rather than a YAML file, in the same way that a .clj file is primarily Clojure code rather than a "s-expression file".

dgelessus:
I'm still interested in this PR. My last comment (https://github.com/github/linguist/pull/4830#discussion_r397577881) was asking for clarification about what the different `type`s mean, because it's not obvious (to me at least) whether a declarative language like Kaitai Struct should be considered `programming` or `data`.

lib/linguist/languages.yml
Alhadis: YAML's already a programming language, though 😉 (kidding, I'm poking fun at the format's complexity…)

> Is there an explanation somewhere what the different `types` mean?

Not really. We don't have any hard-and-fast criteria as to what constitutes `programming`, `markup`, etc. Most of the time it's pretty obvious, but we get cases like this where semantics are hard to define, and we use our best judgement. The `data` type is often used as a catchall for anything that doesn't comfortably fit into one of the other categories; sometimes that leads to [controversy](https://github.com/github/linguist/issues/3805), or recurring requests to redefine XML as `markup`. Eventually, we have to deal with things on a case-by-case basis. User expectations are generally the guiding principle for busting head-scratchers.

That being said, I'd leave Kaiti Struct as `programming`. It's obviously hand-written and users would expect it to contribute to a project's language statistics.

> in the same way that a .clj file is primarily Clojure code rather than a "s-expression file".

Code and data are pretty much one-and-the-same in the Lisp world. Symbolic expressions are pretty much a one-to-one depiction of what you write and what's going through your compiler, so this is a pretty shallow argument. I digress, though.

lib/linguist/languages.yml
Alhadis: Thanks, that saves me 45 minutes staring at the screen waiting for Harvester to finish.

Alhadis:
@dgelessus Keep it as `programming`. Normally I'd recommend adding it as a YAML extension, but reading this

> There are already a few other YAML-based languages that have their own language entry rather than just an extra extension under the YAML language (e. g. Common Workflow Language, LookML, RAML, SaltStack), so I did the same thing here.

… made me realise we're not 100% consistent about differentiating between an extension and a specialised use of a certain file-format. 😜

lib/linguist/languages.yml
dgelessus: > sometimes that leads to [controversy](https://github.com/github/linguist/issues/3805)

Yeah, it's a bit unfortunate that GitHub treats `data` languages differently and excludes them from some statistics/calculations. That's another reason why I don't think `data` is the right choice here - there are a number of "pure Kaitai Struct" repos (e. g. [kaitai-io/kaitai_struct_formats](https://github.com/kaitai-io/kaitai_struct_formats), [RiiConnect24/Kaitai-Files](https://github.com/RiiConnect24/Kaitai-Files), and [others](https://github.com/kaitai-io/awesome-kaitai#formats)), which GitHub at the moment classifies based on the markup language of their docs, or as nothing at all. In order to correctly classify these repos as Kaitai Struct, the language needs to be categorized as `programming` or `markup`. (But since this behavior comes from GitHub and not Linguist, this complaint probably isn't on-topic for this repo...)

> Code and data are pretty much one-and-the-same in the Lisp world. Symbolic expressions are pretty much a one-to-one depiction of what you write and what's going through your compiler, so this is a pretty shallow argument.

That's exactly why I used a Lisp as the example actually ;) In a way every (for example) Clojure program is just data, so it would technically be correct to define a "s-expression" language and classify Clojure code under that. But Linguist doesn't do that and instead has a separate language for Clojure (as well as Common Lisp, Emacs Lisp, Clojure, Hy, EDN, etc.). IMHO the same thing applies to YAML-based DSLs - even though they use YAML syntax, there is a meaningful difference between a generic YAML data file and a file in a specific YAML-based DSL.

dgelessus:
I might have forgotten to reply here, oops.

I think the discussion above is more or less resolved though? Or if there's anything remaining that I should change/adjust, let me know.

PR #4832, opened 2020-03-26
Add jOOQ generated java file detection

simoleone:
Add jOOQ generated java file detection.

## Description
j00Q is a relatively popular (3.7K stars) SQL library for Java, and works by generating `.java` source files. Many people check in these generated files, as [this github search with 38K source file hits](https://github.com/search?q=%22This+file+is+generated+by+jOOQ.%22&type=Code) shows.

The jooq code generator emits a comment at the top of `.java` files which it generates. This detection technique is used by many of the other generated code detectors as well.

## Checklist:

    - Sample source(s):
      - https://github.com/jOOQ/jOOQ/blob/990bfa4663d832695a7baef205e2b532a1d93133/jOOQ-examples/jOOQ-kotlin-example/src/main/java/org/jooq/example/db/h2/tables/BookStore.java
    - Sample license(s): Apache


lib/linguist/generated.rb
```diff
@@ -683,6 +684,14 @@ def generated_html?
       end
     end
 
+    # Internal: Is this a generated jOOQ file?
+    #
+    # Return true or false
+    def generated_jooq?
+      return false unless extname.downcase == '.java'
+      lines.first(3).any? { |l| l.include? 'This file is generated by jOOQ.' }
```
pchaigno: If the header is always the same, let's limit this to the first two lines.

simoleone: done!

PR #4759, opened 2019-12-27
Move Starlark to separate language

arturdryomov:
[Starlark](https://docs.bazel.build/versions/master/skylark/language.html) is a scripting language inspired by Python.

* At this point GitHub recognizes and highlights Starlark files as Python files which is not exactly true. [The Python code is not compatible with Starlark interpreters in full extent](https://docs.bazel.build/versions/master/skylark/language.html#differences-with-python).
* At the same time, it is a bit weird to see repos with Python code when in fact all the code is Starlark.

Starlark is mostly a [Bazel build system](https://www.bazel.build/) child. But it gains a bit of traction. For example, [the Buck build system](https://buck.build/concept/skylark.html) adopts it as well. Interestingly [Drone CI supports Starlark](https://docs.drone.io/starlark/overview/) as a configuration language.

GitHub stats:

* [`.bzl` files](https://github.com/search?q=extension%3Abzl+NOT+nothack&type=Code) (118K hits);
* [`.bazel` files](https://github.com/search?q=extension%3Abazel+NOT+nothack&type=Code) (413K hits);
* [`.star` files](https://github.com/search?q=extension%3Astar+NOT+nothack&type=Code) (20K hits);
* [`.starlark` files](https://github.com/search?q=extension%3Astarlark+NOT+nothack&type=Code) (41 hits);

What this change does:

* defines a new Starlark language based on Python semantics;
* moves everything Starlark, Bazel (`BUILD`, `WORKSPACE`) and Buck (`BUCK`) related from Python to Starlark;
* moves samples and adds a new one with the `.starlark` extension;
* since Starlark is [a part of the Bazel family](https://github.com/bazelbuild/starlark) I’ve defined the color as a [Bazel green one](https://blog.bazel.build/2017/07/05/new-logo-and-homepage.html) ([preview](https://www.colorhexa.com/76d275)). 

CC @jin, @artem-zinnatullin

PR #4840, opened 2020-04-11
Add file extension for Starlark

mcuadros:
## Description
This PR's adds `.star` as an extension to the Starlark language. The `.star` extension is the default extension for Starlark scripts not related to Bazel, this is growing every day since many projects are embedding the language in their own projects.

This extension is used in all official interpreters in the examples and as well in the [testdata](https://github.com/bazelbuild/starlark/tree/master/test_suite/testdata).

Also is the official extension recommended by many opensource projects:
- https://docs.drone.io/pipeline/scripting/starlark/#context
- https://github.com/qri-io/qri/tree/master/startf/testdata
- https://github.com/mcuadros/ascode/tree/master/_examples

 This was discarded initially at https://github.com/github/linguist/pull/4759#discussion_r362961419

The amount of `.star` files currently is [30k](https://github.com/search?q=extension%3Astar+NOT+nothack&type=Code), since the PR instrudicing the language the usage has growth a 50%.
## Checklist:

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Astar+NOT+nothack
    - Sample source(s):
      - https://github.com/dockhippie/nodejs/blob/a04a35ef94418d559b434aa4834b9e71960fd42f/.drone.star
    - Sample license(s):
https://github.com/dockhippie/nodejs/blob/master/LICENSE

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5197,6 +5197,7 @@ Starlark:
   color: "#76d275"
   extensions:
   - ".bzl"
+  - ".star"
   filenames:
   - BUCK
   - BUILD
--- a/samples/Starlark/.drone.star
+++ b/samples/Starlark/.drone.star
@@ -1,0 +1,263 @@
+def main(ctx):
+  config = {
+    'title': 'NodeJS',
+    'image': ctx.repo.name,
+    'arches': [
+      'amd64',
+      'arm32v6',
+      'arm64v8',
+    ],
+    'versions': [
+      {
+        'path': 'latest',
+        'tag': 'latest',
+      },
+    ],
+    'trigger': [
+
+    ],
+  }
+
+  builds = docker(config)
+
+  after = [
+    readme(config),
+    badges(config),
+    trigger(config),
+  ]
+
+  for b in builds:
+    for a in after:
+      a['depends_on'].append(b['name'])
+
+  return builds + after
+
+def docker(config):
+  result = []
+
+  for version in config['versions']:
+    versions = []
+
+    for arch in config['arches']:
+
+      if arch == "amd64":
+        agent = "amd64"
+
+      if arch == "arm32v6":
+        agent = "arm"
+
+      if arch == "arm32v7":
+        agent = "arm"
+
+      if arch == "arm64v8":
+        agent = "arm64"
+
+      versions.append({
+        'kind': 'pipeline',
+        'type': 'docker',
+        'name': '%s-%s' % (arch, version['path']),
+        'platform': {
+          'os': 'linux',
+          'arch': agent,
+        },
+        'steps': [
+          {
+            'name': 'dryrun',
+            'image': 'plugins/docker:latest',
+            'pull': 'always',
+            'settings': {
+              'dry_run': True,
+              'dockerfile': '%s/Dockerfile.%s' % (version['path'], arch),
+              'tags': '%s-%s' % (version['tag'], arch),
+              'repo': 'webhippie/%s' % (config['image']),
+              'context': version['path'],
+            },
+            'when': {
+              'ref': {
+                'include': [
+                  'refs/pull/**',
+                ],
+              },
+            },
+          },
+          {
+            'name': 'publish',
+            'image': 'plugins/docker:latest',
+            'pull': 'always',
+            'settings': {
+              'username': {
+                'from_secret': 'docker_username',
+              },
+              'password': {
+                'from_secret': 'docker_password',
+              },
+              'dockerfile': '%s/Dockerfile.%s' % (version['path'], arch),
+              'tags': '%s-%s' % (version['tag'], arch),
+              'repo': 'webhippie/%s' % (config['image']),
+              'context': version['path'],
+            },
+            'when': {
+              'ref': {
+                'exclude': [
+                  'refs/pull/**',
+                ],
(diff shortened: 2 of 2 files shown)


pchaigno:
I don't think this pull request addresses the concerns expressed in #4759. The issue was not with the in-the-wild usage, but with the presence of many other `.star` files. All these files will be recognized as Starlark files with the pull request as is.

mcuadros:
Sorry, but if I am not mistaken, in linguist are many extension collisions,
that are resolved with the classifier. For example `.md`.

So all this files are not going to be identified as Starlark since a sample
is provided for the classifier. I am wrong?

Usually this can be a problem because this consume more resources than just
check the extension. But given the number of .star is slow, I don't believe
this is going to be a problem.

pchaigno:
> Sorry, but if I am not mistaken, in linguist are many extension collisions,
that are resolved with the classifier. For example `.md`.
>
> So all this files are not going to be identified as Starlark since a sample
is provided for the classifier. I am wrong?

That is correct, but requires us to also have support in Linguist for the conflicting languages. In the case of `.md`, Linguist knows both GCC Machine Description and Markdown and can use the classifier (and other, more precise strategies) to distinguish between the two. So for `.star`, we'll need to identity the conflicting languages and add them to Linguist if they meet usage requirements.

mcuadros:
Well but if the other languages are not identified will remain the same,
but the Starlark files will be identified.

Even if doesn't work like this, the amount files that aren't Starlark it's
minimal. And Starlark is growing everyday.

Please reconsider it.

fkorotkov:
@pchaigno I want to bump this issue. What will require for us to "identity the conflicting languages"?

I made a [search query to look for recently indexed *.star files](https://github.com/search?o=desc&q=extension%3Astar+filename%3A%2A.star+-path%3Apkg+-path%3Ago+-path%3Ainfra&s=indexed&type=Code), checked a few first pages and it seems majority of cases are Startlark cases. The rest is some data files or files without a particular pattern.

mahmoudimus:
bump! seconding @fkorotkov. /cc @lildude  @pchaigno - any way we can get this looked at?

lildude:
> bump! seconding @fkorotkov. /cc @lildude @pchaigno - any way we can get this looked at?

Things are still in the same place as it was... you still need to identify the other users of the `.star` extension. From a very [quick look](https://github.com/search?q=extension%3Astar+loop_), there's one particular use of `.star` files easily identifiable that is pretty active with nearly 8000 files and each file appears to occur only once per repo making it a clear example that qualifies for inclusion else it would be incorrectly identified. I have no idea what language it's written in, but it appears to be related to machine learning from the few READMEs I've looked at so it will be noticed very quickly if it's misidentified.

You don't have to identify _every_ user of the `.star` extension, but an attempt needs to be made to at least identify the most popular (probably just the one I've found), and then add them (it) as part of this PR (I've re-opened it).

fkorotkov:
Just for a reference we at [Cirrus CI](https://github.com/marketplace/cirrus-ci) started using Starlark along side YAML for allowing to [programmatically configure CI builds](https://cirrus-ci.org/guide/programming-tasks/). So there will be more `.cirrus.star` and `lib.star` files on GitHub in the near future. 😅

rohansingh:
@lildude Just so I understand, if we go through and identify all the conflicting languages that meet usage requirements, do they need to be added as new languages in this same PR?

lildude:
@rohansingh I think you may have missed the last sentence in my last response:

> You don't have to identify _every_ user of the `.star` extension, but an attempt needs to be made to at least identify the most popular (probably just the one I've found), and then add them (it) as part of this PR (I've re-opened it).

rohansingh:
Sorry, reading comprehension failure. Anyway, sounds good, I'll try to get to this today.

lildude (changes requested):
Marking this as still needing changes

lildude:
> You don't have to identify _every_ user of the `.star` extension, but an attempt needs to be made to at least identify the most popular (probably just the one I've found), and then add them (it) as part of this PR (I've re-opened it).

I've done this. I've identified the other most commonly used language is [Self-defining Text Archive and Retrieval](https://en.wikipedia.org/wiki/Self-defining_Text_Archive_and_Retrieval) commonly referred to as STAR.

I've pushed changes to your repo to add support to this PR.

Alhadis (commented):
That [`pro-dna.nmr.star`](https://github.com/github/linguist/blob/6ee517cf21dec5c00a2d0893a6ff11e9512aeaaf/samples/STAR/pro-dna.nmr.star) sample is _huge_ (8,613 lines / 145 KBs). Sure we can't trim it down a bit? I noticed that a lot of sections repeat the same data.

lildude:
> That [`pro-dna.nmr.star`](https://github.com/github/linguist/blob/6ee517cf21dec5c00a2d0893a6ff11e9512aeaaf/samples/STAR/pro-dna.nmr.star) sample is _huge_ (8,613 lines / 145 KBs). Sure we can't trim it down a bit? I noticed that a lot of sections repeat the same data.

🤦 And I point this out myself so often 😆 I'll get the ✂️ out.

lildude:
@Alhadis replaced it with a smaller sample.

lildude (approved):

PR #4841, opened 2020-04-13
Add Dafny using grammar from VSCode plugin

robin-aws:
## Checklist:

- [x] **I am adding a new language**
	- [x] The new language is used in hundreds of repositories:
		- `.dfy`: [~11,312 indexed results](https://github.com/search?q=extension%3Adfy+NOT+nothack&type=Code)
	- [x] I have included a real-world usage sample:
		- Source: [`microsoft/Ironclad/…/Math/mul.i.dfy`](https://github.com/microsoft/Ironclad/blob/master/ironfleet/src/Dafny/Libraries/Math/mul.i.dfy)
		- License: [MIT](https://github.com/microsoft/Ironclad/blob/master/ironfleet/LICENSE)
	- [x] I have included a [syntax highlighting grammar](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2FDafnyVSCode%2FDafny-VSCode%2Fblob%2Fdevelop%2Fclient%2Fsyntaxes%2FDafny.tmLanguage&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2FIronclad%2Fblob%2Fmaster%2Fironfleet%2Fsrc%2FDafny%2FLibraries%2FMath%2Fmul.i.dfy&code=)


lib/linguist/languages.yml
```diff
@@ -1099,6 +1099,13 @@ DTrace:
   codemirror_mode: clike
   codemirror_mime_type: text/x-csrc
   language_id: 85
+Dafny:
+  type: programming
+  extensions:
+  - ".dfy"
```
Alhadis, line 1105: According to the [project's wiki](https://github.com/dafny-lang/dafny/wiki/Cautionary-Tales#understand-dafnys-command-line-interface), there's an interpreter named `dafny`. This should be added under `interpreters`.

Add this:

~~~diff
 Dafny:
   type: programming
   extensions:
   - ".dfy"
+  interpreters:
+  - dafny
   tm_scope: text.dfy.dafny
~~~

samples/Dafny/Problem1-SumMax.dfy
```diff
@@ -0,0 +1,26 @@
+// From https://rise4fun.com/Dafny/Problem1-SumMax
+
+// This method computes the sum and max of a given array of
+// integers.  The method's postcondition only promises that
+// 'sum' will be no greater than 'max'.  Can you write a
+// different method body that also achieves this postcondition?
+// Hint: Your program does not have to compute the sum and
+// max of the array, despite the suggestive names of the
+// out-parameters.
```
Alhadis, line 9: This example looks like it was taken from a tutorial. We prefer real-world samples to train our classifier. Could you provide more "realistic" use-cases?

lib/linguist/languages.yml
robin-aws: That seems reasonable. It isn't technically an interpreter since it cross-compiles Dafny to several other target languages instead, but who's counting. :)

What does linguist do with this field exactly?

samples/Dafny/Problem1-SumMax.dfy
robin-aws: For sure. I assume these samples are not intended to be useful for humans to understand the language but only for the classifier? Do we have a good sense of how much code tends to provide adequate training?

lib/linguist/languages.yml
robin-aws: I should say that if you pass the right flag to `dafny` it will compile and then run your program, so it can be used like an interpreter. If the point of this field is to allow the classifier to infer language from shebang lines, then that's valid, even though it's unlikely anyone would use Dafny that way.

robin-aws:
@Alhadis let me know if those samples are adequate. We could also add https://github.com/microsoft/Ironclad/blob/master/ironfleet/src/Dafny/Libraries/Math/mul.i.dfy if it helps to have the common `lemma` keyword in there as well.

lib/linguist/languages.yml
Alhadis: > If the point of this field is to allow the classifier to infer language from shebang lines

Yep, that's pretty much it. 👍

robin-aws:
Thanks @Alhadis! Any concerns @pchaigno?

lib/linguist/languages.yml
```diff
@@ -1099,6 +1099,15 @@ DTrace:
...
+  interpreters:
+  - dafny
+  tm_scope: text.dfy.dafny
+  ace_mode: text
+  language_id: 969323346
```
lildude, line 1110: I see you've not specified a colour for the language. Is this intentional? Without an explicit colour, the language will show up using the default `#cccccc` on some views like the language bar and with no colour at all on others like the repo listing.

robin-aws: No, I didn't include it only because it wasn't strictly required, but I can propose a color. Is there a place I can view all currently used colors, or render them from the library?

lildude: Unfortunately not as it's not simply a matter of picking a free colour as the colour also can't be too similar to another already in use. https://github.com/github/linguist/blob/master/test/test_color_proximity.rb is the test that enforces the proximity.  I'm trying to get the GitHub designers to allow me to drop this but my request is in the backlog of things to consider 😞.

The best thing to do is select a colour from the language's branding (eg logo, website, docs) and then run that test and keep tweaking the colour until you no longer get an error.

By starting with a colour from the branding, you also reduce the chances of someone else wanting to change it on a whim and thus have be directed to the requirements detailed at https://github.com/github/linguist/blob/master/CONTRIBUTING.md#changing-the-color-associated-with-a-language for changing the colour.

robin-aws: Thanks! I managed to wedge our preferred color in-between Javascript and Isabelle with minor not-human-perceptible tweaking. :)

PR #4856, opened 2020-05-01
Updated README to more clearly state that linguist runs in a Git repo

arkban:
Attempt to help others avoid the same mistake I made https://github.com/github/linguist/issues/4854 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ project.languages      #=> { "Ruby" => 119387 }
 
 ### Command line usage
 
+#### Git Repository
+
 A repository's languages stats can also be assessed from the command line using the `github-linguist` executable.
 Without any options, `github-linguist` will output the breakdown that correlates to what is shown in the language stats bar.
 The `--breakdown` flag will additionally show the breakdown of files by language.
@@ -102,7 +104,7 @@ github-linguist
 You can try running `github-linguist` on the root directory in this repository itself:
 
 ```console
-$ bundle exec bin/github-linguist --breakdown
+$ github-linguist --breakdown
 68.57%  Ruby
 22.90%  C
 6.93%   Go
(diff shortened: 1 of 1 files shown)


arkban:
CC @lildude

README.md
```diff
@@ -120,6 +122,19 @@ lib/linguist.rb
...
 ```
 
+#### Single file
+
+Alternatively you can find stats for a single file using the `github-linguist` executable.
+
+You can try running `github-linguist` on files in this repository itself:
+
+```console
+$ bundle exec bin/github-linguist grammars.yml
```
lildude, line 132: Whooops. I've just realised we detail global gem installation instructions and then suggest using bundler to exec which won't work.

Mind changing this to:

```suggestion
$ github-linguist grammars.yml
```

... and whilst you're at it, would you mind doing the same in the "Git Repository" section above? 🙇‍♂️

README.md
```diff
@@ -161,6 +176,15 @@ Linguist detects the language of a file but the actual syntax-highlighting is po
 If you experience an issue with the syntax-highlighting on GitHub, **please report the issue to the upstream grammar repository, not here.**
 Grammars are updated every time we build the Linguist gem so upstream bug fixes are automatically incorporated as they are fixed.
 
+### I get an error when using Linguist on a directory that is not a Git repo
```
lildude, line 179:
```suggestion
### I get an error when using Linguist on a directory that is not a Git repository
```

README.md
```diff
@@ -161,6 +176,15 @@ Linguist detects the language of a file but the actual syntax-highlighting is po
...
+
+When specifying a directory Linguist only works on Git repos -- at the root -- and only on committed changes. It's
+primary use is on GitHub.com which uses bare repositories and thus changes need to be committed as individual files don't show on the filesystem.
```
lildude, line 182: I think we can simplify this. How about a slight wording tweak:

```suggestion
Linguist only works on Git repositories and individual files. Its primary use is on GitHub.com which uses bare
repositories and thus changes need to be committed as individual files don't show on the filesystem.
```

README.md
```diff
@@ -161,6 +176,15 @@ Linguist detects the language of a file but the actual syntax-highlighting is po
...
+
+As a work around you could initialise a temporary Git repo in your directory as demonstrated in this
```
lildude, line 184: We prefer not to abbreviate in documentation outside of code samples.

```suggestion
As a work around you could initialise a temporary Git repository in your directory as demonstrated in this
```

arkban: I can appreciate that.

arkban:
Updated PR, I used your wording verbatim (since this is your repo not mine ; )

PR #4864, opened 2020-05-13
charlock_holmes 0.7.7

mistydemeo:
This bumps charlock_holmes to a minimum of 0.7.7. This version has some improvements to the binary build, but no actual code changes.

/cc @febuiles


febuiles:
@mistydemeo thank you 🙇 

@lildude: To provide a bit more context here: The older `charlock_holmes` version sometimes fails to build on macOS 10.14.6 with the command-line tools ~11.3.1. [This changeset](https://github.com/brianmario/charlock_holmes/compare/0.7.6...v0.7.7) addresses the problem.

Issue #4884, opened 2020-06-11
The state of SQL detection


smola:
## Problem

Linguist's detection of languages in the SQL family is currently quite inconsistent. The classifier seems to be quite biased towards PL/pgSQL, classifying a lot of SQL variants as PL/pgSQL (examples: [mysql](https://github.com/search?o=desc&q=mysql+language%3APLpgSQL&s=indexed&type=Code), [oracle](https://github.com/search?o=desc&q=oracle+language%3APLpgSQL&s=indexed&type=Code)). This has been the subject of a few issues before: [SQLite vs PLpgSQL](https://github.com/github/linguist/issues/4707), [MySQL vs PLpgSQL](https://github.com/github/linguist/issues/4706), [TSQL vs PLSQL](https://github.com/github/linguist/issues/4706) or  [TSQL vs PLpgSQL](https://github.com/github/linguist/issues/4476), [PLpgSQL vs SQLPL](https://github.com/github/linguist/issues/2384).

Furthermore, current heuristics for the `.sql` extension seem to imply that PostgreSQL' dialect of SQL is [PL/pgSQL](https://www.postgresql.org/docs/12/plpgsql.html), IBM DB2's dialect is [SQL PL](https://www.ibm.com/support/knowledgecenter/SSCJDQ/com.ibm.swg.im.dashdb.apdv.sqlpl.doc/doc/c0011916.html) and Oracle's dialect is [PL/SQL](https://docs.oracle.com/database/121/LNPLS/toc.htm). But neither of these are true. All of them are extensions to SQL included that are just a subset of their SQL dialect. In practice, most PostgreSQL code out there does not use PL/pgSQL at all.

It is a bit tricky to disambiguate properly between all of them, but I think we should change towards falling back to SQL more often.

## Possible Solutions

I'm not sure what's the best course of action, but I think it could include some of the following options. I will go ahead and create PRs for some of them, but I'd like to hear your opinions on the bigger picture.

### Heuristics fallback to SQL, no classifier

We currently have a very broad expression to trigger the classifier, as the final rule for `.sql`:

https://github.com/github/linguist/blob/2a6d952df870807034c3830738f95ad33222aca0/lib/linguist/heuristics.yml#L459-L461

Any of these strings may appear in virtually every SQL dialect. In particular `BEGIN`, which is standard SQL. Given that it is unlikely that the classifier does well here, it may be better to just fall back to SQL:

```yaml
  # Fallback to SQL
  - language: SQL
```

[...]

PR #4885, opened 2020-06-11
Relax T-SQL heuristic

smola:
## Description

Relax T-SQL heuristic to avoid a lot of misclassifications of other SQL dialects as T-SQL.

- Remove negative pattern for T-SQL whose only goal or effect is to
    trick linguist tests. This pattern hides misclassifications in the test
    samples that still happen in real world SQL.
- Remove IF/ELSE, which is valid MySQL, Oracle, PL/pgSQL, etc.
- Remove FETCH, which is valid SQL:2011, DB2, Oracle, etc.
- Remove CURSOR, which is valid MySQL, Oracle, etc.
- Remove DEALLOCATE, which is valid MySQL.
- Remove CREATE PROCEDURE, valid in many databases.
- Remove CHOOSE and IIF (too broad).
- Change DECLARE (valid MySQL) to DECLARE\s+@
- Change GO (too broad) to ^\s*GO\b
- Change @@ (too broad) to @@[_A-Z]\w*\b

## Checklist:
    - There are many but without clear license:
      - https://github.com/TrustDB/TrustSQL/blob/e7e9c056b9f60d3bc3e0343390055a9a8854efe4/storage/spider/scripts/install_spider.sql
      - https://github.com/search?l=&o=desc&p=2&q=if+else++mysql+language%3ATSQL&s=indexed&type=Code

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -453,9 +453,7 @@ disambiguations:
     pattern: '(?i:\$\$PLSQL_|XMLTYPE|sysdate|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function)'
   # T-SQL
   - language: TSQL
-    and:
-    - negative_pattern: '(?i:IDENTIFIED|NUMBER|VARCHAR2|REPEAT|UNTIL|IMMEDIATE)'
-    - pattern: '(?i:(GO)|(@@)|(CREATE PROCEDURE)|BEGIN( TRY| CATCH)|OUTPUT( INSERTED)|IF|ELSE|IIF|CHOOSE|CURSOR|FETCH|DEALLOCATE|DECLARE)'
+    pattern: '(?i:^\s*GO\b|@@[_A-Z]\w*\b|BEGIN( TRY| CATCH)|OUTPUT INSERTED|DECLARE\s+@)'
   # Generic SQL
   - language: SQL
     negative_pattern: '(?i:begin|boolean|package|exception)'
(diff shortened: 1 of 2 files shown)


beau-witter:
I am fine with the changes. It was pretty difficult for me to find unique aspects of T-SQL that differentiate it from other flavors, but ultimately I just wanted to add it as a language as it was missing and also a very common one.

I was also coming from knowledge of TSQL but not of many of the other flavors so I am glad someone like smola decided to take a look at this ;)

yourpropertyexpert:
I can give an extreme example of how this (I believe) fixes something.

On the branch running on github's code classifier (which I believe to be "main"):

`SELECT @@version;`

... is identified, not as SQL, but as TSQL. Unless I'm mis-reading the regex in last bullet on the change, that's something that is currently wrong and this would fix.

I understand that @@version isn't in the ANSI standard, and was originally a Microsoft / Sybase thing...

... but I can also run the it on 5.7.28 MySQL Community Server (GPL) and get:

```
mysql> select @@version;
+-----------+
| @@version |
+-----------+
| 5.7.28    |
+-----------+
1 row in set (0.00 sec)

```

I'm afraid I don't have the knowledge of this repo's codebase to try to dive in and add that as a test... but it's at least source data for a simple bit of SQL that is currently mis-classified and this patch should fix.

If it would help, I can give more examples of things currently mis-classified as TSQL which were written for MySQL 5.7

smola:
@pchaigno  Added one sample of code targetting SQLite that uses `ELSE`, maybe other tokens here too. This one is currently marked as TSQL by linguist.

It is extracted from the jOOQ project (Apache 2 license): https://github.com/jOOQ/jOOQ/blob/88aa871b8ed998abd73606395e5acaf9efb0876e/jOOQ-examples/Sakila/sqlite-sakila-db/sqlite-sakila-schema.sql
But the file is actually authored by DB Software Laboratory and licensed under BSD, originally a work MySQL and licensed under New BSD. Anyway, I think all licenses from the original file to its latest form are acceptable here.

There are more interesting samples in the jOOQ project, such as SQL for DB2 identified as TSQL: https://github.com/jOOQ/jOOQ/search?l=tsql&p=1

PR #4887, opened 2020-06-11
Relax SQL PL and PL/pgSQL heuristics

smola:
## Description

For SQL PL:
- Removed BEGIN [NOT] ATOMIC, it is standard and supported in various databases.
- Removed SIGNAL SQLSTATE [0-9]+, it is valid MySQL.
- Added MODE DB2SQL, which is DB2-specific and keeps correct
  classification for our samples.
- Removed LANGUAGE SQL, valid MySQL and PostgreSQL.

For PL/pgSQL:

- Removed  SECURITY DEFINER/CALLER, it is standard and supported in
   various databases.

ALTER MODULE could also be removed, since it is standard and has other implementation (e.g. [Crate](https://crate.io/docs/sql-99/en/latest/chapters/26.html#alter-module-statement)). But it seems that real world usage is mostly DB2, so I've kept it at the moment.

## Checklist:
    - Sample source(s): https://github.com/jOOQ/jOOQ/blob/88aa871b8ed998abd73606395e5acaf9efb0876e/jOOQ-examples/Sakila/mysql-sakila-db/mysql-sakila-schema.sql
    - Sample license(s): New BSD

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -444,10 +444,10 @@ disambiguations:
   rules:
    # Postgres
   - language: PLpgSQL
-    pattern: '(?i:^\\i\b|AS \$\$|LANGUAGE ''?plpgsql''?|SECURITY (DEFINER|INVOKER)|BEGIN( WORK )?;)'
+    pattern: '(?i:^\\i\b|AS \$\$|LANGUAGE ''?plpgsql''?|BEGIN( WORK )?;)'
   # IBM db2
   - language: SQLPL
-    pattern: "(?i:(alter module)|(language sql)|(begin( NOT)+ atomic)|signal SQLSTATE '[0-9]+')"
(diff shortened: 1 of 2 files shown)


smola:
There is an interesting sample for this one:
- Imported from jOOQ, but originally from the Sakila sample database (New BSD, [see here](https://dev.mysql.com/doc/sakila/en/sakila-license.html)): https://github.com/jOOQ/jOOQ/blob/88aa871b8ed998abd73606395e5acaf9efb0876e/jOOQ-examples/Sakila/mysql-sakila-db/mysql-sakila-schema.sql

This sample is for MySQL, and contains constructs catched by the SQL PL heuristic, in theory. But it is still classified as PLpgSQL, so it cannot be included in this PR unless the SQL vs PLpgSQL classification is fixed before.

smola:
Updated with a sample and with a change to the PL/pgSQL heuristic, required in conjunction to the changes to SQL PL to support the added sample.

PR #4888, opened 2020-06-11
.sql heuristic falls back to SQL

smola:
## Description

Do not use the classifier strategy for `.sql`, fallback to SQL. The premise here is that classifying any SQL variant as SQL is less wrong than classifying it as the wrong variant.

Adds some new patterns to SQL PL and T-SQL to keep everything working, at least for current samples.

## Checklist:
    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -452,16 +452,15 @@ disambiguations:
     pattern: '(?i:^\\i\b|AS \$\$|LANGUAGE ''?plpgsql''?|BEGIN( WORK )?;)'
   # IBM db2
   - language: SQLPL
-    pattern: "(?i:alter module|MODE DB2SQL)"
+    pattern: '(?i:alter module|MODE DB2SQL|\bSYSCAT\.\b|\bSYSPROC\.\b|ASSOCIATE RESULT SET|\bEND!\s*$)'
   # Oracle
   - language: PLSQL
     pattern: '(?i:\$\$PLSQL_|XMLTYPE|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function)'
   # T-SQL
   - language: TSQL
-    pattern: '(?i:^\s*GO\b|BEGIN( TRY| CATCH)|OUTPUT INSERTED|DECLARE\s+@)'
-  # Generic SQL
+    pattern: '(?i:^\s*GO\b|BEGIN( TRY| CATCH)|OUTPUT INSERTED|DECLARE\s+@|\[dbo\])'
+  # Fallback to SQL
   - language: SQL
-    negative_pattern: '(?i:begin|boolean|package|exception)'
 - extensions: ['.srt']
   rules:
   - language: SubRip Text
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -449,17 +449,16 @@ disambiguations:
   rules:
    # Postgres
   - language: PLpgSQL
-    pattern: '(?i:^\\i\b|AS \$\$|LANGUAGE ''?plpgsql''?|BEGIN( WORK )?;)'
+    pattern: '(?i:^\\i\b|AS\s+\$\$|LANGUAGE\s+''?plpgsql''?|BEGIN(\s+WORK)?\s*;)'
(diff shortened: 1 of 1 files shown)


lildude:
Makes sense to me. Any thoughts @pchaigno and @Alhadis?

lib/linguist/heuristics.yml
```diff
@@ -453,16 +453,15 @@ disambiguations:
...
-    pattern: "(?i:alter module|MODE DB2SQL)"
+    pattern: '(?i:alter module|MODE DB2SQL|\bSYSCAT\.\b|\bSYSPROC\.\b|ASSOCIATE RESULT SET|\bEND!\s*$)'
   # Oracle
   - language: PLSQL
     pattern: '(?i:\$\$PLSQL_|XMLTYPE|systimestamp|\.nextval|connect by|AUTHID (DEFINER|CURRENT_USER)|constructor\W+function)'
   # T-SQL
   - language: TSQL
-    pattern: '(?i:^\s*GO\b|BEGIN( TRY| CATCH)|OUTPUT INSERTED|DECLARE\s+@)'
-  # Generic SQL
+    pattern: '(?i:^\s*GO\b|BEGIN( TRY| CATCH)|OUTPUT INSERTED|DECLARE\s+@|\[dbo\])'
```
Alhadis, line 462: Here's something a little clearer:

~~~regex
(?i:^\s*GO\b|BEGIN +(TRY|CATCH)|OUTPUT +INSERTED|DECLARE\s+@|\[dbo\])
~~~

Note that if vertical whitespace is permitted where ` +`is used above, then the pattern should instead be:

~~~regex
(?i:^\s*GO\b|BEGIN\s+(TRY|CATCH)|OUTPUT\s+INSERTED|DECLARE\s+@|\[dbo\])
~~~

lib/linguist/heuristics.yml
```diff
@@ -453,16 +453,15 @@ disambiguations:
...
+  # Fallback to SQL
```
Alhadis, line 463: Not sure we even need this comment; it's fairly obvious it's an open-ended fallback.

lib/linguist/heuristics.yml
```diff
@@ -453,16 +453,15 @@ disambiguations:
     pattern: '(?i:^\\i\b|AS \$\$|LANGUAGE ''?plpgsql''?|BEGIN( WORK )?;)'
   # IBM db2
   - language: SQLPL
-    pattern: "(?i:alter module|MODE DB2SQL)"
+    pattern: '(?i:alter module|MODE DB2SQL|\bSYSCAT\.\b|\bSYSPROC\.\b|ASSOCIATE RESULT SET|\bEND!\s*$)'
```
Alhadis, line 456: 1. Since the pattern's case-insensitive, we should uppercase `alter module` so it's consistent with the rest of the pattern. To a casual reader, the discrepancy might be mistaken as deliberate.

2. `\.\b|` is a little vague about its intent, because `\.` is already a word-`\b`oundary, and nothing follows to indicate which sort of character is expected after it. For example `\.\b` would match `A.B`, but not `A. `. If in doubt, remove it entirely.

Suggested replacement:

~~~diff
-pattern: '(?i:alter module|MODE DB2SQL|\bSYSCAT\.\b|\bSYSPROC\.\b|ASSOCIATE RESULT SET|\bEND!\s*$)'
+pattern: '(?i:ALTER +MODULE|MODE +DB2SQL|\bSYS(CAT|PROC)\.|ASSOCIATE +RESULT +SET|\bEND!\s*$)'
~~~

Also, see my below remarks about ` +` being replaced by `\s+` instead.

Alhadis (changes requested):

lib/linguist/heuristics.yml
smola: Removed.

lib/linguist/heuristics.yml
smola: Changed everything to `\s+`, indeed, all kinds of spaces are allowed.

lib/linguist/heuristics.yml
smola: Added  your suggestions (with `\s+`).

Alhadis (approved):

PR #4889, opened 2020-06-12
[docs] Adding note on requirement of path/** for folders

igorsantos07:
## Description
While the docs say "git-style path matchers", that's not so clear; [`.gitignore` matches are different from `.gitattributes` in two aspects](https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Unspecified), and the most important is that it doesn't recursively-match folders. Took me a while to understand the issue :)

As an educated guess, developers are more used to writing `.gitignore` rules, so I included an extra note to explain that knowledge won't translate exactly 100% to `.gitattributes`. I'm not sure how to best format or word it, but that's an idea of what I would do :)

## Checklist

 _I removed the entire section as this is just a docs change and nothing really applied_


igorsantos07:
Ping

Igor Santos
-- Desenvolvedor Web
[enviado do meu celular]

igorsantos07:
No maintainer at all that could merge a simple doc change?

pchaigno (changes requested):
Sorry for the delay to review.

I'm not sure I agree with the change. Not that it is incorrect, but I don't think the right place for this information is Linguist's README. The right place is the git documentation. And I'd expect users who have trouble with their Linguist overrides to check the git documentation.

One thing we can do however is add one such example in the Linguist override examples we already have below. That wouldn't make our already-lengthy README even longer.

pchaigno:
Actually, @Alhadis pointed out in #4992 that `dir/**` is the same as `dir/*`. Since the git documentation was recommending `dir/**`, I tried it out for myself and `dir/**` and `dir/*` are indeed equivalent as far as I can see (checked with `a/* linguist-language=PHP` on file `a/b/c/test.zz`). Was there anything that made you think otherwise?

igorsantos07:
There's quite some time I fumbled with this and I don't want to go all the way back to re-test it - it was quite painful to get it all working.

Thus, all I can say is that it's what worked for my project after a lot of tries. Our particular use was to skip generated build files, instead of fixing languages. 

Would you test **/c/** against a/b/c/d/e/buildstuff.txt, just in case? That's what we ended with in the so-called offending project.

igorsantos07:
Ping

lildude:
I agree with @pchaigno here. This is really about understanding the git documentation and what the developer's goal is with the pattern matching they choose to use.

> Would you test **/c/** against a/b/c/d/e/buildstuff.txt, just in case? That's what we ended with in the so-called offending project.

Sure, but you've not told us what your complete and ultimate goal is as the rule you use is dependent on it.

We'll start by setting up a test repo:

```console
$ git init -q attrib-test
$ cd attrib-test/
$ mkdir -p a/b/c/d/e/f
$ find a/ -type d -exec sh -c 'for d; do touch "$d/file.txt"; done' _ {} +
$ touch .gitattributes
$ git add -A
$ git commit -m 'init'
[main (root-commit) a44907f] init
 7 files changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 .gitattributes
 create mode 100644 a/b/c/d/e/f/file.txt
 create mode 100644 a/b/c/d/e/file.txt
 create mode 100644 a/b/c/d/file.txt
 create mode 100644 a/b/c/file.txt
 create mode 100644 a/b/file.txt
 create mode 100644 a/file.txt
$
```

With this in place, we can start testing.

If your goal is to affect _only_ the files in `a/b/c/`, then `a/b/c/*` does the trick:

```console
$ echo 'a/b/c/* crlf' > .gitattributes && git commit -am 'Add rule'
[main 9192b51] Add rule
 1 file changed, 1 insertion(+)
$ git check-attr crlf a/file.txt
a/file.txt: crlf: unspecified
$ git check-attr crlf a/b/file.txt
a/b/file.txt: crlf: unspecified
$ git check-attr crlf a/b/c/file.txt
a/b/c/file.txt: crlf: set
$ git check-attr crlf a/b/c/d/file.txt
a/b/c/d/file.txt: crlf: unspecified
$ git check-attr crlf a/b/c/d/e/file.txt
a/b/c/d/e/file.txt: crlf: unspecified
$
```

Note how it doesn't affect files in the `d/` and `e/` sub-dirs.

If your goal is to affect _all_ the files in `a/b/c/` and _all_ sub-dirs, ie to recurse, then `a/b/c/**` does the trick:

```console
$ echo 'a/b/c/** crlf' > .gitattributes && git commit -am 'Add rule'
[main 97dcd81] Add rule
 1 file changed, 1 insertion(+), 1 deletion(-)
$ git check-attr crlf a/file.txt
a/file.txt: crlf: unspecified
$ git check-attr crlf a/b/file.txt
a/b/file.txt: crlf: unspecified
$ git check-attr crlf a/b/c/file.txt
a/b/c/file.txt: crlf: set
$ git check-attr crlf a/b/c/d/file.txt
a/b/c/d/file.txt: crlf: set
$ git check-attr crlf a/b/c/d/e/file.txt
a/b/c/d/e/file.txt: crlf: set
$
```

In both cases, the files in the `a/b/c/` dir are considered, but it's only the latter that affects sub-dirs due to the applied recursion.

If your goal is to only affect files in `a/b/c/` and `a/b/c/d/e/f/`, (not `a/b/c/d/` or `a/b/c/d/e/`) the second rule is too aggressive and you may not realise it until a later date and may not know why.

Ultimately, this is really about understanding the git documentation and the developer's end goals, which are both outside the scope of using Linguist. I've deliberately _not_ used Linguist-specific attributes to help illustrate this.

PR #4893, opened 2020-06-16
Support Facility Service Definition

jzbrooks:
## Checklist:

    This language is used heavily internally in my company's GHE instance. It is unclear to me if there's a way to configure linguist for internal use only.
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Afsd+service+NOT+nothack
    - Sample source(s):
      - https://facilityapi.github.io/editor/
    - Sample license(s): MIT

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,6 +28,9 @@
 [submodule "vendor/grammars/Elm"]
 	path = vendor/grammars/Elm
 	url = https://github.com/elm-community/Elm.tmLanguage
+[submodule "vendor/grammars/FacilitySublime"]
+	path = vendor/grammars/FacilitySublime
+	url = https://github.com/FacilityApi/FacilitySublime
 [submodule "vendor/grammars/FreeMarker.tmbundle"]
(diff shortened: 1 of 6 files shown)


jzbrooks:
It's unclear to me how I'm supposed to choose a color. It appears that I can't choose a previously used color and also can't choose an unused color.

lildude:
> It's unclear to me how I'm supposed to choose a color. It appears that I can't choose a previously used color and also can't choose an unused color.

I'm afraid there's no easy way other than trial and error and repeatedly running the colour proximity test until you find a free colour that isn't too close to an already used colour.

That said, as this is a data language, it won't appear in the language bar unless someone expressly forces it to be "detectable" using an override, so a colour isn't really needed.

But the biggest blocker to this PR is going to be the popularity requirment: the in-the-wild usage on GitHub.com appears to be incredibly low so doesn't meet the minimum requirements for acceptance right now anyway. Your search result returns 39 files which is a long way from _"extensions be in use in hundreds of repositories"_.

PR #2954, opened 2016-04-16
Classify builds, pkgproj, resx and sfproj as XML

jasonwilliams200OK:
* `.builds` (MSBuild): https://github.com/search?utf8=%E2%9C%93&q=extension%3Abuilds+xml&type=Code&ref=searchresults

* `.pkgproj` (Nuget): https://github.com/search?utf8=%E2%9C%93&q=extension%3Apkgproj+xml&type=Code&ref=searchresults

* `.resx` (Resource): https://github.com/search?utf8=%E2%9C%93&q=extension%3Aresx+xml&type=Code&ref=searchresults

* `.sfproj` (Azure Service Fabric): https://github.com/search?utf8=%E2%9C%93&q=extension%3Asfproj+xml&type=Code&ref=searchresults

PR #4904, opened 2020-06-26
Add Solidity language heuristic

alcuadrado:
I added an entry to `lib/linguist/heuristics.yml` for `.sol` files. It checks a pragma that the vast majority of Solidity files have.

My understanding is that `.sol` wasn't already associated with Solidity, as it's commonly used for solutions in competitive programming environments. Luckily, most Solidity files have a program indicating the version of the language, as not including it leads to a warning. This makes creating a heuristic simple and safe.

## Checklist:
  - [x] **The new extension is used in hundreds of repositories on GitHub.com**
    - [Search results for each extension](https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3ASOL+Solidity+NOT+nothack)
  - [x] **I have included a change to the heuristics.** ~~to distinguish my language from others using the same extension~~


lib/linguist/heuristics.yml
```diff
@@ -440,6 +440,10 @@ disambiguations:
     pattern: '(?i:\^(this|super)\.|^\s*~\w+\s*=\.)'
   - language: Scala
     pattern: '(^\s*import (scala|java)\.|^\s*class\b)'
+- extensions: ['.sol']
```
pchaigno, line 443: This change currently won't do anything because Linguist doesn't know about the `.sol` file extension. Please see other pull requests trying to add the `.sol` extension for background.

alcuadrado:
Thanks for the quick response, @pchaigno.

My understanding now is that any heuristic that associates `.sol` with Solidity can't be merged until one of these things happens:

1. The `not-a-language` feature is implemented.
2. I add heuristics for every language that uses `.sol` as an extension.

Is guess (1) is a major endeavor, and maybe not a good fit for being implemented by an external contributor. Is it?

If I were to go on with (2):

* Do I have to create heuristics for any kind of `.sol` file that I find on Github, or just the ones that have language definitions in Linguist?
* Is there a way to download every file with a given extension? Using the API maybe?

pchaigno:
> Is guess (1) is a major endeavor, and maybe not a good fit for being implemented by an external contributor. Is it?

Unfortunately, I would tend to think this is true for both (1) and (2). Both will require a lot of work and are likely hard to undertake as a first contribution.

> Do I have to create heuristics for any kind of .sol file that I find on Github, or just the ones that have language definitions in Linguist?

We usually add support for all languages that (1) use an extension we're adding support for and (2) meet our requirement on in-the-wild usage (i.e., hundreds of repositories).

> Is there a way to download every file with a given extension? Using the API maybe?

I don't know of a way to download **all** files with a given extension, but you can use [Harvester](https://github.com/Alhadis/Harvester) to download all files returned by a given search query.

lib/linguist/heuristics.yml
Alhadis: Just tried this for myself to see if there was an easier way to add Solidity support than defining an entirely new strategy for it (following e-mail discussion with.

It, uh, actually works perfectly fine... 😂

lib/linguist/heuristics.yml
```diff
@@ -440,6 +440,10 @@ disambiguations:
...
+  rules:
+  - language: Solidity
+    pattern: 'pragma\s+solidity'
```
Alhadis, line 446: **Nitpick:** I'd add word boundaries to this expression, so something like `not_a_pragma solidity1` won't match:

~~~diff
   rules:
   - language: Solidity
-     pattern: 'pragma\s+solidity'
+     pattern: '\bpragma\s+solidity\b'
~~~

alcuadrado: Makes sense!

alcuadrado:
> 1. Please let me know if there's anything else I could match in older Solidity files that's unambiguous.

I don't think so. The other top-level constructs, apart from `contract`, weren't present in versions of Solidity without `pragma`.

lib/linguist/heuristics.yml
Alhadis: Wait, I may have spoken too soon. Linguist identifies a solitary `.sol` file, but acts strangely when there are several. Specifically, only the last `.sol` file it's given contributes anything to the repository's language statistics. Every other Solidity file gets ignored.

I'll need to study this closer...

lib/linguist/heuristics.yml
Alhadis: Done. 👍 I've pushed the changes to your branch, so this is good to merge, IMO.

Alhadis:
> Throughout Linguist's source code, we assume the list of extensions in `languages.yml` to be the source of truth in terms of what extensions we support. We refer to that list to know what to test, to find conflicts, to determine aliases... and to detect languages of course. I would really prefer to have all extensions we support listed in there.

I don't think that justifies the complexity involved with adding a new strategy or `languages` field. Especially when the underlying problem isn't really solved: Linguist still has no way of explicitly returning nothing.

Solutions I've considered:

* Adding an `isolate` field to `languages.yml` to exclude an entry from candidate lists unless a strategy explicitly requests it. **This is the approach I decided to implement before testing this PR for myself.**
* Adding a new heuristic strategy that *isn't* named `Heuristic` (since that's already taken 😜). Ditched the idea because it would repeat things other strategies already handle. Specifically, matching content and file extensions: that's already the job of our `Heuristic` and `Extension` strategies, respectively.
* Adding a `minimum_certainty` field to `languages.yml`, where matches are required to score highly enough in the Bayessian classifier to be returned. Confidence would be be derived from tokenisation metrics (those displayed when `LINGUIST_DEBUG=2` is set in the environment). Ditched the idea because:
	* Many smaller and less blatant files would slip through the cracks, causing some files to be recognised but not others
	* Setting the threshold requires some knowledge of how the tokenisation process works
	* Come on. We both know why this is a bad idea. 😉

lildude:
Sorry @Alhadis but I agree with @pchaigno here. This doesn't feel like the right solution and whilst solves the immediate problem, it's likely to lead to confusion and inconsistency later.

I'm short on ideas—I've pondered this all day yesterday—but short of implementing the whole `not-a-language` idea (I'm not sure this will really solve our problems now I've thought about it due to all the complications that come with it), the best option I can think of for now might be to add support for the most prominent `.sol` extensions as best as we can.

Linguist has never professed to be perfect in matching languages, but "good enough" should do for most cases. Overrides exist for this reason and most current `.sol` files won't change until the repo changes.

Alhadis:
> I've pondered this all day yesterday

So did I. 😂 It's amazing how difficult it is to think of a satisfactory `.sol`ution to our problem.

I guess a bespoke `Linguist::Strategy::Solidity` is really the least awkward way of fixing this.

lildude:
> I guess a bespoke `Linguist::Strategy::Solidity` is really the least awkward way to do this. I trust you'd have no objections to a PR to add such a thing…?

I think that's starting down a dangerous road of creating a new strategy for each language that uses a very popular or generic extension.

Alhadis:
The only other thing I can think of is a `require_heuristic` field in `languages.yml` that requires a heuristic to match in order for the extension or filename strategies to pass.

Alhadis:
Just for the record, our heuristics are already assigning languages for extensions they're not associated with (e.g., `INI` → `.props`, `.builds` → Text, probably others).

The PR I'm preparing will address this by requiring heuristics to reference extensions in-use by their candidate languages.

Alhadis:
#4936

pchaigno:
> Just for the record, our heuristics are already assigning languages for extensions they're not associated with (e.g., `INI` → `.props`, `.builds` → Text, probably others)

Those are bugs as far as I can see. We should probably have a test to ensure all extensions in `heuristics.yml` are listed in `languages.yml`. I always assumed that was already the case :-/
They were both merged in 2016 when we weren't as careful on these things: #2866 and #2954.

Alhadis:
I already included a test for this as part of #4936

Alhadis:
Closing in favour of #4936 (and so @stalebot will STFU).

PR #4907, opened 2020-06-28
Add Verifpal language extension

kaepora:
This adds the Verifpal language to Linguist.

## Checklist:

  - [x] The new extension is used in about a dozen repositories on GitHub.com
      
      - https://github.com/search?q=extension%3Avp+attacker&type=Code
    - Sample source(s):
      - examples/Verifpal
    - Sample license(s):


pchaigno:
The search query you included has only 11 files. Is that really all the Verifpal files there are on github.com, or is there a way to extend that search query to return more files without matching too many files from other languages? We require new file extensions to be used in hundreds of repositories before adding support to Linguist.

PR #4911, opened 2020-07-02
Add grammer for Singularity definition files

onnovalkering:
## Description
This PR adds a grammer for [Singularity](https://github.com/hpcng/singularity) definition files. These Singularity definition files have two filename conventions, either `Singularity` or `name.def`. This PR only marks the former, since `.def` also seems to be used for arbitrary files and config.

## Checklist:

      -  `Singularity` (filename): https://github.com/search?q=filename%3ASingularity+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/sylabs/examples/blob/master/message-broker/rabbitmq/rabbitmq.def
      - https://github.com/nickjer/singularity-r/blob/master/Singularity
    - Sample license(s):
      - `bsd-3-clause`: https://github.com/sylabs/examples/blob/master/LICENSE.md 
      - `mit`: https://github.com/nickjer/singularity-r/blob/master/LICENSE.txt
  
      - https://github-lightshow.herokuapp.com/?utf8=✓&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fonnovalkering%2Fvscode-singularity%2Fblob%2Fmaster%2Fsyntaxes%2Fsingularity.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fnickjer%2Fsingularity-r%2Fmaster%2FSingularity&code=
     - N/A

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5134,6 +5134,14 @@ Sieve:
   codemirror_mode: sieve
   codemirror_mime_type: application/sieve
   language_id: 208976687
+Singularity:
+  type: programming
+  color: "#202A51"
+  tm_scope: source.singularity
+  filenames:
+  - Singularity
+  ace_mode: text
+  language_id: 987024632
 Slash:
   type: programming
   color: "#007eff"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1022,6 +1022,9 @@
 [submodule "vendor/grammars/vscode-scala-syntax"]
 	path = vendor/grammars/vscode-scala-syntax
 	url = https://github.com/scala/vscode-scala-syntax
+[submodule "vendor/grammars/vscode-singularity"]
+	path = vendor/grammars/vscode-singularity
+	url = https://github.com/onnovalkering/vscode-singularity
 [submodule "vendor/grammars/vscode-slice"]
 	path = vendor/grammars/vscode-slice
 	url = https://github.com/zeroc-ice/vscode-slice
--- a/grammars.yml
+++ b/grammars.yml
@@ -876,6 +876,8 @@ vendor/grammars/vscode-prisma:
 - source.prisma
 vendor/grammars/vscode-scala-syntax:
 - source.scala
+vendor/grammars/vscode-singularity:
+- source.singularity
 vendor/grammars/vscode-slice:
 - source.slice
 vendor/grammars/vscode-vlang:
--- a/samples/Singularity/Singularity
+++ b/samples/Singularity/Singularity
@@ -1,0 +1,57 @@
+# https://github.com/nickjer/singularity-r/blob/master/Singularity (MIT)
+
+BootStrap: shub
+From: nickjer/singularity-r
+
+%labels
+  Maintainer Jeremy Nicklas
+  RStudio_Version 1.2.5033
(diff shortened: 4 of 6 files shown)


eburgueno:
Thank you for this! I came here looking to see if this had been reported already and I'm pleased that there's already a PR for it.

Question, since [Singularity Hub uses `Singularity[.<tag>]`](https://singularityhub.github.io/singularityhub-docs/docs/getting-started/recipes#multiple-recipes), could this look for `Singularity|Singularity\..*` filenames?

onnovalkering:
Hey @eburgueno

I tried to add the `Singularity[.<tag>]` pattern, but it seems that matching based on wildcards / Regex for filenames is not supported by linguist (?). I guess, eventually, [.gitattributes](https://github.com/github/linguist#using-gitattributes) can be used as a per-repo solution for `Singularity.<tag>` files.

While I was at it, I've added the `.def` extension, as advocated by the current Singularity [documentation](https://sylabs.io/guides/3.5/user-guide/definition_files.html).

pchaigno:
> I tried to add the `Singularity[.<tag>]` pattern, but it seems that matching based on wildcards / Regex for filenames is not supported by linguist (?). I guess, eventually, [.gitattributes](https://github.com/github/linguist#using-gitattributes) can be used as a per-repo solution for `Singularity.<tag>` files.

That's correct, we do not currently support this type of prefixes. I would definitely welcome a pull request adding such a new strategy. It is needed for several other patterns, such as `Makefile.xxx`, `Dockerfile.xxx`, `Jenkinsfile.xxx`, etc.

lib/linguist/languages.yml
```diff
@@ -5134,6 +5134,16 @@ Sieve:
   codemirror_mode: sieve
   codemirror_mime_type: application/sieve
   language_id: 208976687
+Singularity:
+  type: programming
+  color: "#64E6AD"
+  tm_scope: source.singularity
+  extensions:
+  - ".def"
```
pchaigno, line 5142: > The `.def` extension doesn't conflict with another language within linguist, but I've added a heuristic as it seems to be a commonly used extension for arbitrary text and config files.

If we add it for Singularity, we will also need to add it for other languages using it. Otherwise, Linguist will associate all `.def` files to Singularity, even if there's a heuristic rule defined for it.

That is likely to be a significant job since a lot of other languages appear to be using it.

lib/linguist/languages.yml
```diff
@@ -5134,6 +5134,16 @@ Sieve:
...
+  filenames:
+  - Singularity
```
pchaigno, line 5144: For other reviewers: it looks like there are a few Bash files with that name, but I'm not overly worried since our shebang strategy should detect them.

fzimmermann89:
@onnovalkering : What's the status on this? Are you still interested in getting this merged?  

Maybe it would be easier to get just the match on 'Singularity' merged ( addressing the requested changes with the least effort on your end) ?

Looking forward to getting this merged :)

onnovalkering:
@fzimmermann89 I was unable to update and reopen this PR, so I created a new PR for just `Singularity` : #5242.

PR #4913, opened 2020-07-05
remove '@@' pattern form T-SQL

smola:
## Description
@@ pattern was associated to T-SQL, but it is used in variables for more
dialects.

Added a simple fixture taken from MySQL 5.7 documentation:

https://dev.mysql.com/doc/refman/5.7/en/set-variable.html

## Checklist:
    - Simple fixture based on https://dev.mysql.com/doc/refman/5.7/en/set-variable.html


smola:
This one is required to add the new sample referenced here: https://github.com/github/linguist/pull/4887#issuecomment-649870166

pchaigno:
Would love a :+1: from @beau-witter if possible :-)

PR #4918, opened 2020-07-07
Documentation change for Mac users "contributing"

yourpropertyexpert:
## Description

For over a year (since May 2019), there has been a bug in the upstream project "Charlock Holmes" that means it will not install as part of the (linguist) scripts/boostrap script.

This was flagged to linguist in issue #4908, and some smart folks found a manual workaround that seems to work.

Sadly, because of the fact it's an upstream dependency, it's not really possible to fix it in code, but this PR updates the "Contributing" documentation to make it easier for folks with modern Macs.

No executable code was changed in the making of this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,13 +18,36 @@ Linguist uses the [`charlock_holmes`](https://github.com/brianmario/charlock_hol
 [Docker](https://www.docker.com/) is also required when adding or updating grammars.
 These components have their own dependencies - `icu4c`, and `cmake` and `pkg-config` respectively - which you may need to install before you can install Linguist.
 
+#### Installing dependencies on MacOS
+
 For example, on macOS with [Homebrew](http://brew.sh/):
 ```bash
 brew install cmake pkg-config icu4c
 brew cask install docker
 ```
 
-On Ubuntu:
+There are issues with the installation of Charlock Holmes on Macs (these have been reported upstream but there has been little progress in over a year.
+
+On Macs with newer versions of xcode, trying to run the bootstreap script below gives an error along the lines of:
+
+```/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/include/ruby-2.6.0/universal-darwin19/ruby/config.h:413:31:
+error: invalid suffix on literal; C++11 requires a space between literal and identifier
+```
+
+If you get this error, then you need to note exactly which file is causing the problem, and edit the lines that are mentioned. (The line numbers will vary depending on your version of MacOS.)
+
+You need to insert a space immediately before the word RUBY that follows the closing of a literal string. There will probably be two of these.
+
+```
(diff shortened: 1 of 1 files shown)


Alhadis:
In all honesty, I think we should be amending the `bootstrap` script somehow to accommodate this, rather than bloat the contributing docs. They're long enough already, and the issue itself is unlikely to be relevant in another 6-8 months (we hope).

yourpropertyexpert:
> In all honesty, I think we should be amending the `bootstrap` script somehow to accommodate this, rather than bloat the contributing docs. They're long enough already, and the issue itself is unlikely to be relevant in another 6-8 months (we hope).

Maybe, but they've had an open issue for over a year, and I don't see how having "shorter but guaranteed to fail" documentation is better than "actually contains what people need, even if it's a bit longer."

CONTRIBUTING.md
```diff
@@ -18,13 +18,36 @@ Linguist uses the [`charlock_holmes`](https://github.com/brianmario/charlock_hol
...
+If you get this error, then you need to note exactly which file is causing the problem, and edit the lines that are mentioned. (The line numbers will vary depending on your version of MacOS.)
+
+You need to insert a space immediately before the word RUBY that follows the closing of a literal string. There will probably be two of these.
+
+```
+OLD:
+#define RUBY_ARCH "universal-"RUBY_PLATFORM_OS
+CORRECTED:
+#define RUBY_ARCH "universal-" RUBY_PLATFORM_OS
+```
```
pchaigno, line 46: If there are currently only two lines that need fixing and they both follow the above pattern, then I think we should just edit them with `sed -i` in the `bootstrap` script as @Alhadis suggested.

Alhadis: > edit them with `sed -i` in

:wink: More portable way of doing that would be `sed -i.bak … && rm -f *.bak`, as some (older) `sed` implementations don't permit optional backup suffixes (meaning `sed -i file` gets interpreted as `sed -i " file"` instead.

(This is only an edge case, but one that's overlooked enough to warrant bringing up anyway)

yourpropertyexpert: OK.

So the project team have now told me two things:

- That they don't want to fix the problem at their end
- That they don't want to update their documentation to show Mac users how to workaround the problem

lildude:
As I understand it, this is a problem with the Ruby as shipped with macOS, not Linguist nor charlock_holmes. This is something Apple should be fixing as they provide the file but I very much doubt they will as they've already [announced plans to stop shipping Ruby](https://developer.apple.com/documentation/macos-release-notes/macos-catalina-10_15-release-notes#Scripting-Language-Runtimes) in future. Yes, charlock_holmes could probably workaround this but I see why they wouldn't want to as it would be for a single platform for a version of Ruby that won't exist for much longer.

I'm reluctant to document changing OS-provided files and definitely against scripting such a change. If we were to document something, I think it should be something along the lines of discouraging the use of the OS supplied Ruby on macOS.

PR #4922, opened 2020-07-10
add support for dotenv (Shell)

### LINKED ISSUE #4873: dotenv syntax highlighting support
Feature request to add syntax highlighting support for `.env` files. This was discussed in #3717 in 2017, but seems to have been closed without reason.

The `.env` syntax is used to define environment variables, and it's very commonly used with Node.js, Docker and probably others.

Although `.env` is often used for passwords, it's also commonly used for general config files which are fine to commit. Some repos have a private `.env` file and one that is committed - For example Vue and React use `.env` and `.env.local`. In addition to that, some repos have `.env-sample`, `.env.example`, or a code block in `README.md`, to show how private `.env` files should look.

References:
- [package.json](https://github.com/mikestead/vscode-dotenv/blob/master/package.json) of the VSCode dotenv extension which lists all the `.env` file extensions it supports.
- [dotenv](https://www.npmjs.com/package/dotenv) npm package

smola:
Fixes #4873

- Associate *.env, .env, .env.example and .flaskenv to Shell.

Add samples:

- .flaskenv from https://github.com/stripe/stripe-payments-demo/blob/70ba859bb68a650b35b6e85ddc889ad143055170/server/python/.flaskenv (MIT License)
- .env.example from https://github.com/proasis/api-rest/blob/670123448b19229677f0710751b2ef4e808f9bd3/.env.example (MIT License)
- .env from
    https://github.com/rabbitmq/tgir/blob/4299d2513dad4bbdae87b107c1f1ec0b14fc04a1/.env
    (Apache 2 License)
- default.env from
    https://github.com/diegoalmesp/trello-eureka/blob/43a4488f6df50be2a17cd9495ce29de6e4192b14/client/default.env
    (MIT License)

## Description

*dotenv files* is a convention to use a subset of shell (commonly known as *dotenv*) to setup an environment.

The modern convention probably originates from the Ruby world in 2012 (see [bkeepers/dotenv](https://github.com/bkeepers/dotenv)) and the supported syntax is roughly the following (although packages for different languages may support a more limited subset):

```shell
# comments
KEY=value
KEY="value"
export KEY=value
KEY=${variable}
KEY=$(command)
```

Anyway, you may find lots of samples in the wild using full Bash syntax (see the links to GitHub Search below).

## Checklist:
      - `.env`: https://github.com/search?q=filename%3A.env
      - `.env.example`: https://github.com/search?q=filename%3A.env.example
      - `*.env`: https://github.com/search?q=extension%3A.env+%28port+OR+host+OR+path+OR+user+OR+pass+OR+export%29
      - `.flaskenv`: https://github.com/search?q=filename%3A.flaskenv
  - ~I have included a real-world usage sample for all extensions added in this PR:~ (added 3 fixtures)

You may find some counter-examples, but not with wide usage as far as I see:
- `Dockerfile.env` (Dockerfile): https://github.com/search?q=filename%3Adockerfile.env
- `Make.env` (Makefile): https://github.com/search?q=filename%3Amake.env&type=Code


test/fixtures/Shell/main.env
```diff
@@ -0,0 +1,2 @@
+# comment
+export KEY=${ENV}
```
pchaigno, line 2: Would it be possible to find sample files from "real" applications here instead of using fixture files? Fixture files are not used to train the Bayesian classifier, so I'd really prefer to avoid them as much as possible.

smola: @pchaigno Ok, I'll add some!

smola:
Please, see the discussion the issue: https://github.com/github/linguist/issues/4873
Some languages may have syntax rules are not strictly a subset of Bash. At least for Node.js, where `VAR={"some": "json"}` is valid, but not in Bash (or any other mainstream shell). So we may still have some `.env` files with some highlighting errors after this change. Unfortunately, I don't see a better alternative, since using a dotenv-specific grammar may incorrectly highlight the files for many other languages that allow different extensions, or for bull Bash itself.

smola:
To add some more fun, the Ruby library supports YAML-like variables: https://github.com/bkeepers/dotenv/blob/master/spec/fixtures/yaml.env

lib/linguist/languages.yml
```diff
@@ -5067,6 +5068,9 @@ Shell:
   - ".bash_profile"
   - ".bashrc"
   - ".cshrc"
+  - ".env"
+  - ".env.example"
+  - ".flaskenv"
```
pchaigno, line 5073: Could you please update the template with a search query for this filename?

smola: Sure. Done.

pchaigno: Thanks!

lildude (approved):

PR #4924, opened 2020-07-11
Split Classic ASP and ASP.NET

### LINKED ISSUE #4867: ASP Classic vs ASP.NET
Hello,

Linguist does not make any distinction between Classic ASP and ASP.NET. That is confusing for both groups of developers.

Is it possible to separate both? Classic ASP files always have the extension ".asp". ASP.NET applications never have the extension .asp. 

Many thanks for taking this into consideration.

Grts, Pieter

smola:
## Description

Fixes #4867

ASP and ASP.NET are different languages. After the emergence of ASP.NET,
ASP has increasingly become known as Classic ASP (even if it was not its
original name).

Splitted the "ASP" language into "Classic ASP" and "ASP.NET". I moved
the assigned color and language id from ASP to Classic ASP, while
ASP.NET got a new id.

"ASP" is now an alias for "Classic ASP". This makes sense conceptually,
but it does not come without disruption for users who were making some
assumptions about alias resolution. Fortunately, ASP/ASP.NET extensions
are pretty unambiguous, so there shouldn't be many users out there using
.gitattributes to override detection.

Added samples for Classic ASP:

- samples/Classic ASP/sendingcontent-xml.asp
  MIT license, imported from https://github.com/superpikar/learn-classic-asp/blob/master/sendingcontent-xml.asp
- samples/Classic ASP/ASPUnitRunner.asp
  MIT license, imported from https://github.com/rpeterclark/aspunit/blob/master/Test/ASPUnitRunner.asp

Added samples for ASP.NET:

- samples/ASP.NET/Login.aspx
  Apache 2 License, imported from https://github.com/aspnet/samples/blob/56bae91f5769a207dcacbf0bc2474d6bd5218171/samples/aspnet/Identity/ChangePK/PrimaryKeysConfigTest/Account/Login.aspx
- samples/ASP.NET/Global.asax
  Apache 2 License, imported from https://github.com/aspnet/samples/blob/095853df8e09572f044084ca18215c7ff8baf019/samples/aspnet/Identity/SQLMembership-Identity-OWIN/SQLMembership-Identity-OWIN/Global.asax
- samples/ASP.NET/OpenAuthProviders.ascx
  Apache 2 License, imported from https://github.com/aspnet/samples/blob/095853df8e09572f044084ca18215c7ff8baf019/samples/aspnet/Identity/ChangePK/PrimaryKeysConfigTest/Account/OpenAuthProviders.ascx
- samples/ASP.NET/EchoSocket.ashx
  Apache 2 License, imported from https://github.com/aspnet/WebSockets/blob/56c5600b20342386de467bb02b7117a785518458/samples/AutobahnTestAppAspNet4/EchoSocket.ashx

## Checklist:
  - [x] I have included a real-world usage sample for all extensions added in this PR. (see description)
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/ *(not sure if I need action here?)
  - ~I have updated the heuristics to distinguish my language from others using the same extension.~ N/A


PieterCooreman (approved):

lib/linguist/languages.yml
```diff
@@ -131,15 +131,13 @@ ASN.1:
   codemirror_mode: asn.1
   codemirror_mime_type: text/x-ttcn-asn
   language_id: 7
-ASP:
+ASP.NET:
   type: programming
-  color: "#6a40fd"
```
lildude, line 136: Given ASP.NET is the language peeps are more likely to use today, we should probably give it a colour instead of falling to the default `#cccccc`

smola: @lildude Which color should it be?

lildude: I have no idea. Normally people aim for a colour related to the language branding.

smola: As far as I know, the current branding color for the dotnet platform is `#9400ff`, although it's the same branding for C#, VB.NET, etc.

lib/linguist/languages.yml
```diff
@@ -764,6 +762,16 @@ Clarion:
   - ".clw"
   tm_scope: source.clarion
   language_id: 59
+Classic ASP:
```
Alhadis, line 765: Wouldn't [`Active Server Pages`](https://en.wikipedia.org/wiki/Active_Server_Pages) be a more appropriate name? I'm only going by a cursory Google search though, so you're probably better-qualified for picking a name than I am.

(For as long as I've worked in web development, I've adamantly refused to touch ASP or anything rhyming with it 😜).

smola: Yes, as much as `Hypertext Markup Language` for `HTML` :wink: 

I would say the full name is not common currency, but I'm not sure, I'm not that much into the Microsoft world either :man_shrugging:

Alhadis: Fair enough. In any case, if there's a more appropriate name than `Classic ASP`, somebody will eventually let us know. 😂

Alhadis (approved):

lildude (approved):

PR #4926, opened 2020-07-20
Improve accuracy of ActionScript heuristic

Ryan-rsm-McKenzie:
## Description

The current Actionscript heuristic flags many Actionscript files as being Angelscript instead. In my test cases, I saw it incorrectly flag 680/1556 (43.70%) of files as being Angelscript, when they were instead Actionscript. By comparison, the new heuristic flags only [a single file](https://github.com/fdorg/flashdevelop/blob/development/FlashDevelop/Bin/Debug/Tools/mtasc/std/ArrayParam.as) as being Angelscript, which despite being Actionscript, is actually also valid Angelscript.

**The test cases for Actionscript included these repositories (1556 files):**
* [as3commons-ui](https://github.com/AS3Commons/as3commons-ui)
* [as3corelib](https://github.com/mikechambers/as3corelib)
* [casalib](https://github.com/aaronclinger/casalib)
* [flashdevelop](https://github.com/fdorg/flashdevelop)

**The test cases for Angelscript included these repositories (670 files):**
* [aatc](https://github.com/Sami-Vuorela/aatc)
* [kag-base-1](https://github.com/ANybakk/kag-base-1)
* [kag-base](https://github.com/ANybakk/kag-base)
* [kag-grassgrowth](https://github.com/ANybakk/kag-grassgrowth)
* [kag-transports](https://github.com/ANybakk/kag-transports)
* [kag-undeadinvasion](https://github.com/ANybakk/kag-undeadinvasion)

Note that it was very difficult to find Angelscript libraries.

The new heuristic makes the distinction between Angelscript/Actionscript files by looking for usage of Actionscript specific keywords, such as `var`, `intrinsic`, `module`, `extends`, `static`, and `function`, as well as Actionscript specific declaration syntax i.e.:
```
function foo(): Void;
var bar: Number = 42;
```
Where the colon is being used to denote types.

## Checklist:

    - Sample source(s):
    - Sample license(s):


lib/linguist/heuristics.yml
```diff
@@ -48,7 +48,13 @@ disambiguations:
 - extensions: ['.as']
   rules:
   - language: ActionScript
-    pattern: '^\s*(package\s+[a-z0-9_\.]+|import\s+[a-zA-Z0-9_\.]+;|class\s+[A-Za-z0-9_]+\s+extends\s+[A-Za-z0-9_]+)'
+    pattern: 'package\s+[\w\.]*\s*[{\n]{1}'
```
smola, line 51: Are these valid only at the beginning of a string? If that's the case, they should be changed to preserve the previous behaviour, which is more restrictive. For example:
`^\s*package\s+[\w\.]*\s*[{\n]{1}`. This will prevent matching in the middle of statements, within quoted strings, comments, etc.

Ryan-rsm-McKenzie: I added anchoring to all the expressions now, so that should avoid matching comments

smola:
@Ryan-rsm-McKenzie We need some new files in the `samples` directory that failed to classify before and work now, for both ActionScript and AngelScript. They need to use a liberal license, so one from https://github.com/AS3Commons/as3commons-ui and one from https://github.com/Sami-Vuorela/aatc. Other than that, looks good, although it still needs to be reviewed by a maintainer.

Ryan-rsm-McKenzie:
I had a sample that demonstrated AS2 package syntax which requires that the package name be equivalent to its folder hierarchy, but it would appear the the test cases don't support sub directories.

Ryan-rsm-McKenzie:
It's probably failing because it needs the multi-line flag, but I don't know enough about ruby to know how to set it

Ryan-rsm-McKenzie:
Yeah I'm going to need some help on that one

Ryan-rsm-McKenzie:
That fixed it

Ryan-rsm-McKenzie:
I rebased and squashed

pchaigno (dismissed):
Thanks a lot for doing this work!

The sample files are also used to train the Bayesian classifier so they need to be from "real" application rather than just "hello world files". Apart from that, it looks good to me!

Ryan-rsm-McKenzie:
Both of these samples are from [CASA Lib](https://github.com/aaronclinger/casalib).
None of the Angelscript files were misclassified as it doesn't actually have a heuristic and is instead the fallback when the Actionscript heuristic fails.

samples/ActionScript/Functional.as
```diff
@@ -0,0 +1,5 @@
+// AS3: Unique (from Angelscript) function declaration syntax using the function keyword
+
+package{
+	function foo();
+}
```
pchaigno, line 5: Could you remove those files as well? If you think they are needed for tests because they match some of the new patterns the larger files don't match, you can move them to `test/fixtures/`.

Ryan-rsm-McKenzie:
There was a case-sensitivity issue with the file name before I rebased

lildude:
This looks good to me. Any other thoughts @pchaigno? @Alhadis these regexes look good to you?

Alhadis:
> @Alhadis these regexes look good to you?

Ehhh... they're excessive and riddled with redundant syntax, but I wouldn't consider them a show-stopper. We can certainly optimise the expressions to be a bit more... well, optimal.

lildude:
> Ehhh... they're excessive and riddled with redundant syntax, but I wouldn't consider them a show-stopper. We can certainly optimise the expressions to be a bit more... well, optimal.

We may as well get em ✨ and 💐  now in this PR.

Alhadis:
> We may as well get em ✨ and 💐 now in this PR.

We get them sparkly and bouquet now in this PR?

I swear I'm getting too old for the Internet…

lildude:
> We get them sparkly and bouquet now in this PR?

Pretty and looking good 😬  I'm too old too, just trying to be 🆒 with all the youngsters 🤣

Alhadis:
This is all a prelude for the real horror awaiting hackers in their old age: angry old men whose opinion rants burn hotter and hotter the older they get, until they collapse in on themselves like a dying star and explode in a supernova of unintended stardom:

rms

Ryan-rsm-McKenzie:
Do you have anything concrete you want to say about this or are you just going to ignore it until it goes away?

lib/linguist/heuristics.yml
```diff
@@ -48,7 +48,13 @@ disambiguations:
...
   rules:
   - language: ActionScript
-    pattern: '^\s*(package\s+[a-z0-9_\.]+|import\s+[a-zA-Z0-9_\.]+;|class\s+[A-Za-z0-9_]+\s+extends\s+[A-Za-z0-9_]+)'
+    pattern: 
+    - '^\s*package(?:\s+[\w\.]+)?\s+[{\n]{1}'
+    - '^\s*import\s+[\w\.\*]+\s*;'
+    - '^\s*intrinsic\s+class\s+[\w<>\.]+'
+    - '^\s*class\s+[\w<>\.]+\s+extends\s+[\w<>\.]+'
+    - '^\s*(?:(?:public|protected|private|static|var|const|local)\s+)+\w+\s*:\s*[\w<>\.]+(?:\s*=.*)?\s*;'
+    - '^\s*(?:(?:public|protected|private|static)\s+)*function\s+[\w]+\s*\((?:\s*\w+(?:\s*:\s*[\w<>\.]+)?,?)*\)(?:\s*:\s*[\w<>\.]+)?'
```
Alhadis, line 57: Here's a more optimal expression...

~~~yaml
pattern: '^\s*(?:package(?:\s+[\w.]+)?\s+(?:{|$)|import\s+[\w.*]+\s*;|(intrinsic\s+)?class\s+[\w<>.]+(?:\s+extends\s+[\w<>.]+)?|(?:(?:public|protected|private|static)\s+)?(?:(?:var|const|local)\s+\w+\s*:\s*[\w<>.]+(?:\s*=.*)?\s*;|function\s+\w+\s*\((?:\s*\w+(?:\s*:\s*[\w<>.]+)?,?)*\)(?:\s*:\s*[\w<>.]+)?))'
~~~

... and the readable version of it is:

~~~regexp
^\s*
(?:
	package (?:\s+[\w.]+)? \s+ (?:{|$)
	|
	import \s+ [\w.*]+ \s* ;
	|
	(intrinsic\s+)? class \s+ [\w<>.]+ (?:\s+extends\s+[\w<>.]+)?
	|
	(?:(?:public|protected|private|static)\s+)?
	(?:
		# Variable
		(?:var|const|local) \s+ \w+ \s*        # var foo
		: \s* [\w<>.]+                         # : Type
		(?:\s*=.*)? \s* ;                      # = ... ;
		
		|
		
		# Function
		function \s+ \w+ \s*                   # function fooBar
		\((?:\s*\w+(?:\s*:\s*[\w<>.]+)?,?)*\)  # (arg1, arg2)
		(?: \s* : \s* [\w<>.]+ )?              # :<returnValue>
	)
)
~~~

Alhadis (changes requested):
Hey @Ryan-rsm-McKenzie, I'm sorry it's taken me this long to get back to you. 👋 I've cleaned up the expressions for you. 👍

Ryan-rsm-McKenzie:
Thank you for taking the time to respond. There were 2 corner cases that the new regex didn't pick up, so I added new tests for them:  
1. Variables in Actionscript can have many keywords attached to them, and the new regex isn't picking up on that.
2. Both Actionscript and Angelscript share the `class` keyword. I try to differentiate between the two by looking for surrounding keywords unique to Actionscript such as `intrinsic` and `extends`. The new regex is flagging all instances of the `class` keyword as being Actionscript.

Alhadis:
> Both ActionScript and AngelScript share the `class` keyword

In that case, it's safer to remove it from the heuristic altogether. I'd also drop the open-ended fallback to AngelScript, so anything that doesn't match the heuristic will fall through to the classifier.

> Variables in ActionScript can have many keywords attached to them

Can they? I realise I've forgotten most of the ActionScript 3.0 I was so fluent in +10 years ago... 😅  (Guess my recollection of it isn't as sharp as I thought 😰). In that case, I'll amend the pattern, hang five…

lib/linguist/heuristics.yml
Alhadis: @Ryan-rsm-McKenzie Righto, I've updated the pattern.

Ryan-rsm-McKenzie:
How do you drop the fallback?

Ryan-rsm-McKenzie:
Also there are no Angelscript samples for the classifier to work with

Alhadis:
Delete this line:

~~~diff
-  - language: AngelScript
~~~

Alhadis:
🤦‍♂️ D'oh. I forgot to remove the `class` keyword… Actually, I thought of something better. I'll push changes to your branch directly, since I feel bad about jerking you around so much…

Alhadis:
Okay, done. Tests are passing locally.

Alhadis (approved):

lildude (approved):

Ryan-rsm-McKenzie:
Thank you for taking the time to review this merge request

PR #4929, opened 2020-07-24
Add extensions for new BuckleScript syntax for Reason

brnrdog:
## Description
This PR adds support for `.res` and ~`.resi`~ (not enough search results for `.resi` extension, so I removed) extensions related to the [new syntax option](https://reasonml.org/blog/bucklescript-8-1-new-syntax) introduced in BuckleScript 8.1 for the Reason programming language.

## Checklist:

      - https://github.com/search?o=desc&q=extension%3Ares+let&s=indexed&type=Code
    - Sample source(s):
      - https://github.com/BuckleScript/syntax/blob/master/benchmarks/data/RedBlackTree.res
    - Sample license(s): MIT
      - https://github.com/BuckleScript/syntax/blob/master/LICENSE

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The language of each remaining file is then determined using the following strat
 The result of this analysis is used to produce the language stats bar which displays the languages percentages for the files in the repository.
 The percentages are calculated based on the bytes of code for each language as reported by the [List Languages](https://developer.github.com/v3/repos/#list-languages) API.
 
-![language stats bar](https://user-images.githubusercontent.com/5887227/91159208-0500af00-e6c8-11ea-810f-9488cd83906e.png)
+![language stats bar](https://user-images.githubusercontent.com/2346707/91533656-9768b300-e953-11ea-808d-994cd50e6273.png)
 
 ### How Linguist works on GitHub.com
 


lib/linguist/languages.yml
```diff
@@ -4575,6 +4575,8 @@ Reason:
   extensions:
   - ".re"
   - ".rei"
+  - ".res"
```
pchaigno, line 4578: @lildude I'm having trouble checking there aren't any other languages using that extension on GitHub, mostly because of the large number of `.res` binary files. Is there a way to exclude binary files in GitHub searches?

lib/linguist/languages.yml
```diff
@@ -4575,6 +4575,8 @@ Reason:
...
+  - ".resi"
```
pchaigno, line 4579: There are only 18 such files, far from the requirement of hundreds of repositories. Please remove it.

lib/linguist/languages.yml
lildude: Unfortunately not, however the search results show quite a few files identified as other languages thanks to the other strategies. Many of those are XML files which means this PR is going to flip them all to Reason because the extension strategy has priority over the XML strategy.

The search result in the OP also returns quite a few files that don't look like Reason to me, though I know nothing about Reason.

lib/linguist/languages.yml
brnrdog: Thanks, I removed the `.resi` extension and its file sample.

lib/linguist/languages.yml
brnrdog: Indeed a few files that are not written in Reason, I don't know how I can filter them out in the search. Also, I think it's worth to say that this new syntax and the new file extensions for Reason/BuckleScript got available at the beginning of July 2020, so I'd say it's in early adoption phase.

ryyppy: Thanks @brnrdog kicking this off, may I chime in here? 
This might sound confusing, but when we introduced the new Syntax for BuckleScript / Reason, we didn't give it a name yet, [until very recently](https://reasonml.org/blog/bucklescript-is-rebranding).

BuckleScript / Reason is rebranding to [ReScript](https://rescript-lang.org/), therefore the `.res` and `.resi` file extension should therefore be called `ReScript` as well.

chenglou:
Thanks @brnrdog. Like Patrick said, this should be labeled as ReScript and distinct from Reason

chenglou:
@brnrdog ah, well in that case, let's just use our fallback TM grammar: https://github.com/rescript-lang/rescript-vscode/blob/af04acbe14aac3b675b73b8b9aea8f2a2da76ed2/rescript.tmLanguage.json

brnrdog:
@chenglou @ryyppy For better clarity, I'm closing this and opening a new PR for the ReScript language: https://github.com/github/linguist/pull/4975

PR #4933, opened 2020-07-26
Add carp language support (extension)

sheerun:
## Description
Add carp language support (extension)

https://github.com/carp-lang/Carp

## Checklist:

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Acarp%20loadq=extension%3AFOOBAR+KEYWORDS+NOT+nothack
    - Sample source(s):
      - https://raw.githubusercontent.com/CaptainLexington/diem/6c9f4150ded5a67548be4061486fffb982c3e9ff/src/resources.carp
    - Sample license(s): wtfpl
    - not applicable

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -602,6 +602,12 @@ C2hs Haskell:
   codemirror_mode: haskell
   codemirror_mime_type: text/x-haskell
   language_id: 45
+Carp:
+  type: programming
+  extensions:
+  - ".carp"
+  tm_scope: none
+  ace_mode: text
 CLIPS:
   type: programming
   extensions:
--- a/samples/Carp/example.carp
+++ b/samples/Carp/example.carp
@@ -1,0 +1,133 @@
+(use IO)
+(use System)
+(use Int)
+(use Double)
+(use Array)
+(use Map)
+
+(load "SDL.carp")
+(load "SDL_image.carp")
+
+(defmodule ResourceManager
+           (deftype ResourceData
+             (None [])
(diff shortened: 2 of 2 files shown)


lildude:
Please follow the instructions on [Adding a language](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language), but before you do, take note of this requirement:

> we prefer that each new file extension be in use in hundreds of repositories before supporting them in Linguist.

A quick search of `.carp` suggests this may not be popular enough yet - you can use [Harvester](https://github.com/Alhadis/Harvester) to perform a more comprehensive search. Unless you expect the popularity to grow rapidly in the next month or so, I'd suggest you close this PR and open a new one, following the instructions, once `.carp` is more popular.

sheerun:
I don't think github search works very well for extensions https://github.com/search?q=extension%3Acarp&type=Code returns zero results but 100% there tenths if not hundreds of repositories with .carp

lildude:
> I don't think github search works very well for extensions https://github.com/search?q=extension%3Acarp&type=Code returns zero results but 100% there tenths if not hundreds of repositories with .carp

That's because you're not using the correct syntax. From the [docs](https://docs.github.com/en/github/searching-for-information-on-github/searching-code):

> Except with filename searches, you must always include at least one search term when searching source code. For example, searching for language:javascript is not valid, while amazing language:javascript is.

You've not included a search term. This is why we use `NOT nothack` in all our search queries.

sheerun:
Okay I think you are right, sorry

PR #4936, opened 2020-07-29
Add support for generic file extensions

Alhadis:
## Description
This pull-request adds non-trivial enhancements to Linguist's `Heuristic` and `Extension` strategies:

1.	__Overly-common file extensions can now be identified as such.__  
	Generic extensions require files to satisfy a heuristic in order to match a candidate language.
2.	__Heuristics can no longer target extensions not listed in a candidate's [`languages.yml`](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) entry__.  
	Linguist's tests will now fail if such a rule is found in [`heuristics.yml`](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml) (see [#4904](https://github.com/github/linguist/pull/4904#issuecomment-665528522) for rationale).
3. __Two redundant heuristics were modified to comply with point \#2:__
	* __Open-ended fallback to `Text` removed from `.builds` heuristic.__  
	It's assumed the author intended for it to return nothing.
	* __`INI` removed from `.props` heuristic.__  
	Only [32 files][1] are affected by this change, so it shouldn't have been added in the first place.
4.	__`.sol` has been added as a generic extension.__  
	Affected languages are `Gerber Images` and `Solidity`. Support for the latter was the original motivator for this pull-request; see [#4904](https://github.com/github/linguist/pull/4904#pullrequestreview-455316253) for prior discussion.

[1]: https://github.com/search?l=INI&p=1&q=extension%3Aprops+NOT+nothack&type=Code

/cc @alcuadrado, @pchaigno


Alhadis:
> ~~~
>   1) Failure:
> TestFileBlob#test_language [/home/runner/work/linguist/linguist/test/test_file_blob.rb:671]:
> No language for /home/runner/work/linguist/linguist/test/fixtures/Solidity/ignored2.sol
>
>  2) Failure:
> TestBlob#test_language [/home/runner/work/linguist/linguist/test/test_blob.rb:276]:
> No language for /home/runner/work/linguist/linguist/test/fixtures/Solidity/ignored2.sol
> ~~~

I'd like to know why these tests *didn't* fail when I tested my changes locally...

lib/linguist/language.rb
```diff
@@ -150,25 +150,31 @@ def self.find_by_filename(filename)
...
-    def self.find_by_extension(filename)
-      # find the first extension with language definitions
-      extname = FileBlob.new(filename.downcase).extensions.detect do |e|
-        !@extension_index[e].empty?
+    def self.find_by_extension(query)
+      unless query.is_a? Array
+        query = FileBlob.new(query.downcase).extensions
       end
 
+      extname = query.detect {|e| !@extension_index[e].empty? }
```
pchaigno: As far as I understand, this is only going to return the first match in case an array is given. Could you add an example of that (array with files/extensions that correspond to different languages) in the comment above and in the tests below?

lib/linguist/heuristics.rb
```diff
@@ -89,6 +97,18 @@ def initialize(exts_and_langs, rules)
       @rules = rules
     end
 
+    # Internal: Return the heuristic's target extensions
+    def extensions
+      @exts_and_langs
```
pchaigno, line 102: Not introduced by this PR, but shouldn't `exts_and_langs` be renamed to `exts` or `extensions`?

lib/linguist/heuristics.rb
```diff
@@ -89,6 +97,18 @@ def initialize(exts_and_langs, rules)
...
+    end
+
+    # Internal: Return the heuristic's candidate languages
+    def languages
+      @rules.map do |rule|
+        [rule['language']].flatten(2).map { |name| Language[name] }
+      end.flatten.uniq
```
pchaigno, line 109: Why are several `flatten` required here?

lib/linguist/strategy/extension.rb
```diff
@@ -1,5 +1,10 @@
 module Linguist
   module Strategy
+    # Extensions considered too common for simple extension lookup.
+    # Languages which use such an extension must instead satisfy a
+    # heuristic (or some other strategy) to be matched on extension.
+    GENERIC = %w[.sol]
```
pchaigno, line 6: This list should be in `languages.yml` instead.

test/test_heuristics.rb
```diff
@@ -494,6 +494,13 @@ def test_sc_supercollider_scala_by_heuristics
     })
   end
 
+  def test_sol_by_heuristics
+    assert_heuristics({
+      "Solidity" => Dir.glob("#{fixtures_path}/Solidity/{legacy,modern}*"),
+      nil        => Dir.glob("#{fixtures_path}/Solidity/ignored*")
```
pchaigno, line 500: If those files are not Solidity, then they should not be in the Solidity directory.

lib/linguist/language.rb
```diff
@@ -150,31 +150,25 @@ def self.find_by_filename(filename)
 
     # Public: Look up Languages by file extension.
     #
-    # query - Either a filename String, or an Array of extension strings
```
pchaigno: Can you squash the a5f66324eb1b85021bcb57618d5a143150774bc6 commit and it's revert?

test/test_blob.rb
```diff
@@ -272,6 +272,8 @@ def test_language
           assert blob.language.nil?, "A language was found for #{filepath}"
         elsif language == 'Generated'
           assert blob.generated?, "#{filepath} is not a generated file"
+        elsif language == 'Solidity' && filename.match(/ignored\d\.sol$/)
+          assert !blob.language, "#{filepath} should not match a language"
```
pchaigno, line 276: I'm assuming these special cases won't be needed once these files are moved out of the Solidity directory.

test/test_file_blob.rb
```diff
@@ -667,6 +667,8 @@ def test_language
           assert blob.language.nil?, "A language was found for #{filepath}"
         elsif language == 'Generated'
           assert blob.generated?, "#{filepath} is not a generated file"
+        elsif language == 'Solidity' && filename.match(/ignored\d\.sol$/)
+          assert !blob.language, "#{filepath} should not match a language"
```
pchaigno, line 671: Ditto.

lib/linguist/heuristics.rb
Alhadis: It should, and I found that name particularly misleading myself. I'll change it.

lib/linguist/heuristics.rb
Alhadis: Because [some](https://github.com/github/linguist/blob/96ad1185828f44bb9b774328a584551ee57ed264/lib/linguist/heuristics.yml#L292) heuristics return an array of candidates, so a second squash is needed to stamp out entries like `[Linux Kernel Module, AMPL]`.

lib/linguist/strategy/extension.rb
Alhadis: Why? I fail to see how this property is semantically different from being `vendored`, save the fact we have enough [vendored patterns](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml) to warrant its own file. Since we only have one extension marked as generic, having its own file would be overkill.

test/test_heuristics.rb
Alhadis: The `fixtures_path` only contains files used by Linguist's test suite. They're ignored by the classifier (and the rest of Linguists runtime, for that matter).

lib/linguist/language.rb
Alhadis: Why? Everything's going to get squash-merged anyway…

test/test_blob.rb
Alhadis: [See above](https://github.com/github/linguist/pull/4936#discussion_r462903775). These files aren't in the `samples/Solidity` directory, only in the `test/fixtures/Solidity` directory (and if that's not a confusing enough distinction for you, our [`test_heuristics.rb`](https://github.com/github/linguist/blob/96ad1185828f44bb9b774328a584551ee57ed264/test/test_heuristics.rb#L20) file refers to the actual "sample" files as "fixtures"...

test/test_file_blob.rb
Alhadis: Ditto ditto.

Alhadis:
> I'd also like to better understand where we use the `samples/` files, apart from end-to-end tests and training of the Bayesian classifier. Those two are fine because we use fixture files for end-to-end tests and we don't use the Bayesian classifier. If we usually rely on the `samples/` files for other things, we might have problems...

We currently don't *have* any Solidity "samples", only test "fixtures" (our nomenclature really needs improvement…). Adding `.sol` files to other sample directories would actually break the tests, as there's a check to make sure only extensions listed in `languages.yml` appear in the `samples/*` directories.

That being said, we can certainly add more real-world examples to the `test/fixtures/Solidity` directory, but not as `samples/`.

lib/linguist/strategy/extension.rb
pchaigno: Given this effectively implements a not-a-language option for the heuristic strategy only, I doubt we'll have to wait long before we see new languages added here. I'm not suggesting a new file, but simply a new entry in `languages.yml`. I don't want this list to be buried where we will forget it. It's also easier to actually use the list (e.g., for additional tests) if it's in `languages.yml`.

test/test_heuristics.rb
pchaigno: Sure, but it doesn't change the fact they don't belong in `fixtures/Solidity`. I think we have `Data/` for fixture files we're unable to classify. I'd also like to see examples of other languages using `.sol`.

lib/linguist/language.rb
pchaigno: So as to not waste the time of reviewers who review commits one by one. I'd also really like if @lildude did not squash all commits of this PR. It makes sense to squash for simple PRs that only add an extension or a language, but not for larger PR such as this one where the different commits can be useful when we git blame later or if we need to revert something.

pchaigno:
> > I'd also like to better understand where we use the `samples/` files, apart from end-to-end tests and training of the Bayesian classifier. Those two are fine because we use fixture files for end-to-end tests and we don't use the Bayesian classifier. If we usually rely on the `samples/` files for other things, we might have problems...
> 
> We currently don't _have_ any Solidity "samples", only test "fixtures" (our nomenclature really needs improvement…). Adding `.sol` files to other sample directories would actually break the tests, as there's a check to make sure only extensions listed in `languages.yml` appear in the `samples/*` directories.

I understand that. My point was that we don't and can't have Solidity files in `samples/` and I'm therefore wondering if that can have a negative impact. I.e., are there tests (or, less likely, features) we will be missing because of the lack of `samples/` Solidity files.

> That being said, we can certainly add more real-world examples to the `test/fixtures/Solidity` directory, but not as `samples/`.

Yes, that's my point 2 above.

lib/linguist/strategy/extension.rb
Alhadis: > but simply a new entry in `languages.yml`

And where do you propose I put it? The top-level entry of the file is an array of language objects…

lib/linguist/language.rb
Alhadis: Following that logic, it would make sense to *keep* a5f6632 in case we have need of it later…

test/test_heuristics.rb
Alhadis: I'd prefer we keep all Solidity-related fixtures somewhere obvious. I'll split `fixtures/Solidity` into two directories:

~~~
./test/fixtures/Solidity/match/
./test/fixtures/Solidity/ignore/
~~~

Alhadis:
I've also added `.sol` as a Gerber Image extension. In-the-wild usage checks out:

* **Files:** [2,389](https://github.com/Alhadis/Silos/blob/sol/urls.log) out of [3,108 total](https://github.com/search?q=extension%3Asol+G75)
* **Repos:** [1,047](https://github.com/Alhadis/Silos/blob/sol/repos.log)
* **Users:** [884](https://github.com/Alhadis/Silos/blob/sol/users.log)

Plus, adding this proves that more than one language can share a generic extension.

lib/linguist/language.rb
pchaigno: Following that latest logic, Linguist would be full of unused code :-p

I don't see the point of adding code we're not using. It's only likely to add complexity and bugs.

lib/linguist/strategy/extension.rb
pchaigno: I think it should be attached to specific languages and we should have a test to check the heuristic rules only pick among those languages.

Alhadis: Can you provide a code sample of what you're talking about? I'm still confused.

lib/linguist/language.rb
Alhadis: Then perhaps we should keep this simple and *not* deviate from our usual procedure of squash-merging pull-requests. If parts of this need to be reverted later (which is already unlikely), the commits made in this PR can be used as reference.

I don't understand why you're making this more complicated than it needs to be...

lib/linguist/strategy/extension.rb
pchaigno: Sure. Something like:
```yaml
Solidity:
  type: programming
  color: "#AA6746"
  extensions:
  - .sol
  generic_extensions:
  - .sol
  ace_mode: text
  tm_scope: source.solidity
  language_id: 237469032
```

Alhadis: Not only is this repeating information we already have, it's visibly confusing to users as to which extensions-list requires updating. Moreover, we need to do the same for *every* language that uses a generic extension — it's more work for users, has an increased risk of complications, and at the end of the day, this approach has **zero benefits** over simply hardcoding the list somewhere else.

Alhadis:
@pchaigno Think we'd be able to include what #4258 attempted to fix? Though we already have a dedicated `Manpage` strategy, it makes more sense to mark numeric file-extensions as "generic" instead.

Alhadis:
@lildude Any input regarding @pchaigno's suggested change? To me, it makes more sense to have generic extensions added to a YAML database named `generic.yml` (a la vendored filenames and extensions). What do you think?

lildude:
Sorry @Alhadis. Things are crazy busy for me at the moment as we're reading the next release of GitHub Enterprise. I'll try take a look when I've got a chance when things quieten down a bit.

Alhadis:
I understand.

@pchaigno What if I added a `generic.yml` file and modified `strategy/extension.rb` to use that instead of an array hardcoded into the strategy? I can always extend the scope of this PR to include other generic file extensions:

~~~yml
# generic.yml
# (Some concise but informative description of what this file does and how to update it)

extensions:
- ".1"
- ".2"
- ".3"
- ".4"
- ".5"
- ".6"
- ".7"
- ".8"
- ".9"
- ".in"
- ".mod"
- ".sol"

filenames:
- unoriginal_but_common_filename
~~~

lib/linguist/language.rb
Alhadis: I think we can meet halfway with a new file. See https://github.com/github/linguist/pull/4936#issuecomment-675875267

alcuadrado:
Hey @lildude, do you have a rough estimation about when you'll be able to review this? Not to pressure you, but just asking so we can better coordinate with @Alhadis. Thanks!

lildude:
I'm looking at it right now. It might take me a few days to process it and respond, but I've got time this week for Linguist stuff.

lib/linguist/strategy/extension.rb
```diff
@@ -15,9 +20,15 @@ class Extension
       # Selected languages must be in the candidate list, except if it's empty,
       # in which case any language is a valid candidate.
       def self.call(blob, candidates)
+        return candidates if generic? blob.name
```
lildude, line 23: We probably want to ensure these are strings, especially if you plan to add numbered extensions in future:

```suggestion
        return candidates if generic? blob.name.to_s
```

lib/linguist/strategy/extension.rb
lildude: I think we're OK keeping this list here for now. We can explore pulling it out to its own file if we see the demand for it. Excluding the numbered files @Alhadis has mentioned, I know `.k` will likely be a candidate for this list.

lildude (commented):
Some initial thoughts from a quick review, but I don't have any major objections from my initial quick review.

Additionally, do we really need quite so many fixtures as they're only used for testing the heuristics and aren't used for generating the `samples.json` used by the classifier?

I'll ponder this a bit more this week.

alcuadrado:
Thanks for the quick answer, @lildude. This is great news :)

Alhadis:
> Excluding the numbered files @Alhadis has mentioned,

I'll leave those for a future PR. At the moment, I'm just anxious to get this under the door in time for the next Linguist release.

lildude:
I've pondered this for the week, and I comfortable with the approach taken here. 

This approach means we lose any further assistance from the XML and classifier strategies but I think this is OK as this we already expect the heuristics to be pretty specific to reduce the chances of falling through to the classifier. I think this will also be a rarely used "exception" (famous last words 😉) so I'm happy to experiment and iterate as needed.

My only request for this PR is to remove the unneeded examples from the test fixtures. I think we only need fixtures to confirm each of the heuristic scenarios, eg we only need one for "ignored" and Gerber and a few to match the various combos of the Solidity regex.

Alhadis:
> My only request for this PR is to remove the unneeded examples from the test fixtures. I think we only need fixtures to confirm each of the heuristic scenarios, eg we only need one for "ignored" and Gerber and a few to match the various combos of the Solidity regex.

That was my line of thinking as well, but @pchaigno [reque](https://github.com/github/linguist/pull/4936#discussion_r463462211)[sted](https://github.com/github/linguist/pull/4936#issuecomment-666992058) the extra samples for some reason, so… 🤷 Getting some mixed signals here.

Will nuke 'em, hang 5.

lib/linguist/strategy/extension.rb
pchaigno: Sure, let's keep here until we need it in a separate file. I doubt it will be long before we need to refer to this list in specific tests though...

Alhadis:
Alright, done. I'm guessing there's nothing left to do here...?

test/test_file_blob.rb
```diff
@@ -667,6 +667,8 @@ def test_language
           assert blob.language.nil?, "A language was found for #{filepath}"
         elsif language == 'Generated'
           assert blob.generated?, "#{filepath} is not a generated file"
+        elsif language == 'Generic'
+          assert !blob.language, "#{filepath} should not match a language"
```
pchaigno, line 671: What's the difference between `Generic` and `Data`? Couldn't these files be moved to the `Data/` directory?

test/test_heuristics.rb
```diff
@@ -61,6 +61,18 @@ def test_detect_still_works_if_nothing_matches
...
+    Heuristics.all.all? do |rule|
+      rule.languages.each do |lang|
+        unlisted = rule.extensions.reject { |ext| lang.extensions.include? ext }
+        assert_equal [], unlisted, (<<~EOF).chomp
+          The extension '#{unlisted.first}' is not assigned to #{lang.name}.
+          Add it to `languages.yml` or update the heuristic which uses it
+        EOF
+      end
+    end
+  end
```
pchaigno, line 74: :+1:

test/test_heuristics.rb
```diff
@@ -100,7 +112,7 @@ def test_bb_by_heuristics
 
   def test_builds_by_heuristics
     assert_heuristics({
-      "Text" => all_fixtures("Text"),
+      nil => all_fixtures("Text"),
```
pchaigno, line 115: Not sure I understand this change. Why did it change?

test/test_samples.rb
```diff
@@ -93,8 +93,9 @@ def test_no_text_samples
 
         # Check for samples if more than one language matches the given extension.
         if language_matches.length > 1
+          samples_dir = Strategy::Extension.generic?(extension) ? "test/fixtures" : "samples"
           language_matches.each do |match|
-            samples = "samples/#{match.name}/*#{extension}"
+            samples = "#{samples_dir}/#{match.name}/*#{extension}"
```
pchaigno, line 98: Nice! :+1:

lib/linguist/heuristics.yml
```diff
@@ -395,12 +394,6 @@ disambiguations:
     named_pattern: key_equals_value
   - language: Java properties
     pattern: '^[^#!][^:]*:'
-- extensions: ['.props']
-  rules:
-  - language: XML
-    pattern: '^(\s*)(?i:<Project|<Import|<Property|<\?xml|xmlns)'
-  - language: INI
-    pattern: '(?i:\w+\s*=\s*)'
```
pchaigno, line 403: Why were these removed (`Text` in `.builds` and all of `.props`)? How is this related to Solidity?

pchaigno (changes requested):
> > My only request for this PR is to remove the unneeded examples from the test fixtures. I think we only need fixtures to confirm each of the heuristic scenarios, eg we only need one for "ignored" and Gerber and a few to match the various combos of the Solidity regex.
> 
> That was my line of thinking as well, but @pchaigno [reque](https://github.com/github/linguist/pull/4936#discussion_r463462211)[sted](https://github.com/github/linguist/pull/4936#issuecomment-666992058) the extra samples for some reason, so… shrug Getting some mixed signals here.

I never requested so much :stuck_out_tongue: (though I tend to be on the side of never too many tests)
I think it's important to cover:
- Different Solidity usages that match the regular expresion we're using.
- Same for Gerber Image.
- A few examples of common `.sol` file patterns.
That last item is pretty important since, for the first time ever, we're going to apply a heuristic rule to an unknown set of languages. We usually know for what candidate languages our heuristic are built. So, I think it's important to properly test that. If you think the current version achieves that, then I'm :+1: 

---

Two concerns below. First, it looks we could reuse `test/fixtures/Data/` instead of a new `test/fixtures/Generic/`.

Second, some of the changes appear to be unrelated to Solidity (or I'm missing some context :-) ). That may be fine (although at the moment I don't understand why these changes have even been applied), but if we keep such unrelated changes in this Solidity PR, can be please please please not squash all commits (:pray: @lildude). We're going to lose all context if we do and we'll have to read the whole PR every time we want to `git blame` something...

lib/linguist/heuristics.yml
Alhadis: These were actually heuristics that didn't have a file extension listed in `languages.yml`. So in order to implement the generic-extension feature correctly, I had to remove a few things.

test/test_heuristics.rb
Alhadis: [See below](https://github.com/github/linguist/pull/4936#discussion_r479645840). Classifying `builds` files as `Text` was presumably somebody's attempt at marking them as having no language. Now we have a more appropriate means to do that.

test/test_file_blob.rb
Alhadis: There's no logical order to the contents of the `test/fixtures/*` subdirectories. In fact, our entire test-suite is a mess of badly-named-everything. Even worse, some of the helper methods **require** the current structure, so, could we just leave this mess as-is…?

Alhadis:
> * A few examples of common `.sol` file patterns.
>
> That last item is pretty important since, for the first time ever, we're going to apply a heuristic rule to an unknown set of languages.

Selected from... what? Randomly-selected samples I blindly plucked from search results, which @lildude just told me to remove? Or the patterns being matched by the relevant heuristics? For the latter case, that's already taken care of. 😕

test/test_file_blob.rb
pchaigno: I'm not sure I understand to what lack of logical order you are referring. Could you give some examples? I was under the impression files were organized in `test/fixtures/[Language]` directories with exceptions for files of unknown/unclassified languages and binary files.

lib/linguist/heuristics.yml
pchaigno: Ah, I see. Makes sense :slightly_smiling_face: 

My main concern is that we will lose the context on this change with the squash (future contributors will have a hard time finding this discussion buried in the PR). Could you maybe send it as a separate PR and I'll be happy to validate it before end of day?

Alhadis: > future contributors will have a hard time finding this discussion buried in the PR

That reminds me, I should rename this PR and clarify that the generic-extension feature is the primary change, with Solidity support being the impetus.

pchaigno:
> > * A few examples of common `.sol` file patterns.
> > 
> > That last item is pretty important since, for the first time ever, we're going to apply a heuristic rule to an unknown set of languages.
> 
> Selected from... what? Randomly-selected samples I blindly plucked from search results, which @lildude just told me to remove? Or the patterns being matched by the relevant heuristics? For the latter case, that's already taken care of. confused

That's a good point :smile: 

I'm not sure why @lildude wanted to see these examples removed, but I think it's fine. I spent some time looking for recurring file patterns in `extension:sol NOT contract` and could find anything that would stand out. Hopefully, if we get too many misclassifications, users will come forward with better examples we can use to improve the heuristic and test it.

test/test_file_blob.rb
Alhadis: `test/fixtures/Data` also contains obvious text-based formats that don't fit anywhere else.

I think we should add a `Generic` subdirectory, though, organised by the actual extensions being tested:

~~~
./Generic
./Generic/sol
./Generic/sol/ignored.sol
./Generic/sol/Gerber.sol
./Generic/sol/…
~~~

This would require the assertions to be written out manually, but at least it's obvious where future fixtures need to be added…

pchaigno: Ah, yes. That would give some structure to this new directory and then it makes sense to keep it separate from `Data` :+1:

Alhadis:
This heuristic was taken from #4904, submitted by an authoritative source (I cleaned the expression up to avoid matching what it shouldn't). Apparently the `pragma` directive is required by all modern Solidity contracts, and only older files have a chance of not matching. I've done what I could to accommodate those too, within reason.

lib/linguist/heuristics.yml
Alhadis: Alright, done. @pchaigno, anything else?

lildude:
> I'm not sure why @lildude wanted to see these examples removed

Only cos they weren't really testing the heuristic so I didn't see much need for them.

> can be please please please not squash all commits (🙏 @lildude).

Only if you say please once more 😜

Alhadis:
> It'll mean I have to do a little more work curating the release notes, but that's fine in this case.

@pchaigno requested this before I renamed and rewrote the PR description to place more emphasis on the generic-extension feature, rather than the reasons for adding it (Solidity). I feel what I've written nicely encapsulates everything that's been changed here… I'm still waiting on @pchaigno to give his approval.

pchaigno:
In general not a big fan of squashing non-trivial PRs, but should be okay here with the new description.

pchaigno (approved):

lildude (approved):
🎉  Lets do this!! I'll make a release next week once all the other PRs I've approved have been checked off by you two too.

Alhadis:
Awesome. 🎉 Linguist finally has Solidity support.

Now all that's left to restore balance to the universe is getting rid of that f\*(&#g colour proximity requirement. 😜

lildude:
Oh. You merged it already. I just allowed merge commits and was going to then merge. Oh well.

Alhadis:
Uh, sorry, I mistook the sudden silence as _"go ahead, you merge it"_. Sorry. 😅

alcuadrado:
Great news!

> 🎉  Lets do this!! I'll make a release next week once all the other PRs I've approved have been checked off by you two too.

Does this mean that in about a week it will be live on github?

Alhadis:
> Does this mean that in about a week it will be live on GitHub?

You bet your sweet arse it does.

lildude:
> Does this mean that in about a week it will be live on github?

Indeed. There are a few more PRs waiting reviews from @pchaigno and @Alhadis that I'd like to include but once approved, I'll get cracking with the release. I'm AFK tomorrow so won't start until Tuesday at the earliest.

alcuadrado:
Awesome! 🥳🥳🥳

alcuadrado:
Will this have any other impact apart from automatically associating the .sol extension to Solidity? I wonder if Linguist is used for other things at GitHub

Alhadis:
No, it won't. I was careful to ensure the existing filename logic wasn't affected.

As for what else Linguist is used for… I'm not staff, so I can't tell you, but I *do* know that Linguist is used by GitLab, and possibly other VCS-hosting services too.

alcuadrado:
Thanks for your reply, @Alhadis

pchaigno:
Yep, GitLab is the only other consumer of Linguist I'm aware of, but they tend to lack a bit behind in terms of releases. So it may take some time before the change is visible on gitlab.com and related software.

PR #4947, opened 2020-08-18
Remove duplicated entry (rhtml)

sheerun:
This PR removes duplicated entry in language files (rhtml vs html+erb)


Alhadis:
First of all, could you provide a bit of detail describing the language(s) involved? Just a brief summary explaining what RHTML is will aide those of us unfamiliar with the format. 😉  (It also improves a PR's visibility in search results, since there's more text to match against a prospective keyword).

Secondly—and this relates to the build errors you're getting—removing a language involves a few other steps. They haven't been documented anywhere, because we generally don't remove languages unless it's for a damn good reason (reasons which you'll also need to provide in the PR description).

## To remove a language from Linguist:

1.	**Delete the language's entry from [`languages.yml`](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml).**
2.	**Delete the `./samples/{LANGUAGE_NAME}/` directory.**  
	If necessary, merge its contents into the samples directory of a more appropriate language.
3.	**Update any [heuristics](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml) that reference the language.**
	Delete any heuristics where the removed language is the only candidate (indicated by entries containing a `rules` list with only one entry:
	~~~yml
	- extensions: ['.ext']
	  rules:
	  - language: Name of Language Being Removed
	    pattern: '…'
	~~~
4.	**Update any [tests](https://github.com/github/linguist/tree/master/test) that reference the language.**  
	Usually, this will include [`test_samples.rb`](https://github.com/github/linguist/blob/58193319b3a1daa8fae0fec897029137a958568e/test/test_samples.rb).
5.	**Unregister any grammar module used to highlight the language.**  
	I'll omit the details here, since they're convoluted and difficult to explain, and the topic is irrelevant to this particular removal.
6.	**Regenerate the [grammar index](https://github.com/github/linguist/blob/d6553e41eeee6254296e74b0ed427d07f2253a5c/vendor/README.md) by running [`script/list-grammars`](https://github.com/github/linguist/blob/master/script/list-grammars).**  

/cc @lildude, who's best-equipped to handle Ruby-related matters.

lib/linguist/languages.yml
```diff
@@ -2066,6 +2066,7 @@ HTML+ERB:
   extensions:
   - ".erb"
   - ".erb.deface"
+  - ".rhtml"
```
Alhadis, line 2069: I recommend adding `rhtml` to `HTML+ERB`'s aliases list:

~~~diff
 HTML+ERB:
   type: markup
   tm_scope: text.html.erb
   group: HTML
   aliases:
   - erb
+  - rhtml
~~~

sheerun:
@Alhadis ready

Alhadis:
@lildude I'm leaving this ball in your court, as you're probably far more familiar with embedded Ruby (or anything Ruby-related, for that matter). Ergo, don't mistake my approval as a sign to merge this: I'm merely acknowledging that an ancillary change I requested was actioned. 😉

lildude:
> @lildude I'm leaving this ball in your court, as you're probably far more familiar with embedded Ruby (or anything Ruby-related, for that matter). Ergo, don't mistake my approval as a sign to merge this: I'm merely acknowledging that an ancillary change I requested was actioned. 😉

NP. It means I've got manual work to do when I deploy this to prod to remove the old name. It's the same thing I do when we rename a language, so nothing unusual.

lildude (approved):

lildude:
@sheerun I've just noticed we don't have any `.rhtml` sample files to train the classifier. Can you find a real-world sample with a liberal license and add it to `samples/HTML+ERB/` please. 🙇

sheerun:
@lildude I've added https://github.com/leej35/Billing-Module/blob/6ec6abc9553d8a9e0797ff0735a83e0b97b12a23/app/views/report/disaggregated_diagnosis.rhtml which is licensed under Mozilla Public License as in the readme: https://github.com/leej35/Billing-Module

lildude:
Lovely. Thanks @sheerun 🙇‍♂️

PR #4935, opened 2020-07-29
Add Clarity smart contract programming language

jdbender66:
## Description
Add Blockstack's smart contract programming language for the Stacks 2.0 blockchain, Clarity 

## Checklist:
- [X] **I am adding a new language.**
  - [X] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

PR #4948, opened 2020-08-18
Create non-fungible-token.clar

jdbender66:
A .clar example file for use with Linguist


Alhadis:
Could you provide a link to where you originally acquired the sample?

lildude:
Was this meant to be a continuation of https://github.com/github/linguist/pull/4935 which is still open?

PR #4957, opened 2020-08-25
Add .coffee.md to Literate Coffeescript

sheerun:
## Checklist:

It seems github search is broken for multiple-part extensions like .coffee.md as it doesn't return any results (even the provided sample): https://github.com/search?q=extension%3Acoffee.md+require&type=Code
    - Sample source(s):
https://github.com/STRd6/pixi-test/blob/master/main.coffee.md
    - Sample license(s):
https://github.com/STRd6/pixi-test/blob/master/LICENSE
Does not apply

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2894,6 +2894,7 @@ Literate CoffeeScript:
   - litcoffee
   extensions:
   - ".litcoffee"
+  - ".coffee.md"
   language_id: 206
 Literate Haskell:
   type: programming
--- a/samples/Literate CoffeeScript/pixi.coffee.md
+++ b/samples/Literate CoffeeScript/pixi.coffee.md
@@ -1,0 +1,113 @@
+Pixi Test
+=========
+
+Testing out Pixi.js
+
+    _ = require "./lib/underscore"
+
+    {applyStylesheet} = require "util"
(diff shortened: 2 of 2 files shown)


lildude:
> It seems github search is broken for multiple-part extensions like .coffee.md as it doesn't return any results (even the provided sample): https://github.com/search?q=extension%3Acoffee.md+require&type=Code

It's not perfect, but we can get closer with [`path:coffee extension:md require`](https://github.com/search?q=path%3Acoffee+extension%3Amd+require&type=Code) which in this case only brings up 28 results.

Switching out `require` for `NOT nothack` returns only 900-odd results, many of which are legit markdown files so I don't think we can include this just yet.

sheerun:
Google Bigquery for:

```
SELECT COUNT(*) FROM (
  SELECT repo_name FROM `bigquery-public-data.github_repos.files`
  WHERE ENDS_WITH(path, ".coffee.md")
  GROUP BY repo_name
)
```

returns 1250 repositories (not files), which have such extensions,  for github in 2015. So I guess it qualifies?

lildude:
Ah yes. Of course. I forgot BigQuery could search for cases like that. Yup. Looks like we've got the numbers.

pchaigno (approved):

PR #4963, opened 2020-08-25
Add .j2 extension to Django+HTML

sheerun:
## Description
Add .j2 extension to Django+HTML

## Checklist:

https://github.com/search?p=3&q=extension%3Aj2+div&type=Code
    - Sample source(s):
https://raw.githubusercontent.com/dedfriedly-inspirdev/oauth-client-helper/master/src/oauth_client/templates/home.j2
    - Sample license(s):
https://github.com/dedfriedly-inspirdev/oauth-client-helper/blob/master/LICENSE
Does not apply

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2027,6 +2027,7 @@ HTML+Django:
   group: HTML
   extensions:
   - ".jinja"
+  - ".j2"
   - ".jinja2"
   - ".mustache"
   - ".njk"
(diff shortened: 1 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -2026,6 +2026,7 @@ HTML+Django:
   tm_scope: text.html.django
   group: HTML
   extensions:
+  - ".j2"
```
lildude, line 2029: The first extension in the list is considered the primary and a search of GitHub suggests it's the most popular, but that doesn't necessarily mean it's the primary extension. 

Is `.j2` is definitely the primary extension for "HTML+Django"?

sheerun: didn't know about that, I'll fix it

lildude (approved):
LGTM

PR #4975, opened 2020-08-28
Add ReScript programming language

brnrdog:
## Description

This pull request adds [ReScript programming language](https://rescript-lang.org/). 
ReScript was born from a new syntax introduced in [Reason BuckleScript 8.1](https://rescript-lang.org/blog/bucklescript-8-1-new-syntax).

## Checklist:
    - [.res](https://github.com/search?o=desc&q=extension%3Ares+let&s=indexed&type=Code)
    - ~[.resi](https://github.com/search?q=extension%3Aresi+let&type=Code) (I understand the lack of results for the signature files)~
    - Sample source(s):
      - ~[module.resi](https://github.com/rescript-lang/syntax/blob/95cd66414cd482a52d742650f862f6446895d8e6/tests/printer/signature/module.resi)~
      - [RedBlackTree.res](https://github.com/rescript-lang/syntax/blob/master/benchmarks/data/RedBlackTree.res)
      - ~[ImmutableArray.resi](https://github.com/reason-association/genType/blob/9e4b921d1141f4a1c5bc0ec733678cbf5d60edc0/examples/typescript-react-example/src/ImmutableArray.resi)~
      - [holobloc-sample.res](https://github.com/PRETgroup/goFB/blob/5b6dd39f4e0a01148a88ea381bb6e6b5242cbcd5/examples/goFB_only/c_tcrest/circuitbreaker_mem/fbt/_CBCore0.res)
    - Sample license(s):
      - ~[module.resi](https://github.com/rescript-lang/syntax/blob/master/LICENSE)~
      - [RedBlackTree.res](https://github.com/rescript-lang/syntax/blob/master/LICENSE)
      - ~[ImmutableArray.resi](https://github.com/reason-association/genType/blob/master/LICENSE)~
      - [holobloc-sample.res](https://github.com/PRETgroup/goFB/blob/5b6dd39f4e0a01148a88ea381bb6e6b5242cbcd5/LICENSE)
  
  - [x] I have included a syntax highlighting grammar: [link to github lightshow](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Frescript-lang%2Frescript-vscode%2Fblob%2Fmaster%2Fgrammars%2Frescript.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Frescript-lang%2Fsyntax%2Fea169976897eb6180af64bae861997461fcb68ec%2Fbenchmarks%2Fdata%2FRedBlackTree.res&code=)


brnrdog:
@chenglou @ryyppy Unfortunately the color `#df4b37` is too close to `#e34c26` and `#df4b37`, which are colours already defined, making the test fail. Do you have another colour suggestion?

chenglou:
What about #e6484f? That's the other better red we use

lib/linguist/languages.yml
```diff
@@ -4554,6 +4554,19 @@ Raw token data:
...
   ace_mode: text
   language_id: 318
+ReScript:
+  type: programming
+  color: "#df4b37"
+  ace_mode: rust
+  codemirror_mode: rust
+  codemirror_mime_type: text/x-rustsrc
+  extensions:
+  - ".res"
```
lildude, line 4564: This extension is incredibly popular on GItHub.com. If we were to merge this PR as-is, they will _all_ be identified as ReScript. As such, if you want to add this extension, you will also need to try and identify the language of as many of the other non-ReScript `.res` files as possible, add them to this PR, and implement a heuristic to differentiate between them.

lib/linguist/languages.yml
```diff
@@ -4554,6 +4554,19 @@ Raw token data:
...
+  - ".resi"
```
lildude, line 4565: This extension can't be added yet as there are only 37 such files which is far from the "hundreds or repositories" requirement.

.gitmodules
```diff
@@ -836,6 +836,9 @@
 [submodule "vendor/grammars/razor-plus"]
 	path = vendor/grammars/razor-plus
 	url = https://github.com/austincummings/razor-plus
+[submodule "vendor/grammars/rescript-vscode"]
+	path = vendor/grammars/rescript-vscode
+	url = https://github.com/rescript-lang/rescript-vscode
```
lildude, line 841: This grammar doesn't have a license so can't be added to Linguist until such time as it does, and it would need to be one of these:

https://github.com/github/linguist/blob/c362ab8b3f4f66cbfa7568cf8cb20f7a560de928/vendor/licenses/config.yml#L3-L14

lildude (changes requested):
Woah! You're removing all the license files in this PR 😱. Please put those back. The only license file change we should see is the addition of the license file for the grammar, which I've noted doesn't appear to have a license.

lib/linguist/languages.yml
ryyppy: Just for clarification: to be able to correctly detect the ReScript language, we need to add lots of example files with different ReScript code scenarios and add some regex rules to [heuristics.yml](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml)?

Should we add non ReScript `.res` files in the examples folder as well and make sure those are not detected?

lildude: > Just for clarification: to be able to correctly detect the ReScript language, we need to add lots of example files with different ReScript code scenarios and add some regex rules to [heuristics.yml](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml)?

Not quite. It's not a case of only correctly identifying ReScript, it's about identifying all the other users too.

If we were to flip the scenario around... hypothetically,  if I beat you to it and added a new language called "Lildude's Cool Language" and associated the `.res` with it, every single ReScript file will be identified as "Lildude's Cool Language" because there is no other language associated with that extension. I'm pretty sure the ReScript users wouldn't be too pleased with this, hence we need to add as many of the other users of `.res` to Linguist at the same time so those users don't suddenly have all their files identified as ReScript.

So in other words, this PR will need to be for the addition of ReScript, _and_ "Language1 Using `.res`", _and_ "Language2 Using `.res`" etc.

.gitmodules
brnrdog: The repository was updated with a MIT license.

lib/linguist/languages.yml
brnrdog: I just added an entry for `.res` in the `heuristics.yml` file, including a pattern for `ReScript`:

@ryyppy can you let me know if it's enough to cover rescript files?https://github.com/github/linguist/pull/4975/commits/74da5fd875c9a5729a4cdbc2b46c5c6dd27bffa6

brnrdog:
> Woah! You're removing all the license files in this PR 😱. Please put those back. The only license file change we should see is the addition of the license file for the grammar, which I've noted doesn't appear to have a license.

Whoops sorry! It should be fixed now!

brnrdog:
I've managed to find a good color match for ReScript: `#ed5051`

lib/linguist/languages.yml
ryyppy: @lildude thanks for the clarification, I think I now understand what you mean.. I guess @brnrdog just started the right actions here :)

@brnrdog I think for the `heuristics.yml` we could also add this (basically like in Reason):

```diff
pattern: 
+- '^\s*module\s+type\s'
 - '^\s*(let|module|type).*='
+- '^\s*(?:include|open)\s+\w+\s*;\s*$'
```

ryyppy:
> I've managed to find a good color match for ReScript: `#ed5051`

That color works for us!

lib/linguist/languages.yml
brnrdog: @ryyppy I updated the patterns, I've changed a little bit since ReScript does not require semi-colon for the statements, (correct me if I'm wrong): https://github.com/github/linguist/pull/4975/commits/5f61ff04844ea2c09092ab61194160aa09d16931

ryyppy: ah right, yeah... semicolons are actually optional here, so leaving them out should be fine.

Not sure what else would be missing now. Do we still need to find colliding file types and disambiguate them separately? I looked through global Github search, and many `.res` files are wildly different plain text formats that sometimes resemble XML flavors, plain text or binary files etc... no idea if those would be classified wrongly, now that we have some heuristics?

ryyppy:
@lildude I think we fulfilled all missing requirements, just wondering if there's anything else left that needs to be done to push this PR further?

lildude:
> @lildude I think we fulfilled all missing requirements, just wondering if there's anything else left that needs to be done to push this PR further?

Yes. We still need to identify as many of the other languages that use the `.res` extension as possible and add support for them to this PR. 

As this PR stands _every_ file on GitHub.com with a `.res` extension would be identified as ReScript because this would become the only language with that extension associated with it which means we'll never get to the heuristic.

bloodyowl:
Is there a way we can make a two-step pass on this? First, merge the PR without extension inference (making ReScript known to GitHub and configurable through `.gitattributes`) and then add some inference?

JT0407:
Good

lildude:
> Is there a way we can make a two-step pass on this? First, merge the PR without extension inference (making ReScript known to GitHub and configurable through `.gitattributes`) and then add some inference?

Unfortunately not as that would require changes to the tests to accept a language without an extension or filename which kind of defeats the purpose.

If we use the search results from the more generic [`search:code extension:res NOT nothack`](https://github.com/search?type=code&q=extension%3Ares+NOT+nothack) search, we can see XML is clearly identified as a popular user of this extension and from a quick calculation, it fits the criteria for inclusion in Linguist, so you could get this moving by adding `.res` to XML and add a sample.

From a search of those files that aren't XML or ReScript, most appear to be "binaries" with the `x-dtbresource+xml` mime type which would be ignored by Linguist and GitHub anyway so wouldn't count to the stats or be identified as ReScript or XML.

ryyppy:
@lildude I added `.res` as an extension to the XML language and and added a [holobloc](https://www.holobloc.com/) XML file as a sample

JT0407 (approved):

JT0407 (approved):

lildude:
Thanks @ryyppy. Can you please update the template in the OP with a link to the source of the XML file and state the license for the file.

Other than that, this is looking good to me.

ryyppy:
Great! I don't have edit rights for the thread (will need to ping the OP to do that).

Will drop the information in this comment for now:

**Sample Source:**
- [holobloc-sample.res](https://github.com/PRETgroup/goFB/blob/5b6dd39f4e0a01148a88ea381bb6e6b5242cbcd5/examples/goFB_only/c_tcrest/circuitbreaker_mem/fbt/_CBCore0.res)

**Sample License:**
- [holobloc-sample.res](https://github.com/PRETgroup/goFB/blob/5b6dd39f4e0a01148a88ea381bb6e6b5242cbcd5/LICENSE)

brnrdog:
@ryyppy @lildude thanks, I just updated the description with the new sample source and license!

lildude (approved):

PR #4863, opened 2020-05-13
Ignore Yarn Plug'n'Play installation dependencies

PatrickShaw:
Yarn Plug'n'Play (a relatively recent package management tool typically used for managing JavaScript dependencies) dependencies and some other files are committed to the repository if you're using [Zero Installs](https://yarnpkg.com/features/zero-installs) guidelines. As such, whatever code you've installed as a dependency (i.e. Code that's not yours) ends up slightly skewing Linguist's analysis statistics.

For example, my [monorepo](https://github.com/PatrickShaw/monorepo/tree/70b24446932fc7f68893f6f500d5ca208be43b12) uses almost exclusively TypeScript, however, ~most of the dependencies in `.yarn/cache/*`~ `.pnp.js` is a giant JavaScript file generated by yarn and so it looks like my monorepo is primarily made up of JavaScript.

This PR should fix this issue for anyone using Yarn Plug'n'Play.

## Checklist:

- [ ] **I am associating a language with a new file extension.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

- [x] **I am fixing a misclassified language**
  - [x] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

[...]

PR #4977, opened 2020-08-29
Add more Yarn 2 folders into vendor list

kidonng:
## Description

Add two Yarn 2 folders to vendor list: `.yarn/sdks` and `.yarn/versions`, accroding to [the list of files used by Yarn 2](https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored).

## Checklist:


Alhadis (changes requested):
According to the link you provided, there's a couple other filenames it recommends be added to one's `.gitignore`:

> * `.yarn/unplugged` and `.yarn/build-state.yml` should likely always be ignored […]
> * `.yarn/cache` and `.pnp.*` may be safely ignored, but you'll need to run `yarn install` to regenerate them

So this brings the complete list of vendored paths to:

~~~yaml
# Yarn 2
- (^|/)\.pnp\.
- (^|/)\.yarn/cache/
- (^|/)\.yarn/plugins/
- (^|/)\.yarn/releases/
- (^|/)\.yarn/sdks/
- (^|/)\.yarn/unplugged/
- (^|/)\.yarn/versions/
~~~

kidonng:
`.pnp.*` is already added in #4863. `cache` is already present at https://github.com/github/linguist/pull/4977/files#diff-7a4aecbd604d08120b884f4369388be6R13. About the `unplugged` one I thought it should be handled by `.gitignore`, but after thinking twice I guess it makes sense to include it as well.

Alhadis:
Users won't always have a `.gitignore` file that covers every commonly-ignored filetype. Some users knowingly track such files in version control (e.g., `node_modules` directories). Whatever the reason, ignoring a file is a different matter to simply marking it as vendored (which only affects Linguist's language usage stats).

Alhadis (approved):

lildude (approved):

yordis:
Hey there,

I am a bit confused about my project, I recently had to https://github.com/straw-hat-team/javascript-monorepo/blob/7569a753a785acb272d4bb0122085862a1f6e176/.gitattributes#L1-L3 which doesn't seem to work perfectly (or it doesn't at all).

Is this particular package being used at GitHub? If so, is it released with such changes or how can I verify that it works since I needed to manually do these things in my project?

Thanks in advanced,

lildude:
> which doesn't seem to work perfectly (or it doesn't at all).

You've needed to implement an override ride because Kurdish only marks these dirs as generated and your `.yarn` contains others:

https://github.com/github/linguist/blob/80f3531e8a1014a23f4606458e5a528053ed3cac/lib/linguist/vendor.yml#L44-L49

This override should do the trick to mark everything as generated, but it won't affect the Search results returned when you click the languages in the sidebar. It only affects the percentages.

What makes you think you override isn't working?

yordis:
> PS This really should have been a new discussion

My apologies, I am not sure it was an issue or something so I asked the question here: https://github.com/github/linguist/discussions/5738

PR #4978, opened 2020-08-30
Abolish restrictions governing colour choices

### LINKED ISSUE #4291: Revisiting language groups

Alhadis:
This is the first of three PRs to fix #4291 and #4506, both of which are blocked on this change. I won't repeat what those threads already do a good job of explaining. 

__Status of related tasks:__
1. [ ] Abolish restrictions governing colour choices (this)
2. [ ] [Add/fix incorrect or missing language colours](https://github.com/github/linguist/pull/4980)
3. [ ] [Ungroup major languages](https://github.com/github/linguist/pull/4979)

# The problem

We're still enforcing colour-difference thresholds, but nobody can explain why. _"Still waiting on GitHub's designers" isn't a sound justification when they've had 1.5 years to raise an objection to the colour-test's removal.

# The solution

Stop giving a crap and just go for it. Chances are the test appears 10 times more important to us than it does GitHub's designers, because they haven't been the ones dealing with it for the past 5 years.


lildude (approved):
Lets do this. 🙇

PR #4979, opened 2020-08-31
Ungroup major languages

Alhadis:
Fixes: #4291 / Depends on: #4978

__Status of related tasks:__
1. [x] [Abolish restrictions governing colour choices](https://github.com/github/linguist/pull/4978)
2. [ ] [Add/fix incorrect or missing language colours](https://github.com/github/linguist/pull/4980)
3. [ ] Ungroup major languages (this)

## Description
This pull-request ungroups languages listed in #4291, from which I'll quote the reason for doing so:

> There are several languages on GitHub which presently fall under the usage statistics of another, "parent" language, which certainly deserve reconsideration — or at the very least, some public discussion for highlighting the reasons why they fall under another language's umbrella.

Note that this PR includes #4978 as part of its changes, so ensure that gets merged first before this one. The reason is that users will notice (and complain) if their repository stats are 60% grey because SCSS/Less are no longer grouped under CSS or what-have-you.

## Checklist:
- [x] **I am ~~changing~~ *adding* the colours associated with languages.**
	- [x] **I have obtained [agreement](https://github.com/github/linguist/issues/4506) from the wider language communities.**  
	_"Agreement"_ isn't the word I'd use, since these colours are already used by each language's official branding, making them canon by proxy.


lildude (approved):
LGTM

PR #4981, opened 2020-08-31
Add .re disambiguation (Reason vs C++)

### LINKED ISSUE #4930: ReasonML files classified as C++
## Preliminary Steps

Please confirm you have...

## Problem Description
Some Reason-files are classified as C++ for no apparent reason (no pun intended). I can't make out any pattern in this, the majority is being classified correctly.
Explicitly mapping *.re files to Reason in .gitignore doesnt seem to make any difference here.

https://github.com/jisensee/charly/
However, I have seen this issue is other repos as well.

### Last modified on:
2020-07-25

### Expected language:
ReasonML

### Detected language:
C++

smola:
## Description

- Add added Reason sample that was previously misclassified as C++:
  - Source: https://github.com/esy/esy/blob/956cf2a6b4ad4db3386f4a081e231b1d2f78bd47/esy-package-config/SourceSpec.re
    License: BSD 2-Clause

- Add heuristics for .re (Reason vs C++).

Fixes #4930.

## Checklist:

    - Sample source(s): https://github.com/esy/esy/blob/956cf2a6b4ad4db3386f4a081e231b1d2f78bd47/esy-package-config/SourceSpec.re
    - Sample license(s): BSD-2

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -416,17 +416,13 @@ disambiguations:
   rules:
   - language: Reason
     pattern:
-    - '^\s*open\s+\w+\s*;$'
-    - '^\s*include\s+\w+\s*;$'
     - '^\s*module\s+type\s'
-    - '^\s*let\s+\w+:\s+.*=.*;\s*$'
-    - '^\s*let\s+module\s\w+\s*=\s*{'
+    - '^\s*(?:include|open)\s+\w+\s*;\s*$'
+    - '^\s*let\s+(?:module\s\w+\s*=\s*{|\w+:\s+.*=.*;\s*$)'
   - language: C++
     pattern:
-    - '^\s*#\s*include\s+<[^>]+>'
-    - '^\s*#(if|ifdef|define|pragma)\s+\w'
+    - '^\s*#(?:(?:if|ifdef|define|pragma)\s+\w|\s*include\s+<[^>]+>)'
     - '^\s*template\s*<'
-    - '\bstd::\w+'
 - extensions: ['.rno']
   rules:
   - language: RUNOFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -412,6 +412,21 @@ disambiguations:
     pattern: '(?i:\bRebol\b)'
   - language: R
     pattern: '<-|^\s*#'
+- extensions: ['.re']
+  rules:
+  - language: Reason
+    pattern:
+    - '^\s*open\s+\w+\s*;$'
+    - '^\s*include\s+\w+\s*;$'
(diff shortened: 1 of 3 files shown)


lib/linguist/heuristics.yml
```diff
@@ -412,6 +412,21 @@ disambiguations:
...
     pattern: '<-|^\s*#'
+- extensions: ['.re']
+  rules:
+  - language: Reason
+    pattern:
+    - '^\s*open\s+\w+\s*;$'
+    - '^\s*include\s+\w+\s*;$'
+    - '^\s*module\s+type\s'
+    - '^\s*let\s+\w+:\s+.*=.*;\s*$'
+    - '^\s*let\s+module\s\w+\s*=\s*{'
```
Alhadis, line 423: This is more optimal (and readable, IMHO):

~~~regexp
^\s*module\s+type\s
^\s*(?:include|open)\s+\w+\s*;$
^\s*let\s+(?:module\s\w+\s*=\s*{|\w+:\s+.*=.*;\s*$)
~~~

Also, unless trailing whitespace after a semicolon and line-break is illegal (which it probably isn't), I'd replace `;$` with `;\s*$`, just to err on the side of caution.

lib/linguist/heuristics.yml
```diff
@@ -412,6 +412,21 @@ disambiguations:
...
+  - language: C++
+    pattern:
+    - '^\s*#\s*include\s+<[^>]+>'
+    - '^\s*#(if|ifdef|define|pragma)\s+\w'
```
Alhadis, line 427:
~~~regexp
^\s*#(?:(?:if|ifdef|define|pragma)\s+\w|\s*include\s+<[^>]+>)
~~~

lib/linguist/heuristics.yml
```diff
@@ -412,6 +412,21 @@ disambiguations:
...
+    - '^\s*template\s*<'
+    - '\bstd::\w+'
```
Alhadis, line 429: I'm a \bit  wary of that word boundary. 🤔 That'll also match `"std::foo"` or `'std::bar'`, which could easily be found in a program that has a string referencing something called `std::whatever`.

Unless there's a more specific and unambiguous way to match this, I'd probably omit this pattern altogether. Or at least replace `\w+` with a list of common C++ classes/functions/etc.

Alhadis (changes requested):

lib/linguist/heuristics.yml
smola: Removed. `std::` can appear in a lot of contexts, so it is a bit hard to be more specific without going wild.

Alhadis (approved):

lildude (approved):

PR #4984, opened 2020-09-01
Remove empty `.gitattributes` file

Alhadis:
[This](https://github.com/github/linguist/blob/9d0ba5801b8acebb0b318eb0405aff2e895f10c0/.gitattributes) has been sitting there for 6 years, but surprisingly, nobody seemed to notice.

*Template removed as it doesn't apply, though it'd be funny if it did.*


lildude (approved):
I've never noticed this file either 😁

PR #4987, opened 2020-09-01
Update CI: Add Ruby 2.7, remove 2.4 and cache dependencies

lildude:
## Description

General CI improvements:

- Added testing against Ruby 2.7 as it is now generally available and in full use on GitHub.com.
- Removed testing against Ruby 2.4 is it now EOL.
- Added dependency caching to speed up CI a little bit in future.

_Template removed as it doesn't apply_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


Alhadis (approved):

lildude:
🤦 I've just realised we don't commit the `Gemfile.lock` file so the caching has no effect. Will sort in another PR after the release.

Issue #4966, opened 2020-08-25
Review most popular languages in lib/linguist/popular.yml


sheerun:
[This file](https://github.com/github/linguist/blob/master/lib/linguist/popular.yml) haven't been update in at least two years and new languages are now popular https://madnight.github.io/githut/#/pull_requests/2020/2 shows that we are missing among others:

- Rust
- Dart
- Kotlin
- Groovy
- Elixir
- PowerShell
- Emacs Lisp
- OCaml
- Erlang

Also it seems some languages are no longer popular, like:

- ActionScript
- CoffeeScript

lildude:
🤔 I'm not sure that file is even used anymore. I'd need to dig through the GitHub.com source code to be sure.

sheerun:
@lildude For sure in "Written in this language" advanced search option

lildude:
Yup. Looks like it is definitely still used. I'll update the list.

PR #2207, opened 2015-03-06
Update popular languages

aroben:
I took all non-fork repositories on github.com, grouped them by their primary language, and took the 25 most popular.

@tnm Is this the criteria you've used in the past when updating this file?

/cc @github/linguist

PR #4988, opened 2020-09-01
Update list of popular languages

lildude:
## Description

As per the title, this updates the list of popular languages to those most popular today. The list was last updated over 6 years ago in https://github.com/github/linguist/pull/2207.

Fixes https://github.com/github/linguist/issues/4966

_Template removed as it doesn't apply_

### DIFF
--- a/lib/linguist/popular.yml
+++ b/lib/linguist/popular.yml
@@ -2,28 +2,28 @@
 #
 # This file should only be edited by GitHub staff
 
-- ActionScript
 - C
 - C#
 - C++
-- CSS
-- Clojure
 - CoffeeScript
+- CSS
+- Dart
+- DM
+- Elixir
 - Go
+- Groovy
 - HTML
-- Haskell
 - Java
 - JavaScript
-- Lua
-- MATLAB
+- Kotlin
 - Objective-C
-- PHP
 - Perl
+- PHP
+- PowerShell
 - Python
-- R
 - Ruby
+- Rust
 - Scala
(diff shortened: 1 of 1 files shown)


Alhadis (approved):
I still long for the day when APL and Brainfuck make it to this list.

sheerun:
DM?

lildude:
Yes, I was surprised by that too. I have no idea why it's showing up as really popular.

sheerun:
maybe it's bug. syntax is similar to c++

Alhadis:
The 'hell is DM?

lildude:
Possibly, though it is a language Linguist supports and has for a very very long time.  The extension is incredibly popular too so I can only think there's a renaissance or something has started using the `.dm` extension and peeps are happy with the syntax highlighting and identification. 🤷

pietroppeter:
is this true about DM language being really popular? I do not see many repos using DM: https://github.com/search?q=language%3Adm

I did see 1 example of a Markdown only repo marked as DM (because of a wrongly named extension) but I am not sure how many more there could be of those: https://github.com/EBB26/datasciencecoursera

also, what is the use of this list of popular languages?

PJB3005:
In case anybody is curious: DM is the language used by the BYOND game engine which hosts Space Station 13, which is open source and one of the forks was #7 most commented on all of GitHub in 2017. 

Just figured I should say that because we just found out about this PR and we're all laughing our asses off about the confusion in here.

PR #4992, opened 2020-09-02
Change `linguist-documentation` example from /* to /**

prototriangle:
Users may find this edit helpful as it is more likely to align with their intentions.

>The rules by which the pattern matches paths are the same as in `.gitignore` files (see [gitignore[5]](https://www.git-scm.com/docs/gitignore), with a few exceptions:
>- negative patterns are forbidden
>- patterns that match a directory do not recursively match paths inside that directory **(so using the trailing-slash `path/` syntax is pointless in an attributes file; use `path/**` instead)**

The above was retrieved from [here](https://www.git-scm.com/docs/gitattributes#main).

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):
  
    - Sample source(s):
    - Sample license(s):

    - [URL to public discussion]

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ Just like vendored files, Linguist excludes documentation files from your projec
 Use the `linguist-documentation` attribute to mark or unmark paths as documentation:
 
 ```gitattributes
-project-docs/* linguist-documentation
+project-docs/** linguist-documentation
 docs/formatter.rb -linguist-documentation
 ```
 


Alhadis (commented):
Attributes are inherited by files inside subdirectories, which means that `foo/*` is, for all intents and purposes, identical to `foo/**`. This might confuse less experienced users, particularly those unfamiliar with how globbing works. At best, it adds no new information. At worst, it's unclear to novices who've seen `*` used, but never `**`, encouraging [cargo-cult](https://en.wikipedia.org/wiki/Cargo_cult_programming) or faulty recollection the next time they need wildcard file-matching.

**TL;DR:** 👎 , sorry.

lildude:
I agree with @Alhadis here. This isn't really necessary, may lead to confusion and may not align with the developer's real intentions.

> Users may find this edit helpful as it is more likely to align with their intentions.

You can't make that assumption. The user may have intended to only apply to the files in that directory and _not_ recursing. See https://github.com/github/linguist/pull/4889#issuecomment-759386273 for another discussion on this.

As with that PR, ultimately, this is really about understanding the git documentation and the developer's end goals, and not making any assumptions, which are outside the scope of using Linguist.

Closing.

PR #4998, opened 2020-09-06
Added FreeBasic language

XusinboyBekchanov:
I added FreeBasic language. Please see this.

## Checklist:

      -  https://github.com/search?q=extension%3Abas+NOT+nothack&type=Code
      - https://github.com/search?q=extension%3Abi+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/XusinboyBekchanov/VisualFBEditor
      - https://github.com/XusinboyBekchanov/MyFbFramework
    - Sample license(s): GPL and LGPL
  

XusinboyBekchanov:
What should I do? I did not understand.

lildude (changes requested):
> What should I do? I did not understand.

1.  fill "I am adding a new language." section of the template in the OP. You can remove the other sections. Pay special attention to providing search results for all the extensions you're adding. If you don't meet the documented popularity requirements, the extension can't be added.

2. review the test failures. Most of these test failures are quite clear as to what the problems are and what you need to do to address them. Start with following those tips and explanations.

3. you're adding the `.inc` extension which, assuming it meets the popularity requirements specifically for FreeBasic, is already extensively used on GitHub and associated with several other languages too, so you will need to write a heuristic (ie a regex) that identifies FreeBasic `.inc` files from the other languages and add it to this section:

https://github.com/github/linguist/blob/8083cb5a89cee2d99f5a988f165994d0243f0d1e/lib/linguist/heuristics.yml#L218-L234

More information is in the CONTRIBUTING.md file

XusinboyBekchanov:
All checks passed, now you can commit. I only left .bi files

lildude:
This PR still needs a sample FreeBasic `.bi` file.

There's also the issue with other `.bi` files on GitHub - https://github.com/search?type=code&q=extension%3Abi+NOT+nothack+size%3A%3E1

Most of those appear to be FreeBasic, but quite a few of those appear to be QBasic which means this PR will also need to add support for QBasic and a heuristic to differentiate between the two.

XusinboyBekchanov:
I did everything

XusinboyBekchanov:
I added .bas extension too.

abc52090241 (changes requested):
No

XusinboyBekchanov:
> No

What for?

lib/linguist/heuristics.yml
```diff
@@ -163,6 +171,11 @@ disambiguations:
   - language: Frege
     pattern: '^\s*(import|module|package|data|type) '
   - language: Text
+- extensions: ['.frm']
+  rules:
+  - language: FreeBasic
+    pattern: '^\s*#(define|endif|endmacro|if|ifdef|ifndef|include|lang|macro)\s'
+  - language: VBA
```
Alhadis, line 178: <a name="yup-this-reason"></a>This should be removed:

```suggestion
- - language: VBA
```

A `language` field without an accompanying `pattern` matches *any* file that doesn't match a preceding pattern. That is to say, it behaves like this:

~~~python
if pattern.match(candidateFile)
	return FreeBasic
else
	return VBA
~~~

It's okay if none of the `rules` match a candidate, so long as our classifier does a good job of analysing samples. 😉

lib/linguist/heuristics.yml
```diff
@@ -217,6 +230,8 @@ disambiguations:
   - language: Slice
 - extensions: ['.inc']
   rules:
+  - language: FreeBasic
+    pattern: '^\s*#(define|endif|endmacro|if|ifdef|ifndef|include|lang)\s'
```
Alhadis, line 234: This is slightly more performant:

```suggestion
    pattern: '^[ \t]*#(?:define|endif|endmacro|ifn?def|if|include|lang)\s'
```

lib/linguist/heuristics.yml
```diff
@@ -232,6 +247,7 @@ disambiguations:
     - '^\s*(?:public\s+|private\s+|\s*)function\s+\w+\s*\([\w\s,]*\)\s*{'
   - language: POV-Ray SDL
     pattern: '^\s*#(declare|local|macro|while)\s'
+  - language: C++
```
Alhadis, line 250: Please remove this. It's out-of-scope for this PR, and will also misclassify files for the [reason I described above](#yup-this-reason).

lib/linguist/languages.yml
```diff
@@ -1596,6 +1596,21 @@ Fortran Free Form:
...
   codemirror_mime_type: text/x-fortran
   language_id: 761352333
+FreeBasic:
+  type: programming
+  color: "#867db1"
+  extensions:
+  - ".bi"
+  - ".bas"
+  - ".frm"
+  - ".inc"
```
Alhadis, line 1606: `.inc` is a very commonly-used file extension, so we'll need a heuristic for matching FormBasic `.inc` files.

Alhadis (changes requested):

lib/linguist/heuristics.yml
XusinboyBekchanov: The test was giving error, so I added these lines
- language: VBA
- language: C++

XusinboyBekchanov:
I returned everything to its previous position

lib/linguist/heuristics.yml
```diff
@@ -82,12 +82,20 @@ disambiguations:
   - language: LTspice Symbol
     pattern: '^SymbolType[ \t]'
   - language: Asymptote
+- extensions: ['.bas']
+  rules:
+  - language: FreeBasic
+    pattern: '^\s*#(define|endif|endmacro|if|ifdef|ifndef|include|lang|macro)\s'
```
lildude, line 88: 🤔 Is this really the only differentiator between VBA and FreeBasic? 

I ask as it'll only match lines that have `#` as the first non-space character and from a quick search on samples of FreeBasic at https://www.freebasic.net/wiki/CodeLibrary, many FreeBasic files don't have lines that start with `#` which means it'll fall through to the classifier which will struggle because your sample looks a lot like VBA, hence the failing test.

Also, why didn't you implement the suggestion from @Alhadis at https://github.com/github/linguist/pull/4998#discussion_r544146635 ? (I've "unresolved" it).

XusinboyBekchanov: I commited the suggestion from @Alhadis. Also in FreeBasic, as opposed to VBA, there are: Dim As, Dim Shared, Common, ...

XusinboyBekchanov: ReDim Shared, ReDim As

anOrdinaryflower: `Declare Operator`  `#lang "FB"`
And the type that VBA does not have: `Short ZString WString`

lildude: Given these additional comments, I think you probably need to extend this regex. Maybe @Alhadis has some more suggestions.

Also, should this be using a different regex from that one defined for `.bi`?

XusinboyBekchanov: #lang exists in my pattern

XusinboyBekchanov: >Also, should this be using a different regex from that one defined for .bi?
There is no difference between .bi and .bas

XusinboyBekchanov: I could not add new changes, all attempts failed, so I left it as the previous one.

XusinboyBekchanov:
How do you fix this conflict? I added BASIC to fix it, but the check does not go (Waiting for status to be reported).

lildude:
I've fixed the merge conflicts for you. We might need to tweak things further, but we'll see once the tests have run.

XusinboyBekchanov:
> I've fixed the merge conflicts for you. We might need to tweak things further, but we'll see once the tests have run.

Thank you.

lib/linguist/languages.yml
```diff
@@ -1650,6 +1650,19 @@ Fortran Free Form:
...
+  extensions:
+  - ".bi"
+  - ".bas"
+  tm_scope: source.vbnet
+  aliases:
+  - fb
+  ace_mode: text
+  codemirror_mode: vb
+  codemirror_mime_type: text/x-vb
+  language_id: 761352334
```
lildude, line 1665:
```suggestion
  language_id: 472896659
```

Just spotted that you set this manually and not using the `script/update-ids` script as detailed in the CONTRIBUTING.md file

test/test_heuristics.rb
```diff
@@ -838,7 +845,7 @@ def test_vba_by_heuristics
       "Vim script" => all_fixtures("Vim script", "*.vba")
     })
   end
-
+
```
lildude, line 848:
```suggestion
 
```

Extra space has snuck in.

lib/linguist/heuristics.yml
lildude: Ok. I think this has gone on long enough. We can always change things later if the current heuristic proves to be problematic.

lildude (approved):

anOrdinaryflower (commented):
Great.

PR #4999, opened 2020-09-06
Add pedantic test for heuristics exhaustiveness

### LINKED ISSUE #4207: Pedantic test to check all heuristic rules are tested
Thanks to #4189, all heuristic rules should now have a corresponding test in `test_heuristics.rb`. We should add a [pedantic test](https://github.com/github/linguist/blob/51feda9e1f3bd1d41f19425e663e9f6f2cb640ed/test/test_pedantic.rb) to make sure it remains true.

*Removed template as it doesn't apply.*

smola:
## Description

Tests that there is a test in test_heuristics.rb for every extension
defined in heuristics.rb.

Add missing heuristic tests and add explicit checks for languages that
miss a rule.

Fixes #4207.

## Checklist:

*Checklist does not apply.*


lildude (approved):
LGTM

lildude:
@smola I've merged `master` and resolved a conflict caused by merging https://github.com/github/linguist/pull/5113. I think I've caught everything, but would appreciate it if you could give is a once-over to be sure I've not missed or messed something. Thanks 🙇

smola:
@lildude Looks good to me!

PR #4589, opened 2019-07-25
Add support for Texinfo documents

Alhadis:
This PR adds support for [Texinfo](https://www.gnu.org/software/texinfo/manual/texinfo/), a lightweight markup language chiefly used by GNU projects.

Fixes: #4479.

## Checklist:
- [x] **I am adding a new language.**
	- [x] **The extensions are used in hundreds of repositories on GitHub.com**
		- `.texinfo`: [~26,875 results](https://github.com/search?q=extension%3Atexinfo+NOT+nothack&type=Code)
		- `.texi`:    [~647,762 results](https://github.com/search?q=extension%3Atexi+NOT+nothack&type=Code)
		- `.txi`:     [~6,509 results](https://github.com/search?q=extension%3Atxi+NOT+nothack&type=Code)
	- [x] **I have included a real-world usage sample**
		- `protocol-spec.texi`: [Source](https://github.com/ruby-debug/ruby-debug-ide/blob/c14e867c33be0d6f5fd7ea0ffab9b0eaf4b24cff/doc/protocol-spec.texi) | [MIT license](https://github.com/ruby-debug/ruby-debug-ide/blob/8add0a9cee3a09ef3996c2b349fdf41df10f5cb7/MIT-LICENSE)
	- [x] **I have included a syntax highlighting grammar:**
		- [`Alhadis/language-texinfo`](https://github.com/Alhadis/language-texinfo)
		- [Lightshow preview](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-texinfo%2Fblob%2Fmaster%2Fgrammars%2Ftexinfo.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fruby-debug%2Fruby-debug-ide%2Fmaster%2Fdoc%2Fprotocol-spec.texi&code=)

PR #5004, opened 2020-09-10
Add common .vscode config filenames as JSON with Comments

dbjorge:
Visual Studio Code uses/generates several JSON with Comments config files under folders named `.vscode` with fixed names and a `.json` extension. This PR registers them as "JSON with Comments" instead of the current "JSON".

## Description
Checking in these `.vscode` configs is extremely common (>100k search results each), but some of these filenames are pretty generic and overlap with non-vscode scenarios that might not support comments, so I don't think it would be unreasonable to push back on this PR looking for it to use something that more specifically looks for these files underneath `.vscode` folders.

I wanted to send out the simple version first to start any discussion about that before writing code for it, since that sort of path-based heuristic for language detection didn't seem to fit the pattern of any of the existing languages (though it does match vendor-detection). I would want to use a path-based heuristic for this rather than a content-based heuristic if we wanted to be more specific; particularly, I would want to avoid adding a `heuristics.yml` entry for `.json` that's based on the presence of `//` or `/*` tokens, because I think "a developer uses a comment in a JSON file that *doesn't* support it" is likely to be such a common error that it makes sense for us to highlight it in pull requests if we aren't extremely confident that the file really does support comments.

Manaully looking over the first ~10 pages of search results for [`extension:json in:path "settings.json" NOT nothack`](https://github.com/search?p=6&q=extension%3Ajson+in%3Apath+%22settings.json%22+NOT+nothack&type=Code), it looks like about 80% of results are VS Code config files, and about 20% are a varied assortment of others. 

## Checklist:

- [x] **I am associating a language with a new ~file extension~ filenames.**
  - [x] The new ~extension~ filenames are used in hundreds of repositories on GitHub.com
    - Search results for each ~extension~ case:
      - All .vscode JSON files: https://github.com/search?q=extension%3Ajson+in%3Apath+.vscode+NOT+nothack&type=Code
      - launch.json: https://github.com/search?p=6&q=extension%3Ajson+in%3Apath+%22launch.json%22+NOT+nothack&type=Code
      - extensions.json: https://github.com/search?p=6&q=extension%3Ajson+in%3Apath+%22extensions.json%22+NOT+nothack&type=Code
      - settings.json: https://github.com/search?p=6&q=extension%3Ajson+in%3Apath+%22settings.json%22+NOT+nothack&type=Code
      - tasks.json: https://github.com/search?p=6&q=extension%3Ajson+in%3Apath+%22tasks.json%22+NOT+nothack&type=Code
  - [x] I have included a real-world usage sample for all ~extensions~ cases added in this PR:
    - Sample source(s):
      - `extensions.json`: default generated by VS Code ([generator source](https://github.com/microsoft/vscode/blob/a45fa6ed50858306e5510742414887371d218a52/src/vs/workbench/contrib/extensions/common/extensionsFileTemplate.ts#L40))
      - `launch.json`: default generated by VS Code for a node project ([generator source](https://github.com/microsoft/vscode/blob/f74e473238aca7b79c08be761d99a0232838ca4c/src/vs/workbench/contrib/debug/common/debugger.ts#L148))
      - `settings.json`: a snippet of the default output from the VS Code "Preferences: Open Default Settings (JSON)" command, contents of which is generated ([generator source](https://github.com/microsoft/vscode/blob/f74e473238aca7b79c08be761d99a0232838ca4c/src/vs/editor/common/config/commonEditorConfig.ts#L457))
      - `tasks.json`: default generated by VS Code for a dotnet build ([generator source](https://github.com/microsoft/vscode/blob/f74e473238aca7b79c08be761d99a0232838ca4c/src/vs/workbench/contrib/tasks/common/taskTemplates.ts#L16))
    - Sample license(s):
      - All samples are generated by default configuration of VS Code, which uses the [MIT license](https://github.com/microsoft/vscode/blob/master/LICENSE.txt)
  - [n/a] I have included a change to the heuristics to distinguish my language from others using the same extension.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2489,8 +2489,12 @@ JSON with Comments:
   - ".jscsrc"
   - ".jshintrc"
   - ".jslintrc"
+  - extensions.json
   - jsconfig.json
   - language-configuration.json
+  - launch.json
+  - settings.json
+  - tasks.json
   - tsconfig.json
   - tslint.json
   language_id: 423
--- a/samples/JSON with Comments/filenames/extensions.json
+++ b/samples/JSON with Comments/filenames/extensions.json
@@ -1,0 +1,13 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		
+	],
+	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+	"unwantedRecommendations": [
+		
+	]
+}
--- a/samples/JSON with Comments/filenames/launch.json
+++ b/samples/JSON with Comments/filenames/launch.json
@@ -1,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Program",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${file}"
+        }
+    ]
+}
--- a/samples/JSON with Comments/filenames/settings.json
+++ b/samples/JSON with Comments/filenames/settings.json
@@ -1,0 +1,16 @@
+{
+	// Controls whether the editor shows CodeLens.
+	"diffEditor.codeLens": false,
+
+	// When enabled, the diff editor ignores changes in leading or trailing whitespace.
+	"diffEditor.ignoreTrimWhitespace": true,
+
+	// Timeout in milliseconds after which diff computation is cancelled. Use 0 for no timeout.
+	"diffEditor.maxComputationTime": 5000,
+
+	// Controls whether the diff editor shows +/- indicators for added/removed changes.
+	"diffEditor.renderIndicators": true,
+
+	// Controls whether the diff editor shows the diff side by side or inline.
+    "diffEditor.renderSideBySide": true
+}
--- a/samples/JSON with Comments/filenames/tasks.json
+++ b/samples/JSON with Comments/filenames/tasks.json
@@ -1,0 +1,24 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build",
+            "command": "dotnet",
+            "type": "shell",
+            "args": [
+                "build",
+                // Ask dotnet build to generate full paths for file names.
+                "/property:GenerateFullPaths=true",
+                // Do not generate summary otherwise it leads to duplicate errors in Problems panel
+                "/consoleloggerparameters:NoSummary"
+            ],
+            "group": "build",
+            "presentation": {
+                "reveal": "silent"
+            },
+            "problemMatcher": "$msCompile"
+        }
+    ]
+}


Alhadis (commented):
These are all _very_ generic filenames likely to match ordinary JSON files in contexts unrelated to VS Code. 👎  from me, sorry.

dbjorge:
Thanks for the review, @Alhadis ! I do agree that the names are concerningly generic. Did you have any thoughts on implementing new support for using a path-based heuristic vs using the existing support for content-based heuristics to make the matching more specific? My inclination is that something path-based ("are the files in a directory named .vscode") makes much more sense for this case, but it doesn't look like any of the existing language-determination heuristics work that way; I don't know whether it's something that just hasn't come up before or if there's an argument against making it context-dependent.

Alhadis:
> Did you have any thoughts on implementing new support for using a path-based heuristic vs using the existing support for content-based heuristics to make the matching more specific?

I believe that's been suggested before, with the consensus being the same as every other sophisticated classification logic — only a small percentage of users would benefit from the change, which doesn't justify the extra complexity added to Linguist's codebase. The fact we've never needed path-scoped heuristics before indicates that such a feature will never see mileage beyond this PR.

The real issue here is Microsoft's continued use of `.json` instead of `.jsonc`. 😜 If you want to "extend" a [standardised file-format](https://www.iso.org/standard/71616.html) with non-standard features, pick a different file-extension.

dbjorge:
I would definitely want any potential new logic to be more broadly useful than just these specific files. I'm confused by the assertion that it has both "been suggested before" and "will never see mileage beyond this PR"; if it's been suggested before, that sounds like an indicator that it *would* be more broadly applicable than this PR.

Can you recall any of the specific cases where something like this has been suggested before? I would be interested to know whether something simpler like path-suffix would be sufficient for most of them or if they'd have been looking for something more powerful (eg, something regex-based like how `vendor.yml` works)

> If you want to "extend" a standardised file-format with non-standard features, pick a different file-extension.

I agree that it would be better to have used an extension that distinguishes the cases more explicitly. This wording feels a little confrontational, though; I am not a member of the VS Code team and have no control over that decision, so I don't think it's constructive feedback for this PR.

Alhadis:
> if it's been suggested before, that sounds like an indicator that it would be more broadly applicable than this PR.

I believe it was used in the context of file proximity; IIRC, it was something like: _"When distinguishing C and Objective-C, check the first two parent directories for a file ending with such-and-such extension."_ Though the context was different, the complications are the same — Linguist can't analyse files one a time, which is how it operates when computing usage metrics.

> This wording feels a little confrontational, though

Sorry. That was me letting off steam in Microsoft's general direction. 😓  As you can probably tell, I'm a wee bit opinionated; none of my remarks were aimed at you, so don't worry. 👍

dbjorge:
> I believe it was used in the context of file proximity; IIRC, it was something like: "When distinguishing C and Objective-C, check the first two parent directories for a file ending with such-and-such extension." Though the context was different, the complications are the same — Linguist can't analyse files one a time, which is how it operates when computing usage metrics.

That's fair, it makes sense that it would cause complications to have to consider files not-one-at-a-time. I'm not sure those same complications would apply to this specific case, though; individual `Blob`s already know their own `path`, so I don't think adding a strategy based on comparison against that `path` would prevent Linguist from being able to process them independently (it actually already does at least one similar path-based-regex check already, via `include_in_language_stats?` -> `vendored?`).

Given that the `path` is all we'd need to filter on here and that it's already present in the data given to each Strategy, I don't think it would require particularly expensive-to-maintain changes to add support for having the language specify `.vscode/settings.json` rather than just `settings.json`. I could imagine:

a. Adding support for directory-qualified filenames like any other filename (eg, enable writing `.vscode/settings.json` in the list instead of just `settings.json`).
b. As a, but rename filenames to `relative_filepaths`/`filepaths`/`filepath_suffixes`/similar.
c. As a, but create a new `languages.yml` property (`relative_filepaths`? `filepaths`? `filepath_suffixes`?) instead of reusing `filenames`.

a is pretty low-churn (mostly updating `language.rb`'s `find_by_filename` to check `@filename_index` a few times with different numbers of ancestor directories), but I'm not thrilled about a because I think it makes the term "filenames" uncomfortably misleading. I'm not thrilled about b because I think it makes the more common case in languages.yml less readable. c feels a little redundant, but less objectionable than a or b; could do c for `languages.yml` for readability there but combine them into one index/strategy during parsing to minimize code redundancy.

I could alternately imagine using a more general pattern-based approach (eg, add a `filepath_pattern` option for `heuristics.yml`), but I think that would be a little more complicated and less performant so I'd probably avoid it unless there were other known cases that would benefit from it.

Alhadis:
> Linguist can't analyse files one a time

Christ. I meant _"Linguist can only analyse one file at a time"_ (I shouldn't be allowed near a computer when this exhausted…)

> Given that the `path` is all we'd need to filter on here and that it's already present in the data given to each Strategy, I don't think it would require particularly expensive-to-maintain changes to add support for having the language specify `.vscode/settings.json` rather than just `settings.json`.

All things considered, it's still better if users add an [override](https://github.com/github/linguist#using-gitattributes) to fix JSONC highlighting. Let's revisit the problem you're attempting to fix — ugly comments. Sure, they're unsightly, but they're also easy to fix, provided they bother a user enough. If possible, we prefer users correct case-specific errors using `.gitattributes`, because Linguist has never purported to be perfect, nor does it strive to be.

dbjorge:
I understand that this is a fix for something that is only a slight inconvenience, and that it's possible to fix it per-repo. I don't think either of those are very compelling arguments; making it slightly less inconvenient to read code on GitHub without having to implement that per-repo is pretty much the selling point of Linguist.

I feel frustrated because I feel that you are applying a double-standard compared to similar Linguist changes:

* You have personally added `filenames` support for many specific cases with <3k usages, rather than asking those users to edit `.gitattributes` because "Linguist has never purported to be perfect". (#4678, #4763, #4764, #4794)
* You have also personally added entirely new grammars for specific languages with usage closer to the same level as we're talking about in this PR, involving substantially more (and more case-specific) code to maintain than the path-filtering proposal; for example, #4589 pulls in about 900 lines specific to one language with ~700k usages.

Based on this and your earlier comment "letting off steam in Microsoft's general direction", I feel concerned that you may have a pre-existing bias against my employer which is affecting the objectivity of your review. I would feel more comfortable continuing discussion with a different reviewer; if that's not possible, please feel free to close the PR.

Alhadis:
Perhaps I was unclear. I don't doubt enough people are using these filenames in their `.vscode` directories. What I *am* concerned about is the trade-off. At best, JSONC users will notice the comment highlighting. At worst, JSON users *won't* notice the error during code review.

See, our usual techniques for distinguishing between two different languages doesn't work here. Linguist normally employs heuristics and/or Bayesian classification to disambiguate filenames/extensions shared by multiple languages. For the most part, the candidates are sufficiently different in syntax that this technique works well. However, for JSON/JSONC, Bayesian analysis is useless, and heuristics will be hit-and-miss.

Hopefully you now understand this isn't about in-the-wild usage, but rather that it's not something Linguist is designed to do. Its purpose is classification of plain-text file formats—only by extension is it responsible for syntax highlighting. Changes to its analysis strategies (which are rarely needed, BTW) should therefore be motivated by something relevant to its purpose.

dbjorge:
Yes, I understand the trade-off you're describing and why using a file-content-based heuristic would not be a good fit for this case. I explained as much in the third paragraph of my original PR description; repeating my own PR description back to me as if *you* haven't made it clear enough is quite condescending and patronizing.

What we're talking about is *exactly* correcting the classification of a few text files that should be classified as jsonc but are instead being classified as json. Again, Linguist *already* uses filenames to classify languages and *already* uses filepaths to classify vendor vs not; this is not an analysis strategy which is fundamentally different from logic Linguist is already willing to use in similar applications.

Again, I feel uncomfortable continuing discussion with you, so I'm going to close this PR for now. @lildude or @pchaigno, if either of you would be willing to give this a look, I'd appreciate it; please feel free to reopen the PR if you'd like to discuss any of the options/proposals further.

pchaigno:
I agree with the gist of John's review:
- The pull request as it stands adds very general filenames that seem likely to cause misclassifications.
- It isn't clear if a path-based classification strategy could be applied to other languages or if it's going to stay `.vscode/`-specific.

Nonetheless, I'm not entirely convinced we should be classifying `.json` files as JSON by default instead of JSON With Comments. One could argue that in most cases, the authors know what they're doing: if they add comments, then we shouldn't highlight as errors; if they want those comments highlighted as errors, they can use Linguist overrides. That's a much bigger discussion though. I'm happy to open it and discuss, but we're unlikely to make much progress without a way to collect more outside opinions.

PR #5009, opened 2020-09-14
Improve classifier performance

smola:
## Description

token_probability dominates the classifier runtime, given it is called
once per token per candidate language. The Math.log is particularly
expensive.

To reduce calls, we first create a hash of token occurrences (they are
usually very repetitive), and then call token_probability a single time
per unique token per language.

Additionally, the value of the unknown token logprobability is cached.
When classifying files outside of the samples database, the unknown
token is usually the most common case.

Overall, these changes boost the classifier performance ~2x.

Quick and dirty script to measure performance:

```ruby
# perf.rb
require 'linguist'

blob = Linguist::FileBlob.new("lib/linguist/classifier.rb")
languages = [Linguist::Language['Ruby'], Linguist::Language['Python']]
(1..5000).each do
  Linguist::Classifier.call(blob, languages)
end

```

## Checklist:

_Checklist does not apply._

### DIFF
--- a/lib/linguist/classifier.rb
+++ b/lib/linguist/classifier.rb
@@ -87,6 +87,7 @@ module Linguist
       @tokens          = db['tokens']
       @language_tokens = db['language_tokens']
       @languages       = db['languages']
+      @unknown_logprob = Math.log(1 / db['tokens_total'].to_f)
     end
 
     # Internal: Guess language of data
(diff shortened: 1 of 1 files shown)


lildude:
Ooo, this is nice 😍. I modified your dirty script slightly to:

```ruby
require "linguist"
require "benchmark"

blob = Linguist::FileBlob.new("lib/linguist/classifier.rb")
languages = [Linguist::Language["Ruby"], Linguist::Language["Python"]]
puts Benchmark.measure { 100000.times { Linguist::Classifier.call(blob, languages) } }
```

... and ran this against `master` and then this branch and the results were:

- **Before**:  58.420874   0.484224  58.905098 ( 62.065967)
- **After**:  34.117573   0.281577  34.399150 ( 36.738602)

That's quite the improvement 👏

smola:
By the way, the trade-off here is that it will consume more memory, but given that the amount of data that the classifier reads from a file is bound, it shouldn't be too bad.

Alhadis (approved):

lildude (approved):

Issue #5012, opened 2020-09-18
Detected language depends on file length or amount of data?


ib:
This seems arbitrary and ridiculous:

`#ifndef FOO_H
#define FOO_H

static const struct foo
{
  int foo1;
  int foo2;
  int foo3;
}
foos[] =
{
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
  { 0x0000, 0x0000, 0 },
};

#endif`

`foo.h: 44 lines (41 sloc)
  type:      Text
  mime type: text/plain
  language:  Objective-C`

and with 1 line of `{ 0x0000, 0x0000, 0 },` less the correct result:

`foo.h: 43 lines (40 sloc)
  type:      Text
  mime type: text/plain
  language:  C`

This should be fixed.

smola:
Yes. Linguist has a set of heuristics to identify a language. When all heuristics fail, it falls back to a Bayesian classifier which performs poorly on short files, and in particular .h files.

There is an open discussion to improve this for C, C++ and Objective-C: https://github.com/github/linguist/issues/1626

ib:
It actually seems to perform worse the longer the file is. The short file is recognized correctly.

[...]

PR #5015, opened 2020-09-20
Add a C sample defining a large initialized array

ib:
This fixes github issue #5012.

  - [X] I have included a new sample for the misclassified language:
    - Sample source(s): `C/foo.h`
    - Sample license(s): file created by me, any license you want


lildude:
🤔 I'm not sure this is really solving the real issue here. The classifier uses real-world samples to try and guess the language based on how it is really used. This contrived sample is being added to deliberately sway the classifier and in turn may start causing legitimate Objective-C or C++ header files (I have no idea if there is such a thing) to be classified as C and thus only add to what is already a tricky problem as detailed in the ongoing discussion in https://github.com/github/linguist/issues/1626

ib:
The sample is based on real-world headers: https://github.com/ib/gucharmap/tree/master/gucharmap: unicode-*.h

lildude:
> The sample is based on real-world headers: https://github.com/ib/gucharmap/tree/master/gucharmap: unicode-*.h

Why not use one of the real-world files?

smola:
The main problem here is that token frequency within documents leads to weird results in classification. For example, if a few more braces appear in C++ files, the classifier will be biased towards C++ when classifying pure C files that have many braces.

I think the right solution is going for the root problem: absolute frequency of tokens is suboptimal, and we should probably move to document frequency. I've been testing this approach and I think it should work decently (and faster). But it still requires some more verification.

lildude:
Closing as this isn't a real world sample and is only being added to improperly influence the classifier. Recent and pending improvements to the classifier will also help towards addressing the original issue.

PR #5018, opened 2020-09-21
Check for generated Pascal _TLB files

JackRainy:
This PR adds one more check for generated files: Pascal _TLB files, including the unit test for it.
Also, the sample for the .pas (primary Pascal extension) is added.

## Description
The Type Library Import tool from Delphi RAD Studio [(see details)](http://docs.embarcadero.com/products/rad_studio/delphiAndcpp2009/HelpUpdate2/EN/html/devwin32/ooctypelibraries_xml.html) is able to generate *_TLB.pas files, which are not meant to be changed after being generated [(see details)](http://docwiki.embarcadero.com/RADStudio/Sydney/en/Code_Generated_When_You_Import_Type_Library_Information).
Thus, this PR makes them to be excluded by default from expanding in the GitHub diffs.

## Checklist:
  - I have included a real-world usage sample for all extensions added in this PR:
  https://github.com/search?q=extension%3Apas+NOT+nothack&type=Code
    - Sample source(s):
      - https://github.com/MartinTrummer/lazarus-com-example/blob/master/LazComDll/lazcomunit.pas
      - https://github.com/MartinTrummer/lazarus-com-example/blob/master/LazComDll/lazcomlib_1_0_tlb.pas
    - Sample license(s): MIT License https://github.com/MartinTrummer/lazarus-com-example/blob/master/LICENSE

- **I am adding new or changing current functionality**
  - I have added or updated the tests for the new or changed functionality.


lib/linguist/generated.rb
```diff
@@ -708,6 +709,16 @@ def generated_jooq?
...
     end
 
+    # Internal: Is this a generated Delphi Interface file for a type library?
+    #
+    # Delphi Type Library Import tool generates *_TLB.pas files based on .ridl files.
+    # They are not meant to be altered by humans.
+    #
+    # Returns true or false
+    def generated_pascal_tlb?
+      name.downcase =~ /_tlb\.pas$/
```
Alhadis, line 719: Shouldn't we match this case-insensitively?

JackRainy: At some point, I thought about that.
In the most cases, the suffix is named with upper cases, however, I found a file (see the link to example source) with lower case. That's why, I decided to switch for both lower and upper cases (as it is now - my assumption here: `.downcase` set all letters in lower case, so I compare with lower case string). Also, the extension itself can be in upper case sometimes.

Alhadis: Sorry, bad choice of words. I meant to ask: _"shouldn't we match this using a case-insensitive regex?"_

The shorter, clearer, and conventional way to do this is to use an `/i` modifier:

~~~ruby
name =~ /_tlb\.pas$/i
~~~

JackRainy: I see now. Yes, I like your suggestion more. Thank you for a good code review.

Alhadis (approved):

lildude (approved):

lib/linguist/generated.rb
```diff
@@ -708,6 +709,16 @@ def generated_jooq?
...
     end
 
+    # Internal: Is this a generated Delphi Interface file for a type library?
+    #
+    # Delphi Type Library Import tool generates *_TLB.pas files based on .ridl files.
+    # They are not meant to be altered by humans.
+    #
+    # Returns true or false
+    def generated_pascal_tlb?
+      name =~ /_tlb\.pas$/i
```
lildude, line 719: 🤦 I completely missed this before, but this doesn't return a boolean which means it'll break the check for generated when the filename does _not_ match as it returns `nil` which is _not_ the same as `false`.

I'll open a PR now to correct this.

lildude: https://github.com/github/linguist/pull/5034

PR #5027, opened 2020-09-29
Remove ocaml from Reason interepreters

sheerun:
It's little stretched to say ocaml is interpreter of ocaml, to say the least...

I'm using  languages.yml for filetype detection of scripts in vim-polyglot and this messes things up. Nobody programming in Reason would set ocaml as interpreter

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4633,8 +4633,6 @@ Reason:
   extensions:
   - ".re"
   - ".rei"
-  interpreters:
-  - ocaml
   tm_scope: source.reason
   language_id: 869538413
 Rebol:


Alhadis (approved):

Alhadis:
Just as a side-note:

> It's little stretched to say ocaml is interpreter of ocaml, to say the least...

What Linguist refers to as an "interpreter" is basically _"anything that runs from `#!/usr/bin/env`"_. Oddly enough, OCaml does:

~~~console
λ cat > script
#!/usr/bin/env ocaml
print_endline "Hello, world!"

λ chmod +x ./script
λ ./script >> script
Hello, world!
~~~

sheerun:
I highly doubt it can interpret any reason code

sheerun:
Besides it seems linguist is using interpreters list to detect languages as well. Because reason has the same syntax as ocaml, your attempts to detect script with `#!/usr/bin/env ocaml` as ocaml are futile, because currently both ocaml and reason match. I think ocaml is better choice here

Alhadis:
> Besides it seems linguist is using interpreters list to detect languages as well

Yes, [that's correct](https://github.com/github/linguist#how-linguist-works).

lildude (approved):

PR #5034, opened 2020-10-01
Ensure generated_pascal_tlb? returns bool

lildude:
I just noticed that the new `generated_pascal_tlb?` method added in https://github.com/github/linguist/pull/5018 doesn't return a boolean which means it'll break the check for generated when the filename does _not_ match as it returns `nil` which is _not_ the same as `false`.

This PR ensures the method returns a bool and switches it to the approach used elsewhere in the file for consistency.

I've also added a test to catch this in future.


JackRainy (commented):
I see the similar problem appears in other functions (e.g. `generated_net_designer_file` and `generated_net_specflow_feature_file`) in this file.
Does it make sense to correct them as well?

lildude:
Oh yes! Good catch @JackRainy.

dbussink (approved):
Makes sense to me!

PR #5035, opened 2020-10-01
Add LabVIEW file extensions and samples

dillonledoux:
## Description

I added the new extensions for LabVIEW code and added real world examples for said code. 

## Checklist:

    - Sample source(s):
      - Samples used are shipped with LabVIEW itself so no URL exists
    - Sample license(s): http://ni.com/samplecodelicense - All code is free to be distributed and used without restriction


lildude (changes requested):
As I mentioned in https://github.com/github/linguist/issues/5033, Linguist can not be used to classify binary files as they are _not_ code and thus not readable. Linguist doesn't even try to analyse them.

Please remove the binary files and the extensions from the list in `language.yml`.

We also only need one sample of each new extension that is a readable text file. Sometimes two samples are useful if they're significantly different but still representative of real-world usage. All of the files with identical extensions in this PR appear to be very similar so only one of each is needed.

Please also update the template to provide a link to each of the new extensions you're adding showing their popularity on GitHub.com.

JT0407 (approved):

JT0407 (approved):

dillonledoux (commented):
I have removed the binary file references so now it should be good

Alhadis (approved):

lildude (approved):

PR #5059, opened 2020-10-22
Mark `.{1..9}` as generic file-extensions

Alhadis:
## Description
Successor of #4258, which makes use of Linguist's new [generic extensions](https://github.com/github/linguist/pull/4936) feature. To recap:

> Many files on GitHub receive an incorrect classification of "Roff", simply because their filenames end in a numeric suffix: 
> 
> ```
> Gemfile-3.0.3
> GFDL-1.2
> AFL-1.2
> mkfontscale-1.1.1
> ```
> 
> Change-logs and version-numbers are particularly problematic examples of these. The [`ValveSoftware/steam-runtime`](https://github.com/ValveSoftware/steam-runtime), for instance, was shown as 72.8% Roff (until [I fixed it](https://github.com/ValveSoftware/steam-runtime/pull/143)), because of [three license files](https://github.com/ValveSoftware/steam-runtime/search?l=roff) with version-numbers. [`haskell-CI/haskell-ci`](https://github.com/haskell-CI/haskell-ci/search?l=roff) was another example (until [fixed](haskell-CI/haskell-ci#223)). You can find those and more by checking GitHub's list of [trending Roff repositories](https://github.com/trending/roff?since=daily).

## Checklist:
- [x] **I am adding new or changing current functionality**
	- [x] I have added or updated the tests for the new or changed functionality.

PR #5070, opened 2020-10-26
Move generic extensions list to a dedicated file

Alhadis:
## Description
This PR moves our nascent list of generic file extensions to its own file, following [approval](https://github.com/github/linguist/pull/5059#issuecomment-716444008) from @lildude in #5059.

### DIFF
--- a/lib/linguist/strategy/extension.rb
+++ b/lib/linguist/strategy/extension.rb
@@ -1,9 +1,7 @@
+require 'yaml'
+
 module Linguist
   module Strategy
-    # Extensions considered too common for simple extension lookup.
-    # Languages which use such an extension must instead satisfy a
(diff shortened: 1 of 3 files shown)


lib/linguist/strategy/extension.rb
```diff
@@ -27,7 +25,17 @@ def self.call(blob, candidates)
 
       # Public: Return true if filename uses a generic extension.
       def self.generic?(filename)
-        GENERIC.any? { |ext| filename.downcase.end_with? ext }
+        self.load()
```
lildude, line 28: Nit: We don't need the `()` here:

```suggestion
        self.load
```

lildude (approved):
One little nit, but looks good to me.

PR #5096, opened 2020-11-25
Added '.comp' extension for GLSL

JonathanWoollett-Light:
Added '.comp' so as to cover all the common extensions and match the khronos.org reference for OpenGL (https://www.khronos.org/opengles/sdk/tools/Reference-Compiler/).
Also reordered extension list to put all the extensions from the reference inorder.

## Checklist:

  - [] The new extension is used in hundreds of repositories on GitHub.com
      
    - Sample source(s):
      - [https://www.khronos.org/opengles/sdk/tools/Reference-Compiler/]
    - Sample license(s):


Alhadis:
> Also reordered extension list to put all the extensions from the reference inorder.

Please don't. The extensions list is supposed to be alphabetised, with the primary (most common) extension being the exception. The order is even checked when running tests (which is causing CI [to fail](https://github.com/github/linguist/pull/5096/checks?check_run_id=1451232076#step:6:67)).

Moreover, the `.comp` extension is fairly generic and likely in-use by a broad variety of formats across GitHub. You also need to provide proof that enough users on GitHub are using GLSL `.comp` files—[`CONTRIBUTING.md`](https://github.com/github/linguist/blob/master/CONTRIBUTING.md) will explain more.

PR #5099, opened 2020-11-30
Add support for Browserslist

voskresla:
## Description
This PR add support for [Browserslist](https://github.com/browserslist/browserslist) configs.

Syntax highlighting based on https://github.com/browserslist/browserslist-vscode.
cc @ai 

## Checklist:
	- [x] **The extension of the new language is used in hundreds of repositories on GitHub.com**
		- `.browserslistrc`: [~360,347 results](https://github.com/search?q=filename%3A.browserslistrc+NOT+nothack&type=code)
		- `browserslist`: [~826,669 results](https://github.com/search?q=filename%3Abrowserslist+NOT+nothack&type=code)
	- [x] **I have included a real-world usage sample for all extensions added in this PR:**
		- Source: [`browserslist/browserslist-vscode`](https://github.com/browserslist/browserslist-vscode/blob/main/browserslist)
		- License: [MIT](https://github.com/browserslist/browserslist-vscode/blob/main/LICENSE)
	- [x] **I have included a syntax highlighting grammar:**
		- [Preview](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fbrowserslist%2Fbrowserslist-vscode%2Fblob%2Fmain%2Fsyntaxes%2Fbrowserslist.tmLanguage.json&grammar_text=&code_source=from-text&code_url=&code=%23+Comment%0D%0Adefaults%0D%0A%0D%0A%23+Not+keyword%0D%0Anot+dead%0D%0A%0D%0A%23+Market+share+with+%25%0D%0A%3E+10%25%0D%0A%3C+1.5%25+in+US%0D%0A%3E%3D+.5%25%0D%0A%0D%0A%23+Rare+queries%0D%0Amaintained+node+versions%0D%0Asupports+es6-module%0D%0A%0D%0A%23+Extend+keyword%0D%0Aextend+%40company%2Fbrowserslist-config%0D%0A%0D%0A%23+Direct+versions%0D%0AChrome+50%0D%0AFirefox+ESR%0D%0ASafari+TP%0D%0AOperaMini+all%0D%0A%0D%0A%23+Block+header%0D%0A%5Bproduction%5D%0D%0A%23+and+keyword%0D%0AEdge+%3E+40+and+Edge+%3C+43%0D%0A%23+or%2F%2C+keywords%0D%0AIE+11+or+IE+10%2C+ie+9)
		- [MIT-licensed](https://github.com/browserslist/browserslist-vscode/blob/main/LICENSE)


lib/linguist/languages.yml
```diff
@@ -6480,6 +6480,15 @@ cURL Config:
   tm_scope: source.curlrc
   ace_mode: text
   language_id: 992375436
+browserslist:
+  type: data
+  aliases:
+  - browserslist
+  filenames:
+  - ".browserslistrc"
```
voskresla, line 6488: **question:** do i need to specify a file `browserlist` without `rc` and dot before filename here ? Or is it enough to specify it in `aliases` ?

ai:
Adding Browserslist config highlighting is important because it is default way to specify target browsers in Autoprefixer and Babel and it is already used in [350K public repositories](https://github.com/search?l=&q=filename%3A.browserslistrc&type=code).

lib/linguist/languages.yml
Alhadis: @voskresla Treat all permutations as separate filenames (provided they meet the usage requirement). That is, like this:

~~~yaml
filenames:
- ".browserslist"
- ".browserslistrc"
- "browserslist"
- "browserslistrc"
~~~

As for `aliases`, they're only used to specify the name of a language in `.gitattributes` files and fenced code-blocks—they have nothing to do with how/what Linguist matches by way of filenames or extensions. For example, [JavaScript has `node`](https://github.com/github/linguist/blob/7c2adbdb15d4efd25d92cd5ed20d0025f0d32d28/lib/linguist/languages.yml#L2593) as an alias, so something like this (normally) receives JS highlighting:

~~~markdown
```node
import {getWindowSize, env, platform} from "./index.mjs";

console.log(getWindowSize());

console.log("gsdgds" in env);
env.gsdgds = 1;
console.log("gsdgds" in env);
```
~~~

The above block isn't showing the highlighted Markdown source, because our Markdown grammar happens to be kinda crap (then again, it's rarely seen or used, so… 🤷‍♂️)

lib/linguist/languages.yml
```diff
@@ -6480,6 +6480,14 @@ cURL Config:
   tm_scope: source.curlrc
   ace_mode: text
   language_id: 992375436
+browserslist:
```
Alhadis: This needs to be capitalised:

~~~diff
-browserslist:
+Browserslist:
~~~

samples/browserslist/.browserslistrc
```diff
@@ -0,0 +1,30 @@
+# Comment
```
Alhadis: The `samples/browserslist` directory also needs to match the capitalisation of the language entry. So with my previous comment in mind, this should be renamed to `samples/Browserslist`.

Alhadis (changes requested):

Alhadis:
Usage checks out (I happen to be familiar with the format, so I'm skipping the usual [sample harvest](http://github.com/Alhadis/Harvester) routine). Just a small nit to correct and we should be good to go. 👍

samples/browserslist/.browserslistrc
voskresla: Screenshot 2020-12-01 at 02 57 18

Hm.. Local test pass, but ci test fail )

lib/linguist/languages.yml
voskresla: ㈽ thx for explanation.

Alhadis (changes requested):

samples/browserslist/.browserslistrc
Alhadis: I assume you're using an OS with a case-insensitive filesystem (macOS or Windows). Our CI tests run on Linux, which uses a case-*sensitive* filesystem. This can cause confusion in version control.

lib/linguist/languages.yml
```diff
@@ -550,6 +550,14 @@ Brightscript:
   tm_scope: source.brightscript
   ace_mode: text
   language_id: 39
+Browserslist:
+  type: data
```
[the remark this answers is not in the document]
voskresla: In `language.yml` exists comments about color:
`CSS hex color to represent the language. Only used if type is "programming" or "markup".`

Our type is 'data'. Is `data` not right choise?

lib/linguist/languages.yml
Alhadis: > ㈽

[What the](https://graphemica.com/%E3%88%BD)… 😀

lib/linguist/languages.yml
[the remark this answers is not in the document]
voskresla: thx for this: `git config --local core.ignorecase false`!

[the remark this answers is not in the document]
Alhadis: That explanation is dated and needs to be revised. Formerly, colours were only used in language stat-bars (where only `programming` and `markup`-type languages contributed towards statistics). However, the addition of the [`linguist-detectable`](https://github.com/github/linguist#detectable) attribute and various design-related tweaks, language colours are popping up everywhere. So the importance of the `type` field is no longer relevant.

Alhadis (approved):

Alhadis:
@voskresla You might want to set `core.ignorecase` globally so it's configured for all future projects:

~~~console
$ git config --global core.ignorecase false
~~~

Or add to your `~/.gitconfig` file:

~~~gitconfig
[core]
	ignorecase = false
~~~

The `--local` switch sets a config setting for one particular repository directory only (stored in `.git/config`). So it won't persist across other projects, including future checkouts of Linguist.

voskresla:
)

Alhadis:
LGTM. 👍

@lildude, I say we :shipit:.

wspwebben:
@ai [mentioned](https://github.com/browserslist/browserslist-vscode/pull/5) `browserslist` config name is deprecated, it makes sense to drop it from samples? We should keep support for existing configs, so it's only about example file.

Alhadis:
@wspwebben We don't judge formats on their deprecation– or recommendation-status, only their real-world distribution. Discouraging the use of a filename won't change the fact there are [over a million files on GitHub](https://github.com/search?q=filename%3Abrowserslist+NOT+nothack&type=code) using that particular format and filename.

wspwebben:
> @wspwebben We don't judge formats on their deprecation– or recommendation-status, only their real-world distribution. Discouraging the use of a filename won't change the fact there are [over a million files on GitHub](https://github.com/search?q=filename%3Abrowserslist+NOT+nothack&type=code) using that particular format and filename.

That's why we should keep support for that format in config files, so syntax highlight would still work for these files, but there's no need to keep that format as an example in samples

Alhadis:
> but there's no need to keep that format as an example in samples

There is. The `samples` directory isn't for human consumption—it's to train our [Bayesian classifier](https://github.com/github/linguist/blob/master/lib/linguist/classifier.rb), which disambiguates files based on the frequency of certain tokens. The more samples Linguist has, the more accurate its predictions are.

wspwebben:
Got it, thank you for clarification :D

Alhadis:
No problem! 😉 That's also the reason we prefer "real-world" samples instead of just _"Hello, world"_ examples, BTW:—they need to be an accurate rendition of the sort of files Linguist will encounter in-the-wild after it starts indexing the new language.

lildude (approved):
LGTM.

This won't go live on GitHub.com until a new release is made which I won't be able to do until mid-to-late January at the earliest.

PR #3753, opened 2017-08-03
Distinction between Component Pascal language and Pascal language and…

leogregianin:
Change "delphi" and "objectpascal" aliases from the "Component Pascal" language to the "Pascal" language because [Component Pascal language](https://en.wikipedia.org/wiki/Component_Pascal) has no relation to Pascal language and their dialects.

[Pascal language](https://en.wikipedia.org/wiki/Pascal_(programming_language)) is the main language of dialects like "Object Pascal", "Oxygene" and "Delphi".

Delphi is a registered product of [Embarcadero Technologies](https://www.embarcadero.com/products/delphi).

Issue #905, opened 2014-01-23
Delphi code misidentified as Pascal


Faq:
Seems all delphi projects show as pascal now
 Example:
https://github.com/Faq/TXTGenerator
https://github.com/Gurux/Gurux.DLMS.Delphi

arfon:
We have a couple of different ways to override language detection now in Linguist which is probably as good as we can do right now. Please take a look at these over here: https://github.com/github/linguist#overrides

EarlGlynn:
I really do not understand this thread.  Why are there summaries by language of GitHub repositories when you refuse to put the right label on the repository?  You even refuse to let the authors put the right labels on the repository.  This "resolution" simply does not make any sense to me.

nunopicado:
Still nothing on this?
arfon's suggestion was tried, but does not work.
If I add 

> *.pas linguist-language=Delphi

to the .gitattributes file, the language changes from Pascal to Component Pascal, which is another completely different language, which is neither Pascal nor Object Pascal/Delphi.

It's funny how people use github statistics to argument that Delphi does not show in the Top languages used today, when github doesn't even recognize the existence of the language.

Would it be so hard to make it work correctly?
I don't mind having to add a .gitattributes file, if it would only work!

arfon:
@nunopicado - this is because `delphi` is [listed as an alias](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L658) for Component Pascal.

Probably the best thing we can do right now is to list Dephi as a language in [`languages.yml`](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) but not add any extensions as we still don't have a way to reliably identify a Delphi project on a _per-file basis_.

Doing this would allow the overrides to work (`*.pas linguist-language=Delphi`) but little else :-\

nunopicado:
Thanks arfon, for your reply.

I do think Delphi should be separated from Component Pascal, the same for Object Pascal.
Delphi and Object Pascal could be marked as alias to one another (one as a language, one as an alias, I don't really mind which is which). Even though there are different flavours of Object Pascal, at least it fits in the description.

There are some file extensions which are exclusive to Delphi. Those could be added.
The problem is with .pas files, which is a common extension.

It was mentioned that there must be something in the code to differentiate standard Pascal from Delphi/Object Pascal. Well, I guess there is.

Standard Pascal is not object oriented, so there are no classes.
[...]

PR #5121, opened 2020-12-25
Fix Pascal aliases

smola:
## Description
Move `delphi` and `objectpascal` aliases to Pascal instead of Component Pascal. Both aliases are used for Object Pascal dialects, which is a Pascal extension, not derived from Component Pascal.

This was discussed in the past (https://github.com/github/linguist/issues/905#issuecomment-216281058 https://github.com/github/linguist/pull/3753). Object Pascal could be introduced as a standalone language, although it would present problems similar to C/C++ distinction. In the absence of a standalone language for Object Pascal, these aliases belong to Pascal, not Component Pascal.

## Checklist:

- [x] **I am fixing a misclassified language** (using a delphi or objectpascal modeline makes linguist to identify the file as Component Pascal)
  - ~I have included a new sample for the misclassified language:~ (probably not required?)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -971,9 +971,6 @@ Component Pascal:
   - ".cp"
   - ".cps"
   tm_scope: source.pascal
-  aliases:
-  - delphi
-  - objectpascal
   ace_mode: pascal
   codemirror_mode: pascal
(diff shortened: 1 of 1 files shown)


lildude (approved):
LGTM

PR #5132, opened 2021-01-08
Shallow-clone modules to optimise bootstrapping process

Alhadis:
## Description
This is an effort to improve the efficiency of our [bootstrap script](https://github.com/github/linguist/blob/47b785a8fda5bf1677bd49d048a1a67e47f2af5a/script/bootstrap), which currently clones every submodule in its entirety. Since we only care about the most recent changes made upstream, we can save on time, bandwidth and disk-space by fetching only the commit we care about. [Here's an error log](https://pastebin.com/raw/LPsm1neb) I recorded this morning before starting work on this PR—notice that all the failed transfers are of fairly large repositories.

I also axed the `fast-submodule-update` script in favour of using `git submodule --jobs` to enable concurrent downloading.

## Known issues
*	__Not every server supports or allows shallow-cloning at arbitrary points in a repo's history.__
	 Currently, this is only an issue for Bitbucket, which hosts only [1 grammar](https://bitbucket.org/bitlang/sublime_cobol) used by Linguist.
*	__Some submodules are cloned with empty working trees.__

/cc @lildude


lildude:
Nice. Have you got timing figures to show the difference?

Also, shallow clones hammer the upstream servers. How do things fair if you use the kinder `git clone --filter=tree:0 <url>` for each submodule?  (See https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/)

script/zap-submodules
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -e
+
+cd "${0%/*}/.."
+rm -rf vendor/grammars vendor/CodeMirror .git/modules
+{ sed -n < .git/config 's/^\[submodule *"\([^"]\{1,\}\)"\]/submodule.\1/p' \
+| xargs -n1 git config -f .git/config --remove-section
+} 2>&1 >/dev/null || :
+
+git checkout -- vendor
```
lildude, line 10: What's this script for and any reason we couldn't use `git rm <submodule path> && git commit` or similar instead?

Alhadis: It's to reset one's working tree to the same state as a freshly checked-out copy of Linguist (save for the fact there's no need to rerun `bundle install`). Basically, it ensures there are no lingering remnants of an earlier submodule install, as submodules tend to get backed up to `.git/modules` when first cloned.

This script was used to test bootstrap timings, but it may also be helpful to users whose working directory has been left in an inconsistent state after botching an install or a particularly messy rebase.

Alhadis:
I neglected to save the most recent installation log, but it took only 13 minutes (as opposed to 278).

I wasn't aware that fetching from an unshallow clone had a performance impact on upstream servers, but it's important to note that the options for `git submodule` are much more limited than they are for `git clone`. In particular, blobless clones and sparse checkouts aren't possible with submodules... short of switching to each submodule directory prior to initialisation and running `git fetch` and what-have-you. Which was actually my initial intended approach before deciding `--depth 1` was a great deal less complex.

lildude:
I've done a few quick tests, each involving completely deleting and re-cloning the repo, and checked the times in the CI builds and I'm not sure this is saving peeps any time. For me this branch is significantly slower and more verbose than the current approach.

Full bootstrap:

```console
$ time script/bootstrap
[...]
Executed in  298.32 secs   fish           external
   usr time  140.18 secs  185.00 micros  140.18 secs
   sys time  371.61 secs  1155.00 micros  371.61 secs

$
```

New shallow clone full bootstrap:

```console
$ time script/bootstrap
[...]
Executed in  582.61 secs   fish           external
   usr time  101.03 secs  145.00 micros  101.03 secs
   sys time  247.49 secs  1122.00 micros  247.49 secs

$
```

One difference could be because the current method uses 8 threads whilst this PR only uses four jobs, so I changed that to 8 and the time didn't really improve much:

```console
Executed in  527.81 secs   fish           external
   usr time  100.31 secs  158.00 micros  100.31 secs
   sys time  251.41 secs  1156.00 micros  251.41 secs
```

I think the added workaround step you've implemented to catch the corner cases as referenced in your TODO comment could be adding to this time too.

CI also isn't showing much of an improvement, if any, and seems to regularly hit a slower build, due to the submodules step, not commonly seen on master or other branches, eg https://github.com/github/linguist/actions/runs/471939106 and https://github.com/github/linguist/actions/runs/471728507 both took over 5mins. I've rerun all the jobs on this PR and almost always see an excessively long build due to the submodules cloning taking many minutes. I wouldn't be surprised if this is due to upsteam load caused by the shallow clones.

Alhadis:
> For me (on my idle MBP) this branch is significantly slower and more verbose than the current approach.

In that case, I'll chalk my installation difficulties up to my shitty Internet connection. I'd explore the possibility of manually cloning each submodule via blobless checkouts, but I don't want to add technical debt over something that's only affecting a handful of users.

I guess I'll close this now.

PR #5142, opened 2021-01-14
Classifier cross-validation script and workflow

smola:
## Description
Added `script/cross-validation` to run leave-one-out cross-validation on the classifier. It will output a list of validated samples along with `GOOD` for successful samples and `BAD (<Language>)` for errors.

By the default it will check only samples with ambiguous filenames and without catch-all heuristics. This behavior can be disabled with the `--all` flag. Languages without, at least, 2 samples, are always ignored.

If the `--test` flag is used, the number of errors will be checked against a hardcoded threshold, and it will exit with 1 status if the number of errors is not less or equal than this threshold.

A new GitHub Actions job is added to run this on every push.

A small optimization was made to Classifier#train!. It speeds it up by ~5x. This will not be a big deal for most users, but it makes a big difference when doing cross-validation.

## Checklist:
_N/A_


script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+STDERR.sync = true
+
+def print_usage(out)
+  out.puts "Usage: #{$PROGRAM_NAME} [--all] [--test]"
+  out.puts ""
+  out.puts "Performs leave-one-out cross-validation of the classifier."
+  out.puts "By default, outputs results only for samples with ambiguous extensions."
+  out.puts "If the --all flag is given, all samples and languages are considered."
+  out.puts ""
+  out.puts "The --test flag can be used to verify that the number of errors is acceptable."
```
lildude, line 22: Quoting consistency:

```suggestion
  out.puts ''
  out.puts 'Performs leave-one-out cross-validation of the classifier.'
  out.puts 'By default, outputs results only for samples with ambiguous extensions.'
  out.puts 'If the --all flag is given, all samples and languages are considered.'
  out.puts ''
  out.puts 'The --test flag can be used to verify that the number of errors is acceptable.'
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+$test = false
+ARGV.each do |arg|
+  case arg
+  when '--all'
+    $all = true
+  when '--test'
+    $test = true
+  else
+    STDERR.puts "Invalid command line argument: ${arg}"
+    STDERR.puts ""
```
lildude, line 40: Quotes:

```suggestion
    STDERR.puts ''
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+$all = false
+$test = false
+ARGV.each do |arg|
+  case arg
+  when '--all'
+    $all = true
+  when '--test'
+    $test = true
+  else
+    STDERR.puts "Invalid command line argument: ${arg}"
```
lildude, line 39: Whoopsie:

```suggestion
    STDERR.puts "Invalid command line argument: #{arg}"
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+  # Train the classifier.
+  db = {}
+  train_samples = $samples.select { |s| s[:path] != sample[:path] }
+  train_samples.each do |train_sample|
+    Classifier.train!(db, train_sample[:language], train_sample[:tokens])
+  end
+
+  # Get result.
+  results = Classifier.classify(db, sample[:data], languages)
+  path = sample[:path].sub("#{Samples::ROOT}/", "")
```
lildude, line 112: Quotes:

```suggestion
  path = sample[:path].sub("#{Samples::ROOT}/", '')
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+end
+
+if $test
+  if $all
+    acceptable_errors = ACCEPTABLE_ERRORS_ALL
+  else
+    acceptable_errors = ACCEPTABLE_ERRORS
+  end
+  STDOUT.flush
+  STDERR.puts ""
```
lildude, line 141: Quotes: 

```suggestion
  STDERR.puts ''
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+
+  # Test only languages with at least 2 samples.
+  n_samples = 0
+  $samples.each do |other_sample|
+    if other_sample[:language] == sample[:language]
+      n_samples += 1
+    end
+  end
+  if n_samples <= 1
+    STDERR.puts "WARNING: No cross-validation possible for #{sample[:language]} (only one sample)"
```
lildude, line 99: As this is just a warning, lets use `warn` instead:

```suggestion
    warn "WARNING: No cross-validation possible for #{sample[:language]} (only one sample)"
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+  else
+    acceptable_errors = ACCEPTABLE_ERRORS
+  end
+  STDOUT.flush
+  STDERR.puts ""
+  if total_errors > acceptable_errors
+    STDERR.puts "Number of errors (#{total_errors}) is above the acceptable threshold (#{acceptable_errors})"
+    exit 1
+  else
+    STDERR.puts "Number of errors (#{total_errors}) is within the acceptable threshold (#{acceptable_errors})"
```
lildude, line 146: We probably don't want this on `STDERR` as this is a "success":

```suggestion
    STDOUT.puts "Number of errors (#{total_errors}) is within the acceptable threshold (#{acceptable_errors})"
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+Samples.each do |sample|
+  sample[:data] = File.read(sample[:path])
+  sample[:tokens] = Tokenizer.tokenize(sample[:data])
+  $samples << sample
+end
+
+def eval(sample)
+  if $skip_extensions.include? sample[:extname]
+    return nil
+  end
```
lildude, line 73: I think we can on-liner this guard:

```suggestion
  return nil if $skip_extensions.include? sample[:extname]
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+  # If --all is set, use all languages. Otherwise, get only languages that are
+  # ambiguous in terms of filename and extension.
+  if $all
+    # nil languages is the same as all languages.
+    languages = nil
+  else
+    languages = Language.find_by_filename(sample[:path]).map(&:name)
+    if languages.length == 1
+      return nil
+    end
```
lildude, line 84: Here too:

```suggestion
    return nil if languages.length == 1
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+    languages = Language.find_by_extension(sample[:path]).map(&:name)
+    if languages.length <= 1
+      return nil
+    end
```
lildude, line 88: Here too:

```suggestion
    return nil if languages.length <= 1
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+results = Parallel.map($samples) { |sample| eval(sample) }
+results.reject! { |s| s.nil? }
+results.sort!
+
+total_errors = 0
+results.each do |res|
+  STDOUT.puts res
+  if res.include? ' BAD '
+    total_errors += 1
+  end
```
lildude, line 131:
```suggestion
  total_errors += 1 if res.include? ' BAD '
```

script/cross-validation
```diff
@@ -0,0 +1,149 @@
...
+end
+
+if $test
+  if $all
+    acceptable_errors = ACCEPTABLE_ERRORS_ALL
+  else
+    acceptable_errors = ACCEPTABLE_ERRORS
+  end
```
lildude, line 139:
```suggestion
  acceptable_errors = $all ? ACCEPTABLE_ERRORS_ALL : ACCEPTABLE_ERRORS
```

lildude (changes requested):
I like this addition and the fact we'll soon be testing the classifier.

Spotted one typo and offered a few style improvements.

script/cross-validation
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env ruby
+
+# Number of acceptable classification errors.
+# It should only be decreased.
+ACCEPTABLE_ERRORS = 61
```
smola, line 5:
```suggestion
ACCEPTABLE_ERRORS = 53
```

lildude (approved):
LGTM. Thanks for this. 😍

PR #5143, opened 2021-01-14
Add Pascal heuristic for .inc extension

### LINKED ISSUE #5115: Delphi file reported as PHP
The following Delphi/Pascal file is being reported a PHP in GitHub
https://github.com/EricGrange/cefHtmlSnapshot/blob/main/cef.inc

## Problem Description

The following Delphi/Pascal file is being reported a PHP in GitHub as seen in
https://github.com/EricGrange/cefHtmlSnapshot/search?l=php

https://github.com/EricGrange/cefHtmlSnapshot/blob/main/cef.inc

### Expected language:
Delphi or Pascal

### Detected language:
PHP

smola:
## Description
Detects some common constructs frequently used in modern Pascal dialects.

Fixes #5115

Thanks to @EricGrange for the hints.

## Checklist:
    - Sample source(s): https://github.com/mike-lischke/GraphicEx/blob/master/GraphicConfiguration.inc
    - Sample license(s): MPL-2.0 https://github.com/mike-lischke/GraphicEx/blob/master/LICENSE


test/test_heuristics.rb
```diff
@@ -439,20 +439,31 @@ def test_ice_by_heuristics
...
+      "#{samples_path}/Pascal/libc.inc",
+      "#{samples_path}/Pascal/program.dpr",
+      "#{samples_path}/Pascal/tw27294.pp",
+      "#{samples_path}/Pascal/large.pp",
+      "#{samples_path}/Pascal/vmops_impl.inc"
+    ]
     assert_heuristics({
       "Motorola 68K Assembly" => all_fixtures("Motorola 68K Assembly", "*.inc"),
       "NASL" => all_fixtures("NASL", "*.inc"),
+      "Pascal" => all_fixtures("Pascal", "*") - ambiguous_pascal,
```
lildude, line 455: This doesn't feel right to me. We're pulling in all Pascal samples, including 3 `.inc` files, only to then remove _all_ but one: `samples/Pascal/GraphicConfiguration.inc`.

This suggests to me the heuristic has scope for improvement. 

Could the heuristic be improved to catch all of the samples we've already got? Maybe matching on the use of `begin` and `end`?

EricGrange: In case that helps, `begin` and `end` would not always be present in .inc files, only through comments or in some FreePascal source, the are however almost present in .pas/.pp/.dpr
`$define` can be present in all Pascal file types.

lildude (changes requested):
I think we need to improve the heuristic as per my comment at https://github.com/github/linguist/pull/5143/files#r563629307

test/test_heuristics.rb
smola: Right. `$define` was already covered. I added an additional pattern that covers all Pascal samples. Also increased the number of samples tested for other languages (not just those originally with `.inc`) to increase coverage of this heuristic.

All of them are fine except C++, which is sometimes classified as NASL, but I'll address that in a separate PR.

lildude (approved):
Thanks @smola. 🙇

PR #5148, opened 2021-01-15
Extend vendor regex for /[Ee]xtern(als?)?/

ligfx:
Extends the vendor regex for `/extern(al)?/` to `/[Ee]xtern(als?)?/`

As used in projects such as [Dolphin](https://github.com/dolphin-emu/dolphin/tree/master/Externals).

## Checklist:


Nixinova:
How about allowing "ext/" as a vendor folder as well? So `/Ee]xt(ern(als?)?)?/`

ligfx:
> How about allowing "ext/" as a vendor folder as well? So `/[Ee]xt(ern(als?)?)?/`

Do you have an example of where that's used?

Nixinova:
>Do you have an example of where that's used?

[Looks like lots of places](https://github.com/search?q=path%3Aext&type=Code), at a glance

ligfx:
> > > How about allowing "ext/" as a vendor folder as well? So /[Ee]xt(ern(als?)?)?/
> >
> > Do you have an example of where that's used?
>
> Looks like lots of places, at a glance

Looks like Ruby uses `ext` as the standard directory for C extensions it will build as part of the project.

I'm going to keep this PR as-is just expanding the `extern(al)?` regex.

Alhadis (approved):

lildude (approved):

PR #5149, opened 2021-01-15
Use JSON instead of YAML for deserialization if Yajl is not available.

tiegz:
Use JSON as deserializer if Yajl gem isn't available.

## Description

History:

* 2013: https://github.com/github/linguist/pull/806  and https://github.com/github/linguist/pull/809 - JSON was added as preferred deserializer
* 2014: https://github.com/github/linguist/pull/1608 - Yajl replaced JSON as a deserializer since it was a development dependency.

The problem is that many repos don't require Yajl, so they fallback to YAML now instead of JSON, which is slower at parsing Linguist's JSON files.

I think JSON was added to Ruby's stdlib [in 1.9.0](https://github.com/ruby/ruby/commit/af1c4167287b9353fec766f932fe4afe97116ad4#diff-0fb7bfe8225acec8067453e88dfd455c166d1822d81c886bb91978ced92d9538), so it should be safe to assume JSON will always be available (?). 

This change falls back to JSON instead of YAML. In my app that doesn't have Yajl, this sped up Rails boot time by ~2 seconds.


lildude:
From a quick glance this makes sense to me, though I'll need to test it to be sure the change doesn't conflict with some GitHub usage (I don't think it will). I'll try test this next week.

> This change falls back to JSON instead of YAML. In my app that doesn't have Yajl, this sped up Rails boot time by ~2 seconds.

Out of curiosity, what effect would using Yajl in your project have on the Rails boot time?

tiegz:
> Out of curiosity, what effect would using Yajl in your project have on the Rails boot time, ie JSON vs Yajl?

When the gem is required, the "samples.json" file is loaded and parsed via the Linguist module:

https://github.com/github/linguist/blob/9edd34988273d307180a1bfc4dfa30a9f2f79105/lib/linguist/language.rb#L498

The improvement here is that the change uses a JSON parser instead of a YAML parser to parse the JSON. That JSON could be parsed either with the `json` library or the `yajl` gem (depends on if the dependent app requires `yajl`), but either one is faster than using the `yaml` library:

```
YAML.load(File.read("lib/linguist/samples.json"))
# => 2.359655

Yajl.load(File.read("lib/linguist/samples.json"))
# => 0.212212

JSON.load(File.read("lib/linguist/samples.json"))
# => 0.173794
```

I also changed the code that loads the languages.{yml,json} files in the Linguist module to be consistent, but that change didn't make much of a speed difference.

tiegz:
Sorry, just realized that I read your question the wrong way 😆 Could've been a shorter answer: it's generally about the same using Yajl vs JSON.

lildude (approved):
I tested this today and as expected, it worked a treat and didn't cause any issues.

PR #5167, opened 2021-01-23
Split Nunjucks into its own language

Nixinova:
This is to be consistent with Liquid and other templating languages (#4979)

<del>Keep the Django syntax highlight references since they work</del>
<ins>Using [alohaas/language-nunjucks](https://github.com/alohaas/language-nunjucks/blob/master/grammars/html%20(nunjucks%20templates).cson) grammar</ins>

Color from official docs: https://mozilla.github.io/nunjucks/templating.html

Related to discussion at #5150


Nixinova:
I've been editing using Github.com because cloning this repo crashes after a bit, but contributing.md says to generate a language_ID with a script, so should I just put a random number in manually?

lildude:
> so should I just put a random number in manually?

Nope. The ID isn't random. It's a SHA digest generated based on the name of the language as it appears in the `languages.yml` file.

You can generate the ID yourself by running this manually in `irb`:

https://github.com/github/linguist/blob/103c43545b9a84cabb3dabc59e3cb9b3cca72e9e/script/update-ids#L9

```console
$ irb
irb(main):001:0> require 'digest'
=> true
irb(main):002:0> Digest::SHA256.hexdigest("Nunjucks").to_i(16) % (2**30 - 1)
=> 461856962
irb(main):003:0>
```

Or copy and paste the ID ☝️ 😄 

> I've been editing using Github.com because cloning this repo crashes after a bit

I hit problems like this with big repos when travelling and using poor conections in the past and found the following solved my problems most of the time:

1. Don't clone recursively. `script/bootstrap` will take care of downloading the submodules.
2. Use SSH
3. Create a `.ssh/config` entry for github.com with at least these params:
```
Host github.com
  Compression yes
  KeepAlive yes
  ServerAliveInterval 60
```

It's the last two that I've found most helpful as they help keep the connection alive.

lib/linguist/languages.yml
```diff
@@ -2074,14 +2074,11 @@ HTML+Django:
...
   - ".j2"
   - ".jinja2"
-  - ".njk"
   aliases:
   - django
   - html+django/jinja
   - html+jinja
   - htmldjango
-  - njk
-  - nunjucks
```
lildude, line 2084: Don't lose me in the trip down south 😁

lildude: You've still forgotten to include me in the new section 😉

Nixinova: Which bit? line 2084? Aren't lowercase forms already automatically aliases?

lildude: 🤦 I forgot about that. Ignore me 😊

lildude (approved):

Alhadis (approved):

Nixinova:
I have found a Nunjucks-HTML-[specific grammar](https://github.com/alohaas/language-nunjucks/blob/master/grammars/html%20(nunjucks%20templates).cson) but using Lightshow it [doesn't seem to work properly](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Falohaas%2Flanguage-nunjucks%2Fblob%2Fmaster%2Fgrammars%2Fhtml%2520%28nunjucks%2520templates%29.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2FNixinova%2Fwebsite%2Ffad13efba3ee6d2822bc52f66b787feaf8951602%2Fassets%2Fimports%2Flayouts%2Fbase.njk&code=%7B%25-+for+item+in+collections.all+%25%7D%0D%0A+%7B%25-+if+not+item.data.sitemapIgnore+%25%7D%0D%0A++%3Curl%3E%0D%0A++++%3Cloc%3Ehttps%3A%2F%2Fnixinova.com%7B%7B+item.data.permalink+%7C+replace%28%27.html%27%2C+%27%27%29+%7D%7D%3C%2Floc%3E%0D%0A++++%3Cchangefreq%3Ealways%3C%2Fchangefreq%3E%0D%0A++++%7B%25+if+item.data.date+%25%7D%3Clastmod%3E%7B%7B+item.data.date+%7D%7D%3C%2Flastmod%3E%7B%25+endif+%25%7D%0D%0A++%3C%2Furl%3E%0D%0A+%7B%25-+endif+%25%7D%0D%0A%7B%25-+endfor+%25%7D). The two grammars in the repo, Nunjucks only and both, each work separately with Linguist but the combined file doesn't load the Nunjucks syntax. Is this an issue with the grammar or is it a limitation of Lightshow?

lildude:
> Is this an issue with the grammar or is it a limitation of Lightshow?

It's a limitation of Lightshow as it doesn't (and can't know) where to pull in the second file sourced from the `html (nunjucks templates).cson` grammar.

Nixinova:
So it should work fine if I switch the grammar to this one?

Nixinova:
Switch the grammar

lildude:
> So it should work fine if I switch the grammar to this one?

Yup. The grammar highlighter knows how to follow the links within grammars as it knows how to find all the sources detailed in the `grammars.yml` file.

grammars.yml
```diff
@@ -554,6 +554,8 @@ vendor/grammars/language-ncl:
 - source.ncl
 vendor/grammars/language-ninja:
 - source.ninja
+vendor/grammars/language-nunjucks:
+- text.html.nunjucks
```
lildude, line 558: We're missing the `source.nunjucks` entry that is also provided by the grammar. You're going to need to run `script/grammar-compiler` or manually add the other source else the syntax highlighter won't be able to find it.

Nixinova: So just that one line needs to be added?

lildude: Yup.

PR #5183, opened 2021-02-05
Ungroup Mustache from HTML

Nixinova:
Ungroups Mustache from HTML for consistency with all other templating languages per #4979.

Set colour to black as that is the only color used in the logo ([site](http://mustache.github.io/)).


Alhadis:
> Set colour to black as that is the only color used in the logo ([site](http://mustache.github.io/)).

Try `#724b3b` instead. I'm not sure I could that a logo—it's merely the Georgia font's `}` glyph rotated 90º with CSS  and presumably made `#000` to replace the `#333` inherited from the `<body>` element:

Figure 1

(I could ramble away about Gestalt's importance in graphic design, or how contrast is the device "black" logos use to establish visual identity, but I think you get the point…)

Nixinova:
Where is `#724b3b` taken from?

Alhadis:
> Where is `#724b3b` taken from?

[Google images](https://www.google.com/search?q=moustache&source=lnms&tbm=isch&sa=X&biw=1792&bih=896). Specifically, some random pixel buried in this bloke's moustache:



(Okay, so it wasn't random. I *did* adjust the saturation and intensity so it was more vibrant…)

Alhadis (approved):

lildude (approved):
LGTM

PR #5184, opened 2021-02-05
Fix misclassified .rs XML files

### LINKED ISSUE #5014: XML files from Katalon Studio (.rs extension) are misidentified
## Preliminary Steps

Please confirm you have...

## Problem Description

XML files from Katalon Studio (.rs extension) are misidentified as Rust or RenderScript.

I did not open a PR directly because I didn't find a sample with an adequate license. We might need to add a synthetic sample.

https://github.com/search?p=1&q=WebElementEntity+extension%3Ars&type=Code

### Last modified on:
2020-09-19

### Expected language:

XML

### Detected language:

Rust / RenderScript

Nixinova:
Classifies .rs as an XML file extension.

Fixes #5014.

This is used for two other languages so I have included a heuristic to match the XML header.

Added a contrived sample for a(n) .rs file; no license.

Consider the checkboxes ticked.


lib/linguist/heuristics.yml
```diff
@@ -476,6 +476,8 @@ disambiguations:
     pattern: '^(use |fn |mod |pub |macro_rules|impl|#!?\[)'
   - language: RenderScript
     pattern: '#include|#pragma\s+(rs|version)|__attribute__'
+  - language: XML
+    pattern: '<\?xml version="[\d.]+" encoding="[\w\d-]+"\?>'
```
Alhadis, line 480: Matching `^<\?xml` should be enough. The [XML specification](https://www.w3.org/TR/REC-xml/#sec-prolog-dtd) only mandates that `<?xml` be used; accurately matching the rest of the header is more complicated than the sample might imply. We can't assume future versions of Katalon Studio will stay consistent about the order or casing of attributes (or even their inclusion, since the `encoding` attribute isn't actually mandatory).

Alhadis (changes requested):

Alhadis (approved):

lildude (approved):

PR #5187, opened 2021-02-05
Ungroup EJS from HTML

Nixinova:
Ungroups EJS from HTML. Continuation of #4979 and similar PRs.

Colour already in place.

EJS is more similar to HTML and is comparable to Nunjucks (#5167) or Liquid's splits. (It's basically inline JS inside HTML files that returns HTML content.)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1298,7 +1298,6 @@ ECLiPSe:
 EJS:
   type: markup
   color: "#a91e50"
-  group: HTML
   extensions:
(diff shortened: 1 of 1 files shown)


Alhadis:
What's EJS?

Nixinova:
Updated just as you commented, lol

Alhadis (approved):

Nixinova:
Didn't know renaming a branch closed the PR...

lildude (approved):

PR #5188, opened 2021-02-05
Ungroup Closure Templates from HTML

Nixinova:
Ungroups Closure Templates from HTML. Continuation of #4979 and similar PRs. Should be the last HTML ungrouping, I think.

This is a language by Google ([repo](https://github.com/google/closure-templates)) that treats HTML as JS-like, it seems.

[Example file](https://github.com/google/closure-templates/blob/master/examples/features.soy); note that the syntax highlighting of inline comments seems broken.

Color `#0d948f` from official site theme: https://developers.google.com/closure

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -852,7 +852,7 @@ Clojure:
   language_id: 62
 Closure Templates:
   type: markup
-  group: HTML
+  color: "#0d948f"
   ace_mode: soy_template
   codemirror_mode: soy
   codemirror_mime_type: text/x-soy


Alhadis (approved):

Nixinova:
Oops, does renaming close stuff?...

Alhadis:
> Oops, does renaming close stuff?...

What are you using as your Git client?

Nixinova:
> What are you using as your Git client?

GitHub dot com

Alhadis:
Oh, you're using GitHub's in-browser editor. Well in that case, you should raise this issue with the [site's support team](https://github.com/contact).

I think I've only edited files on GitHub.com once or twice in my life, so I can't be of much help than that. 😁

lildude (approved):

PR #5192, opened 2021-02-06
Split StringTemplate from HTML

### LINKED ISSUE #4789: Smalltalk files are misidentified as HTML
## Preliminary Steps

Please confirm you have...
this seems to be a general problem of many Smalltalk repositories generated by [FileTree](https://github.com/dalehenrich/filetree).

## Problem Description

Same as #2517, but that has already been closed some years ago.

Without specific knowledge of the linguist implementation, here are some patterns that might indicate Smalltalk instead of HTML:
- The file is not well-formed, so there is no single root tag.
- (Specific to [FileTree](https://github.com/dalehenrich/filetree)-built repositories: The `.st` file paths follow the pattern: `**/<somePackage>.package/<someClass>.(class|trait|extension)/(instance|class)/*.st`)

https://github.com/LinqLover/XmasDecorations/search?l=HTML

Absolutely nothing in this repository should be identified as HTML.

### Last modified on:
2019-11-29

### Expected language:
Smalltalk

### Detected language:
HTML

Nixinova:
https://github.com/github/linguist/issues/4789#issuecomment-774402489: [StringTemplate](https://github.com/antlr/stringtemplate4) was added in #1117 as a raw HTML variant, but per #4979 HTML variants should be split.

It seems to be HTML but with control statements and variables inside $dollars$.

<s>Default to using default HTML grammar, which is not great as it doesn't match the dollar syntax at all, but couldn't find an actual one <ins>that works</ins>.</s>
<s>Found a grammar: [dangmarm/stringtemplate-lang](https://github.com/dangmarm/stringtemplate-lang/blob/master/syntaxes/stringtemplate.tmLanguage.json)</s>
**UPDATE (Alhadis):** Added grammar for StringTemplates that use either [`<…>` delimiters](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-etc%2Fblob%2Fmaster%2Fgrammars%2Fstring-template.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-etc%2Fblob%2Fmaster%2Fsamples%2Ftemplate.default.st&code=) or [`$…$` delimiters](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=cson&grammar_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-etc%2Fblob%2Fmaster%2Fgrammars%2Fstring-template.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2FAlhadis%2Flanguage-etc%2Fblob%2Fmaster%2Fsamples%2Ftemplate.html.st&code=).

Color `#3fb34f` from [logo](https://www.stringtemplate.org/images/st-logo.png).

Closes #4789


Nixinova:
Ah, found a grammar: [dangmarm/stringtemplate-lang](https://github.com/dangmarm/stringtemplate-lang/blob/master/syntaxes/stringtemplate.tmLanguage.json)

Nixinova:
Can't CTRL+F the logs but it's the same issue as #5194's failed run I assume. Don't know how to fix.

Nixinova:
[Lightshow](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fdangmarm%2Fstringtemplate-lang%2Fmaster%2Fsyntaxes%2Fstringtemplate.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fgithub%2Flinguist%2Fblob%2Fe7d32496d008d5194005d96e7b9a2e89980f61c2%2Fsamples%2FStringTemplate%2FApiOverviewPage.st&code=%3Cdiv%3E%3C%2Fdiv%3E) doesn't seem to work..

lildude:
> [Lightshow](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fdangmarm%2Fstringtemplate-lang%2Fmaster%2Fsyntaxes%2Fstringtemplate.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fgithub%2Flinguist%2Fblob%2Fe7d32496d008d5194005d96e7b9a2e89980f61c2%2Fsamples%2FStringTemplate%2FApiOverviewPage.st&code=%3Cdiv%3E%3C%2Fdiv%3E) doesn't seem to work..

Looks to be a problem with the grammar and not Lightshow. Picking another grammar source works.

Nixinova:
Should I just revert the grammar changes then? Can't find a better one.

lib/linguist/heuristics.yml
```diff
@@ -513,6 +513,14 @@ disambiguations:
...
   - language: SubRip Text
     pattern: '^(\d{2}:\d{2}:\d{2},\d{3})\s*(-->)\s*(\d{2}:\d{2}:\d{2},\d{3})$'
+- extensions: ['.st']
+  rules:
+  - language: StringTemplate
+    pattern:
+    - '</\w+>'
+    - '\$\w[($]'
+  - language: Smalltalk
+    pattern: '\A[\[{("\w\d#]|\w+\s*:=\s*\w+|class >> \w+|^\w+\s+\w+:|^Class\s*{|if(?:True|False):\s*\['
```
lildude, line 523: @Alhadis mind casting your 👀 over this? 🙇

Nixinova: Forgot a plus after line 521 \w. Also I assume having a list means "both must match".

lib/linguist/heuristics.yml
```diff
@@ -513,6 +513,14 @@ disambiguations:
...
   - language: SubRip Text
     pattern: '^(\d{2}:\d{2}:\d{2},\d{3})\s*(-->)\s*(\d{2}:\d{2}:\d{2},\d{3})$'
+- extensions: ['.st']
+  rules:
+  - language: StringTemplate
+    pattern:
+    - '</\w+>'
+    - '\$\w+[($]'
+  - language: Smalltalk
+    pattern: '\A\s*[\[{(^"''\w\d#]|\w+\s*:=\s*\w+|class\s*>>\s*\w+|^\w+\s+\w+:|^Class\s*{|if(?:True|False):\s*\['
```
Alhadis, line 523: `\w\d` can be reduced to just `\w`, since the characters matched by `\d` are considered "word" characters (and are thus matched by `\w`).

Having said that, the `\w+` sequences in this

~~~regexp
\w+\s*:=\s*\w+
~~~

should be replaced by `[_A-Za-z]\w*` or `(?!\d)\w+`:

~~~regexp
(?!\d)\w+\s*:=\s*(?!\d)\w+
~~~

Though I'm unfamiliar with Smalltalk, [a BNF on RosettaCode.org](https://rosettacode.org/wiki/Category:Smalltalk#Syntax_.28in_Pseudo_BNF.29) is pretty explicit about what the language considers to be a legal identifier:

~~~ebnf
assignment ::= <variable> ":=" <expression>
<variable> ::= (<underline>|<letter>)[<underline>|<letter>|digit]*
~~~

Alhadis (changes requested):

lib/linguist/heuristics.yml
Alhadis: @lildude Done. Sorry for the unresponsiveness... I haven't been feeling well lately. 😞

Alhadis: @lildude Done. Sorry for the unresponsiveness... I haven't been feeling well lately. 😞

Alhadis: @lildude Done. Sorry for the unresponsiveness... I haven't been feeling well lately. 😞

lib/linguist/heuristics.yml
Nixinova: \w includes numbers?! How have I gone this long without realising that ...

Alhadis: Also underscores. 😁 Roughly speaking, `\w` is shorthand for `[A-Za-z0-9_]`. Modern, Unicode-savvy regex engines broaden that definition to cover the letters and digits of other scripts, too. Oniguruma, for example, [defines `\w`](https://github.com/kkos/oniguruma/blob/0a1d4fbaf2fc9fa3817e783245351a86a727059e/doc/RE#L50-L56) as any character with a `General_Category` of `Letter`, `Mark`, `Number`, or `Connector_Punctuation` (I believe Perl/PCRE's definition is more-or-less identical).

As you might've guessed, this treatment of `\w` is implementation-dependent, albeit a Unicode recommendation. It's also subject to other factors like Perl's `/a` (ASCII) modifier, whether or not the regex engine is running/compiled without Unicode support, and even the system's locale.

**Bottom-line:** When in doubt, specify ASCII "word" characters explicitly as `[A-Za-z0-9_]` or `[[:alnum:]_]`. It's usually easier (and cleaner) to just write `\w` instead and assume the input will never contain identifiers that use illegal characters (and most of the time, this is a pretty sane assumption to make).

Nixinova: Changed to add `[a-zA-Z_]` before the `\w`s

> or (?!\d)\w+

Inline lookarounds don't seem to work (in JS envs at least)

Alhadis: > Changed to add `[a-zA-Z_]` before the `\w`s

_Almost_ there. Just need to change the quantifier from `+` to `*`:

~~~diff
-[A-Za-z_]\w+
+[A-Za-z_]\w*
~~~

Otherwise, it'll miss names that are exactly one character in length, i.e.:

~~~smalltalk
a := "A"
b := "B"
~~~

> Inline lookarounds don't seem to work (in JS envs at least)

What JavaScript environment? Look-aheads have been supported in JS for as long as I remember (and look-behinds were added in more recent ECMAScript versions too, which is also widely-supported nowadays).

Alhadis:
> [StringTemplate] seems to be HTML but with control statements

From a quick glance at their [format's documentation](https://github.com/antlr/stringtemplate4/blob/a914b54f34b6c9944d8559463c8fa5bb5dd8bd73/doc/templates.md), it seems that StringTemplate files can generate any sort of output, similar to Moustache (even if it's most commonly used to generate HTML).

Ergo, a dedicated StringTemplate grammar would be more appropriate than Atom's HTML grammar, and it shouldn't take me too long to smash one together. I'll hold off on it until @lildude's finished deploying [v7.13.0](https://github.com/github/linguist/pull/5209), just in case an  update to [`Alhadis/language-etc`](https://github.com/Alhadis/language-etc) screws with the procedure somehow. 😉

lildude:
All done @Alhadis. Go wild.

lib/linguist/heuristics.yml
Nixinova: *Inline* lookarounds - `(?!a)b=(?!c)d` matches `b=d` and fails on `b=dac` iirc.

Alhadis (approved):

Alhadis:
> Go wild.

Roger. Will push an update to @Nixinova's branch that contains the required grammar update.

I'm wondering if there should be an undocumented `script/bump-grammar` utility to streamline this procedure… 🤔

Nixinova:
> it seems that StringTemplate files can generate any sort of output

Should I improve the StringTemplate heuristic then?

Alhadis:
> Should I improve the StringTemplate heuristic then?

What we have now should be sufficient. The delimiter characters are configurable, so `<word>` is no less valid than `$word$`, `|word|`, `[word]` or `^word^`. StringTemplate's most common use appears to be preparing HTML, and the [docs recommend `$…$ ` for HTML output](https://github.com/antlr/stringtemplate4/blob/a914b54f34b6c9944d8559463c8fa5bb5dd8bd73/doc/templates.md):

> This documentation uses `<...>` to delimit expressions, but you can use any single start and stop character. For HTML, `$..$` is a much better choice obviously.

The only thing I'd suggest adding is StringTemplate's [`<! comment !>` syntax](https://github.com/antlr/grammars-v4/blob/1c2a1367350d330ba96ef085ef7e0510db075bf8/stringtemplate/STLexer.g4#L125-L128):

~~~yaml
pattern: >-
    (?x-m)

    # Symmetrical delimiter: $…$ '…' #…# ~…~
    (.)! .+? !\1

    # Mirrored delimiters: <…> […] {…} “…” «…»
    |  <! .+? !>
    | \[! .+? !\]
    | \{! .+? !\}
~~~

<details><summary><b>Skippable sidenotes that you should read anyway</b></summary>

Now, I'd like to take a moment to explain a small but critical nuance about the behaviour of `.` in Ruby regexes. See, Ruby/Oniguruma's multi-line mode [changes the semantics of `.` so it matches newline characters](https://github.com/kkos/oniguruma/blob/0a1d4fbaf2fc9fa3817e783245351a86a727059e/doc/RE#L268). This differs from other regex engines, which generally use a separate modifier to enable `dotall` mode (`/s` in Perl/PCRE and JavaScript). As a result, this has **severe performance implications** when processing large files with `.+`—_especially_ when it's used repeatedly in an expression.

Now, while possessive matching (`.++`, `(?>.+)`) helps, an even better solution is to double-check if `.` **should** match newlines in the first place. If not, then `[^\r\n]` or `(?-m:.)` can have substantial performance improvements. This is mostly relevant to Ruby/Oniguruma, because the default behaviour of `.` in other engines is *not* to match newline characters.

## Examples of when `.` _should_ match newlines
##### Good
```regexp
<zalgo>.*?</zalgo>
```

##### Bad
```regexp
async function .+\(.*\)
```

##### Slightly less bad
```regexp
async function .+?\(.*?\)
```
Why only slightly? `.+?` only affects the performance of _successful_ matches. A failed match means the rest of the input is scanned, which is as slow as if `.+` were used.
Solution: use possessive quantifiers (`.++`) or atomic groups (`(?>.+)`) to fail faster.

Okay, enough of my rambling bullshit. `^D`
</details>

Nixinova:
Is what I added sufficient?

Alhadis:
@Nixinova Sorry I took so long to write that StringTemplate grammar. The language turned out to be more complex than I assumed it was.

I've pushed an update to your branch to use the new grammar, so, no actions are required on your part. 👍

lildude (approved):

Issue #4409, opened 2019-02-08
Generated ANTLR files


pchaigno:
Just a reminder for myself to do it later. It's also a good first issue for beginners with Linguist.

From [#4404](https://github.com/github/linguist/pull/4404#issuecomment-460610853):

> Many of these files seem to be generated: https://github.com/search?q=extension%3Ag+%22generated+by+xtext%22&type=Code. Maybe we should add that to the `generated.rb` file? (I can do it in a separate pull request if you prefer.)

The solution is to add the appropriate code in `lib/linguist/generated.rb` to recognize these files as generated.

Nixinova:
Should this still be done?

Nixinova:
Added in #5199, can be closed

PR #4404, opened 2019-02-04
improve .g disambiguation (ANTLR, GAP, G-code)

smola:
## Description

Fixes #1528 

- add .g extension to ANTLR (commonly used for ANTLRv3)

- add heuristic for .g extension, disambiguating ANTLR, GAP and G-code.

- add ANTLR sample tinybasic.g4 (BSD-licensed) from https://github.com/antlr/grammars-v4/blob/master/tinybasic/tinybasic.g4

- add ANTLR sample ASN.g (BSD-licensed) from https://github.com/antlr/grammars-v3/blob/master/asn/ASN.g

## Checklist:
- [x] **I am associating a language with a new file extension.**
  - [x] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      - https://github.com/search?q=extension%3Ag+grammar&type=Code
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://github.com/antlr/grammars-v4/blob/master/tinybasic/tinybasic.g4
      - https://github.com/antlr/grammars-v3/blob/master/asn/ASN.g
    - Sample license(s): BSD
  - [x] I have included a change to the heuristics to distinguish my language from others using the same extension.

PR #5199, opened 2021-02-08
Mark generated ANTLR files as such

Nixinova:
Implements #4409, https://github.com/github/linguist/pull/4404#issuecomment-460610853

Matches .g files that have "generated by Xtest" on their second line.

My knowledge of Ruby is nil (null?) so hopefully I did this right.


Nixinova:
Okay think I might be done with the PRs...

lildude (approved):
The `.g` extension isn't currently associated with ANTRL so the language will still be wrong, but peeps are less likely to notice this with this change 😁 

LGTM.

PR #5205, opened 2021-02-09
Truncate long tokens

smola:
## Description

Instead of discarding all tokens beyond 32 bytes, keep all of them, but truncate at 16 bytes.

This makes the vocabulary more compact, since we don't need to store so long tokens in `samples.json`). This ensures that we keep some relevant tokens such as comments used as delimiters (e.g. ///...).

## Checklist:
_N/A_


script/cross-validation
```diff
@@ -6,7 +6,7 @@ ACCEPTABLE_ERRORS = 48
 
 # Number of acceptable classification errors when using --all.
 # It should only be decreased.
-ACCEPTABLE_ERRORS_ALL = 669
+ACCEPTABLE_ERRORS_ALL = 671
```
smola, line 9: Note that this was 672 in master, it seems to have drifted because of new samples.

smola:
No change in cross-validation with default flags.

Changes for cross-validation with `--all`:

```diff
--- before.all.list	2021-02-10 00:06:23.721999209 +0100
+++ after.trunc.all.list	2021-02-09 23:54:18.966015279 +0100
@@ -64 +64 @@
-Apex/BooleanUtils.cls GOOD
+Apex/BooleanUtils.cls BAD (EQ)
@@ -1225 +1225 @@
-NCL/gsn_csm_xy2_time_series_inputs.ncl BAD (Perl)
+NCL/gsn_csm_xy2_time_series_inputs.ncl GOOD
@@ -2011 +2011 @@
-Text/filenames/README.mysql BAD (RUNOFF)
+Text/filenames/README.mysql BAD (Markdown)
@@ -2224 +2224 @@
-XML/wixdemo.wixproj BAD (Oxygene)
+XML/wixdemo.wixproj GOOD
```

test/test_tokenizer.rb
```diff
@@ -170,9 +170,10 @@ def test_operators
     assert_equal %w(~ a), tokenize("~a")
 
     # Edge cases
-    assert_equal %w(-!#$%&*+,.:;<=>?@\\^_`|~), tokenize("-!#$%&*+,.:;<=>?@\\^_`|~")
-    assert_equal %w(-!#$%&*+,.:;<=>?@\\^_`|~*/), tokenize("-!#$%&*+,.:;<=>?@\\^_`|~*/")
-    assert_equal %w(/-!#$%&*+,.:;<=>?@\\^_`|~), tokenize("/-!#$%&*+,.:;<=>?@\\^_`|~")
+    assert_equal %w(-!#$%&*+,.:;<=>), tokenize("-!#$%&*+,.:;<=>")
+    assert_equal %w(-!#$%&?@\\^_`|~), tokenize("-!#$%&?@\\^_`|~")
+    assert_equal %w(-!#$%&*+,.:;<=>), tokenize("-!#$%&*+,.:;<=>")
+    assert_equal %w(/-!#$%&*+,.:;<>), tokenize("/-!#$%&*+,.:;<>")
```
lildude, line 176: 🤔 shouldn't these tests be confirming the truncation behaviour too ie keep the original argument to `tokenize()` but update the assertion?

lildude:
> Changes for cross-validation with `--all`:

Is this expected? For example, why does using truncated tokens result in different and potentially incorrect classification:

```diff
@@ -64 +64 @@
-Apex/BooleanUtils.cls GOOD
+Apex/BooleanUtils.cls BAD (EQ)
```

... and ...

```diff
@@ -2011 +2011 @@
-Text/filenames/README.mysql BAD (RUNOFF)
+Text/filenames/README.mysql BAD (Markdown)
```

Is the "correction" of the others really a correction?

test/test_tokenizer.rb
smola: There's a new test below testing for truncation. This one is meant to test that these characters do not produce undesired token splits.

smola:
> > @smola Changes for cross-validation with `--all`:
> 
> @lildude Is this expected? 

Kind of.

The cross-validation script, for each sample trains a new classifier with all samples excluding the tested sample, and then it classifies the sample with that classifier. By default, without `--all`, the call to the classifier considers only "ambiguous languages" (per filename and extension). This is quite close to real-world conditions. When using the `--all` flag, every language is considered for every sample, regardless of filename and extension.

I try to minimize errors with both settings. While considering all languages is not what linguist does in practice, a good tokenizer and classifier should also be better at classifying languages without previous disambiguation (in theory).

Now, when comparing diffs for the results, either with or without `--all`, changes to the tokenizer and the classifier also lead to some new errors, even if the overall error count decreases. This often happens with Raku and Perl, some assembly samples or some C/C++ samples, where the classification scores are so close that they easily flip between 2 languages.

Sometimes these changes are hard to explain with anything other than "chance". The current classifier is very sensible to 2 parameters: a) total token count across all samples and b) total token count within a language. The former affects the probability of the unknown token and the later affects the token probabilities within a language. Some tokenizer changes lead to significant changes in these counts, so they affect classifications that are not conceptually related.

In this case, this change avoids discarding tokens longer than 32 bytes, which increases the total token counts. I think this might be the reason for the flip in the classification of `Apex/BooleanUtils.cls` (with `--all`). Note that with the master branch, the scores for this sample are: -2740.84 (Apex) and -2741.27 (EQ), and with this branch they are -2741.49 (Apex) and -2741.39 (EQ). The scores are so close that it's easy to flip this classification. Luckily, EQ is not associated with the `.cls` extension.

Regarding `NCL/gsn_csm_xy2_time_series_inputs.ncl`, the new correct classification seems to be because of the `;***************` token (now truncated, previously discarded).

`Text/filenames/README.mysql` has still an incorrect classification, but the new classification does look as an improvement. The file is pretty close to Markdown. This change seems to be causes by the truncation, since the `================` token, usually present in headers for Markdown and others, are now present in the vocabulary, while they were previously discarded if longer than 32 bytes, or scattered across a lot of different tokens.

I'm not really sure about `XML/wixdemo.wixproj`.

test/test_tokenizer.rb
lildude: Yeah, I saw that, but I was wondering if we wanted to check the truncation here too. We'll leave it as-is to keep the test purpose separate.

lildude (approved):
Makes sense to me. Thanks for the great explanation @smola 🙇

PR #5049, opened 2020-10-14
Add support for Lark grammar files

Alhadis:
This PR adds support for [Lark grammar files](https://lark-parser.readthedocs.io/en/latest/grammar.html), a variant of EBNF used by the [Lark parsing library](https://github.com/lark-parser/lark). The format was brought to my attention by @ThatXliner in Alhadis/language-grammars#4; consult that issue for a more detailed overview of the format.

## Checklist:
- [x] **I am adding a new language.**
	-	[x] **The `.lark` extension is used in hundreds of repositories on GitHub:**  
		*	**Total search results:**
			[1,064 indexed](https://github.com/search?q=extension%3Alark+NOT+nothack&type=Code) and [sampled](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/total-results.log)
		*	**Related usage:**
			*	[802 files](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/related-files.log)
			*	[444 repositories](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/related-repos.log)
			*	[386 users](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/related-users.log)
		*	**Unrelated usage:**
			*	[259 files](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/unrelated-files.log)
			*	[3 repositories](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/unrelated-repos.log)
			*	[2 users](https://github.com/Alhadis/Silos/blob/3fc87991215c0a3697d01477a2c1af32f4dbbc4e/unrelated-users.log)
	-	[x] **I have included samples of real-world usage:**
			*	[`ebl_atf.lark`](https://github.com/ElectronicBabylonianLiterature/ebl-api/blob/373a2d6879857c5bf5cfcfc65908767ff2afc411/ebl/transliteration/domain/ebl_atf.lark): [MIT][1]
			*	[`ebl_atf_common.lark`](https://github.com/ElectronicBabylonianLiterature/ebl-api/blob/c270c55dae96e334460714605bc0a4388c120dfd/ebl/transliteration/domain/ebl_atf_common.lark): [MIT][1]
[...]

PR #5206, opened 2021-02-11
Ungroup Lark from EBNF

Nixinova:
Part of language ungrouping a la #4979.

Ungroups Lark from EBNF as its syntax is quite different and other grammar languages like ABNF are left ungrouped.
Differences between Lark and EBNF are described in https://github.com/Alhadis/language-grammars/issues/4#issue-715263376 (I wouldn't use the word "only" to preface that list...)

Change colour from the blackish `#0b130f` to `#2980B9` from [docs](https://lark-parser.readthedocs.io); minor change as it didn't show up anyway due to being grouped. This color was added in OG PR #5049 but without explanation.


Nixinova:
Ping @Alhadis for their opinion on ungrouping as they made the original PR

Alhadis:
Yeah, that description gave me a rather skewed impression of what Lark in relation to EBNF, so I kinda just trusted the author's definition of it. 😁

As for the colour, if I didn't give an explanation, it was probably arbitrary. No idea why I chose black, though.

Alhadis (approved):

lildude (approved):

PR #5212, opened 2021-02-13
Support Ruby 3.0

smola:
## Description
* Update to minitest >= 5.14.2.

* Use `URI.open` instead of plain `open`.

* Add GitHub Actions job for Ruby 3.0.

Note that Dockerfile is not updated yet, since there is a compilation
issue with charlock_holmes in the `ruby:3-alpine` image.

## Checklist:
_N/A_


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ 2.7, 2.6, 2.5 ]
+        ruby: [ '3.0', '2.7', '2.6', '2.5' ]
```
Alhadis, line 11: Any reason these are quoted, as opposed to numerical forms? The diff looks a little weird.

lildude: I think only `3.0` needs quoting as per the comment in the example at https://github.com/ruby/setup-ruby/#matrix-of-ruby-versions referencing https://github.com/actions/runner/issues/849

I'd prefer it if we just quoted `3.0` and put a comment referencing the issue like in the example.

Alhadis: Hold up. How (or why) is the difference between an integral and a decimal numeral significant?

Alhadis (approved):

.github/workflows/ci.yml
lildude: 🤷 From the comments in the issue, it looks to be something the YAML parsing C# code is doing.

lildude (approved):

PR #5225, opened 2021-02-20
Update TeX heuristic to fix #5219

### LINKED ISSUE #5219: VBA appearing as TeX
I have a [repository](https://github.com/sancarn/stdVBA/search?l=tex) which is displaying some `.cls` files as `TeX` file format. I am uncertain why this is the case but generally speaking VBA/VB6 classes all start with a specific header:

```
VERSION 1.0 CLASS
```

and will contain a line as follows:

```vb
Attribute VB_Name = "<<someName>>"
```

For this reason I feel it should be relatively easy to determine between a `TeX` file and a VBA file. Does anyone know the reason why this is being misidentified.

sancarn:
Fix for [VBA appearing as TeX](https://github.com/github/linguist/issues/5219#issuecomment-781983475).

It was identified that all TeX cls/class files require the line `\\ProvidesClass{...}`. This is much less ambiguous than `\\\w+{`, and will likely lead to matching of my VBA files.

## Checklist:
  - I don't believe a sample is required, but if wanted you can use [this](https://github.com/sancarn/stdVBA/blob/940c768326366308c659c4387331f3c047454867/src/stdRegex.cls) sample

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -107,7 +107,7 @@ disambiguations:
 - extensions: ['.cls']
   rules:
   - language: TeX
-    pattern: '\\ProvidesClass{'
+    pattern: '^\s*\\ProvidesClass{'
   - language: ObjectScript
     pattern: '^Class\s'
 - extensions: ['.cmp']
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -107,7 +107,7 @@ disambiguations:
 - extensions: ['.cls']
   rules:
   - language: TeX
-    pattern: '^\s*\\ProvidesClass{'
+    pattern: '^\s*\\(?:NeedsTeXFormat|ProvidesClass){'
   - language: ObjectScript
     pattern: '^Class\s'
 - extensions: ['.cmp']
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -107,7 +107,7 @@ disambiguations:
 - extensions: ['.cls']
   rules:
   - language: TeX
-    pattern: '\\\w+{'
+    pattern: '\\ProvidesClass{'
   - language: ObjectScript
     pattern: '^Class\s'
 - extensions: ['.cmp']


lib/linguist/heuristics.yml
```diff
@@ -107,7 +107,7 @@ disambiguations:
 - extensions: ['.cls']
   rules:
   - language: TeX
-    pattern: '\\\w+{'
+    pattern: '\\ProvidesClass{'
```
Alhadis, line 110: On second thought, it might be wiser to limit this to start-of-line matches only. Just to be on the safe side.

```suggestion
    pattern: '^\s*\\ProvidesClass{'
```

Alhadis (changes requested):

Alhadis (approved):

lib/linguist/heuristics.yml
```diff
@@ -107,7 +107,7 @@ disambiguations:
 - extensions: ['.cls']
   rules:
   - language: TeX
-    pattern: '\\\w+{'
+    pattern: '^\s*\\ProvidesClass{'
```
Alhadis, line 110: Could you include `NeedsTeXFormat` as well, please? I probably should've been clearer in my [earlier comment](https://github.com/github/linguist/issues/5219#issuecomment-782545817).

```suggestion
    pattern: '^\s*\\(?:NeedsTeXFormat|ProvidesClass){'
```

I don't know TeX well enough to know if `ProvidesClass` can or can't be omitted or not, and `NeedsTeXFormat` has virtually zero risk of appearing in a `.cls` file that isn't TeX.

Alhadis (changes requested):

lib/linguist/heuristics.yml
sancarn: Yeah this is a fair point, it generally seemed that NeedsTeXFormat could be ommitted so wouldn't be a requirement.

But yes, not really a problem having it.

Alhadis (approved):

lildude (approved):

PR #5230, opened 2021-02-22
D Language Alias Proposal

sig-aarena:
The D Language is also known as the Dlang, can not find an exact source but the languages homepage is an indicator of it- https://dlang.org/


lildude (approved):
Make sense to me. 👍

PR #4912, opened 2020-07-02
Rebrand "Mathematica" to "Wolfram Language".

jfultz:
Also, fix up the list of appropriate extensions.

## Checklist:

- [ ] **I am associating a language with a new file extension.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

- [ ] **I am fixing a misclassified language**
  - [ ] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am changing the source of a syntax highlighting grammar**
  
  - Old: https://github-lightshow.herokuapp.com/
  - New: https://github-lightshow.herokuapp.com/

- [ ] **I am updating a grammar submodule**
  

- [ ] **I am adding new or changing current functionality**
  
  - [ ] I have added or updated the tests for the new or changed functionality.

- [ ] **I am changing the color associated with a language**
  
  - [ ] I have obtained agreement from the wider language community on this color change.
    - [URL to public discussion]
    - [Optional: URL to official branding guidelines for the language]

PR #5232, opened 2021-02-23
Add Wolfram Language as alias for Mathematica

gwpantazes:
Add "Wolfram Language" and abbreviated "wl" aliases for Mathematica.

## Description

In recent years, the language underpinning Mathematica has been called "Wolfram Language" (https://www.wolfram.com/language/), and Mathematica is just the GUI notebook application.

WL is a common abbreviation for Wolfram Language (as it is the `.wl` file extension, but can also be used as an abbreviated name for the language, so I also put it as an alias).

Disclaimer: I'm currently employed at Wolfram. I'm proposing this change by my own whim, so my actions don't reflect Wolfram as a company. Just a suggestion based on what I know about possible aliases.

## Checklist:

❓ I'm unsure if this is the only applicable item? I'm not sure if my changes apply to this though?

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3299,8 +3299,8 @@ Mathematica:
   aliases:
   - mma
   - wolfram
-  - wolframlanguage
-  - wolframlang
+  - wolfram language
+  - wolfram lang
   - wl
   tm_scope: source.mathematica
   ace_mode: text
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3298,6 +3298,10 @@ Mathematica:
   - ".wlt"
   aliases:
   - mma
+  - wolfram
+  - wolframlanguage
+  - wolframlang
+  - wl
   tm_scope: source.mathematica
   ace_mode: text
   codemirror_mode: mathematica


lib/linguist/languages.yml
```diff
@@ -3298,6 +3298,10 @@ Mathematica:
   - ".wlt"
   aliases:
   - mma
+  - wolfram
+  - wolframlanguage
+  - wolframlang
```
Nixinova, line 3303: It would be better to set these aliases to "wolfram language" and "wolfram lang" as Linguist automatically adds aliases that replace each space with a hyphen and empty string as well: "wolfram lang" will add aliases "wolframlang" and "wolfram-lang" without needing to do that manually. See https://github.com/github/linguist/issues/5135#issuecomment-757868541

lildude:
It's probably worth mentioning an attempt to rename the language was started in https://github.com/github/linguist/pull/4912 but the author didn't respond to the change requests so it went stale.

gwpantazes:
> It's probably worth mentioning an attempt to rename the language was started in #4912 but the author didn't respond to the change requests so it went stale.

Ah yeah, #4912 seems to a be a much more thorough PR regarding the name change. Does it make sense to just add the aliases for now in this PR and maybe someone could have another go at reviving the proposed changes from #4912 some other time?

lib/linguist/languages.yml
gwpantazes: Added the spaces.

lildude:
> Does it make sense to just add the aliases for now in this PR and maybe someone could have another go at reviving the proposed changes from #4912 some other time?

Yeah, lets go with that for now.

lildude (approved):

PR #5233, opened 2021-02-23
Add jq language

wader:
https://stedolan.github.io/jq/

## Description

Language color is based on code example color from jq documentation.

## Checklist:

      -  https://github.com/search?q=extension%3Ajq&type=Code most of the files found seem to be jq and not jsoniq.
    - Sample source(s):
      - https://github.com/stedolan/jq/blob/master/src/builtin.jq
    - Sample license(s):
      - https://raw.githubusercontent.com/stedolan/jq/master/COPYING
      - language-jq submodule is fork of https://github.com/bencampion/language-jq that is under MIT https://github.com/bencampion/language-jq/blob/master/package.json
  
  - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fwader%2Flinguist-language-jq%2Fmaster%2Fgrammars%2Fjq.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fstedolan%2Fjq%2Fmaster%2Fsrc%2Fbuiltin.jq&code=

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2577,7 +2577,7 @@ JSONiq:
   codemirror_mime_type: application/json
   extensions:
   - ".jq"
-  tm_scope: source.jq
+  tm_scope: source.jsoniq
   language_id: 177
 Jasmin:
   type: programming
@@ -6612,7 +6612,7 @@ jq:
   type: programming
   extensions:
   - ".jq"
-  tm_scope: source.jq.2
+  tm_scope: source.jq
   language_id: 905371884
 mIRC Script:
   type: programming
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -64,8 +64,8 @@ ABAP CDS:
   type: programming
   color: "#555e25"
   extensions:
-    - ".asddls"
-  tm_scope: "source.abapcds"
+  - ".asddls"
+  tm_scope: source.abapcds
   language_id: 452681853
   ace_mode: text
 ABNF:
@@ -6606,6 +6606,14 @@ fish:
   tm_scope: source.fish
   ace_mode: text
   language_id: 415
+jq:
+  color: "#c7254e"
+  ace_mode: text
+  type: programming
+  extensions:
+  - ".jq"
+  tm_scope: source.jq.2
+  language_id: 905371884
 mIRC Script:
   type: programming
   color: "#3d57c3"
--- a/.gitmodules
+++ b/.gitmodules
@@ -722,6 +722,9 @@
 [submodule "vendor/grammars/latex.tmbundle"]
 	path = vendor/grammars/latex.tmbundle
 	url = https://github.com/textmate/latex.tmbundle
+[submodule "vendor/grammars/linguist-language-jq"]
+	path = vendor/grammars/linguist-language-jq
+	url = https://github.com/wader/linguist-language-jq
 [submodule "vendor/grammars/linter-lilypond"]
 	path = vendor/grammars/linter-lilypond
 	url = https://github.com/nwhetsell/linter-lilypond
--- a/.gitmodules
+++ b/.gitmodules
@@ -581,6 +581,9 @@
 [submodule "vendor/grammars/language-jolie"]
 	path = vendor/grammars/language-jolie
 	url = https://github.com/fmontesi/language-jolie
+[submodule "vendor/grammars/language-jq"]
+	path = vendor/grammars/language-jq
+	url = https://github.com/wader/language-jq
 [submodule "vendor/grammars/language-jsoniq"]
 	path = vendor/grammars/language-jsoniq
 	url = https://github.com/wcandillon/language-jsoniq
@@ -722,9 +725,6 @@
 [submodule "vendor/grammars/latex.tmbundle"]
 	path = vendor/grammars/latex.tmbundle
 	url = https://github.com/textmate/latex.tmbundle
-[submodule "vendor/grammars/linguist-language-jq"]
-	path = vendor/grammars/linguist-language-jq
-	url = https://github.com/wader/linguist-language-jq
 [submodule "vendor/grammars/linter-lilypond"]
 	path = vendor/grammars/linter-lilypond
 	url = https://github.com/nwhetsell/linter-lilypond
--- a/grammars.yml
+++ b/grammars.yml
@@ -659,6 +659,8 @@ vendor/grammars/latex.tmbundle:
 - text.tex.latex
 - text.tex.latex.beamer
 - text.tex.latex.memoir
+vendor/grammars/linguist-language-jq:
+- source.jq.2
 vendor/grammars/linter-lilypond:
 - source.lilypond
 vendor/grammars/liquid:
--- a/grammars.yml
+++ b/grammars.yml
@@ -531,8 +531,10 @@ vendor/grammars/language-jison:
 - source.jisonlex-injection
 vendor/grammars/language-jolie:
 - source.jolie
-vendor/grammars/language-jsoniq:
+vendor/grammars/language-jq:
 - source.jq
+vendor/grammars/language-jsoniq:
+- source.jsoniq
 - source.xq
 vendor/grammars/language-jsonnet:
 - source.jsonnet
@@ -659,8 +661,6 @@ vendor/grammars/latex.tmbundle:
 - text.tex.latex
 - text.tex.latex.beamer
 - text.tex.latex.memoir
-vendor/grammars/linguist-language-jq:
-- source.jq.2
 vendor/grammars/linter-lilypond:
 - source.lilypond
 vendor/grammars/liquid:
--- a/samples/jq/builtin.jq
+++ b/samples/jq/builtin.jq
@@ -1,0 +1,275 @@
+def halt_error: halt_error(5);
+def error(msg): msg|error;
+def map(f): [.[] | f];
+def select(f): if f then . else empty end;
+def sort_by(f): _sort_by_impl(map([f]));
+def group_by(f): _group_by_impl(map([f]));
+def unique: group_by(.) | map(.[0]);
+def unique_by(f): group_by(f) | map(.[0]);
+def max_by(f): _max_by_impl(map([f]));
+def min_by(f): _min_by_impl(map([f]));
+def add: reduce .[] as $x (null; . + $x);
+def del(f): delpaths([path(f)]);
+def _assign(paths; $value): reduce path(paths) as $p (.; setpath($p; $value));
+def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
+def map_values(f): .[] |= f;
+
+# recurse
+def recurse(f): def r: ., (f | r); r;
+def recurse(f; cond): def r: ., (f | select(cond) | r); r;
+def recurse: recurse(.[]?);
+def recurse_down: recurse;
+
+def to_entries: [keys_unsorted[] as $k | {key: $k, value: .[$k]}];
+def from_entries: map({(.key // .Key // .name // .Name): (if has("value") then .value else .Value end)}) | add | .//={};
+def with_entries(f): to_entries | map(f) | from_entries;
+def reverse: [.[length - 1 - range(0;length)]];
+def indices($i): if type == "array" and ($i|type) == "array" then .[$i]
+  elif type == "array" then .[[$i]]
+  elif type == "string" and ($i|type) == "string" then _strindices($i)
+  else .[$i] end;
+def index($i):   indices($i) | .[0];       # TODO: optimize
+def rindex($i):  indices($i) | .[-1:][0];  # TODO: optimize
+def paths: path(recurse(if (type|. == "array" or . == "object") then .[] else empty end))|select(length > 0);
+def paths(node_filter): . as $dot|paths|select(. as $p|$dot|getpath($p)|node_filter);
+def isfinite: type == "number" and (isinfinite | not);
+def arrays: select(type == "array");
+def objects: select(type == "object");
+def iterables: select(type|. == "array" or . == "object");
+def booleans: select(type == "boolean");
+def numbers: select(type == "number");
+def normals: select(isnormal);
+def finites: select(isfinite);
+def strings: select(type == "string");
+def nulls: select(. == null);
+def values: select(. != null);
+def scalars: select(type|. != "array" and . != "object");
+def leaf_paths: paths(scalars);
+def join($x): reduce .[] as $i (null;
+            (if .==null then "" else .+$x end) +
+            ($i | if type=="boolean" or type=="number" then tostring else .//"" end)
+        ) // "";
+def _flatten($x): reduce .[] as $i ([]; if $i | type == "array" and $x != 0 then . + ($i | _flatten($x-1)) else . + [$i] end);
+def flatten($x): if $x < 0 then error("flatten depth must not be negative") else _flatten($x) end;
+def flatten: _flatten(-1);
+def range($x): range(0;$x);
+def fromdateiso8601: strptime("%Y-%m-%dT%H:%M:%SZ")|mktime;
+def todateiso8601: strftime("%Y-%m-%dT%H:%M:%SZ");
+def fromdate: fromdateiso8601;
+def todate: todateiso8601;
+def match(re; mode): _match_impl(re; mode; false)|.[];
+def match($val): ($val|type) as $vt | if $vt == "string" then match($val; null)
+   elif $vt == "array" and ($val | length) > 1 then match($val[0]; $val[1])
+   elif $vt == "array" and ($val | length) > 0 then match($val[0]; null)
+   else error( $vt + " not a string or array") end;
+def test(re; mode): _match_impl(re; mode; true);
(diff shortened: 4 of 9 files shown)


wader:
Hi, I will probably need some help with heuristics to distinguish between jsoniq and jq, they both use `.jq`. Also i'm not sure I did the correct thing and added jq as `source.jq.2` as `source.jq` is already used by jsoniq. I also did a fork of https://github.com/bencampion/language-jq to https://github.com/wader/language-jq to be able to get scope name `source.jq.2`.

lib/linguist/languages.yml
```diff
@@ -64,8 +64,8 @@ ABAP CDS:
   type: programming
   color: "#555e25"
   extensions:
-    - ".asddls"
-  tm_scope: "source.abapcds"
```
wader, line 68: Not sure why this got changed. Think the update-ids script reindented?

wader:
Nice work on the CI system for this repo 👍  catches a lot of things!

lib/linguist/languages.yml
Alhadis: > Think the update-ids script reindented?

Yes. When it updates `languages.yml`, the entire file is replaced _in situ_ with the YAML-encoded representation of the in-memory database.

Here's what `update-ids` would look like if it were Node.js/JSON instead of Ruby/YAML:

~~~js
#!/usr/bin/env node
import {readFileSync, writeFileSync} from "fs";

const jsonFile = "./lib/linguist/languages.json";
let languages;

// Load database
languages = readFileSync(jsonFile, "utf8");
languages = JSON.parse(languages);

// Generate IDs for any entries that lack one
for(const name in languages){
	if(!("language_id" in languages[name]))
		languages[name].language_id = generateRandomID();
}

// Write updated database back to file
languages = JSON.stringify(languages, null, "\t");
writeFileSync(jsonFile, languages);
~~~

Alhadis: So when you updated the IDs, the script naturally reformatted anything that was added manually in the time since `update-ids` was last run.

lib/linguist/languages.yml
```diff
@@ -2701,6 +2701,14 @@ Jolie:
   ace_mode: text
   tm_scope: source.jolie
   language_id: 998078858
+Jq:
```
Alhadis, line 2704: Shouldn't this be

```suggestion
JQ:
```

instead?

lib/linguist/languages.yml
wader: Ok! should i revert or keep it in the PR?

Alhadis: Keep it in the PR. This happens from time-to-time.

lib/linguist/languages.yml
wader: Can do, what is the convention? i think jq commonly is all lowercase but it seems the yml file uses captial keys?

wader:
Just curious, i haven't specified a path to the syntax file in the submodule, will be derived automatically somehow?

lib/linguist/languages.yml
Alhadis: If a language's name uses non-standard or unusual casing (as [jq](https://stedolan.github.io/jq/) does), then we use that; otherwise, **<ins>L</ins>abel <ins>C</ins>ase** is used (unless it's an obvious acronym, like HTML).

Aside from renaming the YAML entry, the `samples/Jq` directory will need to be renamed to use the updated capitalisation. Now, committing this might be tricky if you're using an OS with a case-insensitive filesystem (e.g., macOS). Running the following commands from the root of your fork should do the trick:

~~~sh
git config --local --add core.ignorecase false
git add --all samples/Jq samples/JQ
~~~

Alhadis:
> Just curious, i haven't specified a path to the syntax file in the submodule, will be derived automatically somehow?

Aye. [This](https://github.com/github/linguist/blob/b2834449f1dd9dd9b76652092bad9fc593f687d3/tools/grammars/compiler/loader.go#L106-L129) should give you the gist of where Linguist looks for grammar files.

> […] as `source.jq` is already used by `jsoniq`

Are you sure that's for a different format also named JQ? If not, we can use the JSONIQ repository's existing `source.jq` grammar.

lib/linguist/languages.yml
Alhadis: Bloody hell, I gave you the wrong capitalisation. 😰 That last line should have been

~~~sh
git mv samples/Jq samples/jq
~~~

So you'll now have to run

~~~sh
git mv samples/JQ samples/jq
~~~

Apologies for the screw-up. 😞

wader:
> > […] as `source.jq` is already used by `jsoniq`
> 
> Are you sure that's for a different format also named JQ? If not, we can use the JSONIQ repository's existing `source.jq` grammar.

They have similar syntax and similar uses so not surprised it's confusing :) It's probably the reason linguist current is mis-detect some jq syntax as JSONIQ.

Compare:
https://github.com/stedolan/jq/blob/master/src/builtin.jq
vs
https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fwader%2Flinguist-language-jq%2Fmaster%2Fgrammars%2Fjq.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fstedolan%2Fjq%2Fmaster%2Fsrc%2Fbuiltin.jq&code=

So not sure what to do about the scope name, not great. Would be good in the future to be able to point the vendor submodule to a syntax file used for vscode/atom etc but extensions for them seem to use `source.jq` for both jq and JSONIQ.

wader:
Do we have to add some heuristics because they are so similar?

lib/linguist/languages.yml
wader: No worries, should be "jq" as yml key also?

Alhadis:
> So not sure what to do about the scope name, not great

I'd submit a PR to the JSONiq repository to change the scope-name to `source.jsoniq`. I don't understand why the author picked `source.jq` as a unique identifier for their JSONiq grammar…

Alhadis:
> Do we have to add some heuristics because they are so similar?

I think so. Try `/^def[ \t]+\w+(?:\(.*?\))?:/` for jq.

wader:
> > So not sure what to do about the scope name, not great
> 
> I'd submit a PR to the JSONiq repository to change the scope-name to `source.jsoniq`. I don't understand why the author picked `source.jq` as a unique identifier for their JSONiq grammar…

Mm 😞  i guess they picked it based on extension name, both use .jq 😄

wader:
Some unique things for JSONIQ, uses `(: ... :)` for comments, `:=` assignment and `{| ... |}` for scopes

Alhadis:
Just make sure the heuristics for each language are matching syntax that's invalid in the other.

Remember, it doesn't have to be foolproof: our new classifier is doing a pretty good job of identifying languages from the `samples/*` it's trained by. :wink:

pjeby:
Yeah, shouldn't the token frequencies be pretty indicative without heuristics?  I mean, you can't write much jq without `def`, and JSONiq comment syntax is invalid syntax in jq.  I think the commonly-used keywords/functions are different in each as well.

wader:
@pjeby @AlaineN ok how to find out if classifier is enough?

Suggest heuristic if needed:
```yml
- extensions: ['.jq']
  rules:
  - language: jq
    pattern: '^def[ \t]+\w+(?:\(.*?\))?:'
  - language: JSONiq
    pattern: '\(:|:\)|:=|{\||\|}'
```

In the patterns does `^` anchor to start of whole source text or line starts?

wader:
@Alhadis Great work, also think it make most sense to rename scope for JSONiq. Should i rename the scope to `source.jq` in https://github.com/wader/linguist-language-jq and update the submodule commit? note that my repo is a fork of https://github.com/bencampion/language-jq but it also has some additional fixes for arrays and keywords, see https://github.com/bencampion/language-jq/issues

Alhadis:
@wader I've submitted a PR to fix the JSONiq grammar's scope. See wcandillon/language-jsoniq#23.

> In the patterns does `^` anchor to start of whole source text or line starts?

In *most* RegExp engines, `^` matches the start of the source-text **unless** multiline mode is active (activated by `/m`). When in multiline mode, `^` will match both the start of a string *and* the source-text (and the converse is also true of `$`). This is complemented by a separate mode called `dotall` (`/s`), which causes `.` to match newlines in addition to non-newlines (by default, `.` won't match a newline character).

Ruby, however, does something of questionable sanity: not only does it enable multiline mode by default, its multiline mode *also* enables `dotall` semantics.

So, to give you the TL;DR answer: yes, `^` matches both newlines and start-of-input. It's worth pointing out that the latter can be explicitly requested by way of the `\A` sequence (which may not behave the way you'd expect if regexes are being tested line-by-line).

Alhadis:
> Should i rename the scope to `source.jq` in https://github.com/wader/linguist-language-jq and update the submodule commit?

Yes, but hold off until @wcandillon approves/merges wcandillon/language-jsoniq#23. In addition to bumping the JSONiq submodule, you'll need to update...

1.	[`grammars.yml`](https://github.com/github/linguist/blob/b2834449f1dd9dd9b76652092bad9fc593f687d3/grammars.yml#L535):
	```diff
	 vendor/grammars/language-jsoniq:
	-- source.jq
	+- source.jsoniq
	```
2.	[`languages.yml`](https://github.com/github/linguist/blob/b2834449f1dd9dd9b76652092bad9fc593f687d3/lib/linguist/languages.yml#L2580)
	```diff
	 JSONiq:
	   color: "#40d47e"
	   type: programming
	   ace_mode: jsoniq
	   codemirror_mode: javascript
	   codemirror_mime_type: application/json
	   extensions:
	   - ".jq"
	-  tm_scope: source.jq
	+  tm_scope: source.jsoniq
	```

wader:
Ok let's wait a bit, thanks for all the help

Alhadis:
> note that my repo is a fork of https://github.com/bencampion/language-jq but it also has some additional fixes for arrays and keywords

If @bencampion isn't maintaining that package/repository anymore, perhaps ask if he'd be willing to transfer ownership of it to you. That way, you can publish improvements that benefit both Linguist *and* Atom users. 👍 Failing that, try submitting a PR. @bencampion appears reasonably active on GitHub, so I doubt contributions will fall on deaf ears.

IMHO, publishing a fork should really only be done when you *know* a grammar repository is dead and unmaintained. Or if Linguist's requirements clash with those of upstream (generally in placement and naming of files, switching to unsupported grammar formats, etc).

wader:
True, maybe i should turn my issues into a PR instead. Sadly it seems like vscode does not have CSON support from what i can tell, but atom and Linguist as JSON support so maybe it should be maintained in that format.

The `source.jq` collision made me do a fork but now i think your approach is better and hopefully we can make the grammar situation a be mess messy.

Alhadis:
<pre language="yaml"><code>- language: JSONiq
  pattern: '<mark><ins>\(:|:\)</ins></mark>|:=|{\||\|}'
</code></pre>

You might want to match `(:…:)` in pairs:

```regexp
(?-m:\(:.*?:\))
```

Otherwise, the heuristic risks matching something like `:)`, which could easily appear alongside a snarky comment:

~~~jsoniq
let $rant := "Back in my day, :) was how you wrote 😄."
    $rant += "Damn kids these days with their newfangled emojis…"
~~~

For performance reasons, I've restricted the regex to single-line searches using `(?-m)`, which disables Ruby's multiline mode (going back to my [earlier comment](https://github.com/github/linguist/pull/5233#issuecomment-789288012) about Ruby's whacky defaults…)

pjeby:
> ```yaml
> pattern: '^def[ \t]+\w+(?:\(.*?\))?:'
> ```

Note that jq is a stream-formatted language (i.e. line feeds and other whitespace are interchangeable), and that `def`  can occur anywhere in a jq pipeline.  e.g. this is valid jq that would not trigger the above heuristic:

```
{x:42} | def x: .x; x
```

Alhadis:
> `def` can occur anywhere in a jq pipeline

That's fine. We don't need heuristics to be exhaustive and airtight, unless we're distinguishing between *very* similar languages (e.g., C / C++ / Objective-C / Objective C++). Otherwise, a heuristic need only match commonly-encountered constructs that cover the lion's share of real-world uses, and do so accurately.

Don't forget, there's still the classifier to fall back on when heuristics (and every other strategy) fails to match a language.

wader:
@Alhadis let me know when i should update things. I guess we should merge a rename and update of JSONiq submodule first?

Alhadis:
@wader See wcandillon/language-jsoniq#24. I couldn't resist fixing some glaring mistakes in the JSONiq grammar (though I left the XQuery grammar alone). We'll wait for it to get merged, then we can bump the grammar. :+1:

wader:
> @wader See [wcandillon/language-jsoniq#24](https://github.com/wcandillon/language-jsoniq/pull/24). I couldn't resist fixing some glaring mistakes in the JSONiq grammar (though I left the XQuery grammar alone). We'll wait for it to get merged, then we can bump the grammar. 👍

Ok 👍 does the jq grammar look ok? i've noticed that it do allow quite a lot of things that are not valid jq, eg `1 || 1` but maybe it's hard to describe syntax exactly.

I noticed a merge commit got pushed, should i stay away from rebasing and just add more commits now?

Alhadis:
> I noticed a merge commit got pushed

Aye, you'll want to update your branch locally by pulling from https://github.com/wader/linguist/tree/jq. 😉

> i've noticed that it do allow quite a lot of things that are not valid jq, eg 1 || 1 but maybe it's hard to describe syntax exactly.

That's fine. 👍 Syntax highlighting isn't the place for validating semantics; only stuff that's clearly illegal syntax really deserves to be marked as an error.

wader:
Created a PR with the fixes https://github.com/bencampion/language-jq/pull/4

wader:
@AlaineN Is linguist used in highlight code in github flavoured markdown?

AlaineN:
> @AlaineN Is linguist used in highlight code in github flavoured markdown?

I think you may be intending to tag someone else.

Alhadis:
> I think you may be intending to tag someone else.

I think I might be that "someone else". 😁 @Alhadis, @AlaineN… close enough for autocomplete to get it wrong. Right, @lildud?

Alhadis:
> ~~\@AlaineN~~ @Alhadis Is linguist used in highlight code in github flavoured markdown?

Alright, two things:

1. Linguist actually doesn't do the highlighting; it simply selects the grammar which GitHub's highlighting engine (something called "PrettyLights") will use to make code look spiffy on GitHub.com
2. I assume that by "GitHub-flavoured markdown", you're referring to fenced code-blocks. In which case, yes, the same grammar is used to highlight embedded code samples. However, this isn't limited to Markdown: other formats supported on GitHub (like [AsciiDoc](https://github.com/Alhadis/Accordion/blob/master/docs/options.adoc)) utilise a similar feature.

lib/linguist/languages.yml
Alhadis: Sorry, I missed that last question.

Yes. Everywhere the language's name appears should match the exact characters used in `languages.yml`: Heuristics, tests, the name of the `samples/` subdirectory, the grammar-index, etc.

The only exception to this is the `F*` language, which necessitated a [new field](https://github.com/github/linguist/blob/76c397c7ad07319e766fb1445b8a2307af2f586b/lib/linguist/languages.yml#L3-L5) that allowed us to use filename-safe characters for [`./samples/Fstar/](https://github.com/github/linguist/blob/76c397c7ad07319e766fb1445b8a2307af2f586b/lib/linguist/languages.yml#L1502). However, this is an edge-case, and one you'll almost certainly never have to worry or care about.

wader: Ok, i assumed so, so already changed it

wader:
> > @AlaineN Is linguist used in highlight code in github flavoured markdown?
> 
> I think you may be intending to tag someone else.

Oj sorry for that, not sure how your name got into the auto completer

Alhadis:
> not sure how your name got into the auto completer

She's a GitHub staff member, and Linguist is hosted on the @github org, so... I imagine that's all it takes to be listed as an autocomplete suggestion.

wader:
Getting close to merge? what happens after merge, will take some time until used by official github and jq-files being detected as jq?

Alhadis:
> Getting close to merge? what happens after merge, will take some time until used by official github and jq-files being detected as jq?

Hey @wader, sorry for the slow response.

Yes, this LGTM. 👍 As for merging, changes made to Linguist only take effect on GitHub once there's been a new release of Linguist. Hence you won't see the improved highlighting on the site straight away.

Alhadis (approved):

wader:
Ok! and as i understand submodules are updated regularly in linguist so no need to send PR if the syntax is updated?

And thanks again for all the help!

Alhadis:
> Ok! and as i understand submodules are updated regularly in linguist so no need to send PR if the syntax is updated?

Grammars are all bumped to the latest versions right before a release is cut. Only in peculiar scenarios (such as this one) is it necessary to manually bump a grammar ourselves.

lildude (approved):
LGTM. Thanks 🙇

PR #5236, opened 2021-02-28
Group Maven POM under XML

Nixinova:
Discussion at #5189. I don't know whether Maven POM should even be it's own language but may as well group it under XML in the meantime.

[The docs](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html) say:
> A Project Object Model or POM is the fundamental unit of work in Maven. It is an XML file


lildude (approved):
Makes sense to me

PR #5241, opened 2021-03-01
Add Euphoria language

ghaberek:
## Description

Add support for the [Euphoria](https://en.wikipedia.org/wiki/Euphoria_(programming_language)) language ([website](https://openeuphoria.org/), [github](https://github.com/OpenEuphoria)). We have been gradually migrating development of Euphoria onto GitHub but the frequent mis-classification by linguist makes it look like we're using a handful of other languages (namely E, Eiffel, and Elixir) due to extension collisions. Euphoria primarily uses .e for include files and .ex for programs. These changes should correct that. I've done some diligence in correcting new false-positives by adding a sample for Elixir and several patterns to the heuristics, although admittedly I am not nearly as familiar with those other languages. I did test linguist against several larger Eiffel and Elixir repositories.

## Checklist:

      - .e - https://github.com/search?q=procedure+extension%3Ae&type=Code&ref=advsearch&l=&l=
      - .ex - https://github.com/search?q=function+-def+-defmodule+-import+-interface+-nil+extension%3Aex&type=Code&ref=advsearch&l=&l=
    - Sample source(s):
      - **Eiffel**
        - [app_embedded_web_execution.e](https://github.com/EiffelWebFramework/EWF/blob/master/examples/desktop_app/src/app_embedded_web_execution.e)
        - [app_embedded_web_service.e](https://github.com/EiffelWebFramework/EWF/blob/master/examples/desktop_app/src/app_embedded_web_service.e)
        - [desktop_app.e](https://github.com/EiffelWebFramework/EWF/blob/master/examples/desktop_app/src/desktop_app.e)
        - [main_window.e](https://github.com/EiffelWebFramework/EWF/blob/master/examples/desktop_app/src/main_window.e)
      - **Elixir**
        - [atom.ex](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/atom.ex)
        - [float.ex](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/float.ex)
        - [function.ex](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/function.ex)
        - [integer.ex](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/integer.ex)
        - [regex.ex](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/regex.ex)
      - **Euphoria**
        - [allsorts.ex](https://github.com/OpenEuphoria/euphoria/blob/master/demo/allsorts.ex)
        - [animal.ex](https://github.com/OpenEuphoria/euphoria/blob/master/demo/animal.ex)
        - [csort.ex](https://github.com/OpenEuphoria/euphoria/blob/master/demo/csort.ex)
        - [debug.e](https://github.com/OpenEuphoria/euphoria/blob/master/include/euphoria/debug/debug.e)
        - [dll.e](https://github.com/OpenEuphoria/euphoria/blob/master/include/std/dll.e)
        - [hash.ex](https://github.com/OpenEuphoria/euphoria/blob/master/demo/hash.ex)
        - [math.e](https://github.com/OpenEuphoria/euphoria/blob/master/include/std/math.e)
        - [mydata.ex](https://github.com/OpenEuphoria/euphoria/blob/master/demo/mydata.ex)
        - [regex.e](https://github.com/OpenEuphoria/euphoria/blob/master/include/std/regex.e)
        - [types.e](https://github.com/OpenEuphoria/euphoria/blob/master/include/std/types.e)
    - Sample license(s):
      - Eiffel: [EFL](https://github.com/EiffelWebFramework/EWF/blob/master/LICENSE)
      - Elixir: [Apache License 2.0](https://github.com/elixir-lang/elixir/blob/master/LICENSE)
      - Euphoria: [MIT license](https://github.com/OpenEuphoria/euphoria/blob/master/License.txt)
  - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2FOpenEuphoria%2Fvscode-euphoria%2Fblob%2Fmaster%2Fsyntaxes%2Feuphoria.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2FOpenEuphoria%2Feuphoria%2Fblob%2Fmaster%2Fsource%2Fmain.e&code=


ghaberek:
Apparently I didn't properly commit the git submodule for [vscode-euphoria](https://github.com/OpenEuphoria/vscode-euphoria), probably because I synced my fork at some point and left that out. I've committed the submodule now, and that corrected the failing Ruby tests.

I'm still not sure how to correct the cross-validation test. I'm not even sure what's wrong. Running `bundle exec script/cross-validation --test` isn't covered in the contributing doc but `bundle exec rake test` completes successfully on my system.

lildude (changes requested):
This is looking great. Thanks. 🙇  One nit about the extensions inline and an explanation for the failing test:

The classifier test uses just the samples in Linguist to test the classifier in isolation. It is failing because the classifier is detecting the `Elixir/function.ex` file as Eurphoria (it's buried [here](https://github.com/github/linguist/pull/5241/checks?check_run_id=2005970554#step:5:175)):

```
Elixir/function.ex BAD (Euphoria)
```

This will be happening because the content of that sample matches more attributes of Euphoria based on the samples provided than Elixir, which isn't really surprising because Elixir has only one other file and it's teeny weeny and really not representative of the language.

The solution is to add another Elixir `.ex` file that is a good representation of the language.

This wasn't an issue before now because Elixir was was added to Linguist many many years before we started curating samples so one wasn't added and it was the only language using `.ex` until now 😁 .

And thanks the for the pointer that the docs are missing mention of running this test.

lib/linguist/languages.yml
```diff
@@ -1484,6 +1485,22 @@ Erlang:
...
+Euphoria:
+  type: programming
+  color: "#FF790B"
+  extensions:
+  - ".e"
+  - ".eu"
+  - ".ew"
+  - ".ex"
+  - ".exu"
+  - ".exw"
```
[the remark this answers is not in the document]
ghaberek: > It's fine to stick with just `.e` and `.ex` as an override can be used for the others until popular enough for inclusion.

I doubt I'll be able to come up with more than a few dozen instances of those other extensions in the wild so I've reduced the list down to just `.e` and `.ex`. It's a lot simpler to just suggest that people use an override in their repos.

ghaberek:
I added several more samples for Elixir and that seems to have pleased the cross-validation script. 👍

lib/linguist/languages.yml
```diff
@@ -1279,6 +1279,7 @@ E:
   color: "#ccce35"
   extensions:
   - ".E"
+  - ".e"
```
lildude, line 1282: Whoops. I missed this yesterday. Linguist is not case-sensitive so we don't need this.

ghaberek: Okay, I removed `".E"` from that list then.

ghaberek:
Oops, didn't realize I need to mark that resolved. Also pulled in the latest change from upstream. 👍

ghaberek:
Just to clarify, is there anything else needed from my end to get this accepted?

lildude:
Not at the moment thanks. @Alhadis may have some suggestions around the regexes when he gets a chance to have a look, but nothing else is needed at the mo.

lildude:
@ghaberek can you please address the conflicts 🙇 

@Alhadis when you've got a mo, do you mind taking a look at these regexes 🙇

lildude:
Nudge

Alhadis:
> @Alhadis when you've got a mo, do you mind taking a look at these regexes 🙇

Crap. 😞 Sorry. I'll have a look now.

lib/linguist/heuristics.yml
```diff
@@ -619,6 +641,11 @@ named_patterns:
   - '^[ \t]*(class|(using[ \t]+)?namespace)\s+\w+'
   - '^[ \t]*(private|public|protected):$'
   - 'std::\w+'
+  euphoria:
+  - '^\s*(namespace)\s+'
```
Alhadis, line 645: Redundant capturing group, and another redundant quantifier:

~~~suggestion
  - '^\s*namespace\s'

lib/linguist/heuristics.yml
```diff
@@ -619,6 +641,11 @@ named_patterns:
...
+  - '^\s*(?:(public)\s+)?(include)\s+'
```
Alhadis, line 646: As above:

~~~suggestion
  - '^\s*(?:public\s+)?include\s'

lib/linguist/heuristics.yml
```diff
@@ -619,6 +641,11 @@ named_patterns:
...
+  - '^\s*(?:(public|export|global)\s+)?(constant|enum|atom|integer|sequence|object)\s+'
+  - '^\s*(?:(public|export|global)\s+)?(function|procedure|type)\s+'
```
Alhadis, line 648: Both of these expressions are very similar, so I recommend merging them into one so we avoid repeating the first half of the regex:

~~~suggestion
  - '^\s*(?:(?:public|export|global)\s+)?(?:atom|constant|enum|function|integer|object|procedure|sequence|type)\s'
~~~

lib/linguist/heuristics.yml
```diff
@@ -169,6 +182,15 @@ disambiguations:
     pattern: '^\s*(?:%%|main\s*\(.*?\)\s*->)'
   - language: JavaScript
     pattern: '(?m:\/\/|("|'')use strict\1|export\s+default\s|\/\*.*?\*\/)'
+- extensions: ['.ex']
+  rules:
+  - language: Elixir
+    pattern:
+    - '^\s*@moduledoc\s+'
```
Alhadis, line 189: There's no need for the trailing quantifier (`\s+`), because the regex need only match a single whitespace character in order to succeed:

~~~suggestion
    - '^\s*@moduledoc\s'
~~~

lib/linguist/heuristics.yml
```diff
@@ -169,6 +182,15 @@ disambiguations:
...
+    - '^\s*(cond|import|quote|unless)\s+'
```
Alhadis, line 190: To improve performance, it's best to use non-capturing groups `(?:foo|bar)` instead of capturing groups `(foo|bar)` unless backreferences are needed:

~~~suggestion
    - '^\s*(?:cond|import|quote|unless)\s'
~~~

lib/linguist/heuristics.yml
```diff
@@ -169,6 +182,15 @@ disambiguations:
...
+    - '^\s*def(exception|impl|macro|module|protocol)(\(|\s+)'
```
Alhadis, line 191: This can be simplified into:

~~~suggestion
    - '^\s*def(?:exception|impl|macro|module|protocol)[(\s]'
~~~

lib/linguist/heuristics.yml
```diff
@@ -157,6 +157,19 @@ disambiguations:
...
+- extensions: ['.e']
+  rules:
+  - language: E
+    pattern:
+    - '^\s*(def|var)\s+(.+):='
+    - '^\s*(def|to)\s+(\w+)(\(.+\))?\s+{'
+    - '^\s*(when)\s+(\(.+\))\s+->\s+{'
+  - language: Eiffel
+    pattern:
+    - '^\s*{/\w+'
```
Alhadis, line 169: This pattern conflicts with the `E` heuristics above. Consider:

~~~E
def foo()
{

}
~~~

Remember, `\s` will match newlines in addition to tabs and spaces. In fact, if matching horizontal whitespace is your intention, it's much better to be explicit (in which case, replace `\s` with `[ \t]` where applicable).

IMHO, this pattern looks incredibly open-ended; such patterns have a high chance of matching something they weren't supposed to. I recommend deleting it, actually.

lib/linguist/heuristics.yml
```diff
@@ -157,6 +157,19 @@ disambiguations:
...
+    - '^\s*(across|elseif|feature|from|inherit|inspect|note|when)\s+'
```
Alhadis, line 170: This pattern can be simplified (see my explanations below):

~~~suggestion
    - '^\s*(?:across|elseif|feature|from|inherit|inspect|note|when)\s'
~~~

lib/linguist/heuristics.yml
```diff
@@ -157,6 +157,19 @@ disambiguations:
...
+  - language: Euphoria
+    named_pattern: euphoria
```
Alhadis, line 172: Do we really need a named pattern? The patterns are only being referenced once, and they aren't hairy enough to warrant burying at the bottom of the file.

Alhadis (changes requested):

lildude:
Nudge @ghaberek 😉

lib/linguist/heuristics.yml
ghaberek: I used a named pattern here because **Euphoria** is referenced in extensions `['.e']` and `['.ex']`. Is that appropriate?

lib/linguist/heuristics.yml
ghaberek: I think I added this pattern because of a weird side-effect I was seeing when testing linguist against some random E, Eiffel, and Elixir projects. Mainly [this file](https://github.com/EiffelWebFramework/EWF/blob/master/tools/estudio_wizard/rootdir/resources/_vars.e) in EWF which linguist thinks is *Euphoria* but it actually looks like a template. I guess it's already misidentified if linguist thinks it's Eiffel when it's not.

I'm not too familiar with these other languages over all and my goal is just to get Euphoria identified without interfering with any other languages. To date, it seems the heuristics for E, Eiffel, and Elixir in linguist have been mostly automatic but since we're sharing the `.e` and `.ex` extensions, I was running into a lot of false positives early in my testing.

I can remove the pattern since it's a fringe case anyway. Aside from that case I think the heuristics are working well.

lib/linguist/heuristics.yml
Alhadis: Ah, so it is. 🤔 Never mind.

lib/linguist/heuristics.yml
Alhadis: Yes, please do remove it. An inappropriate heuristic can be much worse than no heuristic at all.

Alhadis (changes requested):

Alhadis (approved):

lildude (changes requested):
Most of these samples are massive because it appears (from my naïve perspective) that they contain a lot of comments and a fair amount of duplication.

Please swap these for smaller samples with fewer/smaller comments.

ghaberek:
> Please swap these for smaller samples with fewer/smaller comments.

All set. Unless those are still too big? A lot other files in the `samples/` directory are well over 1k lines.

lildude (dismissed):
Much better samples, thanks. 🙇

lildude:
> Much better samples, thanks. 🙇

Oh, maybe not.

ghaberek:
> Tests are failing with these samples.

    TestHeuristics#test_ex_by_heuristics [/home/runner/work/linguist/linguist/test/test_heuristics.rb:23]:
    no fixtures for Euphoria *.ex

So I just need to provide some `.ex` samples in the **samples/Euphoria** folder?

lildude:
> So I just need to provide some `.ex` samples in the **samples/Euphoria** folder?

I don't think so. I think you need to improve your heuristic as it's not detecting one of the files you're supplying correctly so leaving it to chance with the classifier. 

If we're going to add samples, it would be for Eiffel as there are only 3 very old samples, but I think improving the heuristic would be a better approach for now.

ghaberek:
> ~I don't think so.~
> 
> Update: Whoops, yes... for that error.

I figured that so added more Euphoria samples with `.ex` yesterday.

> I think you need to improve your heuristic as it's not detecting one of the files you're supplying correctly so leaving it to chance with the classifier.

The original Eiffel heuristic I added was only looking for left-anchored keywords, but since any Eiffel keyword is also a valid identifier in Euphoria, things like `from = 2` in Euphoria would match Eiffel's keyword pattern as `^\s*(?:from)\s`. I've added two new patterns to the Eiffel heuristic to match variable and function declaration syntax.

> If we're going to add samples, it would be for Eiffel as there are only 3 very old samples, but I think improving the heuristic would be a better approach for now.

I've added a few more Eiffel samples.

lildude (approved):

ghaberek:
Thanks for accepting this! It's going to be a big help to our community. It looks like syntax highlighting is working now on Euphoria files in my projects, but I'm not seeing the "Languages" statistic update on anything yet. What does it take to make that happen?

Nixinova:
Looks fine to me
image
image

ghaberek:
Strange. I went back to the [OpenEuphoria](https://github.com/OpenEuphoria) page and it was correct there as you showed but not on the [Euphoria](https://github.com/OpenEuphoria/euphoria) project itself.

I *just* took this screen shot. But then I did a `Shift+F5` and it updated to what you show. False alarm I guess. Thanks again!

image

lildude:
You need to push a change to the repo that will cause a re-analysis. Changes to `.gitattributes`, the only file changed in that repo since that new release was deployed, does not cause a re-analysis.

Issue #1740, opened 2014-11-18
Gemfile.lock is not Ruby


bkeepers:
`Gemfile.lock` is classified as Ruby, which means it gets syntax highlighted as Ruby.  #1733 removed the generated designation, which means we'll probably see a lot more complains about the syntax highlighting as soon as that gets deployed.

If we remove `Gemfile.lock` from the list of ruby filenames, it will likely [get detected as JSON](https://github.com/github/linguist/blob/df105c0c33dca43533d7724ed7b724d0b80557c3/lib/linguist/languages.yml#L1187). We really want something like what is described in https://github.com/github/linguist/issues/1571, where we can explicitly describe a file as not code.

One option would be to just add a catch-all entry in languages.yml and set the type to `data` (or something that is ignored) where we could define extensions, filenames, and put samples.

Thoughts?

/cc @arfon @pchaigno @vmg @aroben

arfon:
Closing this as stale. (Still think it's a good idea)

PR #5266, opened 2021-03-09
Add grammar for Gemfile.lock files

hmarr:
## Description

I noticed that `Gemfile.lock` files are currently treated as Ruby code (which they're not), which means the syntax highlighting isn't right. [Here's an example](https://github.com/rails/rails/blob/main/Gemfile.lock).

Given the prevalence of `Gemfile.lock` files (GitHub search [reports > 1M](https://github.com/search?q=filename%3AGemfile.lock)) I thought it'd be worth fixing. This PR adds a [TextMate grammar](https://github.com/hmarr/gemfile-lock-tmlanguage) for `Gemfile.lock` files so we can highlight them correctly.

I see this has been [discussed previously](https://github.com/github/linguist/issues/1740), but that issue went stale and was closed. If you'd rather not include a grammar for a generated file format, no hard feelings! But in that case it'd be good to remove `Gemfile.lock` from the file types classed as Ruby, and fix the misclassification problem described in the issue I linked 😄 

Right now this grammar is in its own repo as that was the path of least resistance. However I'd be more than happy to contribute it to [atom/language-ruby](https://github.com/atom/language-ruby) if that'd be a better place for it.

## Checklist:

      -  https://github.com/search?q=filename%3AGemfile.lock&type=Code _(note: as this grammar works only for files with a specific name, I changed "extension" to "filename" in the search query)_
    - Sample source(s):
      - https://github.com/rails/rails/blob/main/Gemfile.lock
    - Sample license(s):
      - https://github.com/rails/rails/blob/main/MIT-LICENSE
  
  - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fhmarr%2Fgemfile-lock-tmlanguage%2Fblob%2Fmain%2Fsyntaxes%2Fgemfile-lock.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Frails%2Frails%2Fblob%2Fmain%2FGemfile.lock&code=
  - [x] I have updated the heuristics to distinguish my language from others using the same extension. _(I don't think this is necessary as it doesn't match any extensions, but let me know if I've misunderstood!)_


hmarr:
I'm not sure what to do about this test failure:

> Ruby Gemfile.lock has a sample with extension (.lock) that isn't explicitly defined in languages.yml

Adding `.lock` to `extensions` in the `languages.yml` file doesn't feel like the right solution – I assume that'd mean all `.lock` files get associated with this grammar?

Nixinova:
> I'm not sure what to do about this test failure:

The sample file must be placed in a "filenames" subfolder.

lib/linguist/languages.yml
```diff
@@ -4995,6 +4994,15 @@ Ruby:
   - Vagrantfile
   - buildfile
   language_id: 326
+Ruby Gemfile.lock:
+  type: data
+  color: "#701516"
+  searchable: false
```
Nixinova, line 5000: @lildude: What does `searchable` do? It's on e.g. JSON but github.com/search?q=language:JSON&type=code works.

hmarr:
Thanks for the tip @Nixinova, that seems to have done the job. ✨

lib/linguist/languages.yml
lildude: https://github.com/github/linguist/issues/5269

lib/linguist/languages.yml
```diff
@@ -4995,6 +4994,15 @@ Ruby:
   - Vagrantfile
   - buildfile
   language_id: 326
+Ruby Gemfile.lock:
```
Alhadis, line 4997: There's no other type of `Gemfile.lock` in existence other than Ruby's, so I'd rename this to

```suggestion
Gemfile.lock:
```

Alhadis (changes requested):

lib/linguist/languages.yml
hmarr: 👍  done

Alhadis (approved):

lildude (approved):

PR #4138, opened 2018-05-24
Restrict Vim modeline parsing to a single line

jatoben:
## Description

Files which have a similar syntax to Vim modelines, such as YAML, can trigger regex matching behavior that makes filetype detection very slow. In this pull request, I've replaced `\s` in the Vim modeline regex (which considers newlines as whitespace) with the POSIX bracket expression `[[:blank:]]`, which doesn't (`[ \t]` should also work, if that's preferred). This constrains the parsing of each modeline to a single line and avoids the pathological case.

This short YAML file can be used to reproduce the behavior:

```
# vim: ft=yaml:ts=2:sw=2:expandtab
defaults:
  milestone: q1
  priority:  2
branches:
  master:
    events:
      push:
        :state: integrate
        :resolution: change
      merge:
        :state: analyze
        :substate: review
```

Before the change, using Ruby 2.4.2 on Linux:

```
$ time bundle exec bin/linguist pathological.yml 
pathological.yml: 15 lines (13 sloc)
  type:      Text
  mime type: text/x-yaml
  language:  YAML

real	1m3.701s
user	1m3.144s
sys	0m0.156s
```

After:

```
$ time bundle exec bin/linguist pathological.yml 
pathological.yml: 15 lines (13 sloc)
  type:      Text
  mime type: text/x-yaml
  language:  YAML

real	0m0.850s
user	0m0.524s
sys	0m0.104s
```

## Checklist:

- [ ] **I am associating a language with a new file extension.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
[...]

PR #5271, opened 2021-03-12
Fix various oversights in modeline regexes

Alhadis:
This PR fixes a handful of oversights and behavioural discrepancies with the regular expressions used by our [modeline](https://github.com/github/linguist/blob/master/lib/linguist/strategy/modeline.rb) strategy (detailed below).

## Description
Whilst reviewing #4138, I [recommended](https://github.com/github/linguist/pull/4138/files/1d4eaf544d9d994b02f668a55525f87d9ea6ba23#diff-2656b80b42a5a301c510517442e97edb5d8cea635d85b750e68befbf47724cd9) `[ \t]` be used instead of `[:blank:]`. However, I didn't think to check if the submitter applied the changes correctly. Only recently did I notice Vim's modeline regex had an erroneous `[^\\[ \t]]` (which should have been `[^\\ \t]` instead).

Realising there might be other things I missed, I recently took the time to [double-check both regexes](https://github.com/Alhadis/language-regexp/blob/d24ea97fbadbecc6ec21de5cd392ed39b346a2b7/test/1-modelines-test.rb) and verify their behaviour matches that of Emacs and Vim. Among the things I found were:

*	__Vim's modelines are case-sensitive.__  
	`:help modeline` says that modelines must be prefixed by _“the string `vi:`, `vim:`, `Vim:` or `ex:`”_. Mixed-case prefixes like `vIM:` or `ViM:` don't work, so `VIM_MODELINE` shouldn't be flagged as case-insensitive.<br/><br/>
*	__Versioned modelines must start with `Vim` or `vim`.__  
	Our current pattern allows `vi>600: ft=ruby` and `# ex>600: ft=ruby`, neither of which actually work in Vim.<br/><br/>
*	__Emacs's modeline pattern allowed newlines.__  
	This was due to `\s` being used instead of `[ \t]`. An embarrassing habit I've picked up from years of writing TextMate grammars is to assume `\s` only matches *horizontal* whitespace, instead of line-breaks as well. Ergo, something like this (obviously) shouldn't match:<br/>
	```
	-*-
	markdown
	-*-
	```
	<br/>
*	__Modeline mode wasn't disabled.__  
	So Ruby's [insidious treatment](https://github.com/github/linguist/pull/5233#issuecomment-789288012) of `.` meant that `.*?` also matched newlines, which it logically shouldn't. I resolved this by unsetting multiline mode with `(?-m)`.<br/><br/>
*	__Carriage returns were allowed inside `vim: set … :`.__  
	A trivial issue that probably won't ever surface in practice; fixed by updating classes to use `\r\n` instead of just `\n`.

_(Template removed as it doesn't apply)_


lildude (approved):
Nice 😍  I'm surprised we've gone for so long without anyone noticing 🤭

PR #5282, opened 2021-03-18
Make 200+ :user/:repo requirement explicit

lildude:
We've been interpreting "hundreds" in our requirements to mean at least 200 unique `:user/:repo` combinations but haven't been explicit about it in the notes. This PR changes that, and makes the wording consistent in the two places we make this specification.

/cc @Nixinova thanks for the nudge.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ bundle exec bin/github-linguist --breakdown
 ## Adding an extension to a language
 
 We try only to add new extensions once they have some usage on GitHub.
-In most cases we prefer that extensions be in use in at least 200 unique user/repo repositories before supporting them in Linguist.
+In most cases we prefer that each new file extension be in use in at least 200 unique `:user/:repo` repositories before supporting them in Linguist.
 
 To add support for a new extension:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -63,7 +63,7 @@ bundle exec bin/github-linguist --breakdown
 ## Adding an extension to a language
 
 We try only to add new extensions once they have some usage on GitHub.
-In most cases we prefer that extensions be in use in hundreds of repositories before supporting them in Linguist.
+In most cases we prefer that extensions be in use in at least 200 unique user/repo repositories before supporting them in Linguist.
```
Nixinova: The statement "user/repo" can be a bit confusing at first glance if you take the slash to mean "or" instead of a URI separator. What would be a non-"user/repo" repository?

lildude: You're looking at an old commit 😉. I thought that moments after I pushed, hence I changed it in the next commit.

Nixinova: Ah, didn't see that, GitHub seems to like deleting the bottom half of the page some of the time...

Alhadis (approved):

PR #5299, opened 2021-03-30
Add .mo file extension to Mustache

Yash-Singh1:
## Description

Add .mo extension for mustache: https://github.com/tests-always-included/mo

## Checklist:

    - Search results for each extension: https://github.com/search?q=extension%3Amo&type=Code
    - Sample source: https://github.com/github/linguist/pull/5299/files#diff-8278b245d656fdfb2fa5d27026505abd2ca20cd2dc57a984bfab582fe3893cb3

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3511,6 +3511,7 @@ Mustache:
   color: "#724b3b"
   extensions:
   - ".mustache"
+  - ".mo"
   tm_scope: text.html.smarty
   ace_mode: smarty
   codemirror_mode: smarty
--- a/samples/Mustache/sample.mo
+++ b/samples/Mustache/sample.mo
@@ -1,0 +1,7 @@
+A: Hey {{NAME}}!
+
+B: Hi!
+
+A: How is life?
+
+B: It is {{FEELING}}


lildude:
🤔 I'm having an incredibly hard time finding _any_ `.mo` files that are not Modelica or binaries (which mostly appear to be GNU message catalog files used for translations) to qualify inclusion of this extension. I've even used Google BigQuery which allows me to easily exclude the binary files and search for the `{{ }}` syntax and it returns zero results. I can't even find a single case of someone using an override to force `.mo` files to be identified as Mustache either.

This doesn't mean the extension isn't in use on GitHub. It does however mean it's not sufficiently popular that enough people have used it in the last year for it to be included in the indices, or have forced the classification, for us to include support for the extension in Linguist.

Anyone that isn't using the default `.mustache` extension and wishes for their files to be identified as such will need to use an override.

Yash-Singh1:
Yes, I also realized that the popularity is a little bit low. The main usage of the `.mo` extension for `mustache` should be those who use `mustache` in `bash` which is not a thing I see too often. I filtered the results even more to contain shell results, and there are 16 files (~6 repositories) on Github: https://github.com/search?q=extension%3Amo+language%3Ashell&type=code. Maybe the popularity should increase (>200) before merging this PR.

lildude:
> Maybe the popularity should increase (>200) before merging this PR.

Yes, this would be needed. Given this is a very long way from this target, may I suggest we close this PR. You can re-opening it at a later date when usage requirements are met.

PR #5316, opened 2021-04-07
Update overrides.md - * -> **

slutske22:
Including all files and subdirectories requires a `**`, not a `*`.  I tested this out, and there's some discussion of that [here](https://stackoverflow.com/a/42545363/12010984).

Thanks for all you do!

### DIFF
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -53,7 +53,7 @@ Just like vendored files, Linguist excludes documentation files from your projec
 Use the `linguist-documentation` attribute to mark or unmark paths as documentation:
 
 ```gitattributes
-project-docs/* linguist-documentation
+project-docs/** linguist-documentation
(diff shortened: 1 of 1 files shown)


lildude:
As with previous attempts to make this change in https://github.com/github/linguist/pull/4992 and https://github.com/github/linguist/pull/4889, you're making assumptions about the developer's intentions and resorting to using the sledgehammer approach. The developer may have intended to only apply to the files in that directory and not recursing. See https://github.com/github/linguist/pull/4889#issuecomment-759386273 for another discussion on this.

Ultimately, this is really about understanding the git documentation and the developer's end goals, and not making any assumptions, which are outside the scope of using Linguist.

Closing.

slutske22:
Thank you for this comment and the links, I think I understand this better now.  Thanks for your time and the work you do.

PR #5330, opened 2021-04-18
Add Valve Data Format language

Nixinova:
## Description
Adds the [KeyValues](https://developer.valvesoftware.com/wiki/KeyValues) language, used in Source engine game config files.
KeyValues is JSON without punctuation or arrays.
Extension `vdf` + filename `gameinfo.txt`.
Colour `#f26025` from the above site's theme.

## Checklist:
    - Search results:
      - `.vdf`: [10K+](https://github.com/search?q=extension%3Avdf&type=code)
      - `gameinfo.txt`: [1K+](https://github.com/search?type=Code&q=filename:gameinfo.txt) -- `gameinfo.txt` is a bit generic but most of the search results are KeyValues
    - Sample source:
      - `gameinfo.vdf`: Custom, CC0
      - `filenames/gameinfo.txt`: [source](https://github.com/Kenzzer/TF2HLCoop/blob/86097fe5c7f097fbfff4b3c394280cc53c2cadf9/game/tf_coop/gameinfo.txt), [permissive](https://github.com/Kenzzer/TF2HLCoop/blob/master/LICENSE)
  
  - [x] I have included a syntax highlighting grammar: [link](https://github-lightshow.herokuapp.com/?utf8=y&scope=from-url&grammar_url=https://github.com/Nixinova/Nova.tmLanguage/blob/main/grammars/keyvalues.yaml-tmLanguage&code_source=from-url&code_url=https://github.com/Nixinova/Nova.tmLanguage/blob/main/samples/gameinfo.txt)


samples/KeyValues/gameinfo.vdf
```diff
@@ -0,0 +1,17 @@
...
+	{
+		SteamAppId	100
+		SearchPaths 	// search for game paths
+		{
+			Game	|default_path|
+			Game	|default_path|../game
+			Game	example
+		}
+	}
+}
```
lildude, line 17: 🤔 this is the same as the gameinfo.txt file. Is this intended?

Also, are these really real world examples? They look like "hello world"-like to me.

lib/linguist/languages.yml
```diff
@@ -2824,6 +2824,20 @@ Kaitai Struct:
   - ".ksy"
   tm_scope: source.yaml
   language_id: 818804755
+KeyValues:
```
lildude, line 2827: I think we might need a more specific name than this, maybe "Valve Data Format"?

My reasoning is "KeyValues" is very generic and I know many examples where people will think this means key value stores which definitely isn't the case.

lildude (changes requested):

samples/KeyValues/gameinfo.vdf
Nixinova: Switched out the gameinfo.txt.
The vdf one is what I used to test the syntax highlighter and includes more language features than any of the real-world examples which are all either tiny or just `key val key {key val key val}` etc with nothing more.

Nixinova:
`gameinfo.txt` search results are a bit generic and there are only 1K of them; should I remove it?

lildude:
> `gameinfo.txt` search results are a bit generic and there are only 1K of them; should I remove it?

Yes, I think that's probably best.

I think you can improve on the samples too. There are quite a few larger samples in your search results.

lildude (approved):

lildude:
Ooof, I've just noticed the license which I don't before. This sample needs changing ASAP @Nixinova

lildude:
NM, I forgot we don’t ship the full samples. Carry on.

Nixinova:
...

Nixinova:
> I think we might need a more specific name than this, maybe "Valve Data Format"?

>My reasoning is "KeyValues" is very generic and I know many examples where people will think this means key value stores which definitely isn't the case.

Only issue is that the actual name of the language is just "KeyValues" and there's another language I'd want to add to linguist called "KeyValues3". When I add KV3 I think it'd be better to rename VDF to KeyValues for consistency with KV3.

PR #5341, opened 2021-04-26
Add support for Fluent, disambiguate from FreeMarker

### LINKED ISSUE #5340: Fluent language files are incorrectly identified as FreeMarker
## Language name

Fluent

## URL of example repository

https://github.com/FanvidDB/fanviddb-server

https://github.com/projectfluent/fluent/tree/master/spec

## Most popular extensions

`.ftl`

## Detected language

FreeMarker

alerque:
Closes #5340, #4206.

Help wanted, please see comments in #5340. My several attempts at disambiguation from FreeMarker have failed.

Coming up with a Github search query that reliably finds Fluent files is pretty hard. [This query](https://github.com/search?q=extension%3Aftl+NOT+xml+NOT+html&type=Code) finds a few, but it is also littered with thousands that are not Fluent at all. It's quite a bit easier with some regular expressions like the heuristic I propose. Here is a [grep.app query](https://grep.app/search?q=%5E%5B%5Cw-%5D%2B%20%3D%20&regexp=true&filter[path.pattern][0]=ftl) that finds a bunch (not exhaustive).


Nixinova:
/`<\/?#?\w+`/ ("</#if>") seems like it would uniquely identify FreeMarker, unless Fluent would contain HTML-like constructs.
/`[\w-]+\s*=`/ ("foo=") seems to be unique to Fluent, as well as /`{\s*$[\w-]+\s*->\s*[`/ ("{$var -> [").

alerque:
@Nixinova Those would be fairly useful heuristics, but they aren't foolproof. The comment syntax for FreeMarker is pretty unique, but there is no garantee a FreeMarker file is going to have a comment at all, particularly not in any kind of snoop distance from the top. Its other syntax features are much harder do disambiguate, and again *ALL* of them are optional. There is no garantee that a freetype file will have ANYTHING at all unique to the syntax because it could be a pass-through template with no features being used.

Likewise on the Fluent side, it's perfectly reasonable to expect XML data (and even FreeMarker data) in a Fluent file. There are is actually a list of grammar features that must be met to be a valid Fluent file, but checking for them is not trivial.

I guess the question is "How accurate do heuristics need to be to be acceptable to Linguist?" We can easily knock out something between 90%-99% of them by checking for XMLish content at the root (e.g. in Fluent no line is going to start with `^\<` because embedded XML content would need to be indented). But that only covers the cases where people have sane XML formatting and use FreeMarker for XML documents. It doesn't help if they happen to use it for something else or have valid but unsanitary indentation.

lildude:
The heuristics don't need to be 100% perfect. The classifier has also improved quite considerably since I made my comment https://github.com/github/linguist/issues/4206#issuecomment-406893073 back in 2018 so we can allow things to fall through to it with a higher degree of confidence that it might get it right. 

As such, the heuristics need only cover the easily identifiable features of most use-cases and then allow the analysis to fall through to the classifier for the tougher cases. In order to do this, the heuristics should not have a "default" language if the heuristics don't match. ie if you think of the heuristics as an if/elif/else statement, we don't want a final "else".  

The heuristics for `.asc` is an example where we don't have a final "else" so if none of the heuristics match, we'll fall through to the classifier:

https://github.com/github/linguist/blob/32ec19c013a7f81ffaeead25e6e8f9668c7ed574/lib/linguist/heuristics.yml#L68-L75

This will also require at least two highly representative real-world samples of each language. The more unique features of each language covered by the samples, the better the classifier will be.

Take a look at the [classifier cross-validation test](https://github.com/github/linguist/blob/32ec19c013a7f81ffaeead25e6e8f9668c7ed574/.github/workflows/ci.yml#L43-L56) we run in CI too. This tests the classifier independently of the heuristics to see if it can successfully classifier the samples itself. Warning: the output is quite verbose but you should be able to find your way through it.

alerque:
I've imported a grammar and setup heuristics for both FreeMarker and Fluent. Let me try to explain the regex logic.

 In both cases I've based the heuristic on Fluent syntax because of the two file formats it is the only one with *required* anything. The [EBNF I based the following expressions on is here](https://github.com/projectfluent/fluent/blob/master/spec/fluent.ebnf). FreeMarker is impossibly loosy-goosy. It is so free it is nothing. I started searching through repositories that use it and found **thousands** of examples where FreeMarker files are used entirely to contain code snippets in other languages and they include **zero** FreeMarker syntax. In other words they are 100% pass through files to other formats. XML, HTML, and Java code are the worst offenders, but I found others too.

Hence the FreeMarker heuristic finds either of two things we know cannot be (valid) Fluent. The Fluent format was designed to handle "junk" content and the spec includes a way to detect and regress without failing to parse anything, but for the sake of this heuristic I've decided that anything that would be considered Junk by Fluent is fair game for assuming it is FreeMarker.

```yaml
- language: FreeMarker
pattern: '^([^ a-zA-Z#\r\n-]+|[^ \n#]+ +[^=])'
```

Both groups are matchin the start of a line. The first potential match finds anything that isn't a valid first character of a line. In Fluent everything is either going to be a comment, an entry, or junk. Entries can oly start with a space (not even a tab!), a letter, or a dash. Anything else and it's not a valid entry. It could contain XML code, but it would **have to be** indented at least 1 space. Notably this bit of expression catches lines starting with `<` (XML, HTML), but only as the first character.

The second group goes for something that turns up in many of the code snippets I found: lines like `import foo`. The "import" bit could be Fluent, but having a space would terminate the Identifier and the only valid thing to follow an identifier is `=`. Anything else and this must not be Fluent.

This heuristic will produce false positives **only** in the event of a Fluent file that contains Junk. It will not catch all potential FreeMarker files, but it should rule out a vast majority of them from being Fluent files.

```yaml
- language: Fluent
pattern: '^-?[a-zA-Z][a-zA-Z0-9_-]* *='
```

The Fluent heuristic is based entirely on the Entry format, specifically valid identifiers. We can't assume anything based of Fluent's comment syntax, but we can make some presumptions Because extended ASCII isn't allowed and Fluent is strict about spaces, we can assume that any Fluent file worth much is going to have at least one Entry with a valid identifier in it.

This is potentially much more aggressive than the FreeMarker heuristic and has the potential for many false positives (e.g. many languages could have `foo = 6`), but given that FreeMarker is the only other thing using `.ftl` extensions and that any language being used inside FreeMarker is highly likely to have other things that trip its own heuristic I think it might be safe to assume that anything that doesn't match FreeMarker and does match at least one valid Fluent Identifier might actually be Fluent.

----

As for testing, I'm not sure how to do that at scale. I've thrown a few of my own project files at this and it seems okay, but that's pretty limited. @lildude the contributor guide mentions pinging you for how to do this more broadly.

lildude (changes requested):
🤔  something's gone awry with the cached license files. I'd expect to only see the addition of the licence for the new grammar and no changes to the others. Please remove the other changes. If there are license changes elsewhere, they'll be updated in the next release.

Is the sample also reflective of real world usage?

Can you please add another real-world Fluent sample (or more, but clearly different, if they're only likely to be very small) so we've got at least two to help the classifier in those cases that fall through the heuristic.

alerque:
The license file and module changes were because I merged from upstream master which has those changes since I branched. It wouldn't have affected anything, but I went ahead and rebased instead so the PR review looks cleaner.

I also added several more samples of Fluent including a longer one extracted from Firefox used across at least half a dozen language implementations for benchmarking, and several of the trickier cases from the upstream fixtures. I didn't include all the fixtures, but I did test them and the only one that fails in one that deliberately has invalid syntax used to test the fault tolerance.

In the process I discovered I had to be less aggressive with the FreeMarker heuristic. I had previously covered all the valid Fluent syntax *except* the embedded FUNCTION syntax. Unfortunately anything inside a FUNCTION is parsed completely differently than normal entries, and a large variety of things might appear at the start of a line. The only one I could find that cannot that seems useful for identifying FreeMarker was `<`, commonly used in XML/HTML.

.gitmodules
```diff
@@ -1051,9 +1051,9 @@
 	path = vendor/grammars/vscode-codeql
 	url = https://github.com/github/vscode-codeql
 	branch = main
-[submodule "vendor/grammars/vscode-cue"]
-	path = vendor/grammars/vscode-cue
-	url = https://github.com/cue-sh/vscode-cue
```
lildude, line 1056: Whoops. The CI failure is happening because you're accidentally removing the Cue grammar.

alerque: Really? Ooops! Stand by...

alerque: Fixed.

lildude:
ProTip: we squash all commits on merge so feel free to push just your incremental changes without rebasing etc.

lib/linguist/heuristics.yml
```diff
@@ -200,6 +200,12 @@ disambiguations:
     pattern: '^\s*(#version|precision|uniform|varying|vec[234])'
   - language: Filterscript
     pattern: '#include|#pragma\s+(rs|version)|__attribute__'
+- extensions: ['.ftl']
+  rules:
+  - language: FreeMarker
+    pattern: '^(\<|[a-zA-Z-][a-zA-Z0-9_-]+ +\w)'
+  - language: Fluent
+    pattern: '^-?[a-zA-Z][a-zA-Z0-9_-]* *='
```
lildude, line 208: @Alhadis when you've got a mo, do you mind taking a look at these regexes 🙇

Alhadis (commented):
I've pushed some additions to @alerque's branch, most of them taken from [FreeMarker's syntax](https://freemarker.apache.org/docs/dgui_template_overallstructure.html) docs.

Alhadis (approved):

alerque:
I appreciate the help getting this read @Alhadis. I'm a little concerned these heuristics don't smell right to me now. Before I detail objectionable cases though let me make sure I understand the goal of heuristics correctly. My understanding is they are a last resort to make a definitive call on whether something is one language or another in cases where the statistical classifier fails to be certain.

As such my understanding was that the goal was to catch very definite clues that something must be or cannot be a language while creating zero false positives. My goal with the expressions was to only create a positive match where the heuristic was 100% sure: preferring to leave a small percentage of files unidentified by heuristics rather than generating false positives.

Is that not the case?

lildude:
I'll let @Alhadis  comment on the regex changes but I can say you've got this the wrong way round:

> My understanding is they are a last resort to make a definitive call on whether something is one language or another in cases where the statistical classifier fails to be certain.

The classifier is the last resort if all the strategies before it, including the heuristics, don't manage to whittle the languages down to a single language.

This is the order of the strategies:

https://github.com/github/linguist/blob/19b3a9bd7d585be8d6459cfab0a51607b2aee68b/lib/linguist.rb#L61-L71

Alhadis:
> Is that not the case?

Nope. Think of heuristics as a nudge in the right direction, rather than definitive, all-encompassing logic that needs to accomodate every possibility. Most of the time, heuristics aren't needed: our classifier does a good job disambiguating most languages (unless our [sample](https://github.com/github/linguist/tree/master/samples) coverage is poor or unrealistic).

However, some languages give the classifier trouble—noticeably those with objectively-similar syntax (C/C++/Objective-C/Objective C++) and those that're mostly prose with very little machine-readable bits (FreeMarker/Fluent and similarly general-purpose templating languages). When healthy sample coverage fails to improve flaky classification, *that's* when a heuristic is needed.

Here's the more readable versions of the changes I made:

<details><summary><b>FreeMarker heuristic</b> <i>(click to show)</i></summary>

~~~regexp
(?x:
	# Case 1: "<…"
	^ <
|
	# Case 2: "foo b…"
	^ [a-zA-Z-][a-zA-Z0-9_-]+ [ \t]+ \w
|
	# Case 3: "${foo…}"
	\${ \w+ [^\n]*? }
|
	^ [ \t]*
	(
	# Default syntax (angle brackets)
	# https://freemarker.apache.org/docs/dgui_template_directives.html
	
		# Case 4: Comment: "<#-- … -->"
		<\# -- .*? -->
		|
		
		# Case 5: FTL tag: "<#foo> … </#foo>"
		<\# (?<tag_name>[a-z]+) (?=\s|>) [^>]* >
			.*?
		</\# \k'tag_name'/>
		|
		
	# Alternative syntax (square brackets)
	# https://freemarker.apache.org/docs/dgui_misc_alternativesyntax.html
		
		# Case 6: Comment: "[#-- … --]"
		\[\# -- .*? -- \]
		|
		
		# Case 7: FTL tag: "[#foo] … [/#foo]"
		\[\# (?<tag_name>[a-z]+) (?=\s|\]) [^\]]* \]
			.*?
		\[\# \k'tag_name' \]
	)
~~~

</details>


<details><summary><b>Fluent heuristic</b> <i>(click to show)</i></summary>

~~~regexp
(?x:
	# Case 1: "-term-definition =…"
	#         "message-text =…"
	^-? [a-zA-Z][a-zA-Z0-9_-]* \x20* =
|
	# Case 2: "{$variable}"
	\{\$
		-?[a-zA-Z][-\w]*      # $variable-name
		(?:\.[a-zA-Z][-\w]*)? # .attribute
	\}
)
~~~

</details>

Now, with @lildude's explanation in mind, you'll notice the patterns this time are much more specific, albeit less likely to cover everything. That's okay: anything *not* matched by a heuristic is passed down to our classifier.

Anything that it fails to recognise is left to the user to correct if they choose.
Linguist doesn't try to be perfect, though part of that is by design—classifying repositories is a *very* resource-intensive process, and with millions of files in need of analysing every day, an inexpensive and user-correctable mistake is more justifiable than an infallable one with an exorbitant CPU cost.

alerque:
Thank you both @lildude and @alhadis for the explanations. I think I'm okay with calling this good and moving on and if you guys are ready to be done indulging my confusion. You both have more experience with linguist and Ruby than I do.

That being said, it might be just my lack of caffeine but your explanations seem to contradict each other:

From Colin [this comment](https://github.com/github/linguist/pull/5341#issuecomment-844975234):

> The classifier is the last resort if all the strategies before it, including the heuristics, don't manage to whittle the languages down to a single language.

From John in [this comment](https://github.com/github/linguist/pull/5341#issuecomment-845148008):

> Most of the time, heuristics aren't needed: our classifier does a good job disambiguating most languages [...] When healthy sample coverage fails to improve flaky classification, that's when a heuristic is needed.

As far as I can get my head around this the order matters as each method whittles down the possibilities and the processing stops when only a single possibility is left at the end of a step. **Hence my concern about false positive heuristics.**

If `Strategy::Extension` brings the possibilities down to just two, FreeMarker and Fluent, then the expression in the `Heuristics` comes back with 100% certainty that it is one language and not the other another, isn't that going to leave `Classifier` out of the loop entirely? That's why I was aiming to catch the bulk of cases correctly but let the remaining uncertain ones fall through rather than generate false positive matches.

If the heuristic was a nudge when the classifier failed to have any certainty then the current expressions might be useful, but if the buck could ever stop there then this won't work super well. I'm pretty certain I can come up with valid test cases in either language that could trip either one of those expressions while actually being the *other* format. Remember both are template systems and either language can be validly nested inside the other.

lildude:
> If `Strategy::Extension` brings the possibilities down to just two, FreeMarker and Fluent, then the expression in the `Heuristics` comes back with 100% certainty that it is one language and not the other another, isn't that going to leave `Classifier` out of the loop entirely?

Yes, and that's fine. The classifier was historically very poor at identifying files so a lot more emphasis was put on getting it right in the heuristics. The classifier is much better these days but as @Alhadis pointed out, it's still not very good in certain circumstances:

> However, some languages give the classifier trouble—noticeably those with objectively-similar syntax (C/C++/Objective-C/Objective C++) and those that're mostly prose with very little machine-readable bits (FreeMarker/Fluent and similarly general-purpose templating languages). When healthy sample coverage fails to improve flaky classification, that's when a heuristic is needed.

As @Alhadis pointed out, we'd still fall through to the classifier if a file doesn't match anything in the heuristics. We also don't need to be perfect, just good enough for most cases. Overrides exist for the outliers that fall through.

lildude (approved):

macabeus:
I don't know where to report it, and so I'm writing here.
On some repositories with Fluent code, GitHub is summarizing it as FreeMaker, for example: https://github.com/projectfluent/fluent
image

Is it expected?

lildude:
The changes in this PR were included in Linguist v7.15.0 that was deployed to GitHub.com on 1 Jun. The last push to your referenced repo @macabeus was 24 May... so before these changes were rolled out. You need to push a change to your repo to force a re-analysis with the newer code as we do not go back and re-analyse all repos everytime we update Linguist as this would be insanely resource intensive for very little gain.

PR #5357, opened 2021-05-06
Classify .h files as C by default

lildude:
## Description

https://github.com/github/linguist/issues/1626 is Linguist's longest running and most commented on issue. It is also effectively the most "duplicated" issue we've ever had.

This all stems from the fact that C, C++, and Objective-C all use `.h`  header files and unless they contain language-specific code to match the heuristics, fall through to the classifier for final classification. This results in quite a few misclassifications which may be unexpected or surprising to some developers.

This PR attempts to resolve this by implementing what feels to be the general consensus of the comments in https://github.com/github/linguist/issues/1626: assume all `.h` header files are C unless they specifically match C++ or Objective-C specific syntax.  This PR does that by adding C as the final option in the heuristics.

This means we'll never fall through to the classifier for `.h` files which I think is a happy compromise and ensures predictable behaviour.  I've added a special section to the troubleshooting docs to explain the behaviour too.

By doing this, I've needed to move two of the ambiguous C++ samples to C - I'm pretty sure no one could determine if they're C, C++ or Objective-C if viewed in isolation.

Fixes https://github.com/github/linguist/issues/1626

## Checklist:

    - Sample source(s):
    - Sample license(s):


MachineThing:
To be honest I like this. .h files are meant for C while .hop files are meant for C++.

Alhadis (approved):

Alhadis:
> To be honest I like this. .h files are meant for C while .hpp files are meant for C++.

Exactly. As for Objective-C, well… that's arguably a lot less common than C, especially nowadays with Swift replacing its use in a lot of macOS/iOS projects.

lildude:
I'm going to merge this now but the changes won't take effect until I make the next release. I've scheduled time for the week of 24 May to make and deploy the next release.

Issue #5347, opened 2021-04-30
Larger Vim plugin help file cannot be highlighted


rhysd:
## Describe the bug

I'm maintainer of [vim-lsp](https://github.com/prabirshrestha/vim-lsp). The documentation is not highlighted on GitHub although its mode line description at bottom of the file is correct.
To reproduce this, open https://github.com/prabirshrestha/vim-lsp/blob/master/doc/vim-lsp.txt in your browser and check if highlight is applied.

### Expected behaviour

Highlight is applied [like this](https://github.com/rhysd/git-messenger.vim/blob/master/doc/git-messenger.txt)

### Related discussion

Today I investigated why the document is not highlighted on GitHub.

[The regex for heuristic](https://github.com/github/linguist/blob/32ec19c013a7f81ffaeead25e6e8f9668c7ed574/lib/linguist/heuristics.yml#L574) matched correctly to modeline at the bottom of the document file.

However, the document file is larger than 51200 bytes. And the heuristic detection has [51200 bytes limit](https://github.com/github/linguist/blob/841803f28777daa3442aa0fab8ca6cd16d674842/lib/linguist/heuristics.rb#L6) on finding the regex pattern. I understood that moving the modeline to top of the document file will make the heuristic detection work, but it's not possible because the format of Vim help file does not allow it. The first line is treated as summary of the document by Vim so modeline cannot be put at the line. On the other hand, Vim's mode line must be put on the first line or the last line of the file (please see `:help modeline` for more details). After all, modeline must be put at the last line in Vim help file. It means that Vim help files larger than 51200 bytes cannot be highlighted currently.

### Additional notes

I think this issue is important for Vim plugin users because larger document is harder to be read and searched. Highlight would help to do them, but currently highlight is not applied due to this issue for larger documents.

obcat:
> Vim's mode line must be put on the first line or the last line of the file

I don't think this is true.

The `:set` line can be put at, for example, line 3 to be recognized as a modeline unless the value of the `modelines` option (default: 5) is less than 3 (please see `:help 'modelines'`).

rhysd:
Yes, that's correct. But it's up to user's configuration. For example, when setting the value to `1`, what I said is true. Since Vim help file cannot assume some user setting is set to specific value, I can't say putting the modeline at line 3 is always ok.

And also, putting modeline at line 3~5 is not good in terms of Vim's help file format. Usually table of contents is put at top of the file. Putting mode line just before the table looks weird.

[...]

PR #5365, opened 2021-05-13
D.R.Y up Vim modeline expression

Alhadis:
## Description
Fixes:#5347 by updating the Vim Help heuristic with fixes made in #5271 (which should've included `heuristics.yml` as well… 🤦). I've added a reminder to `modeline.rb` so something like this doesn't happen again.

*(Template removed as it doesn't apply)*


lildude:
I think we also need to add `help` as an alias for "Vim Help" in order to fix https://github.com/github/linguist/issues/5347

Alhadis:
Derp. 🤦🤦 You're right.

At least one of us is paying attention. 😅

Alhadis:
@lildude We've hit a snag:

~~~
  2) Failure:
TestBlob#test_language [/home/runner/work/linguist/linguist/test/test_blob.rb:266]:
textobj-rubyblock.vba.
Expected: "Vim script"
  Actual: "Vim Help File"
~~~

This is failing because a VimBall file (see #4412) ends with a help file modeline:

https://github.com/github/linguist/blob/19b3a9bd7d585be8d6459cfab0a51607b2aee68b/samples/Vim%20script/textobj-rubyblock.vba#L215

As the name suggests, VimBalls are the Vim equivalent of a tarball: a concatenation of several files with headers betwixt each blob of data:

https://github.com/github/linguist/blob/19b3a9bd7d585be8d6459cfab0a51607b2aee68b/samples/Vim%20script/textobj-rubyblock.vba#L62-L66

I'm not sure what to do here.

lildude:
Oh (vim)balls!!! 

> I'm not sure what to do here.

Yeah, I'm not too sure either. That change is needed in order to fix https://github.com/github/linguist/issues/5347 (cos of the size of the file) but doing so breaks VimBall detection.  I'll  need to have a think about it.

lildude:
Got it!! Vimballs are really easy to find thanks to the first few lines always containing `UseVimball`.

We can use this to our advantage in the modeline strategy with no extra cost. 

I'll push my changes to this branch.

lib/linguist/strategy/modeline.rb
```diff
@@ -144,6 +145,9 @@ def self.call(blob, _ = nil)
         return [] if blob.symlink?
 
         header = blob.first_lines(SEARCH_SCOPE).join("\n")
+        # Return early for Vimball files as their modeline will not reflect their filetype.
+        return [] if header.include?("UseVimball")
+
```
lildude, line 150: What do you think of this @Alhadis?

Alhadis: Looks great! (I was about to suggest that a regression test be added, but remembered that most samples are implicitly used as fixtures anyway... 😅)

lildude (approved):

PR #5415, opened 2021-06-09
Switch out Valve Data Format sample

Nixinova:
@lildude in https://github.com/github/linguist/pull/5330#issuecomment-857465206:
> Ooof, I've just noticed the license which I don't before. This sample needs changing ASAP @Nixinova

Sample from [vdf-rs](https://github.com/LovecraftianHorror/vdf-rs) [file](https://github.com/LovecraftianHorror/vdf-rs/blob/6fc56a024db14c0083d60cbd48e45ad14f74d5ef/keyvalues-parser/tests/assets/app_manifest.vdf) under dual Apache2/MIT.


Nixinova:
Just as I hit enter...
https://github.com/github/linguist/pull/5330#issuecomment-857471617
> NM, I forgot we don’t ship the full samples. Carry on.

PR #5434, opened 2021-06-24
Fix incorrect capitalisation in `.properties` heuristic

Nixinova:
```diff
- - language: Java properties
+ - language: Java Properties
```

## Description

The word "properties" in "Java Properties" was not capitalised in one instance, which leads to a mismatch between the heuristic result and language name.

For me this led to my [Linguist port](https://github.com/Nixinova/Linguist) throwing as it couldn't find an associated language for this heuristic (the port fetches this repo's yml files).

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -445,7 +445,7 @@ disambiguations:
     - pattern: '^[#!]'
   - language: INI
     named_pattern: key_equals_value
-  - language: Java properties
+  - language: Java Properties
     pattern: '^[^#!][^:]*:'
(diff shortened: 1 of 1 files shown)


lildude (approved):
Well spotted!

Alhadis:
That *was* well-spotted, and it's an incredibly easy mistake to miss, too.

We should really add a test (perhaps in [`test_pedantic.rb`](https://github.com/github/linguist/blob/e8de1c0c9ae29984651661e9e242aca3d753dc68/test/test_pedantic.rb)) to ensure a similar mistake doesn't hide in plain sight ever again.

PR #5437, opened 2021-06-27
Rename language entry for e-mail files

Alhadis:
## Description
This PR renames the [`languages.yml`](https://github.com/github/linguist/blob/16c70aef8cd62ca071231a380c69050f5e83c900/lib/linguist/languages.yml#L1382) entry for EML to something more descriptive and self-documenting. The acronym "EML" is only used as a [file-extension](https://www.loc.gov/preservation/digital/formats/fdd/fdd000388.shtml) for messages that've been saved offline. It's been retained as an alias for the benefit of projects whose discussions or readme files contain ` ```eml `-fenced code-blocks.

In addition, I've added CodeBlock's [`mbox`](https://codemirror.net/mode/mbox/index.html) mode to the entry, since there is one.

*(Checklist removed as it doesn't really apply)*


lildude (approved):
Makes sense to me.

PR #5461, opened 2021-07-13
Improve parser for env command's args in shebang

eggplants:
Support the use of `env` commands with arguments in shebang.

## Description
I changed `shebang.rb` to properly parse env commands with arguments when they are used in Shebang.

e.g.)
- `#!/usr/bin/env -vS ruby -wKU`
- `#!/usr/bin/env --split-string sed -f`
- `#!/usr/bin/env -S GH_TOKEN=ghp_*** deno run --allow-net`

## Checklist:


lildude (changes requested):
Can you please add examples to list of `env` examples in the [test_shebangs test](https://github.com/github/linguist/blob/979884174716af23967b5e2a5ef7499f39be40df/test/test_strategies.rb#L128)

lildude (approved):
LGTM. Thanks.

PR #5490, opened 2021-07-29
Add file_version_info.txt to list of Python filenames

eggplants:
I ~moved `.gclient` from `filenames` to `extensions` and~ added `file_version_info.txt` to filenames in Python.

### ~`.gclient`~

~It seems that `.gclient` is more often used as an extension than a file name.~
~See: <https://github.com/search?q=extension%3Agclient&type=Code> (200 hits)~

### `file_version_info.txt`

`file_version_info.txt` is a filename that is often used to add version information to exe programs created by [Pyinstaller](https://github.com/pyinstaller/pyinstaller).
It is the default name of the file to which the version information is extracted from an existing exe program and written, like `pyi-grab_version <exe file>`.

## Checklist:

      - ~`.gclient`: <https://github.com/search?q=extension%3Agclient&type=Code> (200 hits)~
      - `file_version_info.txt`: <https://github.com/search?p=1&q=filename%3A%22file_version_info.txt%22&type=Code> (319 hits)
    - Sample source(s):
      - ~[test.gclient](https://github.com/eggplants/linguist/blob/fix_python_ext/samples/Python/test.gclient)~
      - [OmerFI/PyFlappy/file_version_info.txt](https://github.com/OmerFI/PyFlappy/blob/main/file_version_info.txt)
    - Sample license(s):
      - [MIT License](https://github.com/OmerFI/PyFlappy/blob/main/LICENSE)
  
### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4684,6 +4684,7 @@ Python:
   - SConscript
   - SConstruct
   - Snakefile
+  - file_version_info.txt
   - wscript
   interpreters:
   - python
--- a/samples/Python/filenames/file_version_info.txt
+++ b/samples/Python/filenames/file_version_info.txt
@@ -1,0 +1,43 @@
+# UTF-8
+#
+# For more details about fixed file info 'ffi' see:
+# http://msdn.microsoft.com/en-us/library/ms646997.aspx
+VSVersionInfo(
+  ffi=FixedFileInfo(
+    # filevers and prodvers should be always a tuple with four items: (1, 2, 3, 4)
+    # Set not needed items to zero 0.
+    filevers=(1, 0, 0, 0),
+    prodvers=(1, 0, 0, 0),
+    # Contains a bitmask that specifies the valid bits 'flags'r
+    mask=0x0,
+    # Contains a bitmask that specifies the Boolean attributes of the file.
+    flags=0x0,
+    # The operating system for which this file was designed.
+    # 0x4 - NT and there is no need to change it.
+    OS=0x4,
+    # The general type of file.
+    # 0x1 - the file is an application.
+    fileType=0x1,
+    # The function of the file.
+    # 0x0 - the function is not defined for this fileType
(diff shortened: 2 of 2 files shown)


lildude (changes requested):
> It seems that `.gclient` is more often used as an extension than a file name.

Unfortunately, that's not a reason to remove it from the list of name files... it might not be as common anymore, but it certainly is still used as a standalone file so can't be removed from the list. It's also explicitly called out in the [gclient docs](https://pypi.org/project/gclient/) so we definitely don't want to remove it. In fact, we rarely remove support for something once added. 

Please revert this change.

> See: https://github.com/search?q=extension%3Agclient&type=Code (200 hits)

That's most definitely not going to meet our minimum requirement of 200 unique `:user/:repo` combinations needed to add an extension when I can see three in the same repo on the first page of results 😁 

You're going to have to wait for usage to grow before this extension can be added.

>  * `file_version_info.txt`: https://github.com/search?p=1&q=filename%3A%22file_version_info.txt%22&type=Code (319 hits)

This looks to meet the requirements as it appears to be one file per repo, so we might be able to accept only this change; I'll need to do a proper popularity check when I've got the time.

Please update the OP to link to the source of the sample `file_version_info.txt` file and state the license it is covered by... your fork of Linguist is not the original source of the file 😉

eggplants:
It's too small? If so, please close this PR.

```bash
# count 
PAT='***'
URL="https://api.github.com/search/code?q=filename%3Afile_version_info.txt&per_page=100&access_token=$PAT"
n="$(curl -s "${URL}" | grep -oP '(?<=total_count.: )[^,]+')"
curl -s "${URL}&page=[1-$((n/100+$((n%100==0?0:1))))]" > _
grep -oP '(?<="full_name": ")[^"]+(?=")' _ | sort | uniq | wc -l
#=>119
```

Nixinova:
> 119

Yep that's too small, Linguist needs files that are used in 200+ repos. And there's barely over 200 instances of the file full stop.

PR #5424, opened 2021-06-14
Add ECL language grammar

GordonSmith:
Update samples folder

Signed-off-by: Gordon Smith <GordonJSmith@gmail.com>

## Description
Adding grammar to existing ECL language 
Updating samples for ECL language

Search results:  https://github.com/search?q=extension%3Aecl+OUTPUT+DATASET&type=Code 

## Checklist:

- [ ] **I am associating a language with a new file extension.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  
  - [ ] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

- [ ] **I am fixing a misclassified language**
  - [ ] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am changing the source of a syntax highlighting grammar**
  
  - Old: https://github-lightshow.herokuapp.com/
  - New: https://github-lightshow.herokuapp.com/

- [x] **I am updating a grammar submodule**
  

- [ ] **I am adding new or changing current functionality**
  
  - [ ] I have added or updated the tests for the new or changed functionality.

- [ ] **I am changing the color associated with a language**
  
  - [ ] I have obtained agreement from the wider language community on this color change.
    - [URL to public discussion]
    - [Optional: URL to official branding guidelines for the language]

PR #5500, opened 2021-08-04
Set tm_scope for ECL language

lildude:
## Description

I completely missed that the PR that added the ECL grammar (https://github.com/github/linguist/pull/5424) didn't include an update to the `tm_scope` needed to identify which grammar to use so it continued to use `none` which means no syntax highlighting.

This PR sets the scope to `source.ecl` and also updates the `vendor/README.md` with the missing line that was also missed because the `tm_scope` wasn't updated at the time.

/cc @GordonSmith 

## Checklist:

  - [N/A] I have added or updated the tests for the new or changed functionality.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1411,7 +1411,7 @@ ECL:
   extensions:
   - ".ecl"
   - ".eclxml"
-  tm_scope: none
+  tm_scope: source.ecl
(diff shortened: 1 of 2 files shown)


Alhadis (approved):
Oof.

bossman84 (commented):
#

Issue #4982, opened 2020-08-31
Proposal: Make triaging more welcoming to external contributors


smola:
I've contributed on and off to linguist since 2017 and over time I've observed a quite frustrating pattern with issues and pull requests: legitimate issues and pull requests get labelled quite quickly as stale (30 days) and then closed (60), with no further option to reopen them.

I understand that this workflow may be convenient for maintainers, but it seems discouraging in a few ways:

- For a casual contributor, 60 days may be a too strict deadline for a Pull Request. It is quite frustrating to come back at a Pull Request a couple of months later (holidays or whatever reason for inactivity) to find you need to create a new one. 

- For issues, closing an issue just because there is no PR in 60 days hides them from people who are looking for active ways to contribute. It also tends to promote filing duplicates without reading any of the previous discussion.

I think this could be improved in a way that promotes external contributions while maintaining convenience for maintainers. There are probably many good ways, but I'm thinking about something along the following lines:

- Pull Requests: Do not close them automatically (or extend the grace period). There's 73 PRs closed as Stale over a ~2 years period. A few of them were mostly ready, so the project could probably benefit from keeping them alive for a while.

- Issues: Once an issue is confirmed to be a valid bug or feature request, add a label (e.g. `Help Wanted`) and exclude it from the Stale workflow. This label would be a great way for people looking to contribute, while making it clear to the reporter that it needs someone to work on it. Keeping them open and clearly marked also makes them more easily discoverable by users experiencing the same issue, so they can contribute to the discussion.

What do you think?

Alhadis:
This issue has been automatically marked as stale because it has not had activity in a long time. If this issue is still relevant and should remain open, please reply with a short explanation (e.g. "I have checked the code and this issue is still relevant because ___."). Thank you for your contributions.

Alhadis:
Kidding.

> What do you think?

😜 I think we should torch the annoying robot for good, possibly encasing its corpse in cement so it doesn't come back as a zombie.

Since I'm not the one in charge of making that decision, though, I agree that PRs deserve a longer grace period. Hell, we probably shouldn't close them at all — if a PR needs changes, and the submitter's become unresponsive, it's easier if a maintainer just pushes changes to their branch before merging. If the PR's obviously invalid, and the user doesn't close it themselves, a maintainer can close the issue manually.

lildude:
[...]

PR #5502, opened 2021-08-05
Remove stalebot config

lildude:
## Description

Stalebot's time has come to an end. I liked the idea initially, but it didn't age very well, and with the greater use of labels and encouraging peeps to use discussions it's been of no use so I'm 🔥 it.

Closes https://github.com/github/linguist/issues/4982

/cc @smola


Alhadis (approved):
Good riddance.

Issue #619, opened 2013-08-14
How do I apply vendor.yml to exclude Vagrant and Chef /cookbooks?


Bartvds:
I have some repos that get reported in the wrong language. My projects are JavaScript but my Vagrant + Chef cookbooks makes Github think it's mostly Ruby.

So I look at this FAQ: https://help.github.com/articles/my-repository-is-marked-as-the-wrong-language. And it points me to the vendor.yml file here in this library.

But what do I need do to get my Chef  /cookbooks folder excluded? 

Do I need to fork/merge this and have it applied globally for everybody on Github? That seems a bit drastic.

Do I put a vendor.yml file in my repos? 

Should I pull my /cookbooks out of my repos? But them how do I share them?

I'm a bit lost on what to do with the FAQ item.

PR #5506, opened 2021-08-07
Add typical Gitpod custom Dockerfile name

i-ky:
This adds [custom Gitpod Dockerfile](https://www.gitpod.io/docs/config-docker#configure-a-custom-dockerfile) to the list of vendor files. It is used to setup a development environment and is similar to Vagrantfile in this respect, which already has its own entry in `vendor.yml`. I think it is fair to add an entry for `.gitpod.Dockerfile` too and stop reflecting it in repository's language statistics. Otherwise it may give a false impression that project provides Docker images.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -380,3 +380,6 @@
 
 # Jenkins Pipeline
 - (^|/)Jenkinsfile$
+
+# Gitpod
+- (^|/)\.gitpod\.Dockerfile$


Alhadis:
Would [`documentation.yml`](https://github.com/github/linguist/blob/2d1afbece812f70c62a54b56572a3ca7d0770cb3/lib/linguist/documentation.yml) be a more appropriate place for this pattern? Both "vendored" and documentation files are excluded from statistics in the same way; I'm not sure why/how we differentiate between the two.

Regarding eligibility, usage checks out with [~27,829 `.gitpod.Dockerfile` files](https://github.com/search?q=filename%3A.gitpod.Dockerfile+NOT+nothack&type=Code) currently indexed on GitHub.

i-ky:
> Would [`documentation.yml`](https://github.com/github/linguist/blob/2d1afbece812f70c62a54b56572a3ca7d0770cb3/lib/linguist/documentation.yml) be a more appropriate place for this pattern?

Looking at content of `documentation.yml` I don't think that `.gitpod.Dockerfile` fits there.

> Both "vendored" and documentation files are excluded from statistics in the same way; I'm not sure why/how we differentiate between the two.

If so, maybe they should be merged together into one file (e.g. `ignore.yml`)?

lildude:
> Both "vendored" and documentation files are excluded from statistics in the same way; I'm not sure why/how we differentiate between the two.

Diffs. Vendored are collapsed, docs are not.

Alhadis:
@i-ky Are `.gitpod.Dockerfile` files machine-generated, or are they expected to be written by humans? If it's the latter, then this belongs in `documentation.yml` instead of `vendor.yml`.

However, I'm wondering if excluding them from project stats is even a good idea to begin with. If they're written by hand, then they ought to be reflected in language statistics. Users can always add an [override](https://github.com/github/linguist/blob/master/docs/overrides.md) to their project if they feel `.gitpod.Dockerfile` shouldn't contribute to breakdown graphs:

~~~gitattributes
.gitpod.Dockerfile linguist-documentation
~~~

i-ky:
> Are `.gitpod.Dockerfile` files machine-generated, or are they expected to be written by humans?

The latter. Gitpod has a utility that can generate config templates, but from there user is expected to take over and update it manually.

> If they're written by hand, then they ought to be reflected in language statistics.

If "written by hand" is the yardstick, then why `.vscode/`, `Vagrantfile` and `Jenkinsfile` are in `vendor.yml`?
https://github.com/github/linguist/blob/7efa7b3380ad0221c499f9ef1ca84c4dfcc0b299/lib/linguist/vendor.yml#L353-L354 https://github.com/github/linguist/blob/7efa7b3380ad0221c499f9ef1ca84c4dfcc0b299/lib/linguist/vendor.yml#L170-L171 https://github.com/github/linguist/blob/7efa7b3380ad0221c499f9ef1ca84c4dfcc0b299/lib/linguist/vendor.yml#L381-L382

`.gitpod.Dockerfile` and `Vagrantfile` are very similar in nature. They both describe development environment in a language "foreign" to the rest of repository and don't represent a "typical" use of the language.

> Users can always add an [override](https://github.com/github/linguist/blob/master/docs/overrides.md) to their project if they feel `.gitpod.Dockerfile` shouldn't contribute to breakdown graphs

Sure, but what's the purpose of defaults then?

To be fair, you have almost convinced me that `gitpod.Dockerfile` should not be ignored in stats. What bothers me is inconsistency between `.gitpod.Dockerfile` and other handwritten files in `vendor.yml`.

i-ky:
> [...] why `.vscode/`, `Vagrantfile` and `Jenkinsfile` are in `vendor.yml`?

I did some research on that. VS Code files were added to vendored without any comments in #3723. #619 adds some motivation behind adding `Vagrantfile` to the list:
> My projects are JavaScript but my Vagrant + Chef cookbooks makes Github think it's mostly Ruby.

And finally, https://github.com/github/linguist/pull/2861#discussion_r55047951 provides a good explanation why `Jenkinsfile` needs to be in `vendor.yml` (also touches on `Vagrantfile`):
> [...] by analogy with very domain-specific files like `Vagrantfile` which you would not consider associated with Ruby development. Groovy is used as a host for the EDSL, but examples like this show that the actual `Jenkinsfile` can be made to look more like a config file than a program. [...] the result does not look much like a general-purpose Groovy program.

Alhadis:
Okay, nix the part about being handwritten, it was a shitty (and misleading) explanation on my part. Instead, I should've asked _"Are users responsible for editing/maintaining these files, or is that exclusively a program's job?"_ The question was in the context of deciding whether or not they should appear in diffs; apologies for the poor phrasing.

Alhadis (approved):

PR #5507, opened 2021-08-07
Switch out JSON highlighter

Nixinova:
## Description

JSON files are currently quite difficult to read, as keys and strings are all coloured the same:

It would be much easier to read JSON files if the keys were coloured differently:

This changes the JSON grammar to one that highlights keys separately from strings, making JSON files must easier to read.

## Checklist
  
  - Old: [source](https://github.com/textmate/json.tmbundle/blob/master/Syntaxes/JSON.tmLanguage), [lightshow][light-old]
  - New: [source](https://github.com/Nixinova/NovaGrammars/blob/json@2.0.0/grammars/json.yaml-tmLanguage), [lightshow][light-new]
 - [x] **I am updating a grammar submodule**
### Languages affected
- Ecere Projects
- JSON
- (not) Jupyter Notebook
- Max

[light-old]: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Ftextmate%2Fjson.tmbundle%2Fblob%2Fmaster%2FSyntaxes%2FJSON.tmLanguage&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2FNixinova%2FNovaGrammars%2Fblob%2Fmain%2Fsamples%2Freadable.json&code=

[light-new]: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2FNixinova%2FNovaGrammars%2Fblob%2Fjson@2.0.0%2Fgrammars%2Fjson.yaml-tmLanguage&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2FNixinova%2FNovaGrammars%2Fblob%2Fmain%2Fsamples%2Freadable.json&code=


Nixinova:
Getting a contentless error

lildude:
> Getting a contentless error

No you're not. Here's your error: https://github.com/github/linguist/pull/5507/checks?check_run_id=3271010550#step:4:704

lildude:
🤔 I'm not sure this is the right thing to do. Thoughts @Alhadis as you're more experienced with writing grammars. 

What's more, peeps aren't going to notice when using light mode as you can barely tell the difference between the two when viewing on lightshow

Nixinova:
> I'm not sure this is the right thing to do.

[People really don't like the current syntax highlighting](https://stackoverflow.com/questions/14901245/how-to-style-a-json-block-in-github-wiki) and neither do I. I can try to find a better scope name though if they look too identical on light mode (though who uses that <sup>slight /s</sup>).

Alhadis:
> What's more, peeps aren't going to notice when using light mode

It might be because I'm colour-blind, but I don't see much difference in the dark-mode highlighting, either.

Regarding the grammar, there's a tonne of mistakes and missing syntax that are too numerous to list individually, so I've sent you a [pull-request](https://github.com/Nixinova/NovaGrammars/pull/6) instead.

> (though who uses that <sup>slight /s</sup>).

I do. 😜

Nixinova:
> > Getting a contentless error
> 
> No you're not. Here's your error: https://github.com/github/linguist/pull/5507/checks?check_run_id=3271010550#step:4:704

That should really be more obvious, its an uncoloured unstacktraced error in the sea of gitmodule spam where you can't CTRL+F 'error' because line 704 isn't loaded until you scroll into it and the inbuilt search bar has a bit of a heart attack. I think thats a git thing though so no can do...

lildude:
>  * (not) Jupyter Notebook

Not not Jupyter Notebook 😁. The underlying [raw source](https://github.com/github/linguist/blob/master/samples/Jupyter%20Notebook/JupyterNotebook.ipynb?short_path=2ca7408) looks like JSON to me. It looks good with your grammar, as do the others affected.

> Ecere Projects however (doesn't have a sample here) & [isn't json anyway](https://github.com/LawrenceK/webbrick/blob/cf81416653f091bacfbf29eb6e4507db33ac0ca6/Hardware/SwitchPcbs/2SwitchDocs/2%20layer%20genreic.EPJ) though so unsure why it has the tm_scope (and why it's a language in the first place)...

I can't answer why it's a language, but from a [quick search](https://github.com/search?q=extension%3Aepj+NOT+nothack) of projects with `.ejp` extensions, they do mostly appear to be JSON-like. They also look good with your grammar too. You seemed to have struck lucky with your find.

> That's not entirely true: grammars can import specific rules from other grammars using the form
> 
> ```coffee
> include: "source.json#rule-name"
> ```

Looks like none of the grammars we're using is doing this for the JSON grammar 🎉 .

So all in, I think we're good to switch out the JSON grammar.

PR #5517, opened 2021-08-14
Add `.tcl.in` file extension

xdelaruelle:
Add the `.tcl.in` file extension to the Tcl programming language.

Add `init.tcl.in` sample file, from the Modules project [1], which is
licensed GPLv2+.

This commit also re-order the available extensions for Tcl as `.adp` was
not correctly sorted.

[1] https://github.com/cea-hpc/modules

## Description
This pull request adds the `.tcl.in` extension as a Tcl file. `.in` files are input files: files containing macros that should be translated to obtain an operational file.

## Checklist:

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Atcl.in+proc+NOT+nothack
      - Not sure the search engine correctly handles extension names with a dot character in it. At least when looking at code source a lot of reference to `*.tcl.in` files are found: https://github.com/search?q=tcl.in+NOT+nothack&type=code
    - Sample source(s):
      - https://github.com/cea-hpc/modules/blob/master/tcl/init.tcl.in
    - Sample license(s): GPLv2+


lib/linguist/languages.yml
```diff
@@ -5945,8 +5945,9 @@ Tcl:
   type: programming
   color: "#e4cc98"
   extensions:
-  - ".tcl"
```
Alhadis, line 5948: From [the top of `languages.yml`](https://github.com/github/linguist/blob/7efa7b3380ad0221c499f9ef1ca84c4dfcc0b299/lib/linguist/languages.yml#L17-L19):

> _the first one is considered the primary extension, the others should be listed alphabetically_

I know it looks jarring, but the `extensions` list should be

~~~yaml
   extensions:
   - ".tcl"
   - ".adp"
   - ".tcl.in"
   - ".tm"
~~~

Alhadis (changes requested):

lib/linguist/languages.yml
xdelaruelle: No problem, I will fix that right away. I was confused by the non-regression test suite that told me it was not alphabetically sorted but I miss the point about the primary extension in that list.

Alhadis: You're right, that could probably be made clearer. I'll see if I can improve the feedback generated by `test_pedantic.rb` in a later PR, especially since the _`X should come after 𝑌`_ messages quickly become annoying when multiple extensions aren't sorted.

Alhadis (approved):

Issue #5554, opened 2021-09-03
Incomplete syntax highlighting for the programming language Squirrel (.nut)


TheWitheredStriker:
Hello. I'm a novice GitHub user, and I apologize if this is in the wrong area or if I mislabeled it, but I didn't know where else to post this.

I saw that the syntax highlighting for the Squirrel programming language (file extension `.nut`) appears to be incomplete. 

For example, let's [take this file](https://github.com/MathyFurret/th155-decomp/blob/v1.10d/src/data/script/graphics.nut).

While keywords like `if`, `else` and `return` are highlighted fine, others, like `function` and `local`, are not.

While I'm not fully sure why this happens (I'm a novice coder after all), I _presume_ that this is because the highlighter for Squirrel appears to base itself off C or C++; not only is Squirrel interwoven with these languages in many of its implementations, but the keywords that __are__ highlighted are also found in C(++) whereas the ones that aren't are not found in those languages.

Of course, I can very well be mistaken, this is simply an observation I made. Either way, it does put me off, and I'd appreciate it if someone could look into this. Thanks in adance!

Alhadis:
> I _presume_ that this is because the highlighter for Squirrel appears to base itself off C or C++; not only is Squirrel interwoven with these languages in many of its implementations, but the keywords that **are** highlighted are also found in C(++) whereas the ones that aren't are not found in those languages.

🤔 That code looks nothing like C. If I were to guess, I'd say there wasn't a better grammar for highlighting Squirrel source code at the time when support for it was added to Linguist.

So, your observation is correct—the grammar we're using for Squirrel isn't adequate, and a dedicated one is needed. Do you know of any? We can use any highlighting supported by VS Code, Atom, Sublime Text, or TextMate (technically, they're the same format).

TheWitheredStriker:
After some searching, I found [the repository of a Squirrel TextMate file.](https://github.com/robmerrell/squirrel-tmbundle) Is this what we need, or do we need a different file?

Alhadis:
We need a different file: I can't see a license in that repository, and we require grammars to be permissively-licensed (i.e., MIT, ISC, Apache, etc). The repository also appears to be unmaintained (it was last updated in 2011). This might be the reason it was never added to Linguist in the first place.

[...]

PR #3791, opened 2017-08-23
Added syntax highlighting for Squirrel

mathewmariani:
https://github.com/search?utf8=%E2%9C%93&q=extension%3Anut+NOT+nothack&t
ype=Code

Squirrel is already detected by GitHub but has no syntax
highlighting.

PR #5556, opened 2021-09-04
Fix Squirrel language's `tm_scope`

Alhadis:
## Description
In #3791, the author forgot to update the `tm_scope` for the Squirrel language entry, meaning we've had an unused grammar module sitting around for the past 4 years. This was brought to my attention by @TheWitheredStriker, who pointed out in #5554 that Squirrel was still using C/C++-derived highlighting.

On another note, we should really have a test designed to catch scenarios like this, as it flew under mine and everybody else's radar. That can wait for another day, though.

_(Template removed as it doesn't apply)._


lildude (approved):
Oooof! We missed something similar for ECL fixed in https://github.com/github/linguist/pull/5500 👍 for a test at some point.

PR #4568, opened 2019-07-01
Add Highlighting for Igor

ukos-git:
## Description
The [Igor Pro](https://www.wavemetrics.com/) Syntax Highlighting had been [added in 2013](https://github.com/github/linguist/pull/681) / [2014](https://github.com/github/linguist/pull/1376) and was removed when moving linguist to tm based highlighting due to missing `*.tmlanguage`
file.  The language recognition is still present but it currently does not highlight the code on github.

There was a [demand from the Igor programmers community](https://www.wavemetrics.com/forum/general/igor-procedure-syntax-highlighting-github#comment-form) pending since 2017 and also [stale issues here](https://github.com/github/linguist/issues/3713). 

There are [quite a few Projects on Github](https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Aipf+rtGlobals+NOT+nothack), although the language is very specific to the scientific community. As the program itself is not open source, there are also fewer spread Open Source projects. Nevertheless, it suffices the requirement for adding the language to linguist.

I have added a tm [syntax highlighting under BSD-3 on github](https://github.com/byte-physics/language-igor) so I think it is time to get the highlighting for Igor back to github.

## Checklist:

- [x] **I am changing the source of a syntax highlighting grammar**
  
  - Old: none
  - New: https://github.com/byte-physics/language-igor
- [ ] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Aipf+rtGlobals+NOT+nothack
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [example code](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fbyte-physics%2Flanguage-igor%2Fblob%2Fmaster%2Figorpro.tmLanguage&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fbyte-physics%2Flinguist%2Fadd_igor%2Fsamples%2FIGOR%2520Pro%2FCodeBrowser.ipf&code=)
    - Sample license(s): BSD-3
  
  - [x] I have included a syntax highlighting grammar: 
[...]

PR #5604, opened 2021-10-19
Release v7.17.0

lildude:
It's been a while, but I've finally managed to carve out time to make a new release.

- [x] Gem pushed to RubyGems & GitHub Package Registry 
- [x] Deployed to GitHub.com - this will happen on 20 Oct '21
- [x] Grammars deployed to Syntax highlighting service - highlighting won't be 💯 up-to-date until this happens - this will happen on 20 Oct '21
- [x] Deployed to Lightshow


lildude:
Took advantage of the [new auto-release note generation too](https://github.blog/changelog/2021-10-04-improvements-to-github-releases-public-beta/) using the defaults. Check 'em out - https://github.com/github/linguist/releases/tag/v7.17.0 - very 😍

davidhq:
Hi! Is there approximate date for new release, do grammar authors need to do anything or will latest commits be picked up ? thank you

lildude:
> Hi! Is there approximate date for new release

Early Feb at the earliest, but it depends on my workload at the time.

> do grammar authors need to do anything or will latest commits be picked up ? thank you

Nope. It happens automatically.

PS. 7.18.0 is the latest, released on 15 Dec 2021.

davidhq:
> > Hi! Is there approximate date for new release
> 
> Early Feb at the earliest, but it depends on my workload at the time.

Nice, thank you

> 
> > do grammar authors need to do anything or will latest commits be picked up ? thank you
> 
> Nope. It happens automatically.

great

> 
> PS. 7.18.0 is the latest, released on 15 Dec 2021.

yes, I saw later, 
thank you again

davidhq:
One more question perhaps - is there a 100% faithful way to test the grammar?

I found this https://github-lightshow.herokuapp.com/ and it's ok but for example `entity.name.type.class` should be colored brown instead of red.. I see `entity.name.type.class` in javascript 

```javascript
class Something { }
```

I couldn't find a way to make 100% sure how my grammar will impact minor colors so any advice would be great and then I can double-check in next days.

lildude:
> One more question perhaps - is there a 100% faithful way to test the grammar?

Not really. Lightshow is the closest we've got.

> I found this https://github-lightshow.herokuapp.com/ and it's ok but for example `entity.name.type.class` should be colored orange
> 
> ```js
> class Something { }
> ```
> 
> instead of red (as it is in thos heroku app).
> 
> I couldn't find a way to make 100% sure how my grammar will impact minor colors so any advice would be great and then I can double-check in next days.

I'm not that well versed in writing grammars but @Alhadis is and he's got this [great reference](https://github.com/Alhadis/language-etc/blob/master/samples/lists/scope-previews.nanorc) for how the Textmate scope names match colours on GitHub.

There's also the important notes at https://github.com/github/linguist/pull/4568#issuecomment-513739638 to keep in mind regarding the ordering of the scopes.

If you want further help, please open a new [discussion](https://github.com/github/linguist/discussions) as this topic is no longer related to this release.

Alhadis:
> I found this https://github-lightshow.herokuapp.com/ and it's ok but for example entity.name.type.class should be coloured orange

Lightshow's styling has been out-of-sync with that of GitHub for a while now, something made worse since the introduction of dark mode and deuteranopia support.

There might also be another reason for the discrepancy you're seeing, but it's very convoluted to explain unless you have an understanding of PrettyLights and how TextMate selectors differ from CSS selectors. I actually recently started work on an Atom theme to bring up-to-date styling to my editor so I can preview grammars more accurately as I work on them, although there's nothing to show yet.

Feel free to e-mail me directly if you don't want to open a [new discussion](https://github.com/github/linguist/discussions) for this. As @lildude mentioned, this thread is not the appropriate place for such a topic.

davidhq:
OK thank you very much @lildude and @Alhadis ! Sorry for hijacking this thread a bit.
I might send an email next week when I get into finishing touches for an update.
@lildude this is awesome - https://github.com/Alhadis/language-etc/blob/master/samples/lists/scope-previews.nanorc , thank you

PR #5605, opened 2021-10-19
Document the tree-sitter grammars used on GitHub

lildude:
## Description

GitHub has started to use Tree-sitter based grammars for some languages which means those supplied by Linguist won't be used for those languages, but will still be used for other languages that use the grammar. For example, several languages use the `atom/language-javascript` grammar so they will continue to use the grammar supplied with Linguist whilst JavaScript will use the tree-sitter grammar.

This PR updates the troubleshooting docs to point this out.

[ Checklist removed as it doesn't apply ]

### DIFF
--- a/test/test_grammars.rb
+++ b/test/test_grammars.rb
@@ -37,7 +37,7 @@ class TestGrammars < Minitest::Test
   end
 
   def test_readme_file_is_in_sync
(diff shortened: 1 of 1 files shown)


Matxat (commented):
Duplicate of #

PR #5614, opened 2021-10-22
Add YAML+ERB language

aharpervc:
## Description

This PR add support for ".yaml.erb" extensions

## Checklist:

- [x] **I am adding a new language.** (technically? I guess? situation is unclear)
      
      -  to fulfill the PR template, not great results: https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Aerb+yaml+NOT+nothack
      - better search results: https://github.com/search?q=filename%3A%22.yml.erb%22&type=Code (~33k) and https://github.com/search?q=filename%3A%22.yaml.erb%22&type=Code (~9k)
    - Sample source(s):
      - https://github.com/asrkata/collabnix-dockerlabs/blob/436ed6e4ed16df7949b3d0ac3acbc1ecb2a5ede8/play-with-docker/Portus/spec/integration/fixtures/config.yml.erb
        - this was the first example I noticed in the search results that a) was public b) was fairly short c) had a compatible license, and d) actually used erb templating in the source file
    - Sample license(s):
      - Apache 2.0 https://github.com/asrkata/collabnix-dockerlabs/blob/master/LICENSE
  
    - N/A (?)
    - N/A (?)

----

This is a situation where it's not really a new language, even though it's a new entry in the languages list. I tried to do what HTML+ERB and Javascript+ERB are doing, since they're the most similar entries in the list. For example, Javascript+ERB reports as Javascript for ace & codemirror, so I did the same thing here.


lildude (dismissed):
LGTM. Thanks.

lib/linguist/languages.yml
```diff
@@ -6873,6 +6873,18 @@ YAML:
   codemirror_mode: yaml
   codemirror_mime_type: text/x-yaml
   language_id: 407
+YAML+ERB:
+  type: markup
+  color: "#cb171e"
+  tm_scope: source.yaml.erb
```
lildude, line 6879: We don't have a grammar which has this scope. 

You need to either add such a grammar as part of this PR or change this to match a scope which Linguist already knows about which has support for YAML+ERB syntax.

lildude (changes requested):

lildude:
Closing as I can't find a grammar that supports YAML and erb. We can reopen when such a grammar comes to light.

PR #5617, opened 2021-10-25
Add detection of generated IntelliJ files

shalvah:
Classify the [`.idea/` files](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-) generated by the IntelliJ IDEA family of IDEs (IDEA, PhpStorm, RubyMine, GoLand, etc) as generated.

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -52,6 +52,7 @@ module Linguist
     # Return true or false
     def generated?
       xcode_file? ||
+      intellij_file? ||
       cocoapods? ||
       carthage_build? ||
       generated_graphql_relay? ||
@@ -112,6 +113,13 @@ module Linguist
     # Returns true or false.
     def xcode_file?
       ['.nib', '.xcworkspacedata', '.xcuserstate'].include?(extname)
+    end
+    
+    # Is the blob an IntelliJ IDEA file (IDE-specific files that are sometimes checked in)? 
+    #
+    # Returns true or false.
+    def intellij_file?
+      !!name.match(/.idea\//)
     end
 
     # Internal: Is the blob part of Pods/, which contains dependencies not meant for humans in pull requests.
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -119,7 +119,7 @@ module Linguist
     #
     # Returns true or false.
     def intellij_file?
-      !!name.match(/.idea\//)
+      !!name.match(/(?:^|\/)\.idea\//)
     end
 
     # Internal: Is the blob part of Pods/, which contains dependencies not meant for humans in pull requests.
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -115,7 +115,10 @@ module Linguist
       ['.nib', '.xcworkspacedata', '.xcuserstate'].include?(extname)
     end
     
-    # Is the blob an IntelliJ IDEA file (IDE-specific files that are sometimes checked in)? 
+    # Internal: Is the blob an IntelliJ IDEA project file?
+    #
+    # JetBrains IDEs generate project files under an `.idea` directory
+    # that are sometimes checked into version control.
     #
     # Returns true or false.
     def intellij_file?


Awan27091987:
Hi i m owner developer

lib/linguist/generated.rb
```diff
@@ -113,6 +114,13 @@ def generated?
     def xcode_file?
       ['.nib', '.xcworkspacedata', '.xcuserstate'].include?(extname)
     end
+    
+    # Is the blob an IntelliJ IDEA file (IDE-specific files that are sometimes checked in)? 
+    #
+    # Returns true or false.
+    def intellij_file?
+      !!name.match(/.idea\//)
```
Alhadis, line 122:
```suggestion
      !!name.match(/(?:^|\/)\.idea\//)
```
The regex you have now is too permissive:
* Without an anchor like `^` or `\/` prefixing `.idea`, partial directory names will be matched; e.g., `./my.bright.idea/foo.txt`
* An unescaped `.` will match any single character in a a regular expression, so `/.idea/` will match `1idea/`, `files/Another idea/`, etc.

lib/linguist/generated.rb
```diff
@@ -113,6 +114,13 @@ def generated?
     def xcode_file?
       ['.nib', '.xcworkspacedata', '.xcuserstate'].include?(extname)
     end
+    
+    # Is the blob an IntelliJ IDEA file (IDE-specific files that are sometimes checked in)?
```
Alhadis, line 118: We might as well stay consistent with the other method descriptions:
```suggestion
    # Internal: Is the blob an IntelliJ IDEA project file?
    #
    # JetBrains IDEs generate project files under an `.idea` directory
    # that are sometimes checked into version control.
```

Alhadis (changes requested):
I'm not entirely sure we should be marking *everything* inside an `.idea` directory as generated. Going by GitHub's [`JetBrains.gitignore` template](https://github.com/github/gitignore/blob/cdd9e946da421758c6f42c427c7bc65c8326155d/Global/JetBrains.gitignore#L14-L24) and JetBrains' own [VCS recommendations](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839), it would seem that not everything is generated (and thus subject to high-churn).

Remember, generated files are [hidden](https://github.com/github/linguist/blob/master/docs/overrides.md#summary) from diffs, meaning that important changes still warrant visibility. I'm wondering if it'd be wiser to include `(^|\/)\.idea/` as [`vendored path`](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml) as well. What do you think?

shalvah:
> I'm not entirely sure we should be marking everything inside an .idea directory as generated.

Isn't it the other way around? The files are definitely _generated_ (the IDE creates them automatically when I open a folder), but not necessarily _vendored_ (you might check in some of them)? Or am I mixing up the definitions?

lib/linguist/generated.rb
shalvah: Damn, you're right. Was a bit too hasty with this one. 😅

lib/linguist/generated.rb
shalvah: Thanks. I do wonder, though, do you know the origin of the "Internal:" part? 😄 Is that a holdover from the days when Linguist was internal (maybe)?

Alhadis:
> Isn't it the other way around? The files are definitely generated (the IDE creates them automatically when I open a folder), but not necessarily vendored (you might check in some of them)? Or am I mixing up the definitions?

No, that's correct. The key difference is that generated files don't show up in diffs on GitHub, whereas vendored/documentation files do. However, sometimes changes to machine-generated files need to be visible to reviewers (see #4348 for an example). Hence why I asked if it'd be more appropriate if we mark these files as vendored (and only the high-churn ones as generated). I should also mention I have zero knowledge of JetBrains IDEs, hence why I'm trusting your knowledge more than mine.

lib/linguist/generated.rb
Alhadis: It's a holdover from Linguist's use of [TomDoc](http://tomdoc.org/#the-description-section) to generate API documentation:

> **Internal:** Indicates that the method is part of the project’s internal API. These are methods that are intended to be called from other classes within the project but not intended for public consumption. For example:
> ~~~ruby
> # Internal: Normalize the filename.
> ~~~

shalvah:
> generated files don't show up in diffs on GitHub,

They do, but they're collapsed by default, yes?

Re the yarn reference: yes, that's valid, but all the `.idea` files hold is IDE/project settings. For instance, my `workspace.xml` has personal stuff like my recent commit messages, VCS settings, spellcheck settings, etc.

The files they say you can commit (real examples from one of my JS projects):

- `myproject.iml`:

```xml
<module type="WEB_MODULE" version="4">
  <component name="NewModuleRootManager">
    <content url="file://$MODULE_DIR$" />
    <orderEntry type="inheritedJdk" />
    <orderEntry type="sourceFolder" forTests="false" />
  </component>
</module>
```

- `vcs.xml`:
```xml
<project version="4">
  <component name="VcsDirectoryMappings">
    <mapping directory="$PROJECT_DIR$" vcs="Git" />
  </component>
</project>
```
- `misc.xml`:

```xml
<project version="4">
  <component name="JavaScriptSettings">
    <option name="languageLevel" value="ES6" />
  </component>
</project>
```

Nice-to-haves if you want to share them, maybe, but none of these files really matter—once you open a folder in the IDE, it regenerates the whole `.idea/` folder if it doesn't exist. (In fact, I haven't seen a project where they were consciously shared, though I have seen accidental check-ins from newbie devs, along with `.DS_Store`😅.)

There don't seem to be any malicious opportunities from hidden diffs, except messing with your teammates, and they can easily change the setting back in their IDE.

Alhadis (approved):

Alhadis (approved):

lildude (approved):

PR #5638, opened 2021-11-07
Update julia highlighter

jeremiahpslewis:


Alhadis:
I've read JuliaEditorSupport/atom-language-julia#244 and I see what you're trying to do. However, updating the grammar version manually isn't going to have any effect, because grammar updates are only reflected on GitHub after a new release of Linguist has been published. This happens automagically during the release process too (users don't have to worry about keeping Linguist's version of a grammar up-to-date).

jeremiahpslewis:
Makes sense to pull changes rather than depend on per language pushes. Do you have a rough idea of when the next release will happen?

Alhadis:
Only @lildude can answer that, but there's usually a new release once every 1.5-2 months, depending on his schedule and/or bandwidth. 😉 (Cutting a release is a tedious and fiddly procedure, from what I understand)

jonathantneal:
Thanks, @Alhadis! I was about to do something similar, and this thread saved me time and let me set expectations. My thanks to you and everyone helping out!

PR #4515, opened 2019-04-30
Support for flow

kivikakk:
See https://github.com/area9innovation/flow9.

## Description
- [x] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Aflow+%22import+runtime%22
      - https://github.com/search?q=extension%3Aflow+%22flowutils%22&type=Code
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://github.com/area9innovation/flow9/blob/1a3fb95a1fe0c09d71738064dd6eafda10c51af3/lib/string.flow
    - Sample license(s):
      - https://github.com/area9innovation/flow9/blob/1a3fb95a1fe0c09d71738064dd6eafda10c51af3/LICENSE.txt#L5-L6

PR #5640, opened 2021-11-08
Add AssemblyScript

SebastianSpeitel:
## Description
Adds AssemblyScript language and heuristics.

The heuristics are really rough for now and will be changing.

## Checklist:

      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Ats+assemblyscript+NOT+nothack
    - Sample source(s):
      - https://github.com/AssemblyScript/assemblyscript/tree/main/std/assembly
    - Sample license(s):
      - https://github.com/AssemblyScript/assemblyscript/blob/main/LICENSE
  
    Still discussing

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -567,6 +567,18 @@ disambiguations:
   rules:
   - language: XML
     pattern: '<TS\b'
+  - language: AssemblyScript
+    negative_pattern: 
+    - '\s+undefined\s+'
+    - '\s+any\s+'
+    - '\s+unknown\s+'
+    pattern:
+    - '@inline'
+    - '@final'
+    - '@unmanaged'
+    - '// @ts-ignore: decorator'
+    - '[iuf](8|16|32|64)'
+    - 'usize'
   - language: TypeScript
 - extensions: ['.tst']
   rules:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -387,6 +387,14 @@ Assembly:
   tm_scope: source.assembly
   ace_mode: assembly_x86
   language_id: 24
+AssemblyScript:
+  type: programming
+  color: "#007acc"
+  extensions:
+  - ".ts"
+  tm_scope: source.ts
+  ace_mode: typescript
+  language_id: 841315793
 Astro:
   type: markup
   color: "#ff5a03"
--- a/samples/AssemblyScript/array.ts
+++ b/samples/AssemblyScript/array.ts
@@ -1,0 +1,578 @@
+/// <reference path="./rt/index.d.ts" />
+
+import { BLOCK_MAXSIZE } from "./rt/common";
+import { COMPARATOR, SORT } from "./util/sort";
+import { joinBooleanArray, joinIntegerArray, joinFloatArray, joinStringArray, joinReferenceArray } from "./util/string";
+import { idof, isArray as builtin_isArray } from "./builtins";
+import { E_INDEXOUTOFRANGE, E_INVALIDLENGTH, E_ILLEGALGENTYPE, E_EMPTYARRAY, E_HOLEYARRAY } from "./util/error";
+
+/** Ensures that the given array has _at least_ the specified backing size. */
+function ensureSize(array: usize, minSize: usize, alignLog2: u32): void {
+  // depends on the fact that Arrays mimic ArrayBufferView
+  var oldCapacity = changetype<ArrayBufferView>(array).byteLength;
+  if (minSize > <usize>oldCapacity >>> alignLog2) {
+    if (minSize > BLOCK_MAXSIZE >>> alignLog2) throw new RangeError(E_INVALIDLENGTH);
+    let oldData = changetype<usize>(changetype<ArrayBufferView>(array).buffer);
+    let newCapacity = minSize << alignLog2;
+    let newData = __renew(oldData, newCapacity); // keeps RC
+    memory.fill(newData + oldCapacity, 0, newCapacity - oldCapacity);
+    if (newData !== oldData) { // oldData has been free'd
+      store<usize>(array, newData, offsetof<ArrayBufferView>("buffer"));
+      store<usize>(array, newData, offsetof<ArrayBufferView>("dataStart"));
+    }
+    store<u32>(array, newCapacity, offsetof<ArrayBufferView>("byteLength"));
+  }
+}
+
+export class Array<T> {
+  [key: number]: T;
+
+  // Mimicking ArrayBufferView isn't strictly necessary here but is done to allow glue code
+  // to work with typed and normal arrays interchangeably. Technically, normal arrays do not need
+  // `dataStart` (equals `buffer`) and `byteLength` (equals computed `buffer.byteLength`), but the
+  // block is 16 bytes anyway so it's fine to have a couple extra fields in there.
+
+  private buffer: ArrayBuffer;
+  private dataStart: usize;
+  private byteLength: i32;
+
+  // Also note that Array<T> with non-nullable T must guard against uninitialized null values
+  // whenever an element is accessed. Otherwise, the compiler wouldn't be able to guarantee
+  // type-safety anymore. For lack of a better word, such an array is "holey".
+
+  private length_: i32;
+
+  static isArray<U>(value: U): bool {
+    return isReference<U>() ? builtin_isArray(value) && value !== null : false;
+  }
+
+  static create<T>(capacity: i32 = 0): Array<T> {
+    WARNING("'Array.create' is deprecated. Use 'new Array' instead, making sure initial elements are initialized.");
+    var array = new Array<T>(capacity);
+    array.length = 0;
+    return array;
+  }
+
+  constructor(length: i32 = 0) {
+    if (<u32>length > <u32>BLOCK_MAXSIZE >>> alignof<T>()) throw new RangeError(E_INVALIDLENGTH);
+    var bufferSize = <usize>length << alignof<T>();
+    var buffer = __new(bufferSize, idof<ArrayBuffer>());
+    memory.fill(buffer, 0, bufferSize);
+    this.buffer = changetype<ArrayBuffer>(buffer); // retains
+    this.dataStart = buffer;
+    this.byteLength = <i32>bufferSize;
+    this.length_ = length;
+  }
+
+  get length(): i32 {
+    return this.length_;
+  }
+
+  set length(newLength: i32) {
+    var oldLength = this.length_;
+    if (isManaged<T>()) {
+      if (oldLength > newLength) { // release no longer used refs
+        let dataStart = this.dataStart;
+        let cur = dataStart + (<usize>newLength << alignof<T>());
+        let end = dataStart + (<usize>oldLength << alignof<T>());
+        do __release(load<usize>(cur));
+        while ((cur += sizeof<T>()) < end);
+      } else {
+        ensureSize(changetype<usize>(this), newLength, alignof<T>());
+      }
+    } else {
+      ensureSize(changetype<usize>(this), newLength, alignof<T>());
+    }
+    this.length_ = newLength;
+  }
+
+  every(fn: (value: T, index: i32, array: Array<T>) => bool): bool {
+    for (let index = 0, length = this.length_; index < min(length, this.length_); ++index) {
+      if (!fn(load<T>(this.dataStart + (<usize>index << alignof<T>())), index, this)) return false;
+    }
+    return true;
+  }
+
+  findIndex(predicate: (value: T, index: i32, array: Array<T>) => bool): i32 {
+    for (let index = 0, length = this.length_; index < min(length, this.length_); ++index) {
+      if (predicate(load<T>(this.dataStart + (<usize>index << alignof<T>())), index, this)) return index;
+    }
+    return -1;
+  }
+
+  @operator("[]") private __get(index: i32): T {
+    if (<u32>index >= <u32>this.length_) throw new RangeError(E_INDEXOUTOFRANGE);
+    var value = this.__uget(index);
+    if (isReference<T>()) {
+      if (!isNullable<T>()) {
+        if (!changetype<usize>(value)) throw new Error(E_HOLEYARRAY);
+      }
+    }
+    return value;
+  }
+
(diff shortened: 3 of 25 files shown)


lib/linguist/heuristics.yml
```diff
@@ -567,6 +567,18 @@ disambiguations:
...
+    - '\s+undefined\s+'
+    - '\s+any\s+'
+    - '\s+unknown\s+'
+    pattern:
+    - '@inline'
+    - '@final'
+    - '@unmanaged'
+    - '// @ts-ignore: decorator'
+    - '[iuf](8|16|32|64)'
+    - 'usize'
```
MaxGraey, line 581:
```suggestion
    - '\b[iu]size'
```

lib/linguist/heuristics.yml
```diff
@@ -567,6 +567,18 @@ disambiguations:
...
+    negative_pattern: 
+    - '\s+undefined\s+'
+    - '\s+any\s+'
+    - '\s+unknown\s+'
+    pattern:
+    - '@inline'
+    - '@final'
+    - '@unmanaged'
+    - '// @ts-ignore: decorator'
+    - '[iuf](8|16|32|64)'
```
MaxGraey, line 580:
```suggestion
    - '\b[iuf](8|16|32|64)'
```

lib/linguist/heuristics.yml
```diff
@@ -567,6 +567,18 @@ disambiguations:
...
     pattern: '<TS\b'
+  - language: AssemblyScript
+    negative_pattern: 
+    - '\s+undefined\s+'
+    - '\s+any\s+'
+    - '\s+unknown\s+'
+    pattern:
+    - '@inline'
+    - '@final'
+    - '@unmanaged'
```
MaxGraey, line 578: Also
```yml
- '@operator'
```

SebastianSpeitel: and @unsafe and maybe even @builtin

lib/linguist/heuristics.yml
SebastianSpeitel: why only `\b` on the beginning and not the end?

MaxGraey: because we should accept `isize` or `usize` but not `iusize` or `uisize`

SebastianSpeitel: yes, that's the `\b` at the front, but thy not `\b[iu]size\b`?

MaxGraey: You could add this at the end but usually it's unnecessary. But it's depend on how tokens match - per lines or per words.

SebastianSpeitel: It doesn't match tokens. The regex is matched against the whole file as far as I could figure out.

MaxGraey: well, in this case make sense wrap tokens to `\b`

lildude: The heuristics are only run against the first bytes of the file:

https://github.com/github/linguist/blob/e2f638d17598e84f6f91c2d03c2f3e27f22a2163/lib/linguist/heuristics.rb#L6

lildude (changes requested):
Wow!! Over 20 samples 😱 We really don't need that many, especially the really big files which seems to have a lot of duplicate content from the others. Please cut down the samples to only those that are most representative of the language and real world use. 2-5 are plenty.

Alhadis (commented):
Isn't AssemblyScript quite literally a clean subset of TypeScript? That doesn't make it its own language, and is going to be a _nightmare_ to disambiguate from "regular" TypeScript; i.e., you'll wind up with projects classified as 80% AssemblyScript and perhaps 20% TypeScript due to the presence of `.ts` files that didn't match a heuristic.

lib/linguist/heuristics.yml
Alhadis: `\b[iu]size` is still going to match stuff like `$isize` and `foo.usize`.

SebastianSpeitel: So something like `[\s:<][iu]size[\s>)]` would be more accurate.
(The `:` is for variable defintions like `const ptr:usize`, the `<>` for generics like `Array<usize>` and the `)` for `(ptr: usize)=>void`)

MaxGraey: that's true. It could be false positive. But at the same time this should be also valid: `:isize` for example `function foo():isize`.

MaxGraey: All possible variants: `Array<usize>`, `()=>usize`, `:isize`, `isize(x)`, `<isize>x`

SebastianSpeitel:
@Alhadis Yes it is, except a few very small edge cases.
In some cases the code is valid TS *and* AS (sometimes even by design, because there is portable AS) and it should fallback to TS there.

@lildude I included so many examples because it is so hard to distinguish and I even removed some of them, because they were valid TS and AS even for me.
I want to include a few more TS examples as well to test in that direction.

lib/linguist/heuristics.yml
SebastianSpeitel: It seems like a more appropriate border would be `[^.\w]`, so it doesn't match inside other words and not as a property.

Alhadis:
> @Alhadis Yes it is, except a few very small edge cases.
In some cases the code is valid TS and AS (sometimes even by design, because there is portable AS) and it should fallback to TS there.

This should tell you that no real (language-level) difference exists between AssemblyScript and TypeScript—the former is simply a restrictive and specialised application of the latter, not unlike [asm.js](http://asmjs.org/). From my reading, the only difference between the two technologies is that AssemblyScript [mandates strictly-typed code](https://www.assemblyscript.org/basics.html#strictness) and abstinence from JavaScript's more dynamic features (I'm guessing stuff like `eval()`, `new Function(…)` and so forth).

MaxGraey:
AssemblyScript has not only a stricter type system, but also some semantic features and fixes that TypeScript cannot afford. Some of this diffs are presented in this article: https://blog.suborbital.dev/assemblyscript-vs-typescript

Alhadis:
> Some of this diffs are presented in this article: https://blog.suborbital.dev/assemblyscript-vs-typescript.

All of those differences are semantic in nature, and have nothing to do with syntax. That is, they pertain to the _interpretation_ of TypeScript, and many of these differences can also be enforced by a conventional TypeScript compiler, IIRC.

Sorry, I'm not seeing anything that warrants recognition as a discrete language.

dcodeIO:
Perhaps for reference, the most prominent syntactic difference is that decorators like `@inline` are allowed on functions and variables. The remainder is mostly semantics, like a different set of built-in types, some new, some not supported, availability of operator overloads, type assertions becoming casts, different behavior of triple equals, and such. Not taking a side :)

SebastianSpeitel:
This is what I'm currently at:
```yml
- extensions: ['.ts']
  rules:
  - language: XML
    pattern: '<TS\b'
  - language: AssemblyScript
    and:
    - negative_pattern: 
      # Invalid types in AssemblyScript
      - '[^"`''.\w]undefined[^"`''.\w]'
      - '[^.\w]any[^.\w]'
      - '[^.\w]unknown[^.\w]'
      # No eval in AssemblyScript
      - '[^.\w]eval\s*\('
    - pattern:
      # Builtin decorators
      - '@inline'
      - '@final'
      - '@unmanaged'
      - '@operator'
      - '@global'
      - '@unsafe'
      - '@builtin'
      - '^// @ts-ignore: decorator$'
      # number types only available in AssemblyScript
      - '[^"`''.\w][iuf](8|16|32|64)[^"`''.\w]'
      - '[^"`''.\w][iu]size[^"`''.\w]'
  - language: TypeScript

```

> Sorry, I'm not seeing anything that warrants recognition as a discrete language.

I think AssemblyScript counts as it's own language, but I can see that if there is no way to safely distinguish it, it can't be added here.

Also I think that in cases where a file in an AssemblyScript project is misclassified (which will surely happen) _and_ is valid Typescript, the classification is still correct. For me it would be enough to get the heuristics to a point where no TypeScript file is detected as AS, because that would probably confuse a lot of people, but false-negatives in AS are not that big of a deal (for me).
If false-negatives are a no-go here, I would have no problem with that.

Alhadis:
> I think AssemblyScript counts as it's own language

That's only because the project's name makes it sound like a language. 😉 Would you hold the same perspective if the project were named `ASM.ts` or `ts2wasm`?

> but I can see that if there is no way to safely distinguish it, it can't be added here.

You're right, I'm afraid. However, `assembly script` can still be added as an alias of TypeScript, which will give more meaningful results when [searching by language](https://github.com/search?q=language%3Aassemblyscript+NOT+nothack&type=code), as well as be recognised in fenced code blocks in readme files:

~~~markdown
``` assemblyscript
export function negate(n: i32): i32 {
	return -a;
}
```

``` typescript
export function negate(n: i32): i32 {
	return -a;
}
```
~~~

lib/linguist/heuristics.yml
```diff
@@ -567,6 +567,18 @@ disambiguations:
   rules:
   - language: XML
     pattern: '<TS\b'
+  - language: AssemblyScript
+    negative_pattern: 
+    - '\s+undefined\s+'
+    - '\s+any\s+'
+    - '\s+unknown\s+'
```
MaxGraey, line 574: Btw I'm not sure about `unknown`. It is quite likely that it will still be used at least for variadic function decladarions

SebastianSpeitel:
> However, `assembly script` can still be added as an alias of TypeScript, which will give more meaningful results when [searching by language](https://github.com/search?q=language%3Aassemblyscript+NOT+nothack&type=code), as well as be recognised in fenced code blocks in readme files:

I'm not sure an alias of TypeScript would be correct, because we already have users, who expect to be able to use any TS library in their AS project, which is sadly not the case. And letting them find mostyl TS code when searching for AS wouldn't help with this.
But the recognition in code blocks would be amazing.

Would it be possible to add it, but without detection via heuristics, so it has be set via [overrides](docs/overrides.md)?

lib/linguist/heuristics.yml
trusktr: Also some people in the AS community have implemented various forms of `unknown`, and it could conceivably be supported officially later once semantics are ironed out.

MaxGraey:
> Just to be clear: AssemblyScript is not a language. It's a compilation target for TypeScript with the same syntax, semantics, and compiler.

That's totally false. AssemblyScript is typescript-like language and compiler to WebAssembly which is compiler target. It has own subset syntax which is close to TS but has some limitations and extensions (like global function decorators, operator overloads, native types like i64/u64/i8/u16 which doesn't support by TS). It also has own parser, compiler and emitter. It has similar syntax but totally different semantics sometimes.

Alhadis:
@MaxGraey A compiler and its output are completely independent of its input. [`gcc(1)`](https://man7.org/linux/man-pages/man1/gcc.1.html), for example, can compile at least 9 different languages:

> **`-x` <var>language</var>**  
> Specify explicitly the language for the following input files (rather than letting the compiler choose a default based on the file name suffix).  This option applies to all following input files until the next `-x` option.  Possible values for <var>language</var> are: `c, c++, objective-c, objective-c++, assembler, ada, d, f77, go`.

It can output binaries or shared objects for a variery of architectures, which serves as the analogoue of a compiler targeting JavaScript or WASM as its output.

> It has own subset syntax which is close to TS but has some limitations and extensions (like global function decorators, operator overloads, native types like i64/u64/i8/u16 which doesn't support by TS).

Yes, that ties in with what I said about runtime details:

> _Certain aspects of the runtime may differ (predefined types, a more restrictive standard library, etc), but runtimes **aren't** languages._

MaxGraey:
> A compiler and its output are completely independent of its input

Yes, and how it relate to TS and AS are the same statement? Every language should have at least one compiler. Some of them like Clang or GCC can compile C and C++ at the same time. But I don't really see how this cancels out the fact that AS and TS have pretty serious differences in semantics?

MaxGraey:
For example Flow and TS have much more similarities between each other. Has similar types, syntax, compilation target (JS). Same extension (`.js`). But they syntaxes distinguish in linguist at some points: https://github.com/github/linguist/blob/4fc6808d99980a88465861217db6b74cd0d639d3/lib/linguist/vendor.yml#L229

Alhadis:
When I say "semantics", I'm talking about the exact behaviour and logic of language-level elements: classes, functions, variables, expressions, scoping, etc. Fundamental elements of JavaScript, and by extension, TypeScript.

> But they syntaxes distinguish in linguist at some points:

No, we're not distinguishing syntax here. `vendor.yml` lists filename patterns that are commonly used by vendored code, and are to be excluded from language stats by default.

MaxGraey:
I also meant that cancelled PR #4515

Alhadis:
That's a [different language](https://github.com/area9innovation/flow9) of the same name.

MaxGraey:
Ah, you're right!

munrocket:
I am agree that it's nightmare to detect it. Moreover as a programmer I prefer another extension for AssemblyScript in my personal projects. Here is an issues about supporting different extension that I am using it

1. AssemblyScript/assemblyscript/issues/1003
2. as-pect/as-pect/pull/356 
3. surma/rollup-plugin-assemblyscript/pull/3

AssemblyScript is still pretty young language and follow WebAssembly specification and implemented things like Operator Overloading / SIMD / Reference Types / Own STD Library / Other interfaces that will *never will be* in TypeScript. Also it's *incompatible* with TypeScript and better to write new project from scratch than adopting it. WebAssembly technology is growing and Javascript community is huge. Here statistics in Google Trends, NPM downloads in 5 years and video from Github itself about it https://www.youtube.com/watch?v=97ej9-CE3Gc

|||

Will you add it to github if we change the extension to a unique one? @Alhadis, please give it a try! Because I think we can make this decision and it's right time to change the extension finally (community used .ts just for linting LOL).

romdotdog:
As an AssemblyScript contributor, changing the extension and thus having to replicating the hundreds of errors that TypeScript lints for in the AssemblyScript compiler would be both costly performance-wise and would bloat WASM bootstrapping sizes. I don't intend to hijack this discussion and turn this into a pros-and-cons-of-changing-the-extension discussion so I'd be happy to voice ad reiterate my stance on this in the Discord.

PR #5643, opened 2021-11-10
Add Bikeshed & Brighterscript

paulo-santos-sky:
This PR aims to add brighterscript language support.

It's a superset of Roku's programming language, brightscript that aims to add sugar syntax and lot's of new functionalities
https://github.com/rokucommunity/brighterscript

## Description
- Add the VSCode brightscript language extension, which supports brighterscript on it's own

Please provide support as i'm not fully aware if all requirements are met, or, if i should also aim on this PR to replace the old brightscript language module, which uses "vendor/grammars/BrightScript.tmbundle" and not "vendor/grammars/vscode-brightscript-language"

## Checklist:
      -  https://github.com/search?q=extension%3Abs+function&type=Code
    - Sample source(s):
    - Sample license(s):
  

lib/linguist/languages.yml
```diff
@@ -619,6 +619,14 @@ Brainfuck:
   codemirror_mode: brainfuck
   codemirror_mime_type: text/x-brainfuck
   language_id: 38
+Brighterscript:
```
Nixinova, line 622: is it officially "Brighterscript" or "Brighter<ins>S</ins>cript"?

Nixinova:
Are you planning on undrafting this? it seems complete.

Nixinova:
(Better search query for tallying usage: https://github.com/search?q=extension%3Abs+function&type=Code)

lib/linguist/languages.yml
paulo-santos-sky: the latter actually, you are correct.

paulo-santos-sky: i will update it today. thank you

paulo-santos-sky: changed

Nixinova: the samples folder will also need to be updated

paulo-santos-sky: done, thanks

lib/linguist/languages.yml
```diff
@@ -619,6 +619,14 @@ Brainfuck:
   codemirror_mode: brainfuck
   codemirror_mime_type: text/x-brainfuck
   language_id: 38
+BrighterScript:
+  type: programming
+  color: "#66AABB"
+  extensions:
+  - ".bs"
```
lildude, line 626: 🤔 this is going to result in a lot of files being incorrectly classified as there are a lot of `.bs` files which don't look anything like your samples. As this is the only instance of `.bs` known to Linguist, I think we're going to need to identify at least one of the other popular users of this extension and add support for it too.

paulo-santos-sky: so another language that has .bs as file extension?

lildude: Yup. It can be added as part of this PR.

paulo-santos-sky: i didn't understand, the samples delivered are one from plain brightScript and then an example of what brighterScript bring to the table as a superset of brightScript, classes, ternary operators, etc.

I cannot provide real world usage of the language myself as our source code is private, but in the query search you have fine examples of brighterScript

lib/linguist/languages.yml
philippe-elsass-deltatre: Is there a pattern to have multiple possible syntaxes for the same extension? Some detection mechanism maybe.

lildude: Yup, we have [heuristics](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml) to differentiate. 

We don't need heuristics if the languages are sufficiently different as we can rely on the classifier to detect the language based on the tokens within the samples. In this case, it's best to have at least 2 samples from each language.

paulo-santos-sky: i've added the heuristics regexes.
Should this be enough? 
I didn't add the "Bikeshed" language to the list, as i don't know if there is any syntax highlighting available for it.

paulo-santos-sky: found the syntax highlighting for bikeshed as well, added the language as requested.

lib/linguist/languages.yml
paulo-santos-sky: added Bikeshed language and heuristics.

please have a look when able. 
thank you

paulo-santos-sky:
made changes a few days ago.

any way i can request a review?

lib/linguist/languages.yml
paulo-santos-sky: @lildude can you have a look at this pr when you have some time please and let me know if anything missing? I believe with the adding of bikeshed and heuristics, it should be done right?

lib/linguist/languages.yml
```diff
@@ -550,6 +550,14 @@ Bicep:
   tm_scope: source.bicep
   ace_mode: text
   language_id: 321200902
+Bikeshed:
+  type: programming
```
Alhadis, line 554: This should be `markup`, not `programming`.

```suggestion
  type: markup
```

lib/linguist/languages.yml
```diff
@@ -550,6 +550,14 @@ Bicep:
...
+  color: "#5562ac"
+  extensions:
+  - ".bs"
+  tm_scope: source.csswg
+  ace_mode: text
```
Alhadis, line 559: This grammar is sufficiently similar enough to HTML that we can repurpose [its CodeMirror and ACE editing modes](https://github.com/github/linguist/blob/7503f7588ca2590cebc4a05af7d00a7e94c0d779/lib/linguist/languages.yml#L2415-L2417):

```suggestion
  ace_mode: html
  codemirror_mode: htmlmixed
  codemirror_mime_type: text/html
```

lib/linguist/heuristics.yml
```diff
@@ -98,6 +98,12 @@ disambiguations:
   rules:
   - language: FreeBasic
     pattern: '^[ \t]*#(?:define|endif|endmacro|ifn?def|if|include|lang|macro)\s'
+- extensions: ['.bs']
+  rules:
+  - language: Bikeshed
+    pattern: '^(?i:<\/*pre>)|(?i:<pre class=.*>)'
```
Alhadis, line 104: According to [Bikeshed's spec](https://tabatkins.github.io/bikeshed/#metadata), several metadata fields are required information, stipulating the presence of a `<pre class="metadata">…</pre>` block. We can therefore afford to make this pattern more specific (and ergo, less prone to false matches).

```suggestion
    pattern: '^(?i:<pre\s+class)\s*=\s*('|"|\b)?metadata\1[^>\r\n]*>'
```

lib/linguist/heuristics.yml
```diff
@@ -98,6 +98,12 @@ disambiguations:
...
+  - language: BrighterScript
+    pattern: '(?i:\w+\s*\(((?:[^()]*))?\s*\)(.*\n*\r*\s*)*?(end\s)(function|sub))'
```
Alhadis, line 106: What's going on here? This pattern makes no sense:

~~~regex
.*\n*\r*\s*
~~~

Did you mean to write `(?:.|\n|\r)*` instead? Note that the `dotall` modifier is switched on by default in Ruby (when in Multiline mode, at least), so `.` will also match a carriage return (`\r`) or line-feed (`\n`).

In fact, this entire bit is superfluous and ~~can~~ _should_ be removed for better performance:

```suggestion
    pattern: '(?i:\w+\s*\(((?:[^()]*))?\s*\)(end\s)(function|sub))'
```

**Full disclosure:** I'm entirely unfamiliar with BrighterScript (or its dimmer predecessor, BrightScript).

Alhadis (changes requested):

paulo-santos-sky:
> Can you please update the OP template to add the deets for the Bikeshed files etc

sorry, didn't get that, OP template?

lib/linguist/heuristics.yml
```diff
@@ -98,6 +98,12 @@ disambiguations:
   rules:
   - language: FreeBasic
     pattern: '^[ \t]*#(?:define|endif|endmacro|ifn?def|if|include|lang|macro)\s'
+- extensions: ['.bs']
+  rules:
+  - language: Bikeshed
+    pattern: '^(?i:<pre\s+class)\s*=\s*(|''|\"|\b)?metadata\1[^>\r\n]*>'
```
Alhadis, line 104: Typo:

```suggestion
    pattern: '^(?i:<pre\s+class)\s*=\s*(''|\"|\b)?metadata\1[^>\r\n]*>'
```

(Note that the `\b` capture is to match markup like `<pre class=metadata>` and not `<pre class=metadata_unrelated_tag>`).

Alhadis (changes requested):

Alhadis (changes requested):

Alhadis (approved):

paulo-santos-sky:
```
Error updating vendor/grammars/vscode-brightscript-language
error: pathspec 'vendor/grammars/vscode-brightscript-language' did not match any file(s) known to git
```

does this mean i have to update the grammar repo?

also get on classifier cross validation
```
BrighterScript/Main.bs GOOD
BrighterScript/SimpleGrid.bs BAD (Bikeshed)
```

Alhadis:
@paulo-santos-sky It seems you've tried to point references to the old BrightScript grammar to the newer BrighterScript grammar. That, or you've not added the BrighterScript grammar using the recommended [`script/add-grammar`](https://github.com/github/linguist/blob/cddf7476af4c95d1572956ffc5c0cb84f7e431c5/script/add-grammar) tool.

Is the BrighterScript grammar compatible with the original BrightScript? That is, can we use it to highlight both BrightScript and BrighterScript?

paulo-santos-sky:
> 

yes, the old brightscript grammar of that VSCode is compatible with BrighterScript.
this one:
https://github.com/RokuCommunity/vscode-brightscript-language

paulo-santos-sky:
> nsion, _not_



> I've noticed that both grammars use the `brs` extension instead of `bs`. Is this deliberate? Which is the correct extension?
> 
> Because it seems to me that the BrighterScript entry in `languages.yml` should be using the `.brs` extension, _not_ `.bs`.

brs is for native brightscript
bs is for brighterscript, which is transpiled into plain brightscript

paulo-santos-sky:
> Alright, I _think_ we're on the same page. Give me a minute or two to fix this issue locally, and I'll push a change to your branch.

thank you so much for all the help. \m/

lib/linguist/heuristics.yml
```diff
@@ -98,6 +98,16 @@ disambiguations:
...
+- extensions: ['.bs']
+  rules:
+  - language: Bikeshed
+    pattern: '^(?i:<pre\s+class)\s*=\s*(''|\"|\b)metadata\b\1[^>\r\n]*>'
+  - language: BrighterScript
+    pattern:
+    - (?i:^\s*(?=^sub\s)(?:sub\s*\w+\(.*?\))|(?::\s*sub\(.*?\))$)
+    - (?i:^\s*(end\ssub)$)
+    - (?i:^\s*(?=^function\s)(?:function\s*\w+\(.*?\)\s*as\s*\w*)|(?::\s*function\(.*?\)\s*as\s*\w*)$)
+    - (?i:^\s*(end\sfunction)$)
```
Alhadis, line 110: Given that Bikeshed is a markup format for language specifications, there's no telling what syntaxes might be documented in a `.bs` file. Hence there's a pretty high risk that BrighterScript's heuristics will errantly match something.

I think we're better off removing BrighterScript's heuristics altogether. BrightScript uses a different file extension, so there's no risk of Linguist confusing the two.

```suggestion
```

Alhadis (changes requested):

paulo-santos-sky:
without the heuristics on brighterscript, the samples test fails.

will revert it.

paulo-santos-sky:
Hey @Alhadis,

added another sample on brighterscript, and now it passes the cross validation check locally.
Only thing left should be the license on bikeshed

```
* linguist.git_submodule.bikeshed
  filename: /home/runner/work/linguist/linguist/vendor/licenses/git_submodule/bikeshed.dep.yml
    - license needs review: other
    ```

Alhadis:
> both Bikeshed and BrighterScript share the .bs extension, isn't that the reason behind heuristics? to distinguish between same file extensions?

If none of the heuristics match, then Linguist falls through to its last strategy of Bayesian analysis, where it recognises languages based on the frequency of recurring keywords and syntax. It does this by learning from the sample files we train it on, hence the need for real-world samples for every supported language. The more samples, the better its accuracy.

Sometimes this approach isn't enough, however, especially when disambiguating two very similar-looking languages (say, C and C++). This is where heuristics come in: they give Linguist a nudge in the right direction, so to speak, by searching for patterns that unambiguously identify a certain language. This is why an overly-accepting or unspecific heuristic can do more harm than good.

> added another sample on brighterscript, and now it passes the cross validation check locally.
> Only thing left should be the license on bikeshed

Yup, I noticed. I'm still waiting on @lildude's approval before I update the license whitelist.

lildude:
> I'm assuming it's okay to add it to the license whitelist?

Go for it. Please update the links in the CONTRIBUTING.md whilst you're at it 🙇

paulo-santos-sky:
> > I'm assuming it's okay to add it to the license whitelist?
> 
> Go for it. Please update the links in the CONTRIBUTING.md whilst you're at it 🙇

@Alhadis when you have some time, can you have a look? need me to do anything here?

Alhadis:
> @Alhadis when you have some time, can you have a look?

Fixed, sorry for the wait. 👍

Alhadis (approved):

paulo-santos-sky:
thank you @Alhadis, @lildude can you have a look?

i'm anxious to deliver the good news to the brighterscript community 😸

PR #5645, opened 2021-11-14
Add heuristics for the association of `.frag` extension for JavaScript to deconflict with GLSL fragment shaders

dmitmel:
## Description

This PR removes the association of `.frag` extension for JavaScript, or rather, makes it more specific. This was added in #1178 with the example usage of JS code fragments which would be concatenated with a main program to wrap it into an IIFE (immediately-invoked function expression), but the thing is that JS development has since moved on from source code concatenation, so now it is hard to find legitimate usages of this file extension in contemporary JS projects. Moreover, nothing about `.frag` tells that it is JS-specific - it can stand for "fragment" as in a fragment of some program, but a way more frequently-occurring usage of this extension are GLSL fragment shaders. And so, because of the lack of a rule in `heuristics.yml`, thousands of GLSL files are now very frequently misclassified as JavaScript.

And so, I think that rather than introducing a regex for separating one from the other, association of `.frag` with JS can be removed altogether as an artifact of the past. I did make it more specific though, by changing it to `.js.frag`, as to ensure that the tests still pass.

## Checklist:

    - Sample source(s):
      - [Search results for `extension:frag language:js in:path,file NOT node_modules NOT glslang NOT void NOT struct NOT asm`](https://github.com/search?p=1&q=extension%3Afrag+language%3Ajs+in%3Apath%2Cfile+NOT+node_modules+NOT+glslang+NOT+void+NOT+struct+NOT+asm&type=Code) (~99k results) - this was supposed to exclude GLSL usages of this extension, compiled SPIR-V code and find the JS ones, but you would be hard-pressed to discover the latter ones here. One thing that sometimes pops up is the [node-forge](https://github.com/digitalbazaar/forge) (4k stars) package in committed `node_modules` directories, but [they have moved on to Webpack](https://github.com/digitalbazaar/forge/commit/ff04382e07653aced52be0f63019b66eefea7935).
      - [Search results for `extension:frag language:js in:path,file prototype NOT node_modules NOT glslang NOT void NOT struct NOT asm`](https://github.com/search?q=extension%3Afrag+language%3Ajs+in%3Apath%2Cfile+prototype+NOT+node_modules+NOT+glslang+NOT+void+NOT+struct+NOT+asm&type=Code) (~50 results) - next I tried searching for keywords which are unique to JS and not used in GLSL, e.g. `prototype`. Most of the results are either SPIR-V assembly or the examples from this repository (`samples/JavaScript/intro.js.frag` and `samples/JavaScript/outro.js.frag`) where Linguist itself has been vendored (its heuristics are often used in text editor plugins for file extension detection).
      - [Search results for `extension:frag language:js in:path,file window NOT node_modules NOT glslang NOT void NOT struct NOT asm`](https://github.com/search?q=extension%3Afrag+language%3Ajs+in%3Apath%2Cfile+window+NOT+node_modules+NOT+glslang+NOT+void+NOT+struct+NOT+asm&type=Code) (~600 results) - `window` is another word I thought wouldn't occur randomly in GLSL code. Notable results here include: some C code from a frequently vendored X11 library, some Java sample file from Firefox called `ThemedView.java.frag`, and legitimate JS code fragments for wrapping the program in an IIFE, from ~2014.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2847,6 +2847,7 @@ JavaScript:
   - ".gs"
   - ".jake"
   - ".javascript"
+  - ".js.frag"
   - ".jsb"
   - ".jscad"
   - ".jsfl"
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -190,6 +190,11 @@ disambiguations:
   - language: Frege
     pattern: '^\s*(import|module|package|data|type) '
   - language: Text
+- extensions: ['.frag']
+  rules:
+  - language: JavaScript
+    pattern: '^\s*(?:;\s*)?[(!]function\s*\('
+  - language: GLSL
 - extensions: ['.fs']
   rules:
   - language: Forth


Alhadis:
We don't remove extensions or languages because of a decline in popularity, as that would entail an almost unthinkable amount of on-going surveillance. All that matters is that an extension *was* [sufficiently popular](https://github.com/github/linguist/blob/HEAD/CONTRIBUTING.md#adding-an-extension-to-a-language) once upon a time on GitHub, and I'm certain there are other extensions which are no longer seeing active use.

That being said, all we can really do is refine our heuristics or supply our classifier with more samples.

> so now it is hard to find legitimate usages of this file extension in contemporary JS projects.

Those difficulties might be compounded by the fact that GitHub [no longer indexes](https://docs.github.com/en/search-github/searching-on-github/searching-code#considerations-for-code-search) projects that haven't been updated or searched in the past year:
> * Only repositories that have had activity or have been returned in search results in the last year are searchable.

Hence there may very well be many old projects with JS `.frag` files that aren't listed in search results.

dmitmel:
Very well, I'll add an entry to `heuristics.yml` instead.

dmitmel:
Updated.

lib/linguist/languages.yml
```diff
@@ -2847,6 +2847,7 @@ JavaScript:
   - ".gs"
   - ".jake"
   - ".javascript"
+  - ".js.frag"
```
lildude, line 2850: Whoops, looks like you left this over from your original commits. This isn't needed. It'll be caught by the `.frag` above and your new heuristic.

lildude (changes requested):
Please also update the OP to reflect the actual changes in this PR.

lib/linguist/languages.yml
dmitmel: No, I wanted to make it so that `.js.frag` enforces the usage of JS syntax highlighting, as I imagine old projects would do that (because if you treat the `.frag` extension as meaning "fragment of a file" , then naming your file `start.frag` in a Javascript project is like naming a file `program.txt` and expecting the syntax highlighting to work). Also, I wrote a heuristic only for a start of an IIFE (I imagine that the vast majority of usages of `.frag` for JS are for module wrappers as I have explained in the first post) because, in my opinion, JS should be the fallback choice here. Although GLSL fragment shaders have a few much more recognizable keywords, and a heuristic for them already exists, so it might be more reasonable to use that. What do you think?

lildude: The language detection, by whatever means (ext, heuristic etc) determines the syntax highlighting. We can add this extension but then you need to fulfill all the requirements for adding a new extension.

PR #5567, opened 2021-09-11
Add support for Jest Snapshots

Nixinova:
- Close #5564

## Description

Support [Jest Snapshot](https://jestjs.io/docs/snapshot-testing) files.
Basically JavaScript, but has its own tm-scope to support the highlighting inside \`backticks\`.
Classify these as generated.

## Checklist

- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      - [.snap](https://github.com/search?q=extension%3Asnap+%22jest%22&type=code) -- 680K files
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample: [source](https://github.com/react-native-svg/react-native-svg/blob/develop/__tests__/__snapshots__/css.test.tsx.snap)
    - Sample license: [MIT](https://github.com/react-native-svg/react-native-svg/blob/develop/LICENSE)
  
  - [x] I have included a syntax highlighting grammar: [source](https://github.com/jest-community/vscode-jest/blob/master/syntaxes/jest-snapshot.tmLanguage), [preview](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fjest-community%2Fvscode-jest%2Fblob%2Fmaster%2Fsyntaxes%2Fjest-snapshot.tmLanguage%3Frgh-link-date%3D2021-09-10T06%253A10%253A32Z&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Freact-native-svg%2Freact-native-svg%2Fblob%2Fdevelop%2F__tests__%2F__snapshots__%2Fcss.test.tsx.snap&code=) (slightly broken)
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

PR #5646, opened 2021-11-17
Stop flagging Jest snapshots as generated files

Alhadis:
See [5567](https://github.com/github/linguist/pull/5567#issuecomment-970906650). This is essentially #3579 part 2: Electric Boogaloo.

I've added regression tests to catch any future PRs that add Jest snapshots to the generated files list. I gave Yarn lockfiles the same treatment, since they're in the same boat.

/cc @joscha

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -83,7 +83,6 @@ module Linguist
       generated_apache_thrift? ||
       generated_jni_header? ||
       vcr_cassette? ||
-      jest_snapshot? ||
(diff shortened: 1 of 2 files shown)


joscha (approved):
Nice one, thank you for actioning it so fast!

Alhadis:
A fast response was the very least I could do after such an embarrassing brainfart. 😓

lildude (approved):
Nice one. Thanks @Alhadis. 🙇

PR #5670, opened 2021-12-02
Highlight .gitattributes file should be committed

ekroon:
I added clarification in the documentation about the need to commit `.gitattributes` file while using Linguist locally.

Screenshot 2021-12-02 at 08 12 33


docs/overrides.md
```diff
@@ -8,7 +8,8 @@ Add [a `.gitattributes` file](https://git-scm.com/docs/gitattributes) to your pr
 `.gitattributes` will be used to determine language statistics and will be used to syntax highlight files.
 You can also manually set syntax highlighting using [Vim or Emacs modelines](#using-emacs-or-vim-modelines).
 
-When testing with a local installation of Linguist, take note that the added attributes will not take effect until the `.gitattributes` file is committed to your repository.
+:bangbang: | When testing with a local installation of Linguist, take note that the added attributes will not take effect until the `.gitattributes` file is committed to your repository.
+:---: | :---
```
Alhadis, line 12:
```suggestion
When testing with a local installation of Linguist, **take note that the added attributes will _not_ take effect until the `.gitattributes` file is committed to your repository.**
```

Alhadis (changes requested):
This document is full of important points. What makes this one deserving of a callout?

ekroon:
I found it very unintuitive that my changes to '.gitattributes' weren't picked up. And then this line is very important to pick out of the documentation.

PR #5681, opened 2021-12-10
Exclude generated yarn.lock files

colinwm:


lildude:
We deliberately don't mark these as generated and our test has caught your attempt to change that 😁. See https://github.com/github/linguist/issues/4348 for more details as to why.

Closing.

colinwm:
Oops! Missed that discussion, sorry about that. But it clearly _is_ generated... I guess the purpose of linguist is more to identify what should be shown in diffs than to categorize code?

PR #5700, opened 2021-12-22
Added OpenRA MiniYAML

Mailaender:
@OpenRA uses its own niche YAML variant which causes problems. This adds support for it.

## Description
The MiniYAML variant is older than the YAML standard, and it deviates a quite a bit. Most notably it uses tabs by default https://github.com/OpenRA/OpenRA/issues/2310 which GitHub started to mark as an error which leads to confusion. Attempts to change the file ending https://github.com/OpenRA/OpenRA/pull/13910 have failed. Using SaltStack .yaml turned out to be a workaround, but in https://github.com/OpenRA/OpenRA/pull/19850 we decided to fix this properly with our own parser.

## Checklist:
      -  https://github.com/search?l=YAML&q=WorldInteractionController&type=Code
    - Sample source(s):
      - [OpenRA Mod SDK MiniYAML](https://github.com/OpenRA/OpenRAModSDK)
      - [229Q.yaml](https://github.com/yaml/yaml-test-suite/blob/main/src/229Q.yaml)
    - Sample license(s):
      - Public Domain
      - [MIT](https://github.com/yaml/yaml-test-suite/blob/main/License)
  - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2FOpenRA%2Fatom-miniyaml%2Fblob%2Fmain%2Fgrammars%2Fminiyaml.cson&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2FOpenRA%2FOpenRA%2Fbleed%2Fmods%2Fra%2Frules%2Fdefaults.yaml&code=


lildude (changes requested):

lib/linguist/heuristics.yml
```diff
@@ -618,6 +618,11 @@ disambiguations:
     pattern: '^%(end|ctor|hook|group)\b'
   - language: Linker Script
     pattern: 'OUTPUT_ARCH\(|OUTPUT_FORMAT\(|SECTIONS'
+- extensions: ['.yaml']
+  rules:
+  - language: MiniYAML
+    negative_pattern: '---'
```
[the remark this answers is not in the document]
Mailaender: While spaces are allowed, it is recommended to use tabs so I try to use this as a positive pattern now.

lib/linguist/heuristics.yml
```diff
@@ -618,6 +618,12 @@ disambiguations:
     pattern: '^%(end|ctor|hook|group)\b'
   - language: Linker Script
     pattern: 'OUTPUT_ARCH\(|OUTPUT_FORMAT\(|SECTIONS'
+- extensions: ['.yaml']
+  rules:
+  - language: MiniYAML
+    pattern: '\t\S*.*:'
```
letmaik, line 624: Wouldn't this match a regular YAML file having a string that follows this pattern? Isn't there a `^` missing at the beginning? Something like `^\t+.*:` maybe? I'm not sure why the `\S*` is necessary.

```yaml
foo: "	bar:"
```

Mailaender: Turns out this is valid MiniYAML
```salt
Grenade :
	ValidTargets : Ground, Trees
```
although I have never seen anyone do it.

lib/linguist/heuristics.yml
```diff
@@ -630,6 +630,12 @@ disambiguations:
     pattern: '^%(end|ctor|hook|group)\b'
   - language: Linker Script
     pattern: 'OUTPUT_ARCH\(|OUTPUT_FORMAT\(|SECTIONS'
+- extensions: ['.yaml']
+  rules:
+  - language: MiniYAML
+    pattern: '^\t+.*:'
```
Alhadis, line 636: If you intend to match tab-indented fields like `\tfoo:`, then I recommend matching at least one non-whitespace, non-colon character as part of the field-name:

```suggestion
    pattern: '^\t+.*?[^\s:].*?:'
```

Alhadis (changes requested):

lib/linguist/heuristics.yml
letmaik: Curious, what's the difference between `.*` and `.*?`, aren't they the same given that `*` matches empty as well and `?` just means optional?

Alhadis: Normally, a regex will attempt to match the longest possible string that it can; for example:

~~~perl
/Foo.+/ ~= "Foo_123456789" # => returns the entire line, "Foo_123456789"
~~~

Affixing `?` to a quantifier enables a so-called "conservative" mode (a.k.a., "thrifty", "cautious", "non-greedy", etc). When marked as thrify, a quantifier will match the fewest characters it needs to in order to satisfy the criteria:

~~~perl
/Foo.+?/ ~= "Foo_123456789" # Returns only "Foo_"
~~~

Thrify expressions can be decidedly more performant than greedy ones, but only in the event of a successful match.

Note that you can't quantify a quantifier; an optional `+` or `?` wouldn't make sense. 😉 Hence why the syntax was chosen to avoid conflicting with other regular expression flavours. You can thank Perl for that.

Mailaender: Updated.

Alhadis (approved):

Alhadis:
@lildude Looks good to merge to me. 👍

lildude (approved):

PR #5731, opened 2022-01-15
Change Yarn PnP regex to include all `.pnp.*` files

brc-dd:
## Description

According to Yarn docs, they may generate any `.pnp.*` file ([ref](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored)). Currently only `.pnp.cjs`, `.pnp.mjs`, `.pnp.js` are ignored from stats, but Yarn now also generates `.pnp.loader.mjs`.

With old regex: https://regexr.com/6dchf
With new regex: https://regexr.com/6dchc

## Checklist:

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -413,7 +413,7 @@ module Linguist
     #
     # Returns true or false.
     def generated_yarn_plugnplay?
-      !!name.match(/(^|\/)\.pnp\.(c|m)?js$/)
+      !!name.match(/(^|\/)\.pnp.*$/)
     end
 
     # Internal: Is the blob part of Godeps/,
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -413,7 +413,7 @@ module Linguist
     #
     # Returns true or false.
     def generated_yarn_plugnplay?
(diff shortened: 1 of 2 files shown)


Alhadis (approved):
First of all, the current expression already matches patterns like `.pnp.foo.mjs`, as `.` in a regex will match any single character (including a dot).

Secondly, the current expression is also **too lax**, meaning it matches filenames like `.pnpfoo`. So these changes are still valid.

brc-dd:
@Alhadis I guess you are looking at wrong changes (probably of that typo-fix commit).

Here are two regexes: https://github.com/github/linguist/pull/5731/files#diff-e142be54023b2e4e695deb6f377e16d9f44e6c846a61f88fcec268aaad4af851L416-R416

Earlier there was `/(^|\/)\.pnp\.(c|m)?js$/` which doesn't match `.pnp.foo.mjs` nor `.pnpfoo`. The new expression `/(^|\/)\.pnp\..*$/` matches `.pnp.foo.mjs` but not `.pnpfoo`. I have attached the links in the PR body.

Alhadis:
> I guess you are looking at wrong changes (probably that of the typo-fix commit).

Yes, you're right. 😞 My bad.

lildude (approved):

PR #5742, opened 2022-01-20
Change C++ color to match the official logo

gikari:
## Description
The color is made to be close to the official logo of C++ (instead of the current pink).

See https://isocpp.org/ for the logo.

image

New color:
Screenshot_20220120_221200

Screenshot_20220120_221326

## Checklist:

  - [x] I have obtained agreement from the wider language community on this color change.
    - [Reddit Poll](https://www.reddit.com/r/cpp/comments/s8s6ru/should_the_color_of_the_github_languages_stripe/)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -674,7 +674,7 @@ C++:
   ace_mode: c_cpp
   codemirror_mode: clike
   codemirror_mime_type: text/x-c++src
-  color: "#f34b7d"
+  color: "#004482"
   aliases:
   - cpp
   extensions:


gikari:
From the Reddit poll, it seems like the community response was positive.

Alhadis (commented):
I'm very, _very_ tentatively in favour of the change; however, I've a gut feeling this is going to stir some folks who're used to identifying C++ projects from the garish pink. It's an admittedly ugly colour, but it's one that's been around for as long as [GitHub's been using language colours](https://github.com/github/linguist/commit/c3d6fc5af8cf9d67afa572bba363bf0db256a900#diff-9c337ec09ccaaa1374713aad9ba26cdc1308586a27cb309a883a0b19bde663d8R174). Given the size of C++'s userbase, and the language's widespread popularity, I seriously doubt the change will sit well with every C++ user on GitHub. Even if it's sanctioned by branding [endorsed by the Standard C++ Foundation](https://github.com/isocpp/logos).

@lildude, could I have your input on this, please? I'm not comfortable making a judgement call on this by myself.

Alhadis:
Also, I should point out that not every vote on the aforementioned Reddit poll was positive. In fact, it seems split 50%/50%:

Figure 1

I'm also not convinced a Reddit poll is a faithful reflection on the opinions of the _entire_ C++ community, vast as it is…

gikari:
> Also, I should point out that not every vote on the aforementioned Reddit poll was positive. In fact, it seems split 50%/50%:

98 vs 48 is not a 50/50. It's 2/3 in favor and 1/3 is against. The down votes percentage does not matter, because the down votes could be interpreted as "This post is irrelevant and doesn't matter", i.e. people are downvoting because they do not care.

 
> I'm also not convinced a Reddit poll is a faithful reflection on the opinions of the _entire_ C++ community, vast as it is…

Yes, but what metric could be better? r/cpp has 200k members, so I decided, that it would be a reasonable place to take a sample.

Alhadis:
> 98 vs 48 is not a 50/50. It's 2/3 in favor and 1/3 is against. 

Oops, you're right. 😞 My bad… I'm dyscalculic.

> r/cpp has 200k members, so I decided, that it would be a reasonable place to take a sample.

Yes, but high-traffic subreddits have a habit of burying new polls and threads unless they garner a significant amount of upvotes/discussion. It's possible the poll was only seen by a handful of users... 🤷‍♂️

> Yes, but what metric could be better?

I'm honestly not sure. Literally the only thing I can think of is just running with the change and copping the collective flak after the change goes live. Which happened in #4319 with Rust, and I can't say I'm eager to repeat that debacle…

lildude:
Yeah, I share your concerns @Alhadis. This is going to need way more public votes and buy-in for it to be considered. 147 votes is a long long way from any sort of indication of general community support for such a popular language with over 1M repos on GitHub. I'm also not sure Reddit is really the best place for such a survey either, but I'm not a C++ programmer so I don't know where discussions normally take place.

lildude:
Closing this as there's been no progress on establishing a discussion within formal C++ community channels. Feel free to open a new PR once this has been done and concensus has been agreed.

PR #5743, opened 2022-01-20
Add tinyobjloader as a C sample

XrXr:
## Description
For some reason this file is recognize as C++ in a [repo] of mine, even though it
[compiles] without warnings on -std=c89 with GCC. This code is from
https://github.com/syoyo/tinyobjloader-c

I have not checked whether adding this sample would fix the classification as
the setup process looks too onerous for my slow internet connection.

## Checklist:

    - Sample source(s): [repo]
    - Sample license(s): MIT

[repo]: https://github.com/XrXr/FantasyGolfSimulator
[compiles]: https://godbolt.org/z/d4e71c4bj

### DIFF
--- a/samples/C/tinyobj_loader_c.h
+++ b/samples/C/tinyobj_loader_c.h
@@ -1,0 +1,1288 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016 Syoyo Fujita and many contributors.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


lildude:
I don't think adding this is the right thing to do.

We have a specific section about the C family header files in the [docs](https://github.com/github/linguist/blob/master/docs/troubleshooting.md#my-ccobjective-c-h-header-file-is-detected-as-the-wrong-language). 

If this file isn't correctly identified as C (the default) it's very likely to contain syntax that isn't strictly C and thus is being detected as C++ by the heuristics which means either your file contains C++ syntax or the heuristic needs improving.

XrXr:
Thank you for the guidance @lildude. I did see #5357 so it was puzzling that this file is recognized as C++. I'm pretty sure this file is strictly C89 because both Clang and GCC [agree](https://godbolt.org/z/GdoqPTnEv), so I think the next step is to figure out how to improve the C++ heuristics. I'm going to close this and look into that should I find the time.

lildude:
@XrXr this is why the file is being considered C++: it's the `std::` in the commented out code:

https://github.com/syoyo/tinyobjloader-c/blob/574390baed2a42705850392d0c75c09f8cc980bd/tinyobj_loader_c.h#L1471-L1484

It's commented out so has no effect when you're compiling etc but there is no way for the heuristic regex to know that 😄.

XrXr:
Ah interesting! Thanks for spending time on that. I guess for the C families it's fully heuristics based and Linguist doesn't attempt to parse the files with grammars.

PR #5745, opened 2022-01-21
Slightly adjust bootstrap regex

colinwm:
The bootstrap regex here is a big excessive, e.g. it matches paths like:

`/src/bootstraps/settings.js`

which is human writen code, whereas I think it was intended to match things like

`/src/bootstrap-custom.js`

I made it so that the wildcard (which currently doesn't match dots) also doesn't match slashes, which restrains it a bit.

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -74,7 +74,7 @@
 - ([^\s]*)import\.(css|less|scss|styl)$
 
 # Bootstrap css and js
-- (^|/)bootstrap([^.]*)\.(js|css|less|scss|styl)$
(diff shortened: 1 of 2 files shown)


look:
Please add some tests to this section to cover these cases:

https://github.com/github/linguist/blob/80f3531e8a1014a23f4606458e5a528053ed3cac/test/test_file_blob.rb#L243-L245

colinwm:
@look good idea, done. Let me know if you have more test ideas

look (approved):
I'm a little shocked those tests worked since `sample_blob` usually has a corresponding file, but I guess in this case it doesn't need to read the content.

PR #5769, opened 2022-02-08
Fix brew install command for docker

ADawesomeguy:
"brew cask install" is no longer a command, it's now "brew install --cask"


ADawesomeguy:
Wait someone just submitted a similar PR, I just had an outdated local repo, my bad.

PR #5775, opened 2022-02-10
Add Solidity samples

davidhq:
Adding Solidity language samples from OpenZeppelin

## Checklist:

    - Sample source(s):
      - https://github.com/OpenZeppelin/openzeppelin-contracts
    - Sample license(s):
    - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/LICENSE


lildude:
ProTip: always state why you are doing something when submitting a PR 😉 

Why are you adding these and why so many samples? 2-3 is generally more than enough.

It's worth pointing out that these will have no effect and will cause the tests to fail as Solidity is treated slightly differently. Because the `.sol` extension is so generic we were forced to come up with a whole new approach for such extensions in https://github.com/github/linguist/pull/4936 which means identification is ultimately based on the heuristics. If you're finding files that aren't being classified correctly, then the heuristic needs to be improved with files added to the list of fixtures to test and confirm the heuristic.

davidhq:
> ProTip: always state why you are doing something when submitting a PR 😉
> 
> Why are you adding these and why so many samples? 2-3 is generally more than enough.
> 
> It's worth pointing out that these will have no effect and will cause the tests to fail as Solidity is treated slightly differently. Because the `.sol` extension is so generic we were forced to come up with a whole new approach for such extensions in #4936 which means identification is ultimately based on the heuristics. If you're finding files that aren't being classified correctly, then the heuristic needs to be improved with files added to the list of fixtures to test and confirm the heuristic.

ok, great info and useful pointer to the other thread, will get educated, closing for now

PR #5783, opened 2022-02-14
Fix incorrect capitalisation of 'Unity3D Asset' heuristic

Nixinova:
```diff
-- language: Unity3d Asset
+- language: Unity3D Asset
```

## Description

Fixes a capitalization typo in the heuristics file.
Same problem to #5434 and caused the same issue
I ran a script to check all of the other language names in the file and all other ones should be good - the only values it found didn't have a corresponding entry were `Unity3d Asset`, and `[ 'Linux Kernel Module', 'AMPL' ]` which both have corresponding entries.

## Checklist:

         &uarr; probably should be done

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -335,7 +335,7 @@ disambiguations:
   - language: 'M4'
 - extensions: ['.mask']
   rules:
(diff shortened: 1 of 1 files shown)


lildude (approved):
🤦

PR #5795, opened 2022-02-20
Fix note in docs about which types are included

Nixinova:
## Description

The docs misstated which languages are included in lang stats.
Fixes the line and mentions the other types too.

## Checklist
n.a.


Alhadis (approved):
🤦‍♂️ Hidden right in plain sight…

lildude (approved):
Whoops 😬

PR #5811, opened 2022-03-02
Add LigoLANG

pewulfman:
## Checklist:

      -  https://github.com/search?q=extension%3Aligo&type=code
      - https://github.com/search?q=extension%3Amligo&type=code
      - https://github.com/search?q=extension%3Areligo&type=code
    - Sample source(s):
    - Sample license(s): MIT
  
  - [X] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fpewulfman%2FLigo-grammar%2Fblob%2Fmain%2Fsyntaxes%2Fmligo.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fgithub.com%2Fpewulfman%2FTezos-TZIP-implementation%2Fblob%2Fmain%2FTZIP-12%2520%28FA2%29%2Fcontract%2FFA2_NFT.mligo&code=


lib/linguist/languages.yml
```diff
@@ -3306,6 +3306,30 @@ Lex:
   tm_scope: source.lex
   ace_mode: text
   language_id: 199
+LigoLANG-Cameligo:
```
Nixinova, line 3309: From the [branding](https://ligolang.org/):
```suggestion
CameLIGO:
```

lib/linguist/languages.yml
```diff
@@ -3306,6 +3306,30 @@ Lex:
...
+  type: programming
+  extensions:
+  - ".mligo"
+  tm_scope: source.mligo
+  ace_mode: text
+  group: LigoLANG
+  language_id: 829207807
+LigoLANG-Pascaligo:
```
Nixinova, line 3317:
```suggestion
PascaLIGO:
```

lib/linguist/languages.yml
```diff
@@ -3306,6 +3306,30 @@ Lex:
...
+  type: programming
+  extensions:
+  - ".ligo"
+  tm_scope: source.ligo
+  ace_mode: text
+  group: LigoLANG
+  language_id: 1040646257
+LigoLANG-Reasonligo:
```
Nixinova, line 3325:
```suggestion
ReasonLIGO:
```

lildude (changes requested):
We're going to need a sample for each extension you're adding support for.

pewulfman:
Thanks for the reviews

> We're going to need a sample for each extension you're adding support for.

I have added samples for the other two languages.

> These languages are syntactically similar enough to their namesakes that we can afford to reuse their CodeMirror and ACE modes.

thanks, I committed your suggestions :)

lib/linguist/languages.yml
```diff
@@ -3306,6 +3306,36 @@ Lex:
...
   language_id: 199
+CameLIGO:
+  type: programming
+  extensions:
+  - ".mligo"
+  tm_scope: source.mligo
+  ace_mode: ocaml
+  codemirror_mode: mllike
+  codemirror_mime_type: text/x-ocaml
+  group: LigoLANG
```
Nixinova, line 3317: The group key won't do anything right now as there is no language listed here called 'LigoLANG'.
Are all variants of Ligo co-equal? If so then I'm not sure what would be done here if grouping is wanted, otherwise the other two should be grouped under the first.

pewulfman: So your suggestion is to choose one and make it the "standard". If we don't want to I guess we need to make some change some change in Linguist ot support this :thinking: .

What if we make make a LigoLANG, but we only make the extension of configuration file, like `.json`, would that work then?

lildude: > So your suggestion is to choose one and make it the "standard".

Yes, that would be best. From the look of the grammars you're adding, PascalLIGO may be a good candidate as it's simply named `ligo` in the grammar and has a scope of `source.ligo`; neither of these indicate the dialect.

You can also add an alias for those that want to override or use the name in codeblocks.

Alternatively, pick the most common dialect that people tend to use when referring to "LigoLANG".

> What if we make a LigoLANG language, but we only give the extension of configuration files, like `.json`, would that work then?

This is cheating 😉 ... and whilst it would "work", I don't recommend this as this is likely to upset users as most would expect a JSON file to be identified as such. Adding `.json` to another will also significantly increase the chances of misclassification of non-Ligo related uses, especially if a precise and 100% accurate heuristic can't be written.

lildude: > So your suggestion is to choose one and make it the "standard".

Yes, that would be best. From the look of the grammars you're adding, PascalLIGO may be a good candidate as it's simply named `ligo` in the grammar and has a scope of `source.ligo`; neither of these indicate the dialect.

You can also add an alias for those that want to override or use the name in codeblocks.

Alternatively, pick the most common dialect that people tend to use when referring to "LigoLANG".

> What if we make a LigoLANG language, but we only give the extension of configuration files, like `.json`, would that work then?

This is cheating 😉 ... and whilst it would "work", I don't recommend this as this is likely to upset users as most would expect a JSON file to be identified as such. Adding `.json` to another will also significantly increase the chances of misclassification of non-Ligo related uses, especially if a precise and 100% accurate heuristic can't be written.

Nixinova: (Your comment was sent twice)

Alhadis (approved):

lib/linguist/languages.yml
pewulfman: Following your advice, I put Pascaligo as default. This may change in the future

lib/linguist/languages.yml
```diff
@@ -3306,6 +3306,36 @@ Lex:
   tm_scope: source.lex
   ace_mode: text
   language_id: 199
+LigoLANG:
+  type: programming
```
lildude, line 3310: You probably want to add a `color` else it'll default to a very boring grey.

lildude:
Looks like you've got some test failures to address too.

pewulfman:
> Looks like you've got some test failures to address to

you are right. It should be fix now

lib/linguist/languages.yml
pewulfman: Thanks we are discussing it with the team now. 
I will also probably change the git repo to the official LSP repo (can't at the moment because the directory layout make the import bug)

pewulfman: color added

pewulfman:
Ok. We should be ready to merged

Issue #5808, opened 2022-02-26
Installation points to `libruby.so.2.7` on Linux


zedseven:
## Describe the bug
After a fresh install with `gem install github-linguist` in Arch Linux on a Ruby 3.0.3 installation, running the tool gives the following error:
```
$ github-linguist
<internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require': libruby.so.2.7: cannot open shared object file: No such file or directory - /home/zacc/.local/share/gem/ruby/3.0.0/gems/github-linguist-7.19.0/lib/linguist/linguist.so (LoadError)
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /home/zacc/.local/share/gem/ruby/3.0.0/gems/github-linguist-7.19.0/lib/linguist/tokenizer.rb:2:in `<top (required)>'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /home/zacc/.local/share/gem/ruby/3.0.0/gems/github-linguist-7.19.0/lib/linguist/classifier.rb:1:in `<top (required)>'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /home/zacc/.local/share/gem/ruby/3.0.0/gems/github-linguist-7.19.0/lib/linguist/language.rb:9:in `<top (required)>'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /home/zacc/.local/share/gem/ruby/3.0.0/gems/github-linguist-7.19.0/lib/linguist.rb:5:in `<top (required)>'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
[...]

PR #5818, opened 2022-03-04
Exclude compiled extension from gem

lildude:
## Description

For a long time now we've been accidentally including the compiled C extension used by Linguist for token parsing when creating the gem.

This has gone unnoticed until https://github.com/github/linguist/issues/5808 because I used to build the gem on macOS and the resulting gem included the macOS library and not the Linux library. The Linux library would then have been built at the time of installation, hence no problem.

I've recently switched to using Codespaces and so now the Linux library is being built against the version of Ruby on the Codespace env at the time and then included in the gem, which at the time of building the 7.19.0 gem was running Ruby 2.7.2. This wouldn't be a problem in envs with  Ruby 2.7 installed, but is if the env only has Ruby 3 installed.

This PR fixes that by explicitly excluding the macOS `linguist.bundle` and Linux `linguist.so` files when building the gem.

Once merged, I'll start a new release to get a working gem out.

Fixes https://github.com/github/linguist/issues/5808

_[ Checklist removed as it doesn't apply ]_

### DIFF
--- a/github-linguist.gemspec
+++ b/github-linguist.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
     "github_repo" => "ssh://github.com/github/linguist"
   }
 
-  s.files = Dir['lib/**/*', 'ext/**/*', 'grammars/*', 'LICENSE'].reject {|f| f.match(%r{lib/linguist/linguist\.(so|bundle)}) }
+  s.files = Dir['{lib,ext}/**/*', 'grammars/*', 'LICENSE'] - Dir['lib/linguist/linguist.{so,bundle}']
   s.platform = Gem::Platform::RUBY
   s.executables = ['github-linguist', 'git-linguist']
   s.extensions = ['ext/linguist/extconf.rb']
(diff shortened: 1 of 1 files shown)


github-linguist.gemspec
```diff
@@ -13,7 +13,8 @@ Gem::Specification.new do |s|
     "github_repo" => "ssh://github.com/github/linguist"
   }
 
-  s.files = Dir['lib/**/*'] + Dir['ext/**/*'] + Dir['grammars/*'] + ['LICENSE']
+  s.files = Dir['lib/**/*', 'ext/**/*', 'grammars/*', 'LICENSE'].reject {|f| f.match(%r{lib/linguist/linguist\.(so|bundle)}) }
```
Alhadis, line 16: Not a blocker, but this strikes me as more idiomatic Ruby:

~~~suggestion
  s.files = Dir['{lib,ext}/**/*', 'grammars/*', 'LICENSE'] - Dir['lib/linguist/linguist.{so,bundle}']
~~~

lildude: 🤦 Why didn't I think of that? Thanks. 🙇

Alhadis: Always helps to have another set of eyes. 😉

Alhadis (approved):

PR #5880, opened 2022-04-26
Replace all `git.io` links with their actual URLs

SukkaW:
## Description

All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Replace all git.io links with their actual URLs.


lildude (approved):
Thanks. You beat me to it 😆

PR #5881, opened 2022-04-28
Revert removal of `git.io` links

Alhadis:
There's been an update to GitHub's blog post about `git.io` deprecation:

> **2022-04-27 Update:**
While the git.io URL redirection service is [read-only](https://github.blog/2011-11-10-git-io-github-url-shortener/) and use of the service is limited, we have received feedback from developers and academic researchers who have published git.io links in print documentation and research papers. In order to preserve the integrity of these historical documents, we have decided to archive the current git.io links in a new read-only service that will allow us to serve redirects for those links longer term.

This PR restores three shortlinks that were removed in #5880 following the initial announcement.


lildude:
I'm in two minds about reverting this. Whilst it's no longer necessary now, it'll probably be necessary to change one or more of these links in future. No particularly fussed either way.

Alhadis:
> It'll probably be necessary to change one or more of these links in future

The same holds true of most other URL shorteners too. Bit.ly, for example, doesn't permit modification of existing links, so… one way or the other, a URL is gonna need to be updated. 😜

The reason I originally shortened these URLs is because they point to ancillary resources that are only tangentially related to their context. Ergo, I felt short, unobtrusive links like [`https://git.io/JULye`](https://git.io/JULye) were more appropriate than something as noisy as [`https://github.com/github/linguist/pull/4734#discussion_r479653547`](https://github.com/github/linguist/pull/4734#discussion_r479653547). 😉

lildude (approved):
No compelling reason to change these now git.io is hanging about so lets put them back.

PR #5889, opened 2022-05-10
Mark .github as vendored

daenney:
## Description

Things in `.github/` are used to configure behaviour on GitHub.com, such
as issue templates, or GitHub Actions. They should not be taken into
consideration when trying to figure out the language(s) of a repository.

## Checklist:

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -383,3 +383,6 @@
 
 # Gitpod
 - (^|/)\.gitpod\.Dockerfile$
+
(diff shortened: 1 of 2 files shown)


daenney:
As an example, a small Go package but with a lot of scripts for GitHub Actions in `.github/` is classified as 50% shell: https://github.com/pion/randutill, despite the fact that all code is actually Go. The repo should be classified as ~100% Go.

lildude (approved):
Makes sense to me. 👍

PR #5903, opened 2022-05-21
Add support for Git Revision List files

### LINKED ISSUE #5901: Add git-blame-ignore-revs
## Language name
git-blame-ignore-revs

## URL of example repository
https://github.com/babel/babel/blob/a24c1d047e6db77c32d9505c6876897b4c3689d2/.git-blame-ignore-revs
https://github.com/django/django/blob/19dc3f0f96908898fd62c62517e935331ea29628/.git-blame-ignore-revs
https://github.com/electron/electron/blob/8b5613efcbd322ea19407f8655b4691b09c3b2b4/.git-blame-ignore-revs

Not sure, but the syntax is roughly:

`comment`:
```
^#.*$
```

`sha`:
```
^[0-9a-fA-F]{40}$
```

## Most popular extensions

`.git-blame-ignore-revs`

## Detected language

Doesn't appear to be detected as anything if I'm reading https://cs.github.com/?scopeName=All+repos&scope=&q=path%3A.git-blame-ignore-revs correctly.

Nixinova:
- Resolve #5901

## Description

Adds the simple [git-blame-ignore-revs](https://www.git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt) file format, which is just a list of commit hashes with hash-comments allowed.

## Checklist:

      -  [filename:git-blame-ignore-revs](https://github.com/search?q=filename%3Agit-blame-ignore-revs&type=code): 3.9K+ results, [3.1K](https://github.com/search?q=filename%3Agit-blame-ignore-revs+path%3A%2F&type=code) in root (&plusmn;1K because GitHub search is screwy)
    - Sample: [.git-blame-ignore-revs](https://github.com/babel/babel/blob/ee4576f6c0d25bd0c64a7a891c34e28fc016b974/.git-blame-ignore-revs), license [MIT](https://github.com/babel/babel/blob/ee4576f6c0d25bd0c64a7a891c34e28fc016b974/LICENSE)
    - [grammar repo](https://github.com/Nixinova/NovaGrammars)
    - [Preview](https://novalightshow.netlify.app/?grammar-type=url&grammar=https%3A%2F%2Fraw.githubusercontent.com%2FNixinova%2FNovaGrammars%2Fmain%2Fgrammars%2Fgit-blame-ignore-revs.yaml-tmLanguage&sample-type=text&sample=%23+git-blame+ignored+revisions%0A%23+To+configure%2C+run%0A%23+++git+config+blame.ignoreRevsFile+.git-blame-ignore-revs%0A%23+Requires+Git+%3E+2.23%0A%23+See+https%3A%2F%2Fgit-scm.com%2Fdocs%2Fgit-blame%23Documentation%2Fgit-blame.txt---ignore-revs-fileltfilegt%0AInvalid+content+right+here%0A%0A%23+Apply+prettier%0A5180ecdca48e486b4b0f347fcbd752f865df5e55%0A%0A%23+Update+prettier%0A41085248560b1403b8d0f99f108491e679531c6c) (I revived Lightshow)

Not sure whether the submodule updating was committed.


lib/linguist/languages.yml
```diff
@@ -2176,6 +2176,16 @@ Git Attributes:
   codemirror_mode: shell
   codemirror_mime_type: text/x-sh
   language_id: 956324166
+Git Blame Ignore:
```
Alhadis, line 2179: I'd name this `Git Revision List` instead. Both `git blame --ignore-revs-file`[^1] and the `fsck.skipList` config setting use the same file-format, despite being two fundamentally different operations. From [`Documentation/config/fsck.txt`](https://github.com/git/git/blob/bc42e1e767fa86261fb5fdc5a495ad3ee14a9e0b/Documentation/config/fsck.txt):


> <dl><dt><b>fsck.skipList::</b></dt><dd>The path to a list of object names (i.e. one unabbreviated SHA-1 per line) that are known to be broken in a non-fatal way and should be ignored. On versions of Git 2.20 and later comments ('#'), empty lines, and any leading and trailing whitespace is ignored. Everything but a SHA-1 per line will error out on older versions.</dd></dl>

[^1]: https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt <q>Ignore revisions listed in <var>file</var>, which must be in the same format as an <code>fsck.skipList</code></q>

lib/linguist/languages.yml
```diff
@@ -2176,6 +2176,16 @@ Git Attributes:
...
+  type: data
+  color: "#F44D27"
+  aliases:
+  - Git Blame Ignore Revs
+  filenames:
+  - ".git-blame-ignore-revs"
+  tm_scope: text.git-blame-ignore
```
Alhadis, line 2186: **Nit:** The `scopeName` should (ideally) be prefixed by `source` instead of `text`.

Rationale: Generally speaking, `prose` and `markup`-type languages use grammars with scopeNames beginning with `text`, whereas `programming` and `data` languages are left with `source` instead.

Alhadis (changes requested):

lildude:
🤔 does this really need its own entry? Why not add it to the list of files under the “Ignore List”. From a quick test it highlights fine.

Nixinova:
> does this really need its own entry? Why not add it to the list of files under the “Ignore List”? From a quick test it highlights fine.

I defaulted to using a separate language so it could have the invalid content highlighting but I can change it to be under Ignore-List if you would prefer that.

----

\`\`\`git-blame-ignore
image


\`\`\`ignore-list
image

Alhadis:
> so maybe it's worth updating that instead?

@lildude The grammar used to highlight `.gitignore` files is actually responsible for highlighting [several nearly-identical filetypes](https://github.com/github/linguist/blob/7503f7588ca2590cebc4a05af7d00a7e94c0d779/lib/linguist/languages.yml#L2698-L2726), most of them deliberately modelled off the familiar [`gitignore(5)`](https://git-scm.com/docs/gitignore) format.

For that reason, I'd advise against enforcing any lexical integrity via highlighting, especially since whitespace in a `gitignore(5)` pattern may be intentional (and therefore impossible to disambiguate from a legitimate error in other contexts):

~~~gitignore
pathname with spaces/
~~~

lib/linguist/languages.yml
Nixinova: Does fsck-skiplist have a canonical filename?

Alhadis: Nope, none.

Alhadis (approved):

lildude (approved):

PR #5925, opened 2022-05-31
Add Mule language

mdaloia:
## Description
Adds a new language: `Mule`

Mule applications are written using XML language but unfortunately they don't have a specific file extension and uses `.xml`. But they have at least 2 distinctive characteristics:
- The root node is called `<mule>`
- The default namespace for `<mule>` node is `xmlns="http://www.mulesoft.org/schema/mule/core"`

In order to distinguish this language from other XML file an heuristic was added with a pattern to match those characteristics.

As reference, this was a brief discussion about how to add it: https://github.com/github/linguist/discussions/5922

## Checklist:

      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Axml+mule+%22http%3A%2F%2Fwww.mulesoft.org%2Fschema%2Fmule%2Fcore%22
    - Sample source(s):
      - Not applicable, I created them specially for this contribution.
    - Sample license(s): MIT
  - [ ] I have included a syntax highlighting grammar: Not applicable, reuses the one for XML.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -674,6 +674,10 @@ disambiguations:
     pattern: '^%(end|ctor|hook|group)\b'
   - language: Linker Script
     pattern: 'OUTPUT_ARCH\(|OUTPUT_FORMAT\(|SECTIONS'
+- extensions: ['.xml']
+  rules:
+  - language: Mule
+    pattern:  '<mule([^>]|\s)*xmlns="http:\/\/www\.mulesoft\.org\/schema\/mule\/core"'
 - extensions: ['.yaml']
   rules:
   - language: MiniYAML
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3986,6 +3986,16 @@ Motorola 68K Assembly:
   tm_scope: source.m68k
   ace_mode: assembly_x86
   language_id: 477582706
+Mule:
+  type: programming
+  color: "#00a1df"
+  tm_scope: text.xml
+  ace_mode: xml
+  codemirror_mode: xml
+  codemirror_mime_type: application/xml
+  extensions:
+  - ".xml"
+  language_id: 213085803
 Muse:
   type: prose
   extensions:
--- a/samples/Mule/dataweave.xml
+++ b/samples/Mule/dataweave.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<mule xmlns:http="http://www.mulesoft.org/schema/mule/http" xmlns:ee="http://www.mulesoft.org/schema/mule/ee/core"
+	xmlns="http://www.mulesoft.org/schema/mule/core"
+	xmlns:doc="http://www.mulesoft.org/schema/mule/documentation" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+http://www.mulesoft.org/schema/mule/ee/core http://www.mulesoft.org/schema/mule/ee/core/current/mule-ee.xsd
+http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd">
+	<http:listener-config name="DW_HTTP_Listener_config" doc:name="HTTP Listener config" doc:id="4d265575-e82b-489e-af85-32c28ae51697" >
+		<http:listener-connection host="0.0.0.0" port="8082" />
+	</http:listener-config>
+	<flow name="dataweaveFlow" doc:id="fb9f311f-c951-4459-a094-387c076b2399" >
+		<http:listener doc:name="Listener" doc:id="5dc3e696-2c93-41f0-8952-a9003cee5361" config-ref="DW_HTTP_Listener_config" path="/"/>
+		<ee:transform doc:name="Transform Message" doc:id="b9d228d0-7e8d-46f0-97e4-a9043842fb90" >
+			<ee:message >
+				<ee:set-payload ><![CDATA[%dw 2.0
+output application/json
+---
+classrooms: payload..*teacher groupBy $.subject mapObject ((teacherGroup, subject) -> {
+   class: {
+     name: subject,
+     teachers: { (teacherGroup map {
+       teacher:{
+           name: $.name,
+           lastName: $.lastName
+       }
+     })
+   }
+  }
+})]]></ee:set-payload>
+			</ee:message>
+		</ee:transform>
+	</flow>
+</mule>
--- a/samples/Mule/empty.xml
+++ b/samples/Mule/empty.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<mule xmlns="http://www.mulesoft.org/schema/mule/core" xmlns:doc="http://www.mulesoft.org/schema/mule/documentation"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd"></mule>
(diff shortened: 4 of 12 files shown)


mdaloia:
Hi Team!
I would like to know if there is anything else from my side that I need to do in order to get it reviewed and, eventually, merged and released. How often you usually do that?

Please let me know!

Thanks in advance!

lildude (approved):

vsvipul:
@mdaloia This addition returns Mule for all repos with xml files. For example, for this repo https://github.com/ashwinsangem/spring4-mvc-gradle-xml-hello-world/tree/master/src/main/resources Mule is being returned. 

Is this expected or something wrong. Asking because we have a failing test in scout. https://github.com/github/scout/commit/66ae072026f77cc3152974fbd93989d9d9a8f9ac/checks

lildude:
Oh sheeet 🤦. I completely missed that Mule was a framework which means it doesn't qualify for inclusion in Linguist. I also missed that the heuristic didn't have an alternative 🤦🤦

I'm going to revert this PR as I need to make the release today.

lildude:
Done. Thanks for bringing this up @vsvipul 🙇

mdaloia:
@vsvipul / @lildude Sorry to hear that, clearly wasn't the intention

I thought that the heuristic declaration will differentiate it from other XML files as its regex looks for a `<mule>` node containing a specific attribute `xmlns="http://www.mulesoft.org/schema/mule/core"`.
I asked on [this discussion](https://github.com/github/linguist/discussions/5922) before making the PR.
Also there wasn't any test that failed, neither locally nor on CI.

> I completely missed that Mule was a framework which means it doesn't qualify for inclusion in Linguist.

Mule is not a framework. You construct applications using a XML and those applications are then run in the Mule Runtime.

From: https://docs.mulesoft.com/mule-runtime/4.4/
> Mule runtime engine (Mule) is a lightweight integration engine that runs Mule applications and supports domains and policies. Mule applications, domains, and policies share an XML DSL (domain-specific language).

> Mule applications connect systems, services, APIs, and devices using API-led connectivity instead of point-to-point integrations. Mule applications provide functionality for message routing, data mapping, orchestration, reliability, security, and scalability.

It even has an SDK: https://docs.mulesoft.com/mule-sdk/1.1/

Also, how is Mule different from [XProc](https://github.com/github/linguist/blob/d7799da826e01acdb8f84694d33116dccaabe9c2/lib/linguist/languages.yml#L7239) which is also XML based? ([Web](https://xproc.org)  /  [Article](https://www.xml.com/articles/2019/11/05/introduction-xproc-30/))
The only thing that I see is that it has a specific file extension.

Do you think that there is a way to refactor the contribution to make it work in the right way?

Thanks in advance

mdaloia:
> I also missed that the heuristic didn't have an alternative

@lildude Could you elaborate more on this? How is it specified?

lildude:
> I thought that the heuristic declaration will differentiate it from other XML files as its regex looks for a `<mule>` node containing a specific attribute `xmlns="http://www.mulesoft.org/schema/mule/core"`.

So did I and my initial testing suggested it shouldn't cause any problems.

> Mule is not a framework.

[Wikipedia](https://en.wikipedia.org/wiki/Mule_(software)) suggests it is.

> Also, how is Mule different from [XProc](https://github.com/github/linguist/blob/d7799da826e01acdb8f84694d33116dccaabe9c2/lib/linguist/languages.yml#L7239) which is also XML based?

From a [quick look](https://xproc.org/)…. the first sentence states it's an XML-based language, and it has a specific extension. That said, this isn't really a good comparison as XProc was added over 9 years ago when GitHub was first getting support for distinguishing languages so no thought or consideration was given to these sort of things, including the differentiation between data, prose, markup and programming languages that Linguist now uses to distinguish different languages.

> I also missed that the heuristic didn't have an alternative

[@lildude](https://github.com/lildude) Could you elaborate more on this? How is it specified?

Yes. The heuristic you added only has a single "case":

```yaml
- extensions: ['.xml']
  rules:
  - language: Mule
    pattern: '<mule\s[^>]*xmlns="http:\/\/www\.mulesoft\.org\/schema\/mule\/core"'
```

Heuristics are evaluated _after_ the XML strategy…

https://github.com/github/linguist/blob/d7799da826e01acdb8f84694d33116dccaabe9c2/lib/linguist.rb#L62-L71

… which means all other `.xml` files which weren't caught by the earlier XML strategy or this heuristic would fall through to the classifier which isn't necessary. Adding a final `-language: XML` would prevent this:

```yaml
- extensions: ['.xml']
  rules:
  - language: Mule
    pattern: '<mule\s[^>]*xmlns="http:\/\/www\.mulesoft\.org\/schema\/mule\/core"'
  - language: XML
```

I didn't get the chance to look closer yesterday, but now I do, I can see this is precisely why this causes a problem for other `.xml`. The number of Mule `.xml` sample files far outweighs the number of other `.xml` sample files and as such your files have a much heavier influence on training the classifier than the regular `.xml` samples.

We can see this when analysing the file @vsvipul highlighted:

```shell
$ LINGUIST_DEBUG=1 bundle exec bin/github-linguist tmp/spring4-mvc-gradle-xml-hello-world/src/main/resources/logback.xml --breakdown
tmp/spring4-mvc-gradle-xml-hello-world/src/main/resources/logback.xml: 26 lines (19 sloc)
  type:      Text
  mime type: application/xml
      Mule =   -736.148 +  -5.703 =   -741.851
       XML =   -775.102 +  -3.652 =   -778.755
  language:  Mule

$
```

The fact you see that breakdown indicates we've entered the classifier. Of course, we could add more legit `.xml` samples, but we shouldn't need to.

There's one other reason this has come to light: with this PR Mule is the _only_ `.xml` language marked as a "programming" language. Normal XML is considered "data" and thus doesn't appear in the language statistics by default.

mdaloia:
@lildude

First of all, thank you very much for all the comments and explanations! Really helpful and appreciatted!

I locally added the alternative case in the heuristics and updated the test as you mentioned and worked like a charm. In order to be sure I did the following steps (kind of a regression test):
- Ran the failed example reported here (which failed as shown)
- Ran the tests (ok, as it lacked of the non-Mule XML file tests)
- Updated the heuristics test and ran the tests (failed, as expected as it didn't had the alternative case)
- Added the alternative case and ran the tests (Ok)

Regarding your comments/concerns,

> Mule is the only .xml language marked as a "programming" language. Normal XML is considered "data" and thus doesn't appear in the language statistics by default.

If the focus is strictly on the syntax then yes, Mule is XML using particular namespaces to develop applications.
I agree that XML per se is "data", but a XML that represents an application logic I think that fits as "programming".

Mule DSL is the Mule Language that is represented in the XML format, which lets you build you application logic.

If in the future Mule supports a custom syntax to define application logic, would it then be a candidate for "programming" categorization?


> I don't think it should be considered a programming language.

I think that the line between what is and what is not a programming language is somewhat blurred.

Would it help to clarify what are the "requisites" for a language to be defined as "programming"?
 

> Mule is the engine; XML is the language specifically used to configure/control it (the Mule "eco-system" is the specific domain). From what I can see, these XML files are not hand written and are used as data to describe various services and applications and the connections between them, as such are more befitting of "data" than "programming".

This isn't accurate. It isn't to describe and control the Runtime. Via the XML you can define your application logic using among other things control structures like an if (choice), for loops, data transformations, error handlers, batch processing, and more (https://docs.mulesoft.com/mule-runtime/4.4/about-components).

You can for example implement the endpoints of an OpenAPI definition and expose a HTTP/S listener to serve requests.

For more info you could see: https://docs.mulesoft.com/mule-runtime/4.4/mule-app-dev


> these XML files are not hand written

There is a special IDE (Anypoint Studio, based on Eclipse) to help to assist developing Mule applications in a visual way, but it is technically possible (and sometimes maybe needed) to modify the XML by hand.


I hope that this information provides some clarification to determine which will be its categorization.
If something it is still unclear please let me know and I will happily try to clarify it.

Thanks in advance!

Alhadis:
> I agree that XML per se is "data", but a XML that represents an application logic I think that fits as "programming".

We can't classify projects based on *how* they're using languages, only *what* languages they're using. I agree that the distinction between "data" and "programming" is nebulous at best, particularly with regards to low-level formats generated by another program (e.g., [Ninja](http://ninja.build/)). The best we can do is classify languages based on their most common real-world applications. For XML, that means general-purpose data serialisation.

> Any additional thoughts @Alhadis?

IMHO, Mule should be added as an XML extension, but nothing more specific than that. Going forward, all "XML-based" formats should be added as XML (unless a dedicated grammar is used for highlighting, in which case a separate language entry is justified).

PR #5931, opened 2022-06-08
Add .bb file extension for Babashka to Clojure

not-in-stock:
## Description

* Added new extension to Clojure group in `lib/linguist/languages.yml`
* Added a sample of [Babashka](https://github.com/babashka/babashka) code in `samples/Clojure/tell-number.bb`
* Added heuristic to make linguist distinguish between Babashka, BlitzBasic, and BitBake source files.

## Checklist:

      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Abb+defn+NOT+nothack
    - Sample source(s):
      - [/pathom3-graphql/script/validate-and-format.bb](https://github.com/wilkerlucio/pathom3-graphql/blob/c52a425fa861cc2ad477631b78bae25efc4a6333/script/validate-and-format.bb)
     - Sample license(s):
     - [/pathom3-graphql/LICENSE](https://github.com/wilkerlucio/pathom3-graphql/blob/c52a425fa861cc2ad477631b78bae25efc4a6333/LICENSE) 


lildude (changes requested):
> Sample source(s):
> 
>   * [tell-number.bb](samples/Clojure/tell-number.bb)

You need to link to the _actual_ source of the _real-world_ sample file you have added, not the file within the Linguist PR 😁. We also need to know the license of the source file. Hint: that's what that section of the template that you deleted was for 😬.

The sample is also quite small which means it's not going to perform very well in circumstances where things slip past the heuristic. Please add at least one more sample, ideally a larger one (but not too large) that is more representative of the language.

lib/linguist/languages.yml
```diff
@@ -1036,6 +1036,7 @@ Clojure:
   - ".cljscm"
   - ".cljx"
   - ".hic"
+  - ".bb"
```
lildude, line 1039: You're going to need to move this up to above `.boot` as extensions should be alphabetical after the first extension which is considered the primary extension.

samples/Clojure/validate-and-format.bb
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bb
```
lildude, line 1: The other test failure is because of this line. There are two solutions:

1. Replace the sample with one without a shebang (I see this doesn't appear to be a requirement)
2. Add the following to the Clojure entry in `languages.yml`:

```yaml
interpreters:
- bb
```

lildude (changes requested):
A few tips to get the tests passing

samples/Clojure/validate-and-format.bb
not-in-stock: Thanks for the help!

lildude (approved):

Issue #5756, opened 2022-02-02
FYI: Temporary change in language and extension popularity assessment


lildude:
GitHub's Search is struggling at the moment so all Search requests are being heavily restricted making it almost impossible to count the number of unique `:user/:repo` combinations via the likes of Harvester or the API.

Search is in the process of being rewritten with the Tech Preview available at https://cs.github.com/ (please tinker with it and send GitHub feedback) however it isn't accessible via the API yet and doesn't quite yet meet our needs to determine our current usage requirements so for the foreseeable future I'll be using my judgment to determine popularity until the new Search gains the functionality we need and/or the restrictions are lifted (or we can come up with other qualifying criteria).

I know this is subjective and open to debate so the loose rules I'll be using are along the lines of:

- at least 2000 files per extension indexed in the last year (the number you see at the top of the search results)
- with a reasonable distribution across unique `:user/:repo` combinations assessed by manually and randomly clicking through the results.

If particular users are showing a high proportion of the results, I'll manually filter out those users using `-user:<username>` to reduce their impact on my assessment.

I know this isn't ideal, but I think it's the best option for the moment. I'm open to suggestions too. On the plus side, it does mean a lot more PRs are likely to be merged 😁.

I'll be going back through older PRs in the next week or two and will re-assess based on these notes and merging any that satisfy them.

Alhadis:
> Search is in the process of being rewritten

Might be a good time to request a "search by extension/filename" feature to simplify the task of adding new languages to GitHub... 😉

lildude:
> Might be a good time to request a "search by extension/filename" feature to simplify the task of adding new languages to GitHub... 😉

I think we're covered already by a combination of scopes and more intuitive file path expressions:

![CleanShot 2022-02-03 at 09 26 58](https://user-images.githubusercontent.com/627280/152315729-037b239a-c155-48d0-b032-f0970c378c99.png)

Alhadis:
Wow, regular expressions will be supported? _Now_ we're talking. 😀

Also, I tried to access https://cs.github.com/ but it simply redirected me to my activity feed (i.e., https://github.com/). Is it staff-only or something?

lildude:
Nope. You need to be invited. Join the waitlist at https://cs.github.com/about

Alhadis:
Done. Hopefully this'll make Harvester's rewrite less intimidating. 😅

elimisteve:
[...]

PR #5940, opened 2022-06-13
Add support for Elvish

xiaq:
Usage in the wild:
https://github.com/search?q=extension%3Aelv&type=Code&ref=advsearch&l=&l=. There
are a couple of false positives, but the vast majority of the results are Elvish
scripts.

Code sample is taken from Elvish's standard library:
https://github.com/elves/elvish/blob/master/pkg/mods/epm/epm.elv, license BSD
2-clause.

## Checklist:

      -  https://github.com/search?q=extension%3Aelv&type=Code&ref=advsearch&l=&l=
    - Sample source(s):
      - https://github.com/elves/elvish/blob/master/pkg/mods/epm/epm.elv
  - [x] I have included a syntax highlighting grammar: https://github.com/elves/elvish


xiaq:
@lildude Hi! Thanks for taking a look, I see you added the "Pending Popularity" which I assume means that the new language may not meet the popularity threshold.

Since [the documented criterion](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language) is at least 200 unique repos, I wrote a [program](https://github.com/xiaq/elvish-popularity) to test whether Elvish clears that criterion. The result is [224 repos](https://github.com/xiaq/elvish-popularity/blob/main/result.md#repos) as of today.

There are some false positives - not all `.elv` files are Elvish scripts. Clicking these repos randomly, roughly 1 or 2 out of every 10 repos are false positives. So the actual number of repos may be just a bit lower than 200.

Do you think this is close enough to the threshold, or should I try again the number of unique repos is clearly larger than 200?

lildude:
We're temporarily using the requirements detailed in https://github.com/github/linguist/issues/5756 because GitHub's search is struggling with extension searches which often results in inaccurate, zero or duplicate results.

lildude (approved):

PR #5945, opened 2022-06-20
Add RouterOS Script

zachstence:
## Description

Currently, RouterOS Script is identified as the Rascal programming language, because they share the extension `.rsc`.

The languages look pretty dissimilar, so the samples I provided should be enough to distinguish them without a heuristic.

## Checklist:

      -  https://github.com/search?q=extension%3Arsc+interface+NOT+nothack&type=Code
    - Sample source(s):
      - `BackupAndUpdate.rsc` -> Taken from a [GitHub repo](https://github.com/beeyev/Mikrotik-RouterOS-automatic-backup-and-update) with MIT License
      - `config.rsc` -> Written by me for this contribution
      - `console-dump.rsc` -> Default configuration dump from my Mikrotik router
      - `vlans.rsc` -> Taken from the [Mikrotik Wiki](https://wiki.mikrotik.com/wiki/Manual:Basic_VLAN_switching#Other_devices_with_built-in_switch_chip)
      @lildude has also added two more Rascal samples:
      - https://github.com/usethesource/rascal-core/blob/32f837e07b844ff403f613e3b4d5be1c5c27e72e/src/org/rascalmpl/core/library/lang/rascalcore/compile/Rascal2muRascal/RascalModule.rsc
      - https://github.com/usethesource/rascal-core/blob/32f837e07b844ff403f613e3b4d5be1c5c27e72e/src/org/rascalmpl/core/library/lang/rascalcore/compile/Rascal2muRascal/ModuleInfo.rsc
    - Sample license(s):
      - RouterOS samples: MIT. Rascal samples: BSD-2
  - [x] I have included a syntax highlighting grammar: [URL to grammar repo](https://github.com/devMikeUA/vscode_mikrotik_routeros_script), MIT License

@lildude is removing `samples/Rascal/Rascal.rsc` in this PR as it is not a sample written in Rascal.


zachstence:
> The languages look pretty dissimilar, so the samples I provided should be enough to distinguish them without a heuristic.

Please let me know if a heuristic is needed!

lildude:
> Please let me know if a heuristic is needed!

Yes, it looks like we're going to need one as the classifier test is failing because it's detecting one of the Rascal files as RouterOS:

[592](https://github.com/github/linguist/runs/7074664251?check_suite_focus=true#step:5:593) Rascal/Rascal.rsc BAD (RouterOS Script)

zachstence:
> Yes, it looks like we're going to need one as the classifier test is failing because it's detecting one of the Rascal files as RouterOS:

Added a heuristic here https://github.com/github/linguist/pull/5945/commits/3150ab6e720a3bfa482a991298f048428d6a029f

zachstence:
@lildude I'm a little confused where to go from here... I tried improving the heuristic, but the classifier kept misclassifying `Rascal.rsc`. That prompted me to dig into the Ruby code (I do not know Ruby 😅), and I learned that the classifier doesn't use the heuristics, because they are separate strategies.

So that failure in the action can't be improved by changing the heuristic. Should I instead add more samples?

lildude:
> I learned that the classifier doesn't use the heuristics, because they are separate strategies.

🤔 It does... sort of. It ignores them here:

https://github.com/github/linguist/blob/f3da780afabd10e392901c0940b2a7af1d4b314e/script/cross-validation#L51-L61

But now I look closer, I see it'll only ignore these if there's a final conclusive language, which you've not added for `.rsc`, which isn't a problem.

I've just checked your heuristics and the Rascal one is waaay too liberal. It matches your `samples/RouterOS Script/BackupAndUpdate.rsc` RouterOS sample 🤭 hence the failing tests.

Now I look at it more closely, `samples/Rascal/Rascal.rsc` isn't even written in Rascal; it's a syntax definition of the language so includes tokens not even used in the language, hence the misclassification. I'm going to remove that and push a few more real Rascal samples to this PR and remove the heuristics.

lildude (approved):
💐 👍 Thanks.

zachstence:
Thanks for the help!

Issue #5938, opened 2022-06-10
tokenizer.rb:17: [BUG] Segmentation fault


lildude:
Moving over from discussions as this is definitely a bug.

### Originally discussed in https://github.com/github/linguist/discussions/5876

<div type='discussions-op-text'>

<sup>Originally posted by **ceshiyixiaba** April 25, 2022</sup>
git repository: https://github.com/ceshiyixiaba/languages

[docker enviroment](https://github.com/github/linguist#docker)，error message:

```text
docker run --user=1000 --rm -v $(pwd):$(pwd) -w $(pwd) -t linguist | more
/usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist/tokenizer.rb:17: [BUG] Segmentation fault at 0x0000004700000057
ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-linux-musl]

-- Control frame information -----------------------------------------------
c:0026 p:---- s:0154 e:000153 CFUNC  :extract_tokens
c:0025 p:0007 s:0149 e:000148 METHOD /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist/tokenizer.rb:17
c:0024 p:0043 s:0144 e:000143 METHOD /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist/classifier.rb:106
c:0023 p:0031 s:0136 e:000135 METHOD /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist/classifier.rb:84
c:0022 p:0043 s:0129 e:000128 METHOD /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist/classifier.rb:22
c:0021 p:0011 s:0122 e:000121 BLOCK  /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist.rb:32
c:0020 p:0026 s:0119 e:000118 METHOD /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist.rb:100
c:0019 p:0026 s:0113 e:000112 BLOCK  /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist.rb:31 [FINISH]
c:0018 p:---- s:0108 e:000107 CFUNC  :each
c:0017 p:0020 s:0104 e:000103 BLOCK  /usr/local/bundle/gems/github-linguist-7.20.0/lib/linguist.rb:29
[...]

PR #5060, opened 2020-10-22
Improve punctuation tokenization and word split

smola:
## Description

The tokenizer previously ignored most (but not all) punctuation characters. Now all ASCII printable characters are included in tokens.

These characters are often very salient features when detecting the programming language. In some cases, even more than identifiers.

A sequence of consecutive non-alphanumeric characters will be recognized as a single token (e.g. == != >= <==>) with the following exceptions:

- Comment tokenization behavior is preserved, so `/*` or `<!--` are still ignored as part of comments. As a side-effect, some edge cases are now fixed (e.g. `<!---->`, `<!--Comment-->`).

- SGML tokens are mow more restrictive, so `<a>`, `<?xml>`, `<!DOCTYPE>` are still SGML recognized, but `<!--` or `<--` are not. This also simplifies the corresponding lexer action.

- Parenthesis (`()`), square brackets (`[]`) and curly brackets (`{}`) are tokenized independently.

Word splitting is also altered by this change:

- `[.@#/*]` are not considered part of an identifier anymore. But a single character from [.@#] is recognized as part of a word if it's the first one. So `.globl` in Assembly is still a single token, but `console.log` is now split as `console` and `.log`. In Objective-C, `@interface` will still be a single token, but `mail@example.com` will be 3 tokens. This change leads to less unique tokens in the vocabulary, which helps classification too.

## Checklist:

_Checklist does not apply._

PR #5956, opened 2022-06-30
Fix tokenizer crash when passed very long strings

lildude:
https://github.com/github/linguist/pull/5060 introduced changed to improve the tokenizer to better handle punctuation. What we didn't realise at the time is it introduced a fixed length buffer with these rules as they use trailing context for matching:

```
<INITIAL,punct>[-!#$%&*+,.:;=>?@\\^_`|~]+/{not_punct}   { FEED(); BEGIN(INITIAL); return 1; }
<INITIAL,punct>[-!#$%&*+,.:;=>?@\\^_`|~]+/("<!--"|"/*") { FEED(); BEGIN(INITIAL); return 1; }
<INITIAL,punct>[</]+/{not_punct}          { FEED(); BEGIN(INITIAL); return 1; }
<INITIAL,punct>[</]+/"<!--"               { FEED(); BEGIN(INITIAL); return 1; }
<INITIAL,punct>[</]+/"/*"                 { FEED(); BEGIN(INITIAL); return 1; }
```

The trailing context is the part of each rule that follows after the `/` in the `+/` on the left of each row, ie `"/*"` is the trailing context for `<INITIAL,punct>[</]+/"/*"`.

When a file with a single very long line is fed into the tokenizer, we overrun this buffer and crash with a segmentation fault.

I attempted to find alternate rules to keep this functionality, but I'm really not skilled in Lex enough to do so so I'm reverting this part of #5060. The side effect of this is the classifier may be perform slightly worse, but this is a small price to pay for stability.

Whilst researching this issue, I discovered we hit this problem before and fixed it then so I've added a very crude test to catch changes in the tokenizer that reintroduce the fixed length REJECT buffer.

Fixes https://github.com/github/linguist/issues/5938


Alhadis (approved):

lildude:
Closing in favour of https://github.com/github/linguist/pull/5969

PR #5961, opened 2022-07-04
(Github Actions) Bump versions of actions/checkout in ci workflow and scan for GitHub Actions with dependabot

ArjunSharda:
## Description

This pull request will bump the versions of actions/checkout in the ci workflow and for dependabot adds a package ecosystem for scanning for github actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         ruby: [ '3.0', '2.7', '2.6', '2.5' ]
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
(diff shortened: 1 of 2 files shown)


lildude:
Rejecting this as it has already been tried before and failed so had to be reverted. See https://github.com/github/linguist/pull/5915 

There is also no good reason to change this.

.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     name: Classifier cross-validation
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
```
Zerocoolooog, line 52: https://github.com/github-linguist/linguist/blob/master/.github%2Fworkflows%2Fci.yml

Zerocoolooog (approved):

.github/workflows/ci.yml
Alhadis: > https://github.com/github-linguist/linguist/blob/master/.github%2Fworkflows%2Fci.yml

Please stop doing that.

Zerocoolooog: > https://github.com/github-linguist/linguist/# blob/- [ ] @[]()WDR16BPBridgingtheDisabilityDividethroughDigitalTechnologyRAJA.pdf

- - @ master/.github%2Fworkflows%2Fci.yml

PR #5989, opened 2022-07-22
Add support for HOCON

LaurenceWarne:
## Description

Hi, this PR adds support for [hocon](https://github.com/lightbend/config/blob/main/HOCON.md) files with the `.hocon` file extension.  Also prevalently used with hocon files is the `.conf` extension, though adding this would bring in a lot of false positives, so I'd like to maybe do a follow up PR for that if that's ok :sweat_smile: .  Also let me know if the sample files are too large, thanks.

## Checklist:

      -  https://github.com/search?q=extension%3Ahocon&type=Code
    - Sample source(s):
      - https://github.com/hmrc/penalties-frontend/blob/main/conf/application.conf
      - https://github.com/kazysgurskas/snowplow-realtime-docker/blob/master/config/stream-collector.hocon
    - Sample license(s):
      - https://github.com/hmrc/penalties-frontend/blob/main/LICENSE
      - https://github.com/kazysgurskas/snowplow-realtime-docker/blob/master/LICENSE
  - [x] I have included a syntax highlighting grammar: https://github.com/jacobwgillespie/language-hocon


lildude:
> Also let me know if the sample files are too large, thanks

Oh yes, they most definitely are (if files are not loaded by the UI in the diff you know it's almost certainly too big 😉), and they're mostly filled with comments which are completely ignored so are adding unnecessary bloat.

lildude:
> Also prevalently used with hocon files is the `.conf` extension, though adding this would bring in a lot of false positives, so I'd like to maybe do a follow up PR for that if that's ok 😅 .

Sure, but there be dragons. This is going to be fraught with troubles as this is an incredibly popular extension and getting a performant heuristic for what is effectively a variant of JSON is going to be a nightmare. It's better to leave it for now and peeps can use an override.

> I think a sample like https://github.com/emqx/hocon/blob/master/sample-configs/test01.conf would be a better overall showcase of the syntax, though I think I saw somewhere that you prefer real world examples?

Yup. We want real world usage, not illustrations of usage. The samples here should be go on for now, especially as this is a unique extension.

lildude (approved):

LaurenceWarne:
Cheers!
> This is going to be fraught with troubles as this is an incredibly popular extension and getting a performant heuristic for what is effectively a variant of JSON is going to be a nightmare. It's better to leave it for now and peeps can use an override.

In that case I'll take your advice, I see `conf` is an unfortunate choice for a file extension!

PR #5994, opened 2022-07-26
Option to read file contents from stdin

mschuldt:
## Description

When the `--stdin` option is set, ARGV is cleared and a new Blob is created using the file contents from ARGF.read.
This is useful when calling Linguist from another program when the file contents do not exist on disk.

## Checklist:


lildude (changes requested):
I can see how this could be useful. Can you please add details of this flag to the usage section of the README.md too.

bin/github-linguist
```diff
@@ -32,19 +33,20 @@ def github_linguist(args)
       puts opts
       exit
     end
+    opts.on("-s", "--stdin", "Read file contents from stdin") { read_from_stdin = true }
```
Alhadis, line 36: Conventionally, programs which accept a filename as an argument are made to read from STDIN by passing `-` as a parameter. Linguist should support this idiom as well.

Alhadis (changes requested):

Alhadis:
Hold up. What exactly is Linguist expected to do with the input when it has no way of knowing its origin?

~~~console
λ GitHub-Linguist (read-stdin): bundle exec bin/github-linguist -s < .gitmodules
/Users/Alhadis/Forks/GitHub-Linguist: 1274 lines (1274 sloc)
  type:      Text
  mime type: text/plain
  language:  
λ GitHub-Linguist (read-stdin): bundle exec bin/github-linguist --breakdown -s < .gitmodules
/Users/Alhadis/Forks/GitHub-Linguist: 1274 lines (1274 sloc)
  type:      Text
  mime type: text/plain
  language:  
~~~

bin/github-linguist
mschuldt: I don't see how Linguist can adopt that idiom without making it even more confusing as Linguist still needs the filename for language detection. If we replace the filename with `-` then we'd need to create an additional parameter such as `--filename`

mschuldt:
As I understand it, Linguist doesn't need to know the origin, just the filename and contents. The filename is missing from those examples:
```
 bundle exec bin/github-linguist --breakdown -s .gitmodules < .gitmodules
.gitmodules: 1274 lines (1274 sloc)
  type:      Text
  mime type: text/plain
  language:  Git Config
  appears to be a vendored file
```

```
bundle exec bin/github-linguist -s nonexistentfile.yml < grammars.yml 
nonexistentfile.yml: 1155 lines (1155 sloc)
  type:      Text
  mime type: text/x-yaml
  language:  YAML
```

Alhadis:
@mschuldt I think you have some misconceptions about standard input and how it works, exactly. It won't always come from a _filename_—a lot of the time, it's piped through from a command, which may or may not have consulted the filesystem in order to generate its output. If it's **required** that a filename be present, then you should just… specify a filename. The feature you're proposing goes against user expectations and adds complexity to Linguist that's difficult to explain or justify.

mschuldt:
yeah, I'm expecting this feature to only be useful in those cases where stdin is not redirected from a file, otherwise I see no point. 
I think that any program that can invoke Linguist with file contents will also have access to the file name and not have a problem adding it to the cli command. I've only been thinking about this feature in contexts where there is no access to a filesystem, just access to internal representations of files. I don't imagine this being useful as part of a bash pipeline. 

In any case I don't care enough about this to spend much time modifying it. If it's confusing and and folks don't care then I'm fine just dropping it, it worked ok for me and it's easy enough for others to implement themselves if needed.

Alhadis:
> I've only been thinking about this feature in contexts where there is no access to a filesystem, just access to internal representations of files.

But in order to redirect from a file, you _still_ need access to the filesystem. So this still doesn't make any sense.

I'm still confused as to why you'd redirect from a file instead of passing it as an argument…  The former takes the same number of keystrokes as the latter, plus one extra for the `<` operator. But as far as the shell's concerned, if it can't access the file being redirected, it won't even get as far as `execve(2)`. The command won't run, period:

~~~console
$ cat /tmp/some-cmd.sh
#!/bin/sh
echo >&2 'Called it!'

$ /tmp/some-cmd.sh < /etc/master.passwd
-bash: /etc/master.passwd: Permission denied
~~~

mschuldt:
I'm not redirecting from a file. That is just a convenient way for everyone to test this functionality.
I am invoking Linguist as a subprocess from Python in a task handler running on a random machine.
The file contents are accessed as a byte array from the arguments and sent to the Linguist process as stdin.

Alhadis:
> I'm not redirecting from a file. That is just a convenient way for everyone to test this functionality.

In that case, I recommend implementing this feature as an entirely separate mode of operation:

~~~console
$ # List all languages that would potentially match 'foo.h'
$ github-linguist --test-path="foo.h"
C
C++
Objective-C

$ # As above, but give the classifier something to chow down on
$ github-linguist --test-path="foo.h" --test-string="@interface"
Objective-C
~~~

Additional options could be added to test interpreter names (for hashbangs) and language aliases (for modelines, `linguist-language` attributes, etc):

~~~console
$ github-linguist --test-interpreter="python3"
Python

$ github-linguist --test-name="cpp"
C++
~~~

bin/github-linguist
Alhadis: @mschuldt See my latest comment.

lildude:
Nudge nudge

mschuldt:
The approach proposed by @Alhadis makes sense to me, but I don't think this is important enough for me to put more time into it

lildude:
Thanks for the update. In this case, I think it's best to close this PR. It can be picked up by someone later if there's demand for it.

PR #6002, opened 2022-08-08
Update Ace mode paths for tests

lildude:
The upstream directory structure was changed in https://github.com/ajaxorg/ace/pull/4851 which means we're looking in the wrong place for all the languages supported by the Ace editor. We now need to look in the old location for `jsoniq` and `xquery` and the new location for everything else.

These changes have been made as they plan to release Ace as an NPM.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,8 +35,9 @@ task :fetch_ace_modes do
   File.delete(ACE_FIXTURE_PATH) if File.exist?(ACE_FIXTURE_PATH)
 
   begin
-    ace_github_modes = URI.open("https://api.github.com/repos/ajaxorg/ace/contents/src/mode").read
-    File.write(ACE_FIXTURE_PATH, ace_github_modes)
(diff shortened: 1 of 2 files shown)


Alhadis:
There doesn't appear to be any rationale behind ajaxorg/ace#4851. If upstream makes this sort of change willy-nilly, then it might be wiser to query the [latest tagged release](https://github.com/ajaxorg/ace/releases/latest), as opposed to HEAD.

lildude:
> There doesn't appear to be any rationale behind [ajaxorg/ace#4851](https://github.com/ajaxorg/ace/pull/4851).

There is now as [I asked for one](https://github.com/ajaxorg/ace/pull/4851#issuecomment-1209594416) 😁.

Based on this, I think we're good to stay with using HEAD. This has been the first major change like this since we implemented the test so I think we should be good querying both locations for the future.

Alhadis (approved):

PR #6018, opened 2022-08-15
Add `sol` alias to Solidity

minaminao:
## Description

Solidity source codes have `.sol` extension, and as with other languages (e.g., `rs` in Rust or `cpp` in C++), adding `sol` alias is more convenient when writing markdown.

## Checklist:

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):

    - Sample source(s):
    - Sample license(s):

  - Old: [URL to grammar repo]
  - New: [URL to grammar repo]

    - [URL to public discussion]

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6191,6 +6191,8 @@ Solidity:
   tm_scope: source.solidity
   extensions:
   - ".sol"
+  aliases:
+  - sol
   language_id: 237469032
 Soong:
   type: data


Alhadis:
> adding `sol` alias is more convenient when writing markdown.

Convenient, yes. Unambiguous, no. Unlike `rs` or `cpp`, the abbreviation `sol` has widespread use in contexts unrelated to Solidity; this means there's an elevated risk of Linguist supporting another language in future whose name is commonly abbreviated to "Sol". In fact, as _"Sol"_ is the Latin word for "sun", it wouldn't surprise me if a developer chooses this name for their new programming language (hey, it might be a pun on Lua…).

In other words, it's better if we remain conservative in what we associate with generic-sounding names.

minaminao:
Indeed. I had read #4904 and understood that it is important to be unambiguous, but I did not think we should be conservative in adding an alias that is the same as the extension. One technical reason to avoid adding is that it is difficult to apply heuristics to code blocks, right? You may close this PR.

Alhadis:
> One technical reason to avoid adding is that it is difficult to apply heuristics to code blocks, right?

Actually, heuristics _don't_ apply to code blocks: in fact, no classification is performed on highlighted code-blocks whatsoever. That's why it's important we avoid naming conflicts here, for the concerns I described earlier.

PR #6039, opened 2022-08-26
Update /lib/linguist/languages.yml

VitalikLevin:
### Added
- `package.json` in "JSON" section


lildude:
This is not needed as the file will be classified as JSON thanks to the `.json` extension. Closing.

PR #6048, opened 2022-09-01
Add .tag to Java Server Pages

lildude:
## Description

This PR takes advantage of the generic extension functionality and adds `.tag` to Java Server Pages with a heuristic to correctly detect it.

I've gone with matching any of the possible directives that will appear in the JSP tag files as detailed at https://docs.oracle.com/javaee/1.4/tutorial/doc/JSPTags5.html

Fixes https://github.com/github/linguist/issues/3094, the oldest open issue in this repo right now.

## Checklist:

      - https://github.com/search?q=extension%3Atag+tag&type=Code
    - Sample source(s):
      - https://github.com/bzennn/WavyArch/blob/f6a3df9b2e714cc9948e473e20c3da7954653d1f/WavyArch/src/main/webapp/WEB-INF/tags/skeleton_page.tag
    - Sample license(s): MIT


lib/linguist/heuristics.yml
```diff
@@ -621,6 +621,10 @@ disambiguations:
     pattern: '^\s*(?:use\s+v6\b|\bmodule\b|\bmy\s+class\b)'
   - language: Turing
     pattern: '^\s*%[ \t]+|^\s*var\s+\w+(\s*:\s*\w+)?\s*:=\s*\w+'
+- extensions: ['.tag']
+  rules:
+  - language: Java Server Pages
+    pattern: '\s*<%@.?(taglib|tag|include|attribute|variable)\s'
```
Alhadis, line 627:
```suggestion
    pattern: '<%[@!=\s]?\s*(taglib|tag|include|attribute|variable)\s'
```

1. `\s*` here is redundant, since the rest of the heuristic will match with or without leading whitespace.

2. Based on my reading of JSP's [syntax reference](https://www.oracle.com/technetwork/java/syntaxref12-149806.pdf), it appears as if `<@= … =>`, `<@! … =>` and `<@ … >` are also supported.

3. `.?` will match virtually anything, including `<%@etag` or `<%@>include`.

Alhadis (changes requested):
Some discrepancies picked up from a cursory scan of Wikipedia[^1] and one of its cited sources[^2].

Also, it appears as if the language is known as _"**Jakarta** Server Pages"_ now…?

[^1]: https://en.wikipedia.org/wiki/Jakarta_Server_Pages#cite_note-syntax-5
[^2]: https://www.oracle.com/technetwork/java/syntaxref12-149806.pdf

Alhadis (approved):

PR #6067, opened 2022-09-14
add cairo v0.10.0 support

0xHyoga:
Replaced grammar for Cairo with the new v0.10.0 syntax. This new version of Cairo was released in September 2022 and introduced a few breaking features. 
 
## Description

The grammar file is the same as the old file except it updates 3 main things:
1. Comments are no longer `#` and replaced it with `//`
2. Blocks start wtih `{` instead of `:` and end with `}` instead of `end`
3. Add `l1_handler` decoration

I migrated sample files to match new syntax.

## Checklist:
  - Old: https://github.com/0xChqrles/atom-language-cairo
  - New: https://github.com/0xHyoga/atom-language-cairov0.1

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -275,9 +275,9 @@
 [submodule "vendor/grammars/atom-language-1c-bsl"]
 	path = vendor/grammars/atom-language-1c-bsl
 	url = https://github.com/xDrivenDevelopment/atom-language-1c-bsl.git
-[submodule "vendor/grammars/atom-language-cairo"]
-	path = vendor/grammars/atom-language-cairo
-	url = https://github.com/xshitaka/atom-language-cairo.git
+[submodule "vendor/grammars/atom-language-cairov0.1"]
+	path = vendor/grammars/atom-language-cairov0.1
+	url = https://github.com/0xHyoga/atom-language-cairov0.1.git
 [submodule "vendor/grammars/atom-language-clean"]
 	path = vendor/grammars/atom-language-clean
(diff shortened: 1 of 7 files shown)


0xHyoga:
> Wouldn't it be better to submit a PR to [`0xChqrles/atom-language-cairo`](https://github.com/0xChqrles/atom-language-cairo) that adds support for _both_ old and new features?

The old features are pretty much obsolete at this point, but yes maybe it is best actually. What would be the best way to preserve the old but add the new features in this case? Just merge both files in 1 cairo.cson or could we not have 2 repos instead for Cairo?

Alhadis:
> What would be the best way to preserve the old but add the new features in this case

Try this:

~~~diff
--- grammars/cairo.cson	2022-09-15 05:50:06.000000000 +1000
+++ grammars/cairo.cson	2022-09-15 05:53:36.000000000 +1000
@@ -35,9 +35,9 @@
   'comment':
     'patterns': [
       {
         'name': 'comment.line.number-sign'
-        'begin': '//'
+        'begin': '#|//'
         'beginCaptures':
           '0':
             'name': 'punctuation.definition.comment'
         'end': '$'
@@ -116,9 +116,9 @@
           '1':
             'name': 'storage.type.function.cairo'
           '2':
             'name': 'entity.name.function.cairo'
-        'end': '\\{'
+        'end': '(?<=:)|\\{'
         'patterns': [
           {'include': '#comment'}
           {'include': '#function-implicit'}
           {'include': '#function-params'}
~~~

PR #6118, opened 2022-10-14
Add repo "language used support" for INI/CFG.

yoidog:
## Description
Hello, I have been seen by someone called "lildude", Since i did 2 pull requests about my language, This time I'm trying to add something for some projects I'm trying to do, The reviewer of this pr maybe will not understand "language used support" but here's an example:

something

You can notice a brown color and next to it a text "Assembly" under the description of the repo, this indicates which language is used the most OR if it's the only used language.

I want to add the exact same thing but with INI/CFG. It will look like this:

INI

If you think this is some joke or some stupid idea, you're right, But I get insecure when I don't get the "language used" below my repo (probably won't make sense at all). Thanks!

- Waled Ali / yoidog
## Checklist:

- [x] **Adding "language used" support**

### DIFF
--- a/samples/INI/serv.cfg
+++ b/samples/INI/serv.cfg
@@ -1,0 +1,6 @@
+version=0.1
+url=localhost
+port=27043
+ascii=0
+host=1
+allowed_ips=10


Alhadis:
If I understand this correctly, you'd like `.cfg` to be reflected in a project's language stats. Yes?

If so, add a `.gitattributes` to the root of your project with the following [override](https://github.com/github/linguist/blob/master/docs/overrides.md#detectable):

~~~gitattributes
*.cfg linguist-detectable
~~~

This will instruct Linguist to consider `.cfg` files the next time a change is pushed to the repository.

yoidog:
> If I understand this correctly, you'd like `.cfg` to be reflected in a project's language stats. Yes?
> 
> If so, add a `.gitattributes` to the root of your project with the following [override](https://github.com/github/linguist/blob/master/docs/overrides.md#detectable):
> 
> ```gitattributes
> *.cfg linguist-detectable
> ```
> 
> This will instruct Linguist to consider `.cfg` files the next time a change is pushed to the repository.

oop, haha, I'm really stupid, thanks dude.

Issue #5824, opened 2022-03-14
Visual Basic 6.0 or VB6


Gagniuc:
## Language name
Visual Basic 6.0

## URL of example repository
https://github.com/Gagniuc/Markov-Chains-Simulation-framework
https://github.com/Gagniuc/Markov-Chains-Prediction-framework
https://github.com/Gagniuc/Discrete-Probability-Detector-in-VB6
https://github.com/Gagniuc/Micro-chart-in-VB
https://github.com/Gagniuc/Visual-Sequence-Alignment-in-VB6
https://github.com/Gagniuc/Simple-sequence-alignment-in-VB6
https://github.com/Gagniuc/Markov-Chains-The-weather
https://github.com/Gagniuc/Markov-Chains-detector-in-VB6
https://github.com/Gagniuc/Genomin
https://github.com/Gagniuc/Mapping-pixels-to-LEDs-for-the-Photon-pixel-coupling-method
https://github.com/Gagniuc/WebCam-software-sampling
https://github.com/Gagniuc/VB6-add-GUI-objects-at-runtime
https://github.com/Gagniuc/Motif-detection-VB6
https://github.com/Gagniuc/PI-laboratory-in-VB6
https://github.com/Gagniuc/PromKappa-3.0-Objective-Digital-Stains-in-VB6

## URL of syntax highlighting grammar
https://github.com/Gagniuc/Micro-chart-in-VB
However, an extra color for the inbuilt functions will be extraordinary.

## Most popular extensions
Notice that the <b>.vbp</b> and <b>.vbw</b> extensions are VB6 specific.

The extensions commonly used by this language:
   - ".vbp" 
   - ".vbw"
   - ".bas" 
   - ".cls" 
   - ".frm" 
   - ".frx"

## Detected language
VBA

zspitz:
I would say the primary difference on the code level between VBA and VB6 is the different global objects available to them. But that is also true of VBA when running under different host environments -- the global objects available when VBA is running under Microsoft Word are different from those available when running under Microsoft Excel.

Gagniuc:
> I have no experience in either language, but from what I've read this is going to be a challenge to implement if you plan to have all those extensions supported by both languages as I believe they have similar (identical?) syntax, which means how would one differentiate programmatically between a VB6 `.cls` file and a VBA `.cls` file? The same thing would apply to all the other shared extensions.
> 
[...]

PR #6124, opened 2022-10-18
Add Visual Basic 6.0 as its own language

DecimalTurn:
## Description
This pull request adds VB6 as its own language and removes the aliases for VBA that were related to VB6.
This PR aims at solving issue #5824 and takes major inspiration from PR #4725.

For the color, I went with a green close to the green from the [VB6 logo](https://en.wikipedia.org/wiki/Visual_Basic_(classic)#/media/File:Visual_Basic_6.0_logo.png).

The extensions provided aren't shared with any languages in languages.yml, so no additional grammar or heuristic is necessary. The VB.NET grammar should suffice for now as it was also the grammar used when adding VBA.

## Checklist:

      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Actl+End+Sub+NOT+nothack
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3ADsr+End+Sub+NOT+nothack
    - Sample source(s):
      - https://github.com/kellyethridge/SimplyVBUnit/blob/master/Source/Component/UIListBox.ctl
      - https://github.com/jaygadhave/BCA-BCS-MANAGEMENT-SYSTEM/blob/ebaed2ee5141dff5a3573a358d5440eb1dc0cd70/DataReport1.Dsr
      - https://github.com/bmbodj/ECSE437-Project/blob/fd28fdbe07bd1c7a1ea60650fc91ef18e49d1bef/main/migrationanalysis/src/wizard/Wizard.Dsr
    - Sample license(s): 
      - kellyethridge/SimplyVBUnit: MIT
      - jaygadhave/BCA-BCS-MANAGEMENT-SYSTEM: MIT
      - bmbodj/ECSE437-Project: Apache License 2.0


lib/linguist/languages.yml
```diff
@@ -6784,6 +6784,21 @@ V:
   codemirror_mode: go
   codemirror_mime_type: text/x-go
   language_id: 603371597
+VB6:
```
lildude, line 6787: I think we should spell this out as it appears to be commonly referred to in full from a quick search.

```suggestion
Visual Basic 6.0:
```

lib/linguist/languages.yml
```diff
@@ -6784,6 +6784,21 @@ V:
...
+  type: programming
+  color: "#2c6353"
+  extensions:
+  - ".ctl"
+  - ".dsr"
```
lildude, line 6792: This extension isn't popular enough for inclusion.

```suggestion
```

Please also remove the `.dsr` sample.

lildude (changes requested):

lib/linguist/languages.yml
DecimalTurn: I don't have a strong opinion about this personally and I can indeed see that the spelled out version is indeed used often.
I've made the change accordingly. https://github.com/github/linguist/pull/6124/commits/d443d5283bbdb8ca37a14a38ad1203437f3f240f

lib/linguist/languages.yml
DecimalTurn: Actually, the version with the first letter in upper case is more common in the wild (`.Dsr`). I used the lower case version since that's what [Microsoft documentation](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-basic-6/aa241721(v=vs.60)) seems to suggest, but I can make the change.

DecimalTurn:
And regarding the `.Dsr` sample, I was a little bit limited by the fact that a license is needed. The problem with the current sample is that it has no sub procedures defined in it. We only have designer metadata. This is representative of a good portion of `.Dsr` files, but for completeness, I think might be useful to add another sample with a few sub procedure definitions.

lildude (approved):

DecimalTurn:
Thanks @lildude!

Gagniuc:
> 

Does the new modification make an association between ".vbp" and Visual Basic 6.0, or ".vbw" and Visual Basic 6.0?
If it doesn't, can you please make this possible? Because if you look at my own Visual Basic 6.0 projects, there are no ".ctl" or ".dsr" files at all. But all the Visual Basic 6.0 projects that I have seen so far have something in common: files with the extension ".vbp" and ".vbw". More than likely for this reason we do not see any change.

lildude:
> Does the new modification make an association between ".vbp" and Visual Basic 6.0, or ".vbw" and Visual Basic 6.0?

No. You'll need to use a manual override. I see this has also been asked and answered in https://github.com/github/linguist/issues/5824

Gagniuc:
> > Does the new modification make an association between ".vbp" and Visual Basic 6.0, or ".vbw" and Visual Basic 6.0?
> 
> No. You'll need to use a manual override. I see this has also been asked and answered in #5824

Thanks for the replay. In my view it is not ok to use manual override for VB6 projects on GitHub. I personally am curious how many VB6 projects are on GitHub. This language is among the most important languages produced by our civilization. I mean, there are new languages that no one has ever heard of that are listed by the linguist, while VB6 is not. VB6 was the most popular computer language in the world for a long time and today it is used by a large mass of scientists and engineers for prototyping projects. Recognizing the VB6 language by default is a decent thing to do on GitHub. I am certain in the end this will be resolved.

lildude:
> Recognizing the VB6 language by default is a decent thing to do on GitHub. I am certain in the end this will be resolved.

Indeed. As I mentioned in the issue, a heuristic would be needed to do this. Please feel free to submit a PR that implements such a heuristic.

Locking this as the discussion should be taking in place in issues now this PR has been merged.

PR #6127, opened 2022-10-18
Add WORKSPACE.bazel to Starlark filenames

Alphadelta14:
Add WORKSPACE.bazel as another Starklark matching case.

## Description

Both WORKSPACE and WORKSPACE.bazel are valid names for the WORKSPACE file. The latter takes precedent, even though it is an alias because other projects may have a similarly named file.

Alternatively, *.bazel could be added to extensions, but .bzl is the recommended extension.

## Checklist:

    - Sample source(s):
      - [[skia/WORKSPACE.bazel]](https://github.com/google/skia/blob/main/WORKSPACE.bazel)
    - Sample license(s):
      - BSD-3-Clause

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6311,6 +6311,7 @@ Starlark:
   - BUILD.bazel
   - Tiltfile
   - WORKSPACE
+  - WORKSPACE.bazel
   aliases:
   - bazel
   - bzl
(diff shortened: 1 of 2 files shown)


lildude (changes requested):
Going for the named file is fine, however...

>  I have included a new sample for the misclassified language: 

Please do this. We need the sample added to the actual repo, not just referenced.

lildude (approved):

PR #6139, opened 2022-10-26
Add ".plt" to Prolog extensions along with sample

GeoffChurch:
.plt files are used by Gnuplot for saved plots, and by SWI-Prolog (and probably SICSTUS Prolog) for PlUnit tests. Most seem to be misclassified as Gnuplot.

I can add a simple ":-\s?begin_tests" heuristic, but hopefully the classifier will pick up on the vastly different syntax.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5127,6 +5127,7 @@ Prolog:
   color: "#74283c"
   extensions:
   - ".pl"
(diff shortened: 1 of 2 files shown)


GeoffChurch:
Hi @lildude, here's the updated PR: #6141

I wrote the sample just for this PR so I think it can just fall under linguist's license.

PR #6142, opened 2022-10-26
Man pages

Alhadis:
This PR adds man pages for the two executables installed by Linguist: `github-linguist(1)` and `git-linguist(1)`. [They look like this when displayed in the terminal](https://codepen.io/alhadis/full/jOKEwmO).

I still haven't worked out how to include man pages as part of a RubyGem installation. Some cursory googling convinces me that there's no "official" means of including a manual page:

* https://stackoverflow.com/q/25458485
* https://github.com/defunkt/gem-man

Summoning @lildude, because he knows a hell of a lot more about Ruby's ecosystem than I ever will.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ $ github-linguist
 #### Additional options
 
 ##### `--rev REV`
-The `--rev REV` flag will change the git revision being analyzed to any [gitrevisions(1)](https://git-scm.com/docs/gitrevisions#_specifying_revisions) compatible revision you specify.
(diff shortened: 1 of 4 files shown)


lildude:
> I still haven't worked out how to include man pages as part of a RubyGem installation.

Yeah, installing man pages with a gem so they can be easily found by the OS `man` is a PITA. I think going the route of allowing `gem-man` to find and display them is a good start and we can leave it as an exercise for those that really want to ship the gem in a pkg.

PR #6155, opened 2022-11-04
Add .frm and .cls extensions to VB6

DecimalTurn:
Following up on discussions in issue https://github.com/github/linguist/issues/5824, I'm proposing to add .frm and .cls as extensions for VB6 with heuristics to distinguish it from VBA.

## Description
The heuristics added to distinguish VBA from VB6 were suggested by @XusinboyBekchanov :
* [frm](https://github.com/github/linguist/issues/5824#issuecomment-1296763172) 
VBA:
image
VB6:
image

* [cls](https://github.com/github/linguist/issues/5824#issuecomment-1296774239)
VB6 :
image
VBA : If the pattern above doesn't match, we default to VBA.

The key of these are based on the way the VB6 and VBA IDEs encode information at the top of the file when exporting the forms or class module as text files. From what I know, there isn't really a good reason for developers to alter this information as this could cause issues when trying to import back the text file into their IDE. Hence, those differences should be sufficiently reliable to distinguish between VBA and VB6.

PS : Due to the way I pulled changes from master, this pull request appears to contain more commits that it really does. Commits before [2d3547](https://github.com/github/linguist/pull/6155/commits/2d35478a676817d5b94f2949b1d3a3c2adb41f80) can be ignored.

### VB6
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Afrm+"Begin+VB.form"+NOT+nothack
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Acls+Persistable+NOT+nothack
    - Sample source(s):
      - ~~https://raw.githubusercontent.com/dragokas/hijackthis/devel/src/frmMain.frm~~
      - https://raw.githubusercontent.com/dragokas/hijackthis/devel/src/frmProcMan.frm (GPL-2.0 license)
      - https://raw.githubusercontent.com/tannerhelland/PhotoDemon/main/Forms/Adjustments_BlackAndWhite.frm (BSD-icense)
      - https://raw.githubusercontent.com/dragokas/hijackthis/devel/src/clsDataPack.cls (GPL-2.0 license)
      - https://raw.githubusercontent.com/tannerhelland/PhotoDemon/main/Classes/clsMenuImage.cls   (BSD-icense)
    - Sample license(s): Specified after the link

### VBA
    - Sample source(s):
      - https://raw.githubusercontent.com/ivankokan/Excel2LaTeX/master/src/frmConvert.frm (LaTeX public license)
      - https://raw.githubusercontent.com/sancarn/stdVBA/a6068a7d0176eb55d3095d1b12450daa108d0c38/src/stdExcelLibraries/stdSaveHandler.frm (MIT)
      - https://raw.githubusercontent.com/rubberduck-vba/Battleship/master/src/WorksheetView.cls (MIT)
      - https://raw.githubusercontent.com/krijnsent/crypto_vba/master/WorkbookReporter.cls (MIT)
    - Sample license(s): Specified after the link


lib/linguist/languages.yml
```diff
@@ -6946,7 +6946,6 @@ VBA:
   - ".bas"
   - ".cls"
   - ".frm"
-  - ".frx"
```
lildude, line 6949: Whoops. You don't want to remove this extension. You don't want to remove the `.frm` extension either.

lib/linguist/languages.yml
```diff
@@ -7104,8 +7103,10 @@ Visual Basic 6.0:
   type: programming
   color: "#2c6353"
   extensions:
+  - ".cls"
```
lildude, line 7106: The first extension should be the primary extension for the language. This is tricky in this case but as the file can contain no VB6 code, I don't think it should be first. Move it down one.

lildude (changes requested):
Why are you removing the `samples/VBA/cApplication.cls` sample? E we don't remove samples if we can avoid it.

`samples/Visual Basic 6.0/frmMain.frm` is massive. Please replace it with a smaller sample.

I also recommend adding two distinct samples for each of these extensions to help the classifier in the event things slip through the heuristics.

lib/linguist/languages.yml
DecimalTurn: `.frx` should not have been added in the first place since it is a binary data file and not VB6 code. And the format is proprietary.

> An .FRX file is a binary data file used to store binary information for the form (.FRM) file with the same file name.

> 

> MORE INFORMATION
> Some controls, such as the PictureBox and Image controls, have properties that use binary data as their values. Visual Basic stores the binary data in an .FRX file with the same file name as the Form file. The .FRX file is a proprietary format.
> 
> If your project contains .FRX files, you should include these files with the corresponding .FRM files. A form containing binary information but lacking the associated .FRX files will not function correctly.

[source](https://www.vbforums.com/showthread.php?115463-What-is-an-frx-file&p=617730&viewfull=1#post617730)

lib/linguist/languages.yml
DecimalTurn: Oh, I was going for alphabetical order 😅. In that case, it's a bit tricky indeed, but a `.cls` file is much more likely to include VB6 code than a `.ctl` or `.Dsr` file. In the case of the `.Dsr` file, I already mentionned it [here](https://github.com/github/linguist/pull/6124#issuecomment-1284650876). And for the `.ctl` file, it the same issue and could contain a lot of designer metadata and little to no VB6 ([example 1](https://github.com/Jucko13/KeyboardLights/blob/e5b17df9fd3460a44bb48911197327177937c0d7/Program/bColor.ctl) and [example 2](https://github.com/Cth1003/BonziBuddy/blob/4164bd069b9f59ed629282c307c588a419397d27/src/CalendarVB.ctl)).

In this regard, `.cls` is the primary extension for VB6 (along side `.bas`) since its only purpose is to contain VB code.

DecimalTurn:
Regarding the `.bas` extension, I've intentionally kept it out of this PR since I'm still not sure about the heuristic for it when it's time to distinguish between VBA and VB6.

lildude:
> Because this sample is a VB6 sample (based on the heuristics this PR introduced). My understanding is that in the PR https://github.com/github/linguist/pull/4725, where this sample was added, the intention was to put VBA and VB6 under the same language "VBA/VB6", but when it changed to just VBA, the sample for VB6 was left there by mistake.

Cool. In that case, please move it rather than remove it. We've already approved its inclusion so there's no need to find the source etc.

You can still add another if it'll help the classifier which is falling on this extension.

DecimalTurn:
Tthe Classifier test didn't succeed with 100%, but it's already better:

**.cls** (5/6)
VBA/GridCoord.cls BAD ❌(Visual Basic 6.0)
VBA/WorkbookReporter.cls ✅ 
VBA/dictionary.cls ✅
Visual Basic 6.0/cApplication.cls ✅
Visual Basic 6.0/clsDataPack.cls ✅
Visual Basic 6.0/clsMenuImage.cls ✅


**.frm** (4/4)
VBA/frmConvert.frm ✅
VBA/stdSaveHandler.frm ✅
Visual Basic 6.0/Adjustments_BlackAndWhite.frm ✅
Visual Basic 6.0/frmProcMan.frm ✅

Considering that the heuristics saould be pretty reliable, can this suffice?

lildude:
> The Classifier test didn't succeed with 100%, but it's already 

It needs to. Try removing the sample or adding another sample.

DecimalTurn:
Alright, I've switched the problematic sample with another one from the same project. Now, all the tests passed and I updated the PR description at the top with the correct samples.
Looks like we are good to go unless there is something I missed.

lib/linguist/heuristics.yml
```diff
@@ -138,6 +138,12 @@ disambiguations:
     pattern: '\/\* |\/\/ |^\}'
 - extensions: ['.cls']
   rules:
+  - language: Visual Basic 6.0
+    and:
+    - named_pattern: vb-class
+    - pattern: '^\s*BEGIN(\r\n|\n)\s*MultiUse\s*=.*(\r\n|\n)\s*Persistable\s*='
```
Alhadis, line 144: Using `\R` is more idiomatic and concise:

```suggestion
    - pattern: '^\s*BEGIN\R\s*MultiUse\s*=.*\R\s*Persistable\s*='
```

Also, I noticed the pattern only matches `BEGIN\R`, without accounting for trailing whitespace (`BEGIN\s*\R`). Was this intentional? (I'm not at all familiar with these languages).

Alhadis (changes requested):

lib/linguist/heuristics.yml
DecimalTurn: Thanks for the review. I'm less familiar with Ruby-style regex, so that `\R` trick is good to know!

Regarding the absence of trailing spaces after `BEGIN`, this is indeed intentional because I'm relying on the VB6 IDE which doesn't add trailing spaces. Since this part of the text file is basically metadata generated by the IDE, there's no reason for developers to alter these lines.

Alhadis (approved):

lildude (approved):

lib/linguist/heuristics.yml
bzz: > Using \R is more idiomatic and concise:

... but it also breaks matches that use other regex engines, like RE2 😢

DecimalTurn: I could always revert to what I had originally (`(\r\n|\n)`), but maybe it's better to take this opportunity to a find a good way to convert `\R` for other engines in case this happens again in the future.

From what I can see in the [documentation](https://ruby-doc.org/3.2.1/Regexp.html#:~:text=/%5CR/%20%2D%20A%20linebreak%3A%20%5Cn%2C%20%5Cv%2C%20%5Cf%2C%20%5Cr%20%5Cu0085%20(NEXT%20LINE)%2C%20%5Cu2028%20(LINE%20SEPARATOR)%2C%20%5Cu2029%20(PARAGRAPH%20SEPARATOR)%20or%20%5Cr%5Cn.), the long way to express `\R` would be something like: `(\n|\v|\f|\r|\u0085|\u2028|\u2029|\r\n)`.

@bzz Would that longer expression work with RE2?

Alhadis: @bzz I fail to see how RE2 is relevant here, as these heuristics are only consumed by Linguist (a Ruby program, which uses [Oniguruma](https://github.com/kkos/oniguruma/blob/master/doc/RE)-flavoured regexes).

@DecimalTurn Please stick with `\R`; it's already in-use elsewhere in `heuristics.yml` ([exempli gratia](https://github.com/DecimalTurn/linguist/blob/d790d3f4311ac80aabbcfd7b178793a1eba76da2/lib/linguist/heuristics.yml#L661)), and its inclusion isn't going to break anything.

DecimalTurn: @Alhadis - Compatibilty with RE2 is relevant insofar as we care about Code Search to be in line with Linguist. As described in this [blog post](https://github.blog/2021-12-15-a-brief-history-of-code-search-at-github/), GitHub now uses [Enry ](https://github.com/go-enry/go-enry)(which itself uses RE2) for language detection in Code Search. 

Divergences between Linguist and Enry can lead to some weird situations. For example, Linguist might say that a language is present in a repo when looking at the language statistics bar, but Code Search could then find no files using that language.

In the case of `\R`, I don't think it is really problematic since Enry can likely solve the issue with a simple substitution (as I mentioned above). Nonetheless, I do believe that it's a valid point of discussion to raise here. Ideally, we want to keep [the list of divergences](https://github.com/go-enry/go-enry#divergences-from-linguist) as small as possible.

Alhadis: @DecimalTurn Ah. I wasn't aware that the new Code Search was being powered by Go. That changes things.

Regarding `\R`, I'm actually wondering if it wouldn't make more sense to have Linguist replace `\n` with `(?:\r?\n|\r)` when loading `heuristics.yml`: it's unlikely that a meaningful heuristic will ever depend on one's choice of line terminator. If I'm wrong, then `\\x0A` or `\\x0D\\x0A` could be used as a workaround to circumvent the "auto-newline" feature.

PR #6159, opened 2022-11-08
Update Ruby versions for CI: drop 2.5 & add 3.1

chenrui333:
## Description
Ruby 2.5 is already EOL, https://endoflife.software/programming-languages/server-side-scripting/ruby

Ruby 3.1 is now available too.

## Checklist:
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ '3.0', '2.7', '2.6' ]
+        ruby: [ '3.1', '3.0', '2.7', '2.6' ]
     steps:
     - uses: actions/checkout@v1
     - name: Set up Ruby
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ '3.0', '2.7', '2.6', '2.5' ]
+        ruby: [ '3.0', '2.7', '2.6' ]
```
lildude, line 21: We may as well add 3.1 to the list whilst we're at it.

```suggestion
        ruby: [ '3.1', '3.0', '2.7', '2.6' ]
```

lildude:
I've updated the samples test code as were relying on broken functionality which was fixed in  https://github.com/ruby/ruby/pull/4583 to bring the `Dir.glob` behaviour inline with what the [docs](https://docs.ruby-lang.org/en/3.1/File/File/Constants.html#FNM_CASEFOLD) have been saying for years.

lildude (approved):

PR #6160, opened 2022-11-11
Unvendor the `vignettes/` directory for R packages

jennybc:
## Description

In an R package, the `vignettes/` directory holds executable documents that are rendered into documentation at package build time. Currently linguist incorrectly labels this folder as **vendored** code, which has multiple downstream downsides, including causing exclusion of the files below `vignettes/` from code search. (It seems like linguist might use the "vendored" label pragmatically, but then other projects use this data and take the label at face value?)

This was introduced in https://github.com/github/linguist/commit/9ae0bdbb434aa8f2a92b4c805712fa0f99e03b1c, which was wrong IMO. (Or, at least, has become wrong as other folks started to use linguist's vendored list as if it really means vendored).

The executable documents in `vignettes/` tend to be RMarkdown documents, a file type that that is already (correctly) handled elsewhere in linguist:

In `languages.yml`:

https://github.com/github/linguist/blob/16a7a528ab840dd24188edda671d31881a67e527/lib/linguist/languages.yml#L5428-L5439

The executable documents we keep in `vignettes/` are compiled to PDF or html files that live in `inst/doc/`, which is already handled (by which I mean "excluded"), thanks to #4068:

https://github.com/github/linguist/blob/16a7a528ab840dd24188edda671d31881a67e527/lib/linguist/documentation.yml#L19

linguist is correctly ignoring the rendered documentation but is incorrectly labelling the source as vendored code, which it's not.

## Checklist:
  
It's not clear to me that this functionality is tested. Sorry if that's just user error.


jennybc:
If you want to see a bunch of wild-caught R packages, we have a read-only mirror of CRAN here: <https://github.com/cran>.

Some packages, but not all, have `vignettes/` and that mostly contains `.Rmd` files. When `vignettes/` is present, the corresponding rendered documentation is found int `inst/doc/`.

lildude (approved):
I know nothing about R but your explanation makes sense and checks out.

hadley:
@lildude any sense of how long it will take this change to percolate in code indexing on github.com?

hadley:
@lildude thanks!

PR #6166, opened 2022-11-18
Add support for flake.lock

figsoda:
Make `flake.lock` files JSON and mark them as generated.

## Description
`flake.lock` is the lock file for [Nix flakes](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#lock-files) generated by the `nix` command.

## Checklist:
- [x] **I am associating a language with a new file extension**
	- [x] **The new extension is used in hundreds of repositories on GitHub.com:**  
	[24.7k results for `path:**/flake.lock`](https://github.com/search?q=path%3A**%2Fflake.lock&type=code)
	- [x] **I have included a real-world usage sample:**  
		- Sample source:
		- Sample license(s):
	- [ ] ~~I have included a change to the heuristics~~
	- [x] **I have added or updated the tests for the new or changed functionality.**


lib/linguist/generated.rb
```diff
@@ -466,6 +467,13 @@ def cargo_lock?
       !!name.match(/Cargo\.lock/)
     end
 
+    # Internal: Is the blob a generated Nix flakes lock file?
+    #
+    # Returns true or false
+    def flake_lock?
+      !!name.match(/flake\.lock/)
```
Alhadis, line 474: `/flake\.lock/` will match filenames that merely contain `flake.lock` as a substring; i.e., “<samp>Backup‑of‑<ins><b>flake.lock</b></ins>files.tgz</samp>”. To match the complete filename only, use:

```suggestion
      name.downcase == 'flake.lock'
```

Remove the `.downcase` part if the match is supposed to be case-sensitive.

Alhadis (changes requested):

Jennifejpugh9091:
STOP

lib/linguist/generated.rb
figsoda: fixed it, since it is case sensitive, I removed the downcase

Alhadis (approved):

Alhadis:
Looks great. Real-world distribution checks out. Just to confirm the origin of the sample you've added, this _is_ one that you've created, correct? I saw your username [listed](https://github.com/github/linguist/blob/262f4ad00ffe4f4271805bbeb1877110635227d0/samples/JSON/filenames/flake.lock#L35) in one of the fields, but I'd rather not go off an assumption.

figsoda:
It is, it is copied from https://github.com/figsoda/cfg/blob/main/flake.lock, probably one commit or so behind main

figsoda:
The test fail seems to be because that `name` is the full path of the file, maybe something like `!!name.match((^|\/)flake.lock$)` would work?

Alhadis:
> The test fail seems to be because that `name` is the full path of the file

Shit, [you're right](https://github.com/github/linguist/blob/086adcaa8e6671d0130c572b673b064d524b4adf/lib/linguist/generated.rb#L396). 🤔 So many of these tests are interrogating only the `basename(3)` of a file path, it didn't occur to me that there might be a `dirname(3)` component. And, of course, it doesn't help that the property in question has a misleading name/description…

In any case, these oversights are beyond the scope of this PR. For now, we'll just get these tests passing.

lildude (approved):
LGTM. Thanks.

PR #6167, opened 2022-11-18
Amend Just language entry

casey:
Sorry for not catching this in the initial review!

- Remove `Justfile` alias. `just` is the name of the tool and the language.
- Just looks for files called `justfile` in any case, so add the three most common cases: `JUSTFILE`, `Justfile`, and `justfile`.
- Add the `.just` extension, which is in common use. See this search: https://github.com/search?q=extension%3Ajust


mihaigalos (commented):
@casey, what would you say if we refactor this to `Justfile` instead of `just` as file naming?
The way it currently shows up is this:
image

mihaigalos:
Or capitalize the 1st letter?

Alhadis:
> Or capitalize the 1st letter?

Yes, that's supposed to be the language's **human-readable name**, not its file extension or interpreter. Note that exceptions are made for non-standard casing if that's how the language's name is canonically spelt (e.g., `reStructuredText`, `mIRC Script`, `cURL Config`, et al).

> Remove `Justfile` alias. `just` is the name of the tool and the language.

Don't. This is a valid addition: it enables users to highlight Markdown code-blocks with `` ```justfile `` as well as `` ```just ``.

lib/linguist/languages.yml
```diff
@@ -7829,12 +7829,14 @@ jq:
...
-  aliases:
-  - Justfile
   color: "#384d54"
   tm_scope: source.just
   filenames:
+  - JUSTFILE
   - Justfile
+  - justfile
+  extensions:
+  - ".just"
```
Alhadis, line 7839: > Add the `.just` extension, which is in common use. See this search: https://github.com/search?q=extension%3Ajust

Using GitHub's new Code Search engine, I only found 88 results for [`path:**/?*.just`](https://github.com/search?q=path%3A**%2F%3F*.just&type=code), which falls far short of [our real-world usage criteria](https://github.com/github/linguist/blob/59b2d88b2242e6062384e5fb876668cc30ead951/CONTRIBUTING.md#adding-an-extension-to-a-language) (at least 200 unique users/repos that use the new extension).

Alhadis (changes requested):

mihaigalos:
@casey: Can you please allow others' contributions to this PR? I wanted to implement some of the suggestions but couldn't since it's read-only.

lib/linguist/languages.yml
casey: Done! I removed the `.just` extension.

Is there a way for me to access the new Code search engine? The link `https://github.com/search?q=path%3A**%2F%3F*.just&type=code` doesn't show any results.

casey:
@mihaigalos I think of the name of the language as being `just`, as in `the just language`, and not `the justfile language`. Doing a quick search on google, "the make language" seems to be more common than "the makefile language", so this is consistent with make. Although, I note that `Makefile` is used by linguist. Overall I prefer `just`, and capitalizing it seems to be the convention with linguist, so I changed it to `Just`.

> Don't. This is a valid addition: it enables users to highlight Markdown code-blocks with ```justfile as well as ```just.

@Alhadis I'm not sure I see the value of adding an additional alias. Wouldn't it be better to have a single way of annotating markdown code blocks?

Alhadis:
> Wouldn't it be better to have a single way of annotating markdown code blocks?

No, because this is the same lookup machinery responsible for interpreting `language:` operators in code-searches, or the value of `linguist-language=` in a `.gitattributes` file. I think we can all agree that `js` and `javascript` should both be mapped to the same language entry. Likewise for Just.

Remember, users who type `` ```justfile `` in the middle of a long-winded comment won't always catch themselves writing the "wrong" name. Some might even fail to notice that their annotated code-block is missing highlighting altogether. We should strive for intuitive, "DWIM" behaviour instead of being needlessly restrictive.

Alhadis:
@casey FYI, I pushed a fix to your branch that updated the capitalisation of the `samples/just` directory—something that's notoriously fiddly to do on a case-insensitive filesystem (so, macOS and Windows). The `languages.yml` list is also meant to be ordered case-sensitively, so I fixed that as well.

Alhadis (approved):

casey:
> @casey FYI, I pushed a fix to your branch that updated the capitalisation of the samples/just directory—something that's notoriously fiddly to do on a case-insensitive filesystem (so, macOS and Windows). The languages.yml list is also meant to be ordered case-sensitively, so I fixed that as well.

Thanks!

> Remember, users who type ```justfile in the middle of a long-winded comment won't always catch themselves writing the "wrong" name. Some might even fail to notice that their annotated code-block is missing highlighting altogether. We should strive for intuitive, "DWIM" behaviour instead of being needlessly restrictive.

Fair enough! I added the alias back.

Alhadis:
Ace. Well, this looks good to me. Now we just need to wait for @lildude to give the GitHub Staffer's Seal-of-Approval™ so we can merge this. :shipit:

lib/linguist/languages.yml
lildude: Yup: https://github.com/features/code-search/

lildude (approved):
LGTM. Thanks.

lildude:
Don’t worry about merging in master. I’ll do it again before merging.

lib/linguist/languages.yml
```diff
@@ -3331,6 +3331,18 @@ Jupyter Notebook:
...
+Just:
+  type: programming
+  aliases:
+  - Justfile
+  color: "#384d54"
+  tm_scope: source.just
+  filenames:
+  - JUSTFILE
+  - Justfile
+  - justfile
```
mihaigalos, line 3343: @casey:  [This](https://github.com/casey/just#hiding-justfiles) talks about `.justfile` naming. Would it not make sense to also include it in this list?

nk9: Note that in the wild, users also frequently use `.just` as a file suffix for Justfiles.

lildude: Neither of these extensions are sufficiently popular on GitHub for inclusion right now. The `.just` extension was discussed earlier in this PR [here](https://github.com/github/linguist/pull/6167#pullrequestreview-1185649435).

mihaigalos: I understood the popularity requirement as something to be fulfilled for new languages.
The PR is not producing support for a new language (the language is already known to Linguist), but instead adding file aliases to an already known language.

lildude: The popularity requirement applies for all new extensions and filenames, regardless of whether it's a new or known language. See the details in CONTRIBUTING.md

Issue #6180, opened 2022-11-21
Svelte highlighting seems to be not working


stevepopovich:
## Describe the bug

Hi folks, hubber here from the mobile team. Svelte highlighting seems to be broken and has been for months. I am trying to get to the bottom of what is going on and need some direction, hoping y'all can help.

Highlighting in this https://github.com/Lleweraf/supachat/blob/latest/src/components/ChatInput.svelte is broken, both on dotcom and on mobile.

### Expected behaviour

We should see svelte code highlighted properly 

### Related discussion

[This](https://github.com/github/mobile/issues/3068) is the original issue we are tracking on mobile.

lildude:
@stevepopovich I recently got to the bottom of this in https://github.com/github/linguist/discussions/6164

In short, the grammar is using regex matches in the scope names but GitHub's very old syntax highlighter doesn't expand these so subsequent rules aren't found.

The old highlighter isn't going to be fixed so the options are:

- update the grammar to take this limitation into account, or
- confirm the tree-sitter grammar is at least as complete as the Textmate-based grammar and renders correctly and then I can lobby to have that added to the new highlighter instead.

lildude:
I've done some quick testing in dev using the tree-sitter grammar and it looks promising and much better than the Textmate grammar:

![CleanShot 2022-11-22 at 09 45 37](https://user-images.githubusercontent.com/627280/203281409-8d8a2cd2-8dd5-4e32-b6ee-3771953e69b8.png)

I'm going to submit a PR proposing to add the Svelte tree-sitter grammar. We'll see how it goes.

stevepopovich:
Amazing! Thank you. Sorry I was in the wrong spot, things have gotten confusing for me trying to figure out who owns the problem.

lildude:
So it turns out my local dev env was not actually pulling in the treesitter grammar when testing. It was actually pulling in the TextMate grammar from Linguist v7.20.0 which means the TextMate grammar was working properly then and started failing after that.

More deets in my response at https://github.com/github/linguist/discussions/6164#discussioncomment-4215301

stevepopovich:
So once https://github.com/github/github/pull/248502 is deployed, this is expected to be fixed?

lildude:
Yup.

lildude:
And we're back to working Svelte syntax highlighting (below should look like the screenshot above):

```svelte
<script>
  const notWorking = 'I am not highlighted';
</script>

{#each something as s}
[...]

PR #5963, opened 2022-07-05
Change source for svelte

3ventic:
## Description
Changes the source for Svelte from a seemingly abandoned fork to the original, active, repository. This change is prompted by a bug in the grammar; the fork has disabled issues, and the original repository has since added a compatible license (MIT) and commits, which possibly fix my issue.

## Checklist:

- [x] **I am changing the source of a syntax highlighting grammar**
  - Old: https://github.com/umanghome/svelte-atom
  - New: https://github.com/sveltejs/svelte-atom

PR #6184, opened 2022-11-24
Revert "Change grammar source for Svelte"

lildude:
Reverts github/linguist#5963

This broke the syntax highlighting between `<(script|style|template)>` tags due to a really old limitation in the old prettylights highlighting engine which won't be fixed.

See https://github.com/github/linguist/discussions/6164 for more details.

This can be reverted back to the sveltejs/svelte-atom grammar once it has been updated to address this limitation or a production-ready tree-sitter grammar becomes available.

This PR will be merged early next week (w/c 28 Nov 2022) and a patch release made, unless the sveltejs/svelte-atom grammar is updated by then, in which case I'll pull the latest version of that grammar.

Fixes https://github.com/github/linguist/issues/6180

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1108,7 +1108,7 @@
 	url = https://github.com/tenable/sublimetext-nasl
 [submodule "vendor/grammars/svelte-atom"]
 	path = vendor/grammars/svelte-atom
-	url = https://github.com/sveltejs/svelte-atom.git
+	url = https://github.com/umanghome/svelte-atom
 [submodule "vendor/grammars/swift.tmbundle"]
 	path = vendor/grammars/swift.tmbundle
 	url = https://github.com/textmate/swift.tmbundle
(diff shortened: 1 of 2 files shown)


Alhadis:
> due to a really old limitation in the old prettylights highlighting engine

Could you elaborate on the specifics of said limitation? This is relevant because I still use Atom for writing, previewing and testing grammars, and I need to know of any discrepancies between it and GitHub/PrettyLights.

lildude:
See https://github.com/github/linguist/discussions/6164#discussioncomment-4178987

Alhadis:
> The problem is it doesn't expand scope matches and instead treats the `$1` in places like [this](https://github.com/sveltejs/svelte-atom/blob/865a55f096ed0d60a316234f55f6f356abd5cd1f/grammars/svelte.cson#L606) literally:

Oh, that. I always assumed that was an intentional optimisation due to the rarity of meaningful scopes getting captured by a grammar. It's bitten me several times when abusing injections (most recently in [`language-mermaid`](https://github.com/Alhadis/language-mermaid/blob/1cd1997642b11a551bc415e8833f33c55850158f/grammars/all.cson)).

(Sorry for the noise…)

lildude:
> Oh, that. I always assumed that was an intentional optimisation due to the rarity of meaningful scopes getting captured by a grammar.

It's certainly possible. I can't find any explanation of why this wasn't implemented other than a "Expand?" comment in the code and the author left GitHub a long time ago - I doubt they could remember why they wrote that over 8 years ago too 😁.

PR #6187, opened 2022-11-24
Add clarification for override using alias + case-insensitive

DecimalTurn:
I noticed that it was possible to use an alias of the language for overrides [here](https://github.com/fafalone/LongLongHelper/commit/0e91ab04833e3d8236ea53eef94174b0363c8b1a), so thought it would be good to mention it in the docs.

### DIFF
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -19,6 +19,9 @@ File and folder paths inside `.gitattributes` are calculated relative to the pos
 # Replace any whitespace in the language name with hyphens:
 *.glyphs linguist-language=OpenStep-Property-List
 
+# Language names are case-insensitive. Hence, the line above is equivalent to:
+*.glyphs linguist-language=openstep-property-list
+
 # You can also use an alias of the language (eg. js instead of JavaScript)
 *.cshtml linguist-language=js
 ```
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -18,6 +18,9 @@ File and folder paths inside `.gitattributes` are calculated relative to the pos
 
 # Replace any whitespace in the language name with hyphens:
 *.glyphs linguist-language=OpenStep-Property-List
+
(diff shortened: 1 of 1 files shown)


docs/overrides.md
```diff
@@ -18,6 +18,9 @@ File and folder paths inside `.gitattributes` are calculated relative to the pos
 
 # Replace any whitespace in the language name with hyphens:
 *.glyphs linguist-language=OpenStep-Property-List
+
+# You can also use an alias of the language (eg. js instead of JavaScript)
+*.cshtml linguist-language=js
```
Alhadis, line 23:
```suggestion
# Language names are case-insensitive and may be specified using an alias:
*.cshtml linguist-language=js
```

Alhadis (changes requested):
We might as well take this opportunity to point out that language names are matched case-insensitively as well.

docs/overrides.md
DecimalTurn: Good point! But I'd prefer to keep this info in a seperate line for more clarity. I've updated the code accordingly.

Alhadis: This is an introductory section, so it behoves us to keep examples as concise as possible. If the wording I proposed earlier was unclear, perhaps try this instead:

~~~gitattributes
# Language names are case-insensitive and may be specified using an alias:
# the following three lines are all functionally equivalent:
*.es linguist-language=js
*.es linguist-language=JS
*.es linguist-language=JAVASCRIPT
~~~

DecimalTurn: Ok, that multi-line example is clear enough. I've pushed this change with a slight punctuation adjustment to avoid using the colon twice.

Alhadis (approved):
Works for me. I personally would've opened the second sentence with `Ergo` instead of `So` (it's a bit more formal), but that's pretty much nitpicking at this point.

lildude (approved):

PR #6221, opened 2022-12-20
register .vscode/{extensions.json,settings.json,launch.json} as `JSON with Comments`

gfx:
As you folks may know, `.vscode/*.json` are JSON with Comments (jsonc), not JSON. That is, when I open, e.g., `.vscode/settings.json` on GitHub Web, I'd expect it is rendered as jsonc. 

If I add a git attribute to these files, I can see those files are rendered as jsonc (like `.vscode/*.json linguist-language=jsonc` => https://github.com/msgpack/msgpack-javascript/blob/main/.vscode/launch.json), but I believe it should be the default.

## Description

I'm adding a new lookup path to `Language.find_by_filename` to see the entire given filename, as well as its basename. I'm not sure this is an ideal implementation. The implementation might be arguable.  It's still O(1) but sounds too ad-hoc and it cannot handle deep paths such as `some/other/dir/.vscode/settings.json`. Another algorithm (e.g. fnmatch(3)-like behavior + cache) would be better but I'd like to discuss the new functionality before implementing such a smart algorithm. 

## Checklist:

### DIFF
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -144,7 +144,13 @@ module Linguist
     # Returns all matching Languages or [] if none were found.
     def self.find_by_filename(filename)
       basename = File.basename(filename)
-      @filename_index[basename]
+      languages = @filename_index[basename]
+
+      if languages.empty? && @filename_index.key?(filename)
+        @filename_index[filename]
+      else
+        languages
+      end
     end
 
     # Public: Look up Languages by file extension.
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3072,6 +3072,9 @@ JSON with Comments:
   - ".jscsrc"
   - ".jshintrc"
   - ".jslintrc"
+  - ".vscode/extensions.json"
+  - ".vscode/launch.json"
+  - ".vscode/settings.json"
   - api-extractor.json
   - devcontainer.json
   - jsconfig.json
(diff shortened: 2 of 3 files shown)


Alhadis:
This has been suggested and requested before, and as it stands, such a feature would require Linguist to operate on pathnames, rather than filenames. This alone would involve disruptive changes to the codebase, as methods—including public APIs—can no longer assume their input will be a simple `basename(3)` (file name + file extension). Such headaches need to be justified in the long-term… but here, the only benefit is defaulting to nicer-looking highlighting for VSCode users who haven't added a `.gitattributes` to their project yet.

This is a problem specific to this one particular scenario: in its 11 years of existence, Linguist hasn't ever needed to consider parent directories as part of its filename/format mapping.

I'd also like to stress that part of this JSON/JSONC problem rests with Microsoft (or whoever's responsible for steering VSCode development). This issue can—and *should*—be resolved by using the `.jsonc` extension instead of `.json`, because it's in open conflict with a standardised data format (JSON) which is guaranteed never to change. IMO, if a change is to be made, it should be to VSCode.

gfx:
@Alhadis

Thanks for your quick response. I perfectly agree with you. It is the vscode team that violates the guideline of the world 😓 

I'm not sure vscode should support `.vscode/*.jsonc` as well as `.vscode/*.json` because I believe the JSON spec will have comments and trailing commas in the future. Hope the vscode team proposes a new JSON spec to make it compatible with JSONC.

Alhadis:
> I believe the JSON spec will have comments and trailing commas in the future.

It won't. Having two different versions of the JSON spec would create interoperability issues between programs that support different versions of the spec; that is, a spec-compliant JSON parser that's worked perfectly for the past 12 years suddenly breaks because it's fed JSONC — syntax it's not designed to deal with, per the original spec.

PR #6275, opened 2023-02-04
Add Sway language

sdankel:
## Description

Sway is a Rust-like programming language for smart contracts.

The examples are from the official [sway examples](https://github.com/FuelLabs/sway/tree/master/examples) ([Apache](https://github.com/FuelLabs/sway/blob/master/LICENSE) license).

### Color

I chose the same color used by Rust, since Sway is based on Rust syntax.

## Popularity

It's hard to tell exactly how many repositories are using sway because of the current limitations of github search. Using [sourcegraph](https://sourcegraph.com/search?q=context:global+file:Forc.toml%24+count:2000&patternType=standard&sm=0&groupBy=repo), I can see that there are at least 910 sway projects on github (each sway project must have a `Forc.toml` in it's root directory). 

Based on the temporary requirements outlined in https://github.com/github/linguist/issues/5756, I believe there are over 2000 sway files, as sway projects usually have multiple `.sw` files, ([example](https://github.com/nfurfaro/Optimistic-Chess/tree/5748a23a802853c91a55c3417e3e44ff120f2380/project/lib_chess/src)).

There are [9,312](https://github.com/search?q=extension%3Asw&type=code) files with the `.sw` extension, but some of these are not sway files. `.sw` is also used for [Signed Word Audio Files](https://pc.net/extensions/file/sw#:~:text=An%20SW%20file%20is%20a,files%2C%20meaning%20they%20are%20uncompressed.). Those are already correctly highlighted by github as XML ([example](https://github.com/br-automation-com/brdkVision/blob/7f8a7040775bb14a8da281d0ba87ed6ea519d930/project/Physical/ARM/PLC/Cpu.sw)) or not highlighted ([example](https://github.com/Gao-Group/lammps-3Mar20-atomdnn/blob/43a56e29f9be83f0624e95cac656763c4275f403/src/USER-INTEL/TEST/in.intel.sw)). Added heuristics for sway files to disambiguate.

## Checklist:
       https://github.com/search?q=extension%3Asw&type=Code

    - Sample source(s) above
    - Sample license(s): above
  - [x] I have included a syntax highlighting grammar: https://github.com/FuelLabs/sway-vscode-plugin/blob/master/syntaxes/sway.tmLanguage.json


lib/linguist/heuristics.yml
```diff
@@ -672,6 +672,10 @@ disambiguations:
   rules:
   - language: STL
     pattern: '\A\s*solid(?=$|\s)(?m:.*?)\Rendsolid(?:$|\s)'
+- extensions: ['.sw']
+  rules:
+  - language: Sway
+    pattern: '^(dep |fn |mod |pub |trait |impl |abi #!?\[)'
```
lildude, line 678: I think this can definitely be optimised, especially the final `abi #!?\[` part as I don't think this will match what I think you're hoping it'll match based on the samples included.

/cc @Alhadis as the resident regex guru.

lildude (changes requested):
> It's hard to tell exactly how many repositories are using sway because of the current limitations of github search. Using [sourcegraph](https://sourcegraph.com/search?q=context:global+file:Forc.toml%24+count:2000&patternType=standard&sm=0&groupBy=repo), I can see that there are at least 910 sway projects on github (each sway project must have a `Forc.toml` in it's root directory).

You're in luck. The new [beta search](https://github.com/features/code-search/) can take your exact heuristic and tell us 😁.

[This search](https://github.com/search?q=path%3A*.sw+AND+%2F%5E%28dep+%7Cfn+%7Cmod+%7Cpub+%7Ctrait+%7Cimpl+%7Cabi+%23%21%3F%5C%5B%29%2F&type=code) results in 2.7k files and from it we can see there are a few heavy influencers, so if we [exclude the 4 most prominent users](https://github.com/search?q=path%3A*.sw+AND+%2F%5E%28dep+%7Cfn+%7Cmod+%7Cpub+%7Ctrait+%7Cimpl+%7Cabi+%23%21%3F%5C%5B%29%2F+NOT+%28user%3AFuelLabs+OR+user%3Alibsway+OR+user%3ABlockchainCap+OR+user%3Asway-gang%29&type=code) from a quick look, we're down to 2.3k which appear to be well spread across repos so I think we're good to include this language.

Buuuuuttttt, as you've pointed out:

> There are [9,312](https://github.com/search?q=extension%3Asw&type=code) files with the .sw extension, but some of these are not sway files. .sw is also used for [Signed Word Audio Files](https://pc.net/extensions/file/sw#:~:text=An%20SW%20file%20is%20a,files%2C%20meaning%20they%20are%20uncompressed.). Those are already correctly highlighted by github as XML ([example](https://github.com/br-automation-com/brdkVision/blob/7f8a7040775bb14a8da281d0ba87ed6ea519d930/project/Physical/ARM/PLC/Cpu.sw)) or not highlighted ([example](https://github.com/Gao-Group/lammps-3Mar20-atomdnn/blob/43a56e29f9be83f0624e95cac656763c4275f403/src/USER-INTEL/TEST/in.intel.sw)). Added heuristics for sway files to disambiguate.

As this will be the only language Linguist knows about with this extension, your heuristic won't be used. In order to get the heuristic to correctly identify the Sway files, you'll need to add `.sw` to at least another language. As you have clearly identified XML files, we can add `.sw` to XML's list of extensions and add a sample.

So with that in mind...

1. Please add `.sw` and a sample to XML
2. Update the heuristic test to be sure it doesn't misclassify the XML file as Sway
3. Replace your samples with more real world looking uses of the language from the examples you got these from. These samples look like contrived demonstrations of the syntax rather than illustrations of real world usage.

lib/linguist/heuristics.yml
Alhadis: This regex will match any of the given strings:

1. `dep `
2. `fn `
3. `mod `
4. `pub `
5. `trait `
6. `impl `
7. `abi #[`
8. `abi #![`

I don't see any `#` characters in the samples, but without making assumptions about author intent, a more robust regex would be

~~~regexp
^\s*(?:(?:dep|fn|impl|mod|pub|trait)\s|abi\s#!?\[)
~~~

sdankel: Updated it, I'm expecting it to support:
1. `dep` 
1. `fn`
1. `mod` 
1. `pub` 
1. `trait` 
1. `impl` 
1. `abi`
1. `#[`

Thanks for pointing this out!

sdankel: Updated the samples as well. Let me know what you think.

Alhadis: > I'm expecting it to support:

Is this is an exclusive list? As in, `abi` should be matched separately to `#[`? If so, then a better regex would be

~~~regexp
^\s*(?:(?:abi|dep|fn|impl|mod|pub|trait)\s|#\[)
~~~

test/test_heuristics.rb
```diff
@@ -903,6 +903,12 @@ def test_stl_by_heuristics
     })
   end
 
+  def test_sw_by_heuristics
+    assert_heuristics({
+      "Sway" => all_fixtures("Sway", "*.sw"),
+    })
```
lildude, line 909: We'll need a test here to test the XML heuristic for this extension too.

lildude:
Looking good. One final thing, could you please state in the OP the motivation/influence for the colour you've chosen. I plan to add this to the template to help give some clarity in future if/when peeps submit a PR to change a language's colour.

sdankel:
> Looking good. One final thing, could you please state in the OP the motivation/influence/inspiration for the colour you've chosen. I plan to add this to the template to help give some clarity in future if/when peeps submit a PR to change a language's colour.

Done ✅

lildude (approved):

PR #6297, opened 2023-02-23
Add Erlang application resource file extension

gionea:
## Description

An application resource file (`*.app`) contains an application specification which defines an Erlang application. It is officially documented in the [`app`](https://www.erlang.org/doc/man/app.html) Erlang manual page. The `*.app` file is similar to the `*.app.src` file (which is correctly categorized by the Linguist library; see #2964) in that an `*.app.src` file is used as input by various build systems to generate an `*.app` file. It is, however, also common to manually write an `*.app` file in which case no `*.app.src` file is present in the source tree.

The syntax of an `*.app` file (as that of an `*.app.src` file) is standard Erlang syntax.

Due to the generic nature of its name we may need some heuristics to distinguish this particular usage from others.

## Checklist:

- [x] **I am associating a language with a new file extension**
      
      - [`path:*.app`](https://github.com/search?q=path%3A*.app&type=code)
      - [`path:*.app language:Erlang`](https://github.com/search?q=path%3A*.app+language%3AErlang&type=code&l=Erlang)
    - Sample source(s):
      - [`otp/bootstrap/lib/compiler/ebin/compiler.app`](https://github.com/erlang/otp/blob/master/bootstrap/lib/compiler/ebin/compiler.app)
      - [`otp/bootstrap/lib/kernel/ebin/kernel.app`](https://github.com/erlang/otp/blob/master/bootstrap/lib/kernel/ebin/kernel.app)
    - Sample license(s):
      - Apache version 2
  - [x] I have included a change to the heuristics to distinguish my language from others using the same extension


lildude (changes requested):
> Due to the generic nature of its name we may need some heuristics to distinguish this particular usage from others.

Most definitely and it will need to be 100% accurate. There is no room for misclassification of generic extensions. There are over 27k files with this extension so there is massive scope for misclassification. 

If you can come up with a heuristic that precisely matches these Erlang files and only these files, you'll also need to add the extension to the [`generic.yml`](https://github.com/github/linguist/blob/master/lib/linguist/generic.yml) file and add tests like for other heuristics and generic extensions.

If you can't come up with a 100% precise heuristic, it's best not to pursue this PR and let users use an override if they really want these files to be classified as Erlang.

> 
>     * Sample source(s):
>       
>       * [`otp/bootstrap/lib/kernel/ebin/kernel.app`](https://github.com/erlang/otp/blob/master/bootstrap/lib/kernel/ebin/kernel.app)
>       * [`cowboy/ebin/cowboy.app`](https://github.com/ninenines/cowboy/blob/master/ebin/cowboy.app)

No you haven't. 😁 You've included a very "hello world" looking sample. If you're going to go ahead with this PR, please remove that sample and add the two you've referenced.

lib/linguist/languages.yml
```diff
@@ -1819,8 +1819,9 @@ Erlang:
   type: programming
   color: "#B83998"
   extensions:
-  - ".erl"
+  - ".app"
```
lildude, line 1822: You don't want to do this 😉 

The only time the alphabetisation doesn't apply is for the first extension. This is the primary extension for the language and I doubt it's changed from `.erl` to `.app` 😉

gionea: Ah, sorry, I seem to have missed exactly the sentence about the primary extension in the guide.

lib/linguist/heuristics.yml
```diff
@@ -61,6 +61,10 @@ disambiguations:
     - pattern: '\b(?i:(CODEUNIT|PAGE|PAGEEXTENSION|PAGECUSTOMIZATION|DOTNET|ENUM|ENUMEXTENSION|VALUE|QUERY|REPORT|TABLE|TABLEEXTENSION|XMLPORT|PROFILE|CONTROLADDIN|REPORTEXTENSION|INTERFACE|PERMISSIONSET|PERMISSIONSETEXTENSION|ENTITLEMENT))\b'
   # Open-ended fallback to Perl AutoLoader
   - language: Perl
+- extensions: ['.app']
+  rules:
+  - language: Erlang
+    pattern: '^{\s*(?:application|''application'')\s*,\s*(?:[a-z]+[a-z\d_@]*|''[\s\S]+'')\s*,\s*\[[\s\S]*\]\s*}\.$'
```
lildude, line 67: This doesn't look right to me with all those unescaped quots and `\s` and feels like it could be improved too.

@Alhadis over to you 🙇

lib/linguist/languages.yml
lildude: Top of the file:

https://github.com/github/linguist/blob/4101b6b6c118c0d978b1f4cd4c5bc66bd81a82d3/lib/linguist/languages.yml#L17-L19

gionea: I know, was not contradicting you. It is also mentioned in the contribution guide.

lib/linguist/heuristics.yml
gionea: An application specification is an Erlang term, more specifically, a triple. White space is not significant in Erlang and any of the terms can be padded with an arbitrary number of white space characters.

lildude: @Alhadis when you've got a mo, could you take a squizz as this regex. Thanks 🙇

Alhadis, line 67: 1. The sequence `'[\s\S]+'` will match stuff like `'Foo's Bar'`, which I'm assuming isn't valid Erlang (I know nothing of the language). I'm assuming you meant `'.+'` where `.` includes newlines as well (i.e., the "dotall" modifier typical of most regex flavours). If so, I recommend using `'[^']+'` instead (which doesn't account for escape sequences, if such a thing is supported).

2. Did you intend for `[a-z]` to match only *lowercase* characters? If not, `[a-z\d_@]` can be simplified into `[\w@]`, as (IIRC) GitHub's Linguist process runs without full [CLDR](https://cldr.unicode.org/) support, meaning `\w` will match its traditional POSIX definition (`[[:alnum:]]`, or `[A-Za-z0-9_]`). @lildude will correct me if I'm wrong, I'm sure. 😉

3. `\.$` doesn't account for trailing whitespace. Unless that's a deliberate omission, I recommend using `\.[ \t]*$` to match EOL.

Putting the changes proposed by all three points together, we get:

```suggestion
-    pattern: '^{\s*(?:application|''application'')\s*,\s*(?:[a-z]+[a-z\d_@]*|''[\s\S]+'')\s*,\s*\[[\s\S]*\]\s*}\.$'
+    pattern: '(?m)^{\s*(?:application|''application'')\s*,\s*(?:[a-z]+[\w@]*|''[^'']+'')\s*,\s*\[.*?\]\s*}\.[\t]*$'
```

<table width="700"><tbody><tr><td valign="top" width="350"><details><summary><b>Source (readable)</b></summary>

~~~regexp
(?xm) ^
{ \s*
(?:
	application
	|
	'application'
)
\s* , \s*
(?:
	[a-z]+[\w@]*
	|
	'[^']+'
)
\s* , \s*
\[ .*? \]
\s* }
\. [ \t]* $
~~~

</details></td><td valign="top" width="350"><details><summary><b>Source (YAML string)</b></summary>

~~~yaml
'(?m)^{\s*(?:application|''application'')\s*,\s*(?:[a-z]+[\w@]*|''[^'']+'')\s*,\s*\[.*?\]\s*}\.[\t]*$'
~~~

</details></td></tr></tbody></table>

Alhadis (changes requested):

lib/linguist/heuristics.yml
Alhadis: @lildude Done. 👍

gionea: 1. Your assumption is correct.
2. The first `[a-z]` occurrence should only match lowercase letters. However, `[a-z\d_@]` can be reduced to `[\w@]` since, although uncommon, that particular segment of an atom _can_ include uppercase characters.
3. Not deliberate.

Thank you for your time. I have updated the code to reflect your suggestions.

Alhadis (approved):

lildude (approved):

devilshezi0:
> 
> 
> ## Description
> 
> An application resource file (`*.app`) contains an application specification which defines an Erlang application. It is officially documented in the [`app`](https://www.erlang.org/doc/man/app.html) Erlang manual page. The `*.app` file is similar to the `*.app.src` file (which is correctly categorized by the Linguist library; see #2964) in that an `*.app.src` file is used as input by various build systems to generate an `*.app` file. It is, however, also common to manually write an `*.app` file in which case no `*.app.src` file is present in the source tree.
> 
> The syntax of an `*.app` file (as that of an `*.app.src` file) is standard Erlang syntax.
> 
> Due to the generic nature of its name we may need some heuristics to distinguish this particular usage from others.
> 
> ## Checklist:
> 
> 
> 
> - [x] **I am associating a language with a new file extension**
>   - [x] The new extension is used in hundreds of repositories on GitHub.com
>     - Search results for each extension:
>       
>       - [`path:*.app`](https://github.com/search?q=path%3A*.app&type=code)
>       - [`path:*.app language:Erlang`](https://github.com/search?q=path%3A*.app+language%3AErlang&type=code&l=Erlang)
>   - [x] I have included a real-world usage sample for all extensions added in this PR:
>     - Sample source(s):
>       - [`otp/bootstrap/lib/compiler/ebin/compiler.app`](https://github.com/erlang/otp/blob/master/bootstrap/lib/compiler/ebin/compiler.app)
>       - [`otp/bootstrap/lib/kernel/ebin/kernel.app`](https://github.com/erlang/otp/blob/master/bootstrap/lib/kernel/ebin/kernel.app)
>     - Sample license(s):
>       - Apache version 2
>   - [x] I have included a change to the heuristics to distinguish my language from others using the same extension

Issue #6293, opened 2023-02-17
better syntax highlighting for console snippets


mcandre:
## Describe the enhancement

The current console language highlighter has really strange tokenization, for example when used inside of a README.md (GitHub flavored Markdown) document.

The highlighter really seems to highlight text at random. It's so strange and distracting, that even no colorization would be easier to read.

### New tokenization scheme

* 99% of text: A boring, consistent foreground color.
* Standard prompts: A bare dollar (`$`) or greater than (`>`) prompt leading the line. We could extend this very slightly to handle UNIX and NTFS file paths preceding these characters. But beyond that complexity, anything else in a snippet prompt is likely going to be wasteful, distracting text for the human reader, and difficult to establish a performant parser ROC curve. We can color the prompt classic green, and dim the luminosity, in order to emphasize that the user is not expected to literally type out the prompt when following console snippets.
* Standard line comments: A sharp (`#`) or a double colon (`::`) leading the line. Any obscure comment syntax, or multiline comment syntax, is better suited to placing above or below a snippet (in the case of a documentation), or highlighted with a more specific parser, for script files. Color in the area of brown. Like the prompt, dim to denote its parenthetical nature.

And that's it. Do not attempt to highlight other tokens specially, because they're very likely to be arbitrary output from unknowable programs.

### Don't highlight built-in shell commands like cd

Don't try to highlight shell built-ins specially. Given that built-in commands and other executables will be interleaved in console snippets, all that does is make a distracting, unpredictable view.

And which exact built-ins would we even support exactly? Because ls is a built-in for POSIX sh and PowerShell but not for Command Prompt. where may be a built-in on some systems but not others. find is a built-in for Command Prompt and PowerShell but not UNIX sh. ll may be a distribution-provided alias for ls with certain flags enabled, though not technically a shell built-in, in for troubleshooting purposes.

curl/wget are built-ins for PowerShell but not for Command Prompt or UNIX sh. which is a built-in for UNIX sh but not for Command Prompt. Most things are built-ins in OS distributions based on Busybox/Toybox sh.

Such applies not only to UNIX sh implementations, but note that there are also non-POSIX shells like csh, tcsh, fish, ...

pushd/popd are built-ins for bash and zsh but not for ksh. Some distributions shim pushd/popd on behalf of ksh, mimicking a built-in. For troubleshooting reasons, should we treat these commands as built-ins or non-built-ins? You can see what a headache highlighting built-ins becomes.

[...]

PR #6302, opened 2023-03-01
Add support for Elvish transcripts.

xiaq:
This adds support for highlighting transcripts of Elvish sessions.

## Description

Support for Elvish was accepted in https://github.com/github/linguist/pull/5940. Because Elvish is a shell language, it is very frequent to not show just Elvish code, but also its outputs. Instead of just:

```elvish
echo foo
```

One often needs to show:

```elvish-transcript
~> echo foo
foo
```

These _transcripts_ are most often used inside the Markdown of (for example) issue comments, and rarely saved to files, so there aren't a lot of those files on GitHub today.

However, my primary goal is to make it possible to use the `elvish-transcript` language tag on GitHub, and since `elvish` is already usable as a language tag, I'd argue that it should be supported too. The notability test was passed when Elvish itself was accepted.

I did define a file extension for Elvish transcripts, `.elvts`, but I am fine if I should drop that.

## Checklist:

    - The `.elvts` extension is **not** widely used, see above for explanation.
    - Sample source(s):
      - Examples given for the `echo` command (https://elv.sh/ref/builtin.html#echo, [source](https://github.com/elves/elvish/blob/5c835b1c5f5c9576b0d55f001e354f785202a9cf/pkg/eval/builtin_fn_io.d.elv#L160-L167))
  - [x] I have included a syntax highlighting grammar: https://github.com/elves/elvish, already included


lib/linguist/languages.yml
```diff
@@ -1778,6 +1778,16 @@ Elvish:
...
+Elvish transcript:
+  type: programming
+  ace_mode: text
+  extensions:
+  - ".elvts"
+  interpreters:
+  - elvish-transcript
+  tm_scope: source.elvish-transcript
+  color: "#99FF99"
+  language_id: 452025714
```
lildude, line 1790: This can be simplified:

```suggestion
Elvish Transcript:
  type: programming
  group: Elvish
  ace_mode: text
  tm_scope: source.elvish-transcript
  color: "#55BB55"
  language_id: 452025714
```

Notable changes:

- we don't need an extension as it'll only ever be used in markdown files
- we group it under Elvish as it is essentially the results of processing that language, like `pycon` for Python
- we don't need an interpreter as this is only used for shebang matching which will never happen
- the colour, whilst never likely to be seen, should probably be the same as Elvish

lildude (changes requested):
We really don't need an extension or sample in this case and can simplify the entry (see inline comment). 

Widespread usage is still a consideration which is quite easy to determine as you've stated this will only be used in markdown files or issue comments. 

[This search](https://github.com/search?q=language%3Amarkdown+elvish-transcript&type=code) shows it is only being used in 135 markdown files so falls a little short on the files side.

Searching [issue comments or bodies](https://github.com/search?q=%27elvish-transcript%27+in%3Abody%2Ccomments&type=issues) doesn't seem to indicate it's used that much with only 8 results.

So whilst there is usage, it doesn't appear widely adopted on GitHub yet, though I guess the lack of highlighting might have something to do with that in issues.

Let me sleep on this.

xiaq:
Thanks, applied your suggested change. Also regenerated vendor/README.md and removed the sample.

I wonder though - how does Linguist know that the `elvish-transcript` language tag in Markdown maps to the `source.elvish-transcript` scope? Is it just based on the name of the scope? (Also FWIW, the reason I added an `interpreter` was based on the misunderstanding that it is also used as a language tag.)

> So whilst there is usage, it doesn't appear widely adopted on GitHub yet, though I guess the lack of highlighting might have something to do with that in issues.

I definitely agree that the lack of usage currently is _because_ there's no syntax highlighting yet, so people don't bother.

Looking at the first page of open issues in https://github.com/elves/elvish/issues, a lot of comments containing transcripts, but not using the `elvish-transcript` tag: https://github.com/elves/elvish/issues/1599#issue-1335425183, https://github.com/elves/elvish/issues/1603#issue-1356716036, https://github.com/elves/elvish/issues/1625#issue-1456592731, https://github.com/elves/elvish/issues/1631#issue-1489888756, https://github.com/elves/elvish/issues/1641#issue-1529522473, https://github.com/elves/elvish/issues/1648#issue-1551873274, https://github.com/elves/elvish/issues/1653#issue-1556086571, https://github.com/elves/elvish/issues/1654#issue-1559100573, https://github.com/elves/elvish/issues/1656#issue-1573626194 (the last code block), https://github.com/elves/elvish/issues/1662#issuecomment-1442799510. That is about 1/3 of all Elvish issues that has at least one transcript and could benefit from `elvish-transcript` support. At least in the context of issues the use of Elvish transcripts is already widespread, just not the use of the `elvish-transcript` tag.

lildude:
> I wonder though - how does Linguist know that the `elvish-transcript` language tag in Markdown maps to the `source.elvish-transcript` scope? Is it just based on the name of the scope?

Linguist doesn't directly handle the codeblocks, [markup](https://github.com/github/markup) does, but markup uses the languages and grammars supplied by Linguist. 

I gave a brief run down of how this works just yesterday at https://github.com/github/linguist/issues/6293#issuecomment-1451627502 but in short it knows because users tell it and thanks to the entry in the `languages.yml` file it knows to tell the syntax highlighting engine to use the grammar which has the `source.elvish-transcript` scope to highlight the blocks:

_"The file/codeblock has been identified as 'Elvish Transcript', so Mr Syntax Highlighting Engine, please use the `source.elvish-transcript` grammar to make it pretty"_ 😁

lildude (approved):
Lets go for it.

PR #6319, opened 2023-03-13
Update WDL's description

aofarrel:
* ~Fix name to show [Workflow Description Language](https://openwdl.org/) rather than just wdl to put it in line with other languages (compare cwl/Common Workflow Language)~ See comments below. This does raise the question if CWL should be changed since it's usually called CWL in my experience, but I'll leave that to people more familiar with that language to make that call.
* Add two interpreters
  * [miniwdl](https://github.com/chanzuckerberg/miniwdl)
  * [Cromwell](https://github.com/broadinstitute/cromwell) 
* Moved it to maintain alphabetical ordering 

WDL wasn't misclassified per say, so none of the checklists seem to apply?


lib/linguist/languages.yml
```diff
@@ -8077,6 +8069,19 @@ wisp:
...
+Workflow Description Language:
+  aliases:
+  - wdl
+  type: programming
+  color: "#42f1f4"
+  extensions:
+  - ".wdl"
+  interpreters:
+  - miniwdl
+  - Cromwell
```
lildude, line 8081: Are these really shell interpreters, ie commonly used in shebang lines? From a quick search for the `.wdl` extension, I can't find either of these in shebang lines.

```suggestion
```

lildude (changes requested):
> Fix name to show [Workflow Description Language](https://openwdl.org/) rather than just wdl to put it in line with other languages (compare cwl/Common Workflow Language)

Is this language really more commonly referred to as "Workflow Description Language" in full than the "wdl/WDL" acronym? 

We try to use the most commonly use term/name for languages. For example we use XML, HTML, CSS etc... as these are more commonly referred to by their acronyms than their full names and from a quick look of https://openwdl.org/ even they seem to refer to it more as wdl/WDL... the big fat logo at the top of the page is even "wdl".

If we were to go with this change, you'll also need to rename the samples directory.

lib/linguist/languages.yml
aofarrel: My understanding is that they are interpreters, but shebangs are almost never used in workflow languages. In fact, strictly speaking, the WDL spec seems to imply they should not be used, [as the first line of the file should instead be used to declare what version of WDL is being used](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#versioning).

In other words, you can do this, but no one ever does:
```
#! miniwdl run

version 1.0 

[insert valid WDL 1.0 here]
```

aofarrel:
> > Fix name to show [Workflow Description Language](https://openwdl.org/) rather than just wdl to put it in line with other languages (compare cwl/Common Workflow Language)
> 
> Is this language really more commonly referred to as "Workflow Description Language" in full than the "wdl/WDL" acronym?

This is a fair point! I was extrapolating from how this repo handles Common Workflow Language, usually shortened to CWL but showing as Common Workflow Language here. WDL is almost always referred to as WDL, so I'll revert that part.

lib/linguist/languages.yml
lildude: Yeah, this is a bit unclear in the comments. Neither linguist nor GitHub will be using these to actually interpret the files. These are used to help the shebang strategy when identifying the language [here](https://github.com/github/linguist/blob/9358f090deb76511b613ab024b704c1c0f13a4c2/lib/linguist/shebang.rb#L20) so extensionless files, common to interpreted languages like shell scripts, can be correctly identified.

Alhadis: > In other words, you can do this, but no one ever does:
>
>     #! miniwdl run

Slightly off-topic, but a portable interpreter directive should use [`env(1)`](https://man.openbsd.org/env.1) to resolve the interpreter (in this case, `miniwdl`). As for the version-on-first-line requirement… well, a lot of languages mandate similar constructs, but make exceptions for hashbangs and/or modelines. So I wouldn't take the implied limitation too literally.

lib/linguist/languages.yml
```diff
@@ -8060,11 +8060,16 @@ sed:
   ace_mode: text
   tm_scope: source.sed
   language_id: 847830017
-wdl:
+WDL:
```
Alhadis, line 8063: You'll need to move this under `Vyper`, as language names are alphabetised case-sensitively (i.e., by code point). The conversion to uppercase (which is correct, BTW) invalidates the entry's current position in `languages.yml`.

Alhadis (changes requested):

lib/linguist/languages.yml
aofarrel: Since miniwdl is installed by pip and Cromwell is a jar file, I'm actually not quite sure how to make them more portable.

lildude: I don't think it really matters in the context of this PR. These interpreters are not used in shebang lines so these don't belong in the `languages.yml` file for this language. Please remove them as requested earlier.

Alhadis (approved):

lib/linguist/languages.yml
```diff
@@ -7317,6 +7317,19 @@ Wavefront Object:
...
+  aliases:
+  - Workflow Description Language
+  type: programming
+  color: "#42f1f4"
+  extensions:
+  - ".wdl"
+  interpreters:
+  - miniwdl
+  - Cromwell
```
lildude, line 7329:
```suggestion
```

Suggesting again.

lildude (changes requested):
You're also going to need to address the test failures, namely changing the case of the directory that hold the samples and fixing the order in the README.

lildude (approved):

PR #5166, opened 2021-01-23
Add support for the BASIC family of programming languages

telnet23:
## Description

As I reported in #5156, many BASIC programs have been incorrectly labeled as VBA (or Visual Basic .NET if they have not been updated since #4725). There are numerous dialects of [BASIC](https://en.wikipedia.org/wiki/BASIC) (e.g. AppleSoft, Commodore, GW-BASIC), but most dialects share similar syntax and reserved words, and all dialects are collectively known as BASIC. At the very least, a file with the extension `.bas` and line numbering should be labeled as BASIC rather than Visual Basic .NET or VBA. Therefore, I have added a BASIC language along with an appropriate grammar and sample, and I have added a heuristic to check for line numbering.

## Checklist

- [x] **I am adding a new language**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com
    * https://github.com/search?q=10+20+30+GOTO+extension%3Abas&type=Code
  - [x] I have included a real-world usage sample for all extensions added in this PR
    * https://github.com/telnet23/bas/blob/master/mandelbrot.bas
  - [x]  I have included a syntax highlighting grammar
    * https://github.com/telnet23/language-basic
  - [x] I have updated the heuristics to distinguish my language from others using the same extension
    * Added a heuristic to check for line numbering

PR #6320, opened 2023-03-14
Edit BASIC heuristic to avoid false positives + new sample

DecimalTurn:
There is currently at least 2 occurences of false positives with the BASIC heuristic (`'^\s*\d+'`) introduced in https://github.com/github/linguist/pull/5166/:

1) If a VBA file contains line numbers, it will automatically be identified as BASIC. It is not that common for VBA code to have line numbers, but some people like to use them to improve error reporting.
eg. https://github.com/Sven-Bo/Integrate-ChatGPT-in-Excel-using-VBA/blob/master/mChatGPT.bas#L34

2) In VBA (or VB6), an underscore can be used as a line continuation character which means that lines can be split and there is the possibility that a number ends up being the first thing on the second line thus matching with the heuristic.
eg. https://github.com/dragokas/hijackthis/blob/7560e67b5e0d14f53ea25bbfb88587cbfadca1c8/src/modPermissions.bas#L905

The good thing is that both of these issues can't happen at the start of the file because you can't have line numbering outside of a procedure in VBA and it can't be a line continuation if it's the start of the file (obviously).

Hence, I'm proposing to add `\A` at the start of the heuristic (which gives `^\A\s*\d+`) and I also added a sample so that we have two samples for BASIC with the `.bas` extension.

## Checklist:
    - Sample source(s):
      - [GPIOGW.BAS](https://github.com/fdivitto/FabGL/blob/61eeaf90715fa2070df86c9342ec3c189c3ad1e0/Extras/PC%20Emulator%20BASIC%20Programs/GPIOGW.BAS
) 
    - Sample license(s): [GPL v3](https://github.com/fdivitto/FabGL/blob/master/LICENSE)

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -87,7 +87,7 @@ disambiguations:
   - language: FreeBasic
     pattern: '^[ \t]*#(?:define|endif|endmacro|ifn?def|if|include|lang|macro)\s'
   - language: BASIC
-    pattern: '^\s*\d+'
+    pattern: '^\A\s*\d+'
 - extensions: ['.bb']
   rules:
   - language: BlitzBasic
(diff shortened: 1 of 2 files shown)


lildude (approved):
Good explanation and makes sense to me. Thanks.

PR #6322, opened 2023-03-15
Use ghcr.io for grammar compiler docker container

lildude:
## Description

Docker has release a very confusing message about the sunsetting of Free Team organizations. We use such a team organisation for storing the grammar compiler docker image and our choice to use this predates the introduction of the GitHub Packages registry.

It's not 100% sure how we'd be affected by this change, but suffice to say it's the motivator we need to stop using it. We don't really need to be using dockerhub, so this PR switches to using using ghcr.io.

This change will be hidden to most users.

[ Checklist doesn't apply ]


Alhadis:
What's the reason for using Docker anyway? Wouldn't it have been easier to implement the grammar compiler in Ruby?

(I might be asking a dumb question… it's been a while since I've gone near that part of Linguist's codebase…)

PR #6334, opened 2023-03-21
Add `proto` alias for protobuf

dimitropoulos:
proto is the extension for protobuf files, and is used in [other](https://github.com/shikijs/shiki/blob/44b9b0c458744267d824a8283eac4d6393f65013/packages/shiki/src/languages.ts#L104) syntax highlighters as the language id.

## Description

Just adding the alias so that the `proto` language id can be used for code fences.

## Checklist:

None of the below checklist items seem to apply

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5282,6 +5282,7 @@ Propeller Spin:
 Protocol Buffer:
   type: data
   aliases:
+  - proto
   - protobuf
   - Protocol Buffers
   extensions:


lildude (approved):
LGTM. Thanks.

dimitropoulos:
@Alhadis I don't think so, but I'm glad you mention it.

For one thing, many other highlighters already have the `proto` alias (including `rouge`, `pigments`, `chroma`, `lowlight`, `refractor`, `prism`, `highlight.js`).  For another, the extension for Cap'n Proto is `capnp` anyway.  Of those that also have Cap'n Proto highlighting, they seem to use `capnp`.

Alhadis:
@dimitropoulos Fair enough. Well, that answers my question nicely, then. 👍

Alhadis (approved):

PR #6339, opened 2023-03-22
Add MDX language

wooorm:
## Description

MDX is a literate programming language that combines markdown with several JavaScript language features, such as JSX, expressions, and ESM.

*   markdown: <https://commonmark.org/>
*   JavaScript: <https://tc39.es/ecma262/>
*   JSX: <https://facebook.github.io/jsx/>

Previously, there was some temporary duct-tape in place, to map `.mdx` extensions to markdown: GH-4416.
This improved the situation from nothing to something, but has issues, because the syntax that MDX brings wasn’t highlighted.

With this grammar, those can be properly highlighted.

Related-to: GH-4416.

## Checklist:

- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?q=path%3A*.mdx+NOT+nothack&type=code
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://wooorm.com/markdown-tm-language/, based on examples from tests of the MDX parser, and on https://mdxjs.com
    - Sample license(s): ?
  - [x] I have included a syntax highlighting grammar: https://github.com/wooorm/markdown-tm-language
      
  - [x] I have added a color
    - Hex value: `#fcb32c`
    - Rationale: color from the MDX logo :) https://mdxjs.com
  - [ ] ~~I have updated the heuristics to distinguish my language from others using the same extension.~~ (not needed?)

---

P.S. the URL in the issue template doesn’t work with the new GH search:

```
https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AMDX+KEYWORDS+NOT+nothack
```

PR #6341, opened 2023-03-23
Replace markdown grammar with maintained version

wooorm:
## Description

This replaces the grammar used to highlight markdown with a maintained one.

I’ll add more sample files later. First wanted to bring this to your attention.

Related to: GH-6339.

## Checklist:

  - Old: https://github.com/atom/language-gfm
  - New: https://github.com/wooorm/markdown-tm-language

---

Important: the current scope is `source.gfm`. The new scope is `text.md`.
There seems to be some Go code to handle `text.html.markdown` and `text.md`, which I changed.
The reasons why I prefer changing `source.gfm` to `text.md` is that:
a) markdown is closer to a text, markup, language, than to a source language that is “evaluated” or so
b) markdown isn’t a superset of HTML, it has its own complex rules for that, so `text.html.markdown` isn’t great IMO
c) Since it was (probably) added, GFM now bases itself on CommonMark (CM), bringing it much closer to normal markdown (for example, fenced code blocks are now “normal” markdown). This grammar, other than GFM features, also includes some other extensions (frontmatter, GitHub-specific features, directives, math), which aren’t in GFM

But, I wanted to bring this to attention because:
a) it’s still early days, I can be convinced to use something else
b) `source.gfm` is used in other existing grammars.

Notably, I see these grammars that use `source.gfm`:

*   `jawee/language-blade` (not updated in 2 years)
*   `atom/language-ruby` (archived)
*   `language-viml` (@Alhadis)
*   `wooorm/markdown-tm-language` (that’s me, I can do that!)

I imagine @Alhadis to be able to update this, and I can raise an issue with `jawee/language-blade`, although it’s not actively maintained, so I’m not sure that can be solved.


Alhadis:
> I prefer changing `source.gfm` to `text.md` is that:
> a) markdown is closer to a text, markup, language, than to a source language that is “evaluated” or so

Correct. A language's `type` field is a good indicator of what prefix a TextMate grammar's scope should use: `programming` and `data`-type languages should use `source.*`-scoped grammars, while `markup` and `prose`-type languages should use `text.*`-scoped grammars instead.

It's unclear why the Atom team chose `source.gfm` for their markdown grammar, but it—along with virtually every other choice of scope-names they used—is an objectively poor choice of scopes that conflicts with established [naming conventions](https://macromates.com/manual/en/language_grammars#naming_conventions).

>  a) it’s still early days, I can be convinced to use something else

I started a clean rewrite of a markdown grammar several years ago, but never published it (as I was reluctant to share a crappy 2-hour project stub with the world…). I'm sure I'll finish it eventually: I simply wanted to indent code-blocks without screwing up the rest of the document's highlighting:
Figure 1

> b) `source.gfm` is used in other existing grammars.

Atom packages are likely to reference this scope for compatibility with the editor's default Markdown package (`source.gfm`), while others may support `source.gfm` and `text.md` in tandem, whichever the user has available.

> Notably, I see these grammars that use `source.gfm`:
> […]
> * `language-viml` (@Alhadis)

Only in VimL's [`injectionSelector`](https://github.com/Alhadis/language-viml/blob/5030985ab9f5a84a3a18d999d7c3a82f581b0283/grammars/viml.cson#L3) field, which does nothing on GitHub because injections only affect grammars bundled in the same package/submodule (a wise restriction, IMHO). Ergo, it's use in `language-viml` is benign and can be ignored. 👍

lildude (approved):

PR #6351, opened 2023-03-31
Add Nim filter grammar

AmjadHD:
## Description

Adds Support for Nim's [Source Code Filters](https://nim-lang.github.io/Nim/filters.html), the grammar was included in [grammars.yml](https://github.com/github/linguist/blob/79c4aa4706e08f9eb36e46535f0e909f51f3b74d/grammars.yml#L66) but the extension wasn't included in [languages.yml](https://github.com/github/linguist/blob/79c4aa4706e08f9eb36e46535f0e909f51f3b74d/lib/linguist/languages.yml#L4452-L4457) nor was a sample file.
## Checklist:

      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Animf+stdtmpl+NOT+nothack
    - Sample source(s):
      - [src](https://nim-lang.github.io/Nim/filters.html#available-filters-stdtmpl-filter)
    - Sample license(s):
      - I guess it's Nim's [license](https://github.com/nim-lang/Nim#license).
  - [x] I have included a syntax highlighting grammar: [Nimlime](https://github.com/nim-lang/Nimlime).
      
  - [ ] I have added a color
    - Hex value: `#RRGGBB`
    - Rationale: 

### DIFF
--- a/grammars.yml
+++ b/grammars.yml
@@ -63,7 +63,7 @@ vendor/grammars/NSIS:
 - source.nsis
 vendor/grammars/NimLime:
 - source.nim
-- source.nim_filter
+- source.nimscf
 - source.nimcfg
 vendor/grammars/NovaGrammars:
 - source.git-revlist
--- a/grammars.yml
+++ b/grammars.yml
@@ -63,7 +63,7 @@ vendor/grammars/NSIS:
 - source.nsis
 vendor/grammars/NimLime:
 - source.nim
-- source.nimscf
+- source.nim_filter
 - source.nimcfg
 vendor/grammars/NovaGrammars:
 - source.git-revlist
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4455,6 +4455,7 @@ Nim:
   - ".nimble"
   - ".nimrod"
   - ".nims"
+  - ".nimscf"
   filenames:
   - nim.cfg
   ace_mode: text
--- a/samples/Nim/bar.nimf
+++ b/samples/Nim/bar.nimf
@@ -1,0 +1,23 @@
+#? stdtmpl | standard
+#proc generateHTMLPage(title, currentTab, content: string,
+#                      tabs: openArray[string]): string =
+#  result = ""
+<head><title>$title</title></head>
+<body>
+  <div id="menu">
+    <ul>
+  #for tab in items(tabs):
+    #if currentTab == tab:
+    <li><a id="selected"
+    #else:
+    <li><a
+    #end if
+    href="${tab}.html">$tab</a></li>
+  #end for
+    </ul>
+  </div>
+  <div id="content">
+    $content
+    A dollar: $$.
+  </div>
+</body>


grammars.yml
```diff
@@ -63,7 +63,7 @@ vendor/grammars/NSIS:
 - source.nsis
 vendor/grammars/NimLime:
 - source.nim
-- source.nim_filter
+- source.nimscf
```
lildude, line 66: Don't manually change this file. It's automatically updated based on the files in each grammar.

lildude (changes requested):
> … extension wasn't included in [languages.yml](https://github.com/github/linguist/blob/79c4aa4706e08f9eb36e46535f0e909f51f3b74d/lib/linguist/languages.yml#L4452-L4457) nor was a sample file.

And it won't be any time soon 😉; the number of files with that extension is a long way from meeting it usage requirements, even the temporally relaxed requirements.

Users can use an override in the mean time.

AmjadHD:
> Users can use an override in the mean time.

@lildude, Can you explain ?

lildude:
> > Users can use an override in the mean time.
> 
> @lildude, Can you explain ?

Yup… https://github.com/github/linguist/blob/master/docs/overrides.md

AmjadHD:
I added 
```gitattributes
*.nimf linguist-language=Nim
```
to my Nim fork
https://github.com/AmjadHD/Nim/blob/a92a19efa6e8dc273e2133e9b992126bdd057445/.gitattributes#L10
Now `.nimf` files are highlighted as Nim, whereas they should be highlighted with their own grammar included in the same package https://github.com/nim-lang/NimLime/blob/c733fb3e173c3342f0aad8794a7e04cb99ef5d7e/Syntaxes/Nim_Filter.tmLanguage.

grammars.yml
AmjadHD: done.

lildude:
> Now .nimf files are highlighted as Nim, whereas they should be highlighted with their own grammar included in the same package

In this case, it means the Nim grammar doesn't source this other grammar and your PR won't help.

If this file uses a completely different grammar, it needs to be its own language entry or the current grammar needs to source the other grammar and handle this file type too.

Linguist includes all grammar files in a repo to allow the grammar to be spilt into multiple files and to allow other grammars to reuse parts of other grammars.

lildude:
> https://github.com/nim-lang/NimLime/blob/c733fb3e173c3342f0aad8794a7e04cb99ef5d7e/Syntaxes/Nim_Filter.tmLanguage

This doesn't exist in the master branch of the grammar repo so wouldn't be available to Linguist.

AmjadHD:
I'm not sure I understand, The grammar is included in the same package, Is there no way to highlight `.nimf` files with their correct grammar ?

lildude:
> I'm not sure I understand, The grammar is included in the same package, Is there no way to highlight `.nimf` files with their correct grammar ?

No. The `source.nim` grammar will be applied to all Nim files because of this line:

https://github.com/github/linguist/blob/79c4aa4706e08f9eb36e46535f0e909f51f3b74d/lib/linguist/languages.yml#L4461

The only way this could highlight using the separate grammar is if `source.nim` sourced the grammar or if there is a specific language entry which specifies the grammar like ☝️ .

As I said before, Linguist will pull in all grammar files in a repo because the grammar itself may be sourcing them and other grammars might be too. For example, many grammars source `source.c` or `source.javascript` without actually including the C/JavaScript grammars in their repos.

Alhadis:
> or `source.javascript`

`source.js`, actually. 😁

AmjadHD:
`.nim.cfg` files are highlighted with their own grammar. I can't find `tm_scope: source.nimcfg` nor it's own separate language entry.
What does "sourcing a grammar" mean ?

lildude:
> `.nim.cfg` files are highlighted with their own grammar. I can't find `tm_scope: source.nimcfg` nor it's own separate language entry.

The `source.nim` grammar as it stood the last time the grammars were updated will be used to hightlight these files:

https://github.com/github/linguist/blob/5a0c74277548122267d84283910abd5e0b89380e/lib/linguist/languages.yml#L4449-L4462

> What does "sourcing a grammar" mean ?

Whoops. That's a little vague. Maybe I should have used "including". It means including the definitions from one grammar in another.

For example, many grammars pull in the `source.js` grammar by using something similar to this within their grammar rather than redefining all the same patterns:

```
'patterns': [
  {
    'include': 'source.js'
  }
]
```

PR #6355, opened 2023-04-02
Add `.bas` extension to VB6 and heuristics for VB6/VBA

DecimalTurn:
This PR: 
- Adds the extension `.bas` as an extension for VB6
- Adds samples for VB6 and VBA to help with disambiguation
- Implements heuristics for VBA and VB6 with respect to the `.bas` extension

## Description
This is the 3rd instalment of a 3 part series related to VB6 (and VBA).
For more context:
Part 1: [Add Visual Basic 6.0 as its own language](https://github.com/github/linguist/pull/6124)
Part 2: [Add .frm and .cls extensions to VB6](https://github.com/github/linguist/pull/6155)

The reason why I've waited a bit to add the `.bas` extension was because it is more tricky than for the extensions in part 2. Indeed, `.bas` is used for a module file and the attributes at the top of those files are identical between VBA and VB6.
ie.:
`Attribute VB_Name = "ClassOrModuleName"`
So, we can't use it to separate the two, but we can at least use it to distinguish VBA/VB6 from other languages using the `.bas` file extension (see new heuristic with named pattern `vb-module`).

Hence, the general approach that I'm proposing is the same idea as what was used for the `.h` extension as suggested by @lildude:

> This whole VB-family of languages classification issue sounds like it's going to be a lot like the .h extension used by the C-family of languages which was a long battle until we finally resolved on the strategy in use and detailed in the [troublshooting doc](https://github.com/github/linguist/blob/master/docs/troubleshooting.md?rgh-link-date=2022-10-31T09%3A53%3A35Z#my-ccobjective-c-h-header-file-is-detected-as-the-wrong-language). If a similar approach can be found for these VB languages [and] their shared extensions, then that would be great. If not, the only option is manual overrides.

From https://github.com/github/linguist/issues/5824#issuecomment-1296850510

This comment refers to part of the `troubleshooting.md` file reproduced here with the appropriate substitution for the case of `.bas` : 

> Correctly detecting the language for the [Visual Basic .bas module] files is tough because Linguist detects the languages of files in isolation when analysing repositories and these [module] files, especially the smaller ones, can be used across [the 2] languages without using any language-specific content. To try and reduce the number of false positives and keep some degree of predicatability (sic) for users, Linguist will assume all [.bas module files with a VB_NAME Attribute] are [VB6] by default and will only identify a file as [VBA] if the content matches the specific heuristic for that language. This will mean you will need to implement an [override](https://github.com/github/linguist/blob/master/docs/overrides.md) for some of your [module] files if you wish for them to be classified as [VBA] if they do no contain language-specific content.

**Difference between VBA and VB6**
Alright, now let's get to the meat of it: what are the identifiable differences between VBA and VB6?

1. [VBA7 has new 64-bit features that were not implemented in VB6](https://learn.microsoft.com/en-us/office/vba/language/concepts/getting-started/64-bit-visual-basic-for-applications-overview)
2. VBA has some unique top module declarations
	a. [Option Private Module](https://learn.microsoft.com/en-us/office/vba/language/reference/user-interface-help/option-private-statement#:~:text=Option%20Private%20is,in%20Visual%20Basic.)
	b. [Option Compare Database](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-basic-6/aa266181(v=vs.60)#:~:text=OptionCompareDatabase%20can%20only%20be%20used%20within%20Microsoft%20Access)
3. VBA has preloaded libraries with objects that are common in VBA code

These 3 sections appear in the `vba` named pattern at the bottom of the heuristic file. I don't think there is a need for debate on (1) and (2) as those differences between VBA and VB6 are directly based on syntax and are well documented. However, I could see some disagreement regarding section (3) as there is a bit more of a grey area there.

Regarding the choices made for (3), I've included the invocation of a VBA library. This means that just having "Word" in the code is not enough to trigger the regex, you need to invoke a method/property/constant on it like (e.g. `Word.Application`). I've also included multiple common objects that are part of the DLLs VBA has natively access to.

**Potential issues with section 3**
The issue with this is that some VB6 code can reference the DLLs in question and then be "wrongly" labelled as VBA. I'm using quotes around "wrongly" because a case could be made that if you are using VB to automate an office application, you are basically writing VBA code at this point 😅.
	
Anyhow, I'm open to tweaking the list of objects included in case there is too many false positives, but since this approach gives VB6 as the default outcome for vb-like modules, it seems fair to give some extra chances to VBA.

@fafalone, @XusinboyBekchanov, @Gagniuc and @zspitz, I would be interested to have your opinion on this since you've been involved in discussions regarding this topic in the past. Cheers!

## Checklist:

- [x] **I am adding a new extension to a language. (VB6)**
      
      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3Abas+public%20sub+NOT+nothack
    - Sample source(s):
      - https://raw.githubusercontent.com/wqweto/VbAsyncSocket/master/test/Basic/Module1.bas
Licence: [MIT](https://github.com/wqweto/VbAsyncSocket/blob/master/LICENSE)
      - https://raw.githubusercontent.com/dragokas/hijackthis/7560e67b5e0d14f53ea25bbfb88587cbfadca1c8/src/tools/RegTLib/Module1.bas (renamed to Module 2)
License: [GPL2](https://github.com/dragokas/hijackthis/blob/devel/LICENSE.md)
      - https://github.com/ImortisInglorian/fbrtLib/blob/ecfda75a82ad928220dea28f7d477787f89e7456/array_clearobj.bas
License: [GNU Lesser General Public](https://github.com/ImortisInglorian/fbrtLib/blob/master/license.txt)
      - [freebasic/fbc@e43074e/src/compiler/ir.bas](https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/src/compiler/ir.bas)
License: GNU GPLv2
      - https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/contrib/nsis-installer/makescript.bas
License: GNU GPLv2

- [x] **I am fixing a misclassified language (VBA)**
    - Sample source(s):
      - https://raw.githubusercontent.com/hilkoc/vbaDeveloper/master/src/vbaDeveloper.xlam/ErrorHandling.bas
License: [MIT](https://github.com/hilkoc/vbaDeveloper/blob/master/LICENSE)


DecimalTurn:
Looks like with the new samples, the FreeBasic samples are now detected as VB6:

> FreeBasic/Plasma Generation.bas BAD (Visual Basic 6.0)
> FreeBasic/try_catch_throw.bas BAD (Visual Basic 6.0)

@XusinboyBekchanov, are there any good samples you'd suggest?

zspitz:
I would argue that the situation around VBA/VB6 is more akin to JavaScript. The language is the same language regardless of runtime environment (browser, Node.js) or loaded libraries and global objects.

My original intention was to use VBA/VB6 as the language name, but [Linguist doesn't like slashes in language names](https://github.com/github/linguist/pull/4725/files/3cb595f4a40217c831bb472cea1db207c7f6eefd#r350686432) so I went with VBA.

AFAICT the only argument for VB6 as a separate language is that there are some file extensions and file formats unique to VB6.

Alhadis:
> My original intention was to use VBA/VB6 as the language name

That strikes me as a combination of two equally-valid names for the same technology. In which case, either `VBA` or `VB6` should be added as an alias instead. Otherwise, if you *really* must include a slash, then this is something that the [`fs_name` field](https://github.com/github/linguist/blob/79c4aa4706e08f9eb36e46535f0e909f51f3b74d/lib/linguist/languages.yml#L3-L5) should (ideally) accommodate.

zspitz:
> Early versions of Internet Explorer actually _did_ support VBA (or something with "Visual Basic" in its name)

That would be VBScript, which (I would argue) is indeed a separate language. Most notably, because -- like JavaScript -- it has no notion of compile-time types.

Alhadis:
> like JavaScript -- it has no notion of compile-time types.

[Static typing is overrated anyway](http://www.jsfuck.com/).

DecimalTurn:
VBA and VB6 do have a different syntax when it comes to handling 64-bit and there is also module declarations that you won't find in VB6. 

This means that unlike `javascript`, you can't just copy-paste code across and add some librairies and expect the code to work. It won't compile in VB6.

From the PR description:
> **Difference between VBA and VB6**
> Alright, now let's get to the meat of it: what are the identifiable differences between VBA and VB6?
> 
> [VBA7 has new 64-bit features that were not implemented in VB6](https://learn.microsoft.com/en-us/office/vba/language/concepts/getting-started/64-bit-visual-basic-for-applications-overview)
> VBA has some unique top module declarations
> a. [Option Private Module](https://learn.microsoft.com/en-us/office/vba/language/reference/user-interface-help/option-private-statement#:~:text=Option%20Private%20is,in%20Visual%20Basic.)
> b. [Option Compare Database](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-basic-6/aa266181(v=vs.60)#:~:text=OptionCompareDatabase%20can%20only%20be%20used%20within%20Microsoft%20Access)

lib/linguist/heuristics.yml
```diff
@@ -840,3 +846,20 @@ named_patterns:
...
+  - '^\s*Option\sPrivate\sModule\b'
+  - '^\s*Option\sCompare\sDatabase\b'
+  # Libraries/Objects preloaded with VBA
+  - '(?:\s|\()(?:VBA|Access|Excel|Outlook|PowerPoint|Visio|Word)\.\w+' # VBA Libraries invocation
+  - '\b(?:ThisDrawing|AcadObject)\b' # AutoCAD Objects
+  - '\b(?:ThisWorkbook|ActiveWorkbook|ActiveSheet|ActiveChart|ActiveCell|WorksheetFunction)\b' # Excel Objects
+  - '\bRange\("(?:[a-zA-Z]+\d+|[a-zA-Z]+\d+:[a-zA-Z]+\d+|[a-zA-Z]+:[a-zA-Z]+|\d+:\d+)"\)' # Excel range
+  - '\b(?:ActiveExplorer|ActiveInspector)\b' # Outlook Objects
+  - '\b(?:ActiveWindow\.Presentation|ActivePresentation)\b' # PowerPoint Objects
+  - '\b(?:ActiveDocument|Selection\.Find|Selection\.Paragraphs)\b' # Word Objects
```
Alhadis, line 865: We can combine most of these patterns, including their comments:

```suggestion
  # Objects used in AutoCAD, Outlook, PowerPoint, and Word
  - '\b(?:AcadObject|Active(?:Cell|Chart|Document|Explorer|Inspector|Presentation|Sheet|Window\.Presentation|Workbook)|Selection\.(?:Find|Paragraphs)|This(?:Drawing|Workbook)|WorksheetFunction)\b'
  # Excel range
  - '\bRange\("(?:[A-Za-z]+\d+(?::[A-Za-z]+\d+)?|[A-Za-z]+:[A-Za-z]+|[A-Za-z]*\d+(:\d+)?)"\)'
```

XusinboyBekchanov:
> Looks like with the new samples, the FreeBasic samples are now detected as VB6:
> 
> > FreeBasic/Plasma Generation.bas BAD (Visual Basic 6.0)
> > FreeBasic/try_catch_throw.bas BAD (Visual Basic 6.0)
> 
> @XusinboyBekchanov, are there any good samples you'd suggest to add for FreeBasic?

Plasma Generation.bas uses a unique keyword for FreeBasic Shared: https://freebasic.net/wiki/KeyPgShared

The try_catch_throw.bas uses the #include keyword as specified in the heuristic: https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml#L116

Alhadis (changes requested):
I had a brief flashback to high school accounting classes (my only experience with Microsoft Excel) and recall formulae being case-insensitive. A quick Google search suggests Visual Basic—or one of its confuseables[^1]—[is case-insensitive](https://learn.microsoft.com/en-us/dotnet/visual-basic/misc/bc40055).

[^1]: These technologies wouldn't be Microsoft if they weren't given confusing, forgettable and/or easily-rebranded names. Or in the case of Excel's Turing-complete formula functions, [no programming language name whatsoever](https://www.infoq.com/articles/excel-lambda-turing-complete/).

XusinboyBekchanov:
> I had a brief flashback to high school accounting classes (my only experience with Microsoft Excel) and recall formulae being case-insensitive. A quick Google search suggests Visual Basic—or one of its confuseables[1](#user-content-fn-1-7622e9861d85b61b7a33a161bbe7ae69)—[is case-insensitive](https://learn.microsoft.com/en-us/dotnet/visual-basic/misc/bc40055). If so, these regexes should begin with `(?i)` to mark them as case-insensitive matches. I'll leave this up to you to decide; I'm out-of-my-depth when it comes to Microsoft technologies named /Visual\s*(Studio|Basic)/.
> 
> ## Footnotes
> 1. These technologies wouldn't be Microsoft if they weren't given confusing, forgettable and/or easily-rebranded names. Or in the case of Excel's Turing-complete formula functions, [no programming language name whatsoever](https://www.infoq.com/articles/excel-lambda-turing-complete/). [↩](#user-content-fnref-1-7622e9861d85b61b7a33a161bbe7ae69)

This needs to be added for FreeBasic.

XusinboyBekchanov:
For VBA, you also need to add: Attribute VB_Name =

DecimalTurn:
> > Looks like with the new samples, the FreeBasic samples are now detected as VB6:
> > > FreeBasic/Plasma Generation.bas BAD (Visual Basic 6.0)
> > > FreeBasic/try_catch_throw.bas BAD (Visual Basic 6.0)
> > 
> > 
> > @XusinboyBekchanov, are there any good samples you'd suggest to add for FreeBasic?
> 
> Plasma Generation.bas uses a unique keyword for FreeBasic Shared: [freebasic.net/wiki/KeyPgShared](https://freebasic.net/wiki/KeyPgShared)
> 
> The try_catch_throw.bas uses the #include keyword as specified in the heuristic: [`master`/lib/linguist/heuristics.yml#L116](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml?rgh-link-date=2023-04-03T02%3A14%3A38Z#L116)

Thanks for the clarification @XusinboyBekchanov, so my undestanding is that the supplied sample `Plasma Generation.bas` is not really code that can be found "in the wild" but more of an example to illustrate a feature of the language.

What if we added this sample below instead which is longer and more in line with a good sample to illustrate the use of the "shared" keyword?
https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/src/compiler/ir.bas

XusinboyBekchanov:
> > > Looks like with the new samples, the FreeBasic samples are now detected as VB6:
> > > > FreeBasic/Plasma Generation.bas BAD (Visual Basic 6.0)
> > > > FreeBasic/try_catch_throw.bas BAD (Visual Basic 6.0)
> > > 
> > > 
> > > @XusinboyBekchanov, are there any good samples you'd suggest to add for FreeBasic?
> > 
> > 
> > Plasma Generation.bas uses a unique keyword for FreeBasic Shared: [freebasic.net/wiki/KeyPgShared](https://freebasic.net/wiki/KeyPgShared)
> > The try_catch_throw.bas uses the #include keyword as specified in the heuristic: [`master`/lib/linguist/heuristics.yml#L116](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml?rgh-link-date=2023-04-03T02%3A14%3A38Z#L116)
> 
> Thanks for the clarification @XusinboyBekchanov, so my undestanding is that the supplied sample `Plasma Generation.bas` is not really code that can be found "in the wild" but more of an example to illustrate a feature of the language.
> 
> What if we added this sample below instead which is longer and more in line with a good sample to illustrate the use of the "shared" keyword? https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/src/compiler/ir.bas

The one provided in your example has the same #include. In this case, it is determined by this feature.

Alhadis:
Folks, let's try to avoid nested quote-blocks (preferably by quoting only the relevant parts of a comment). It makes this discussion much harder to follow, and makes searching the page for text more cumbersome.

lib/linguist/heuristics.yml
```diff
@@ -840,3 +846,20 @@ named_patterns:
   raku: '^\s*(?:use\s+v6\b|\bmodule\b|\b(?:my\s+)?class\b)'
   vb-class: '^\s*VERSION\s+[0-9]\.[0-9]\s+CLASS'
   vb-form: '^\s*VERSION\s+[0-9]\.[0-9]{2}'
+  vb-module: '^\s*Attribute\s+VB_Name\s+=\s+'
```
Alhadis, line 849: Redundant qualifier that I missed earlier before:

```suggestion
  vb-module: '^\s*Attribute\s+VB_Name\s+=\s'
```

DecimalTurn:
@XusinboyBekchanov, from what I undestand, the main goal of the samples are to help the classifer make a guess when the heuristics aren't conclusive. So, it doesn't really matter if the samples contains features that the heuristics are targetting as long as they represent FreeBasic code "in the wild".

For this reason, I've changed the sample `Plasma Generation.bas` for this sample [freebasic/fbc@e43074e/src/compiler/ir.bas](https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/src/compiler/ir.bas) (License GNU GPLv2).

Alhadis (changes requested):

lib/linguist/heuristics.yml
```diff
@@ -840,3 +846,20 @@ named_patterns:
...
+  vba:
+  # VBA7 new 64-bit features (https://learn.microsoft.com/en-us/office/vba/language/concepts/getting-started/64-bit-visual-basic-for-applications-overview)
```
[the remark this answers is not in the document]
DecimalTurn: No worries, I can remove it.

lib/linguist/heuristics.yml
```diff
@@ -840,3 +846,20 @@ named_patterns:
...
+  - '^\s*(?:Public|Private)?\s+Declare\s+PtrSafe\s+(?:Sub|Function)\b'
+  - '^\s*#If\s(:?VBA7|Win64)\b'
```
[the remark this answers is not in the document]
DecimalTurn: The VBA IDE will always remove they excess spaces as soon as you exit the line, so yes that was intentional to use one space.

lib/linguist/heuristics.yml
```diff
@@ -840,3 +846,20 @@ named_patterns:
...
+  - '^\s*Dim\s[0-9a-zA-Z_]*\sAs\s(?:LongPtr|LongLong)\b'
```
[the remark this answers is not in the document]
DecimalTurn: 1. Again intentional to use only a single space.
2. Sure, why not!

lib/linguist/heuristics.yml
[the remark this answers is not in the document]
Alhadis: In that case, am I right to assume that this whitespace character will always be [<samp>U+0020</samp>](https://graphemica.com/%20) (as opposed to potentially being a tab, form feed, or newline character)?

If so, then `\s` should be replaced with simply ` `.

lib/linguist/heuristics.yml
DecimalTurn: I'm thinking this section could be contentious and subject to future fine-tuning, so I'd keep them seperate for now just to keep things easier to edit.

DecimalTurn:
> For VBA, you also need to add: Attribute VB_Name =
@XusinboyBekchanov  I've added this heuristic in the form of the named pattern "vb-module" as it is both used by VBA and VB6

lib/linguist/heuristics.yml
Alhadis: Fair enough. In that case, I recommend moving the comments (`# … Objects` and friends) to their own lines, above the lines they're documenting, so they won't be missed by users reading the file in a split-pane editor window (or with soft-wrapping disabled).

lib/linguist/heuristics.yml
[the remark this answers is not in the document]
Alhadis: /cc @DecimalTurn, in case you missed this.

DecimalTurn:
> [...] If so, these regexes should begin with (?i) to mark them as case-insensitive matches.

@Alhadis, thanks for letting me know this is possible. While it is true that Excel formulas are case-insensitive, it is not always true for VBA. For instance, the reserved keyword and compiler constants will have a specific capitalization that the VBA IDE will enforce as soon as you exit the line in the editor.



For this reason, I don't think we should worry about the case insensitivity of VBA. The only place it can play a role is with variable names sometimes, but that's won't apply to the heuristics used here.

Alhadis:
> specific capitalization that the VBA IDE will enforce as soon as you exit the line in the editor.

Seriously? (An editor that does that would shit me right up the wall…)

> For this reason, I don't think we should worry about the case insensitivity of VBA

Agreed, provided there's no substantial risk of these files being authored in other editing environments. However, @XusinboyBekchanov [mentioned earlier](https://github.com/github/linguist/pull/6355#issuecomment-1493560944) that FreeBASIC's heuristics should be marked case-insensitive (where it makes sense to do so), although that might warrant a separate PR.

lib/linguist/languages.yml
```diff
@@ -7262,6 +7262,7 @@ Visual Basic 6.0:
   type: programming
   color: "#2c6353"
   extensions:
+  - ".bas"
```
lildude, line 7265: Is `.bas` the primary extension for Visual Basic 6.0 files? (The first extension is considered the lanuage's primary extension).

lildude (changes requested):
Please find smaller samples. Hint: if the PR is not displaying the full file content in the diff, it's probably unnecessarily large.

I think we're also going to need another FreeBasic sample or two (depends on how things behave after you replace the large samples) as the classifier is now detecting one of the current FreeBasic samples as Visual Basic 6.0:

```
FreeBasic/try_catch_throw.bas BAD (Visual Basic 6.0)
```

lib/linguist/languages.yml
DecimalTurn: I would say yes. Like most languages derived from BASIC, the `.bas` is the default extension where you can find standalone code. And that's actually the reason why it took longer to include, because it's the same extension as VBA and I had to make sure that good heuristics were put in place to make this work.

DecimalTurn:
> Please find smaller samples. Hint: if the PR is not displaying the full file content in the diff, it's probably unnecessarily large.

@lildude, thanks for your patience on this one. At least I didn't go with a 8000+ lines sample this time, so there's progress 😅.
I'm working on some more adjustements to the heuristics and I'll make the changes for the samples this week.

DecimalTurn:
Classifier cross-validation is still falling for these:
```
BASIC/mandelbrot.bas BAD (Visual Basic 6.0)
FreeBasic/try_catch_throw.bas BAD (Visual Basic 6.0)
```

DecimalTurn:
Adding the following sample for BASIC fixed the issue with `BASIC/mandelbrot.bas`
https://raw.githubusercontent.com/DualBrain/gw-basic/032cbf70bcddd5a971a81d7e56e795e61fe028f0/spacesc.bas
License: [MIT](https://github.com/DualBrain/gw-basic/blob/master/LICENSE)

lib/linguist/heuristics.yml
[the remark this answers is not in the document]
DecimalTurn: I've changed `\s` to  an actual space (` `) when permitted. I've also double-checked in which cases the space can be non-unique.

The only case where there can be more than one space and the VB Editor won't remove it is when the extra-spaces appear before the `as` of a `dim` statements:

```vba
Dim x           As LongPtr
Dim something   As String
```

So, this line is the only one making use of `[ ]*` : 
```diff
- '^\s*Dim [0-9a-zA-Z_]*[ ]*As Long(?:Ptr|Long)\b'
+ '^\s*(?:Dim|Const) [0-9a-zA-Z_]*[ ]*As Long(?:Ptr|Long)\b'
```

lib/linguist/heuristics.yml
DecimalTurn: Ok, I've moved the comments and I've also merged Outlook, PowerPoint and Word objects, but kept CAD objects seperate for now as it seems justified to keep it seperate from other Office apps.

DecimalTurn:
@lildude I'd be interested to know if you have any more suggestions regarding this issue with samples for FreeBasic (FB) and the failing classifier tests. Just to recap : 

**What did the original impact of the PR?**:
- Both samples already present in Linguist for FB started to classify as VB6 (`try_catch_throw.bas` ❌ and `Plasma Generation.bas` ❌). Note that when you made your previous comment, it was right after I had temporarily removed the sample `Plasma Generation.bas` to see if that would help.

**Which samples I've tried to add**:
- `ir.bas` => is correctly classifed as FB ✅
- `Picture.bas` => is classified as VB6 ❌

So, after adding 2 more samples, we only have 1/4 passing the test. What should be done in this case?

XusinboyBekchanov:
You have provided several different heuristics. They have Attribute VB_Name = "ClassOrModuleName" inside them too.
But each provided heuristic must contain this check.

DecimalTurn:
@XusinboyBekchanov, I'm not sure I understand your comment. With the current heuristics a file will need to match the `vb-module` pattern AND any of the heuristics listed inside the `vba` pattern to be labelled as VBA. Hence, the presence of `Attribute VB_Name =` is made mandatory by the use of the `vb-module` pattern for VBA (and VB6).

```
  - language: VBA
    and:
    - named_pattern: vb-module
    - named_pattern: vba
  - language: Visual Basic 6.0
    named_pattern: vb-module
```
```
vb-module: '^\s*Attribute VB_Name = '
```

XusinboyBekchanov:
Maybe OR?
In your case, the FreeBasic examples should not have been marked as VB6.
Then you can combine all heuristics into one heuristic and check.

XusinboyBekchanov:
This and does not work well, in my opinion:
 and:
    - named_pattern: vb-module
    - named_pattern: vba

DecimalTurn:
> In your case, the FreeBasic examples should not have been marked as VB6.
@XusinboyBekchanov, the reason why they have been marked as VB6 has nothing to do with the heuristics. The FreeBasic samples were classified as VB6 during the **Classifier cross-validation test**. This test completely ignores the heuristics and performs a statistical Bayesian analysis on the content of the files to guess the language of the samples.

XusinboyBekchanov:
> Agreed, provided there's no substantial risk of these files being authored in other editing environments. However, @XusinboyBekchanov [mentioned earlier](https://github.com/github/linguist/pull/6355#issuecomment-1493560944) that FreeBASIC's heuristics should be marked case-insensitive (where it makes sense to do so), although that might warrant a separate PR.

I added a separate PR for this:
https://github.com/github/linguist/pull/6367

zapuly76 (approved):

DecimalTurn:
Alright, all tests are now passing 🚀

To avoid problems errors with the samples, I've had to try a lot of sample for FreeBasic and I unfortunatly couldn't keep the 2 that we already there.

Here are the new samples for FreeBasic:
- https://github.com/ImortisInglorian/fbrtLib/blob/ecfda75a82ad928220dea28f7d477787f89e7456/array_clearobj.bas
	- License : [GNU Lesser General Public](https://github.com/ImortisInglorian/fbrtLib/blob/master/license.txt)
- [freebasic/fbc@e43074e/src/compiler/ir.bas](https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/src/compiler/ir.bas)
	- License: GNU GPLv2
- https://github.com/freebasic/fbc/blob/e43074e5aa97a904a31c00d9917bc952308bddc0/contrib/nsis-installer/makescript.bas
	- License : GNU GPLv2

@lildude please let me know if this works.

DecimalTurn:
@Alhadis, I believe that the suggestions you've made were addressed. Can you please have a look and let me know if there is something I missed or something else that could be improved?

Alhadis (approved):

Alhadis:
Hey @DecimalTurn, sorry about the wait. 👋 You've been very patient.

Yup, this LGTM. @lildude, any thoughts?

lildude (approved):
This all looks good to me. Thanks for your effort and patience.

lildude (approved):

PR #6364, opened 2023-04-11
Add global.json to the JSONC files

Porges:
## Description

Specifies that `global.json` allows comments, per [its documentation](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#comments-in-appsettingsjson).

## Checklist:

  - [x] The new extension is used in ~~hundreds~~ tens-of-thousands of repositories on GitHub.com
      - https://github.com/search?type=code&auto_enroll=true&q=path%3A**%2Fglobal.json
      
    - Sample source(s):
      - me
    - Sample license(s):
       - public domain 

    - Sample source(s):
    - Sample license(s):
      
  - [ ] I have added a color
    - Hex value: `#RRGGBB`
    - Rationale: 

    - Sample source(s):
    - Sample license(s):

  - Old: [URL to grammar repo]
  - New: [URL to grammar repo]

    - [URL to public discussion]

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3136,6 +3136,7 @@ JSON with Comments:
   - ".jslintrc"
   - api-extractor.json
   - devcontainer.json
+  - global.json
   - jsconfig.json
   - language-configuration.json
   - tsconfig.json
--- a/samples/JSON with Comments/filenames/global.json
+++ b/samples/JSON with Comments/filenames/global.json
@@ -1,0 +1,7 @@
+{
+    "sdk": {
+        "version": "7.0.100",
+        // allows 7.0.xxx
+        "rollForward": "latestFeature"
+    }
+}


Alhadis:
This name strikes me as far too generic; `global.json` could refer to almost anything.

👎 from me, sorry.

Porges:
Is there a way to add an additional simple heuristic such as if the file contains the JSON key `"sdk"` it is very likely to be the `global.json` for .NET Core?

lildude:
I agree with @Alhadis here. This is waaaaay too generic to start making this JSONC for every instance of this file.

> Is there a way to add an additional simple heuristic such as if the file contains the JSON key `"sdk"` it is very likely to be the `global.json` for .NET Core?

Nope. Heuristics only consider files based on extensions, and I'm not sure it's worth extending the current generic extensions or heurstic implementations for files at this point as I don't think it would be that useful beyond maybe this filename. 

Another possible solution would be for a new strategy to detect JSON files with `//` and mark them as JSONC, like we do with XML, however this could be incredibly expensive as we'd need to scan the entire content of every single JSON file in a repo... and there are some _massive_ JSON files out there.

This is also making the assumption _all_ users will be .NET Core users, which you can't be certain of, so the heuristic would need to be more complex with broaching nigh on perfect matching due to the generic nature of this file.

As such, I think the best thing here is for peeps to use an override to force "JSON with Comments" if they need to use comments in `.json` files and don't like the red highlighting and they can't rename the extension to `.jsonc`. And yes, I know, Microsoft is notorious for abusing JSON for JSONC 😞.

Alhadis:
> Another possible solution would be for a new strategy to detect JSON files with `//` and mark them as JSONC, like we do with XML, however this could be incredibly expensive as we'd need to scan the entire content of every single JSON file in a repo... and there are some *massive* JSON files out there.

This solution suffers from the same issue that prohibits us from using one grammar/filetype for both JSON and JSONC files: it **assumes** that `.json` files containing `// …` lines are actually valid in the context of whatever application reads their contents. Obviously, this isn't something Linguist has any way of knowing, so the lesser evil is to force users to explicitly declare JSONC files that use `.json` extensions for some reason, rather than implicitly assuming `// …` is valid JSON _everywhere_.

Having said that, I'm gonna close this PR in favour of recommending [good ol' fashioned overrides](https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes):

~~~gitattributes
global.json linguist-language=JSONC
~~~

PR #6365, opened 2023-04-11
Fix typos in troubleshooting.md

DecimalTurn:
Fix typos ~~+ Add reference to go-enry's divergences from Linguist.~~

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -21,8 +21,8 @@ If you have not made any changes to your repository in a while, you may find pus
 ## My C/C++/Objective-C `.h` header file is detected as the wrong language
 
 Correctly detecting the language for the C-family `.h` header files is tough because Linguist detects the languages of files in isolation when analysing repositories and these header files, especially the smaller ones, can be used across all three languages without using any language-specific content.
-To try and reduce the number of false positives and keep some degree of predicatability for users, Linguist will assume all `.h` header files are C by default and will only identify a file as C++ or Objective-C if the content matches the specific heuristic for that language.
-This will mean you will need to implement an [override](/docs/overrides.md) for some of your header files if you wish for them to be classified as C++ or Objective-C if they do no contain language-specific content.
+To try and reduce the number of false positives and keep some degree of predictability for users, Linguist will assume all `.h` header files are C by default and will only identify a file as C++ or Objective-C if the content matches the specific heuristic for that language.
+This will mean you will need to implement an [override](/docs/overrides.md) for some of your header files if you wish for them to be classified as C++ or Objective-C if they do not contain language-specific content.
 
 ## My repository isn't showing my language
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -70,4 +70,4 @@ Grammars will only be updated in patch releases if the patch release is specific
 
 Note: New languages will not appear in GitHub's search results for some time after the pull request has been merged and the new Linguist release deployed to GitHub.com.
 This is because GitHub's search uses [go-enry](https://github.com/go-enry/go-enry) for language detection which tends to lag behind Linguist by a few weeks to months.
-This in turn requires an update to the underlying search code once go-enry is inline with Linguist.
+This in turn requires an update to the underlying search code once go-enry is inline with Linguist (See also go-enry's list of [divergences from Linguist](https://github.com/go-enry/go-enry#divergences-from-linguist)).
```
Alhadis, line 73: I'm not sure this is relevant enough to Linguist to warrant inundating readers with the gory details of portability differences between various regular expression engines. Keep in mind, most users perusing this document are probably in a rush, so they're more likely to expect missing details to be included in pages linked to from a troubleshooting answer.

Alhadis (changes requested):

docs/troubleshooting.md
DecimalTurn: Maybe there is too many elements in that list that won't be relevant to most people,  but I think there should be at least a mention that go-enry doesn't support `linguist-language` overrides in the `.gitattrubutes`.

I know that I was confused about this multiple time until I discovered that in the list of divergences.

It's especially weird when you click on a language featured in the language stats bar and code search tells you there is no files with that language in the project :/

Alhadis: > It's especially weird when you click on a language featured in the language stats bar and code search tells you there is no files with that language in the project :/

Wait, so *that's* the culprit behind all those flaky, confusing code-listings seen when clicking certain languages in project stat-bars? 😕 Finally, I now know where years of [“missing” INI files](https://github.com/search?q=repo%3AAlhadis%2F.files++language%3AINI&type=code) went in my dotfiles repository…

> but I think there should be at least a mention that go-enry doesn't support `linguist-language` overrides in the `.gitattributes`

This would be more suited to a separate troubleshooting entry explaining why some files aren't being listed when filtering by language (ideally after the [_“My repository is detected as the wrong language”_](https://github.com/github/linguist/blob/9b53b4b0450194fb7990bda3823d982ccb7bd9b3/docs/troubleshooting.md#my-repository-is-detected-as-the-wrong-language) entry). Such an addition warrants its own PR, however.

In the mean time, we should avoid repeating ourselves; this troubleshooting entry already includes a link to the [`go-enry`](https://github.com/go-enry/go-enry) repository; users can find their own way to the [_“Divergences from Linguist”_](https://github.com/go-enry/go-enry#divergences-from-linguist) section if they so choose to.

Alhadis (approved):

lildude (approved):

PR #6367, opened 2023-04-11
Added case-insensitive matches for FreeBasic.

XusinboyBekchanov:
## Description

Added case-insensitive matches for FreeBasic heuristics.

## Checklist:

- [x] **I am fixing a misclassified language (FreeBasic)**
    - Sample source(s):
    - Sample license(s):

  - Old: [URL to grammar repo]
  - New: [URL to grammar repo]

    - [URL to public discussion]

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,3 @@
 # All pull requests need to be reviewed by someone from GitHub.
-# @github/linguist is a GitHub maintained team.
 #
-*     @github/linguist
+*     @github-linguist/github
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,7 +11,7 @@
   - [ ] The new extension is used in hundreds of repositories on GitHub.com
     - Search results for each extension:
       <!-- Replace FOOBAR with the new extension, and KEYWORDS with keywords unique to the language. Repeat for each extension added. -->
-      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
+      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
   - [ ] I have included a real-world usage sample for all extensions added in this PR:
     - Sample source(s):
       - [URL to each sample source, if applicable]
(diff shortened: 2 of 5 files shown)


DecimalTurn:
After looking a little more into it, I don't think it's a good idea to make the heuristics case-insensitive for FreeBasic mainly for these 2 reasons:
- When I look at [FreeBasic code](https://github.com/search?q=path%3A*.bas+AND+%2F%5E%5B+%5Ct%5D*%23%28%3F%3Adefine%7Cendif%7Cendmacro%7Cifn%3Fdef%7Cif%7Cinclude%7Clang%7Cmacro%29%5Cs%2F+language%3AFreeBasic&type=code&l=FreeBasic) on GitHub, I see that the use of lowercase is prevalent and I actually found no use of uppercase in the first five pages of results.
- VBA and VB6 use `#If` and so the case-insensitive version of the heuristic would match that and cause false positive for FreeBasic.

XusinboyBekchanov:
> Do you have any examples of FreeBasic repos or files missclassified because they were using a different casing convention than the norm?

For example, here is a capital letter:
https://github.com/bob-the-hamster/ohrrpgce/blob/f8cb11c86de7eb1c9c32ae60bb7bd40f131ab97a/miditest.bas

DecimalTurn:
> For example, here is a capital letter:
> [bob-the-hamster/ohrrpgce@f8cb11c/miditest.bas](https://github.com/bob-the-hamster/ohrrpgce/blob/f8cb11c86de7eb1c9c32ae60bb7bd40f131ab97a/miditest.bas)

I see that this file still makes use of lowercase `#include`, so the lowercase heuristics would still work.

However, after looking further, I've found the following file here:
https://github.com/doranchak/azdecrypt/blob/a9dbde3adc451c1aaccb4b4339f81e1ab5dafc0d/FreeBASIC-1.09.0-winlibs-gcc-9.3.0/examples/GUI/GTK%2B/FB_Calc/FB_Calc.bas#L3

It does only include the use of `#INCLUDE` and is indentified by Linguist as v7.25 as `VBA`, so I guess this would be a good case to justify the use of case-insensitive matches.

In that case, I'm in favor of this change.

lib/linguist/heuristics.yml
```diff
@@ -89,7 +89,7 @@ disambiguations:
 - extensions: ['.bas']
   rules:
   - language: FreeBasic
-    pattern: '^[ \t]*#(?:define|endif|endmacro|ifn?def|if|include|lang|macro)\s'
+    pattern: '^[ \t]*#(?i)(?:define|endif|endmacro|ifn?def|include|lang|macro)\s'
```
Alhadis, line 92:
```suggestion
    pattern: '^[ \t]*#(?i)(?:define|endif|endmacro|ifn?def|include|lang|macro)(?:$|\s)'
```

Alhadis (changes requested):
While we're here, we might as well harden these heuristics to deal with files that aren't terminated by a trailing newline (something that's more common on Windows). For example:

~~~ruby
regex = /^[ \t]*#(?i)(?:define|endif|endmacro|ifn?def|include|lang|macro)\s/
"Foo\nBar\n#endif\n" =~ regex # => 8
"Foo\nBar\n#endif"   =~ regex # => nil
~~~

Alhadis (approved):

lildude (approved):
LGTM. Thanks.

XusinboyBekchanov:
Ok

lildude (approved):

PR #6372, opened 2023-04-14
Add Kind language support

aripiprazole:
## Checklist:

      - https://github.com/search?q=path%3A*.kind2&type=code
    - Sample source(s):
      - https://github.com/HigherOrderCO/Wikind
    - Sample license(s):
      - https://github.com/HigherOrderCO/Wikind/blob/master/LICENSE
  - [x] I have included a syntax highlighting grammar: https://github.com/HigherOrderCO/tree-sitter-kind
      
  - [x] I have added a color
    - Hex value: `#2f4aab`
    - Rationale: I choose this color, because of the language's logo background color

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1293,3 +1293,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/tree-sitter-kind"]
+	path = vendor/grammars/tree-sitter-kind
+	url = https://github.com/HigherOrderCO/tree-sitter-kind.git
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3488,6 +3488,15 @@ Kickstart:
   - ".ks"
   tm_scope: source.kickstart
   language_id: 692635484
+Kind:
+  type: programming
+  color: "#329866"
+  extensions:
+  - ".kind"
+  - ".kind2"
+  tm_scope: source.kind
+  ace_mode: text
+  language_id: 356227571
 Kit:
   type: markup
   ace_mode: html
--- a/samples/Kind/Kindelia/io.kind
+++ b/samples/Kind/Kindelia/io.kind
@@ -1,0 +1,11 @@
+Example.Kindelia.io : Kindelia.IO U60 {
+  do Kindelia.IO {
+    Kindelia.IO.save U60 1337
+    ask x = (Kindelia.IO.take U60)
+    return (+ x x)
+  }
+}
+
+Main {
+  Example.Kindelia.io
+}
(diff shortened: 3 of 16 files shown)


.gitmodules
```diff
@@ -1293,3 +1293,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/tree-sitter-kind"]
+	path = vendor/grammars/tree-sitter-kind
+	url = https://github.com/HigherOrderCO/tree-sitter-kind.git
```
lildude, line 1298: This won't work. We can only accept TextMate-compatible grammar.

lib/linguist/languages.yml
```diff
@@ -3488,6 +3488,15 @@ Kickstart:
   - ".ks"
   tm_scope: source.kickstart
   language_id: 692635484
+Kind:
+  type: programming
+  color: "#329866"
+  extensions:
+  - ".kind"
+  - ".kind2"
```
lildude, line 3496: You've not included any samples for this extension.

lildude (changes requested):
See inline comments. We also don't need that many samples, 2-3 of the most representative real world samples are enough, and be considerate of size; if the diff suppresses the content, it's too big.

lildude:
Oh yes, and don't use any structure for the samples. There must be directly in the language directory.

lib/linguist/languages.yml
algebraic-sofia: This extension is used by the older versions of the language, we are going to change all of the `.kind2` files to `.kind` so it's just a anticipation. Can we just add an PR in the future if we change `.kind2` to `.kind` definitely or we can maintain that antecipation?

lildude: > we are going to change all of the `.kind2` files to `.kind` so it's just a anticipation. 

You'll only be able to change your files. Any other files owned by other users will still remain until they change the extension 😁.

> Can we just add an PR in the future if we change `.kind2` to `.kind` definitely or we can maintain that antecipation?

Once support for an extension is added, it remains. We never remove or change extensions as this will remove recognition for files previously recognised.

So you can start off with adding just the most used extension now (assuming it meets our popularity requirements) and submit a new PR later when the other extension has met our usage requirements.

On that topic, I've taken a quick look at the popularity, and I don't think either are popular enough for inclusion right now which means we won't merge this PR. The `HigherOrderCO` user is disproportionately influencing the numbers as they seem to own most of the files currently indexed.

algebraic-sofia: > On that topic, I've taken a quick look at the popularity, and I don't think either are popular enough for inclusion right now which means we won't merge this PR. The HigherOrderCO user is disproportionately influencing the numbers as they seem to own most of the files currently indexed.
Yeah, HigherOrderCO owns most part of the ecosystem of the language as it is widely used for internal projects, so I guess that we can wait a little bit more time in order to add the language to this repo :)

PR #6403, opened 2023-05-04
Add Nushell

hustcer:
## Checklist:

      - https://github.com/search?q=path%3A*.nu&type=code
     
    - Sample source(s):
      - https://github.com/nushell/nushell/blob/main/.github/workflows/release-pkg.nu
      - https://github.com/nushell/nu_scripts/blob/main/make_release/this_week_in_nu_weekly.nu
    - Sample license(s): [MIT](https://github.com/nushell/nushell/blob/main/LICENSE)
  - [x] I have included a syntax highlighting grammar: [Nushell Grammar](https://github.com/hustcer/nu-grammar/blob/main/grammars/tmLanguage.json)
      
  - [x] I have added a color
    - Hex value: `#4E9906`
    - Rationale: It's the main theme color of Nushell logo from: https://github.com/nushell
    
### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -147,7 +147,7 @@ disambiguations:
     - named_pattern: vb-class
     - pattern: '^\s*BEGIN\R\s*MultiUse\s*=.*\R\s*Persistable\s*='
   - language: VBA
-    named_pattern: vb-class  
+    named_pattern: vb-class
   - language: TeX
     pattern: '^\s*\\(?:NeedsTeXFormat|ProvidesClass){'
   - language: ObjectScript
@@ -258,7 +258,7 @@ disambiguations:
   - language: Visual Basic 6.0
     and:
     - named_pattern: vb-form
-    - pattern: '^\s*Begin\s+VB\.Form\s+'  
+    - pattern: '^\s*Begin\s+VB\.Form\s+'
 - extensions: ['.fs']
   rules:
   - language: Forth
@@ -478,6 +478,10 @@ disambiguations:
   - language: NL
     pattern: '^(b|g)[0-9]+ '
   - language: NewLisp
+- extensions: ['.nu']
+  rules:
+  - language: Nushell
+    pattern: '^\s*(import|export|module|def|let|let-env) '
 - extensions: ['.odin']
   rules:
   - language: Object Data Instance Notation
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -482,6 +482,7 @@ disambiguations:
   rules:
   - language: Nushell
     pattern: '^\s*(import|export|module|def|let|let-env) '
+  - language: Nu
 - extensions: ['.odin']
   rules:
   - language: Object Data Instance Notation
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4526,7 +4526,7 @@ Nunjucks:
   language_id: 461856962
 Nushell:
   type: programming
-  color: "#5EA602"
+  color: "#4E9906"
   extensions:
   - ".nu"
   interpreters:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4524,6 +4524,21 @@ Nunjucks:
   tm_scope: text.html.nunjucks
   ace_mode: nunjucks
   language_id: 461856962
+Nushell:
+  type: programming
+  color: "#5EA602"
+  extensions:
+  - ".nu"
+  interpreters:
+  - nu
+  aliases:
+  - nu-script
+  - nushell-script
+  tm_scope: source.shell
+  ace_mode: sh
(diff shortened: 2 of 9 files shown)


lildude:
How is Nushell different from Nu which Linguist already supports and has done so for many many years? From a cursory naïve look, it appears to be the same thing with Nushell often being referred to as Nu, even in the [Nushell book](https://www.nushell.sh/book/).

hustcer:
> How is Nushell different from Nu which Linguist already supports and has done so for many many years? From a cursory naïve look, it appears to be the same thing with Nushell often being referred to as Nu, even in the [Nushell book](https://www.nushell.sh/book/).

They are totally two different language with the same name, and they are created by two different team. That's why I updated `lib/linguist/heuristics.yml` to distinguish Nushell from Nu using the same extension.

test/test_heuristics.rb
```diff
@@ -705,6 +705,12 @@ def test_nl_by_heuristics
     })
   end
 
+  def test_nu_by_heuristics
+    assert_heuristics({
+      "Nushell" => all_fixtures("Nushell", "*.nu")
+    })
+  end
```
lildude, line 712: This needs to also confirm the heuristics fallback to Nu for the Nu files.

lib/linguist/languages.yml
```diff
@@ -4524,6 +4524,22 @@ Nunjucks:
   tm_scope: text.html.nunjucks
   ace_mode: nunjucks
   language_id: 461856962
+Nushell:
+  type: programming
+  color: "#5EA602"
+  extensions:
+  - ".nu"
+  - ".nush"
```
lildude, line 4532: There are no `.nush` files currently indexed so we can't include it:

```suggestion
```

lib/linguist/heuristics.yml
```diff
@@ -478,6 +478,10 @@ disambiguations:
   - language: NL
     pattern: '^(b|g)[0-9]+ '
   - language: NewLisp
+- extensions: ['.nu']
+  rules:
+  - language: Nushell
+    pattern: '^\s*(import|export|module|def|let|let-env) '
```
lildude, line 484: We know the only other match is going to be Nu so we can shortcut things and save the classifier:

```suggestion
    pattern: '^\s*(import|export|module|def|let|let-env) '
  - language: Nu
```

lildude (changes requested):
Thanks for the clarification that this is a completely different language.

We need a few changes though:

1. Please update the template in the OP to state the licenses of the two samples and add a reason for why you have chosen the colour you have.
2. You need to add the cached license file for the grammar. This should have been created when you run `script/add-grammar` (you can manually run the command [here](https://github.com/github-linguist/linguist/blob/5a0c74277548122267d84283910abd5e0b89380e/script/add-grammar#L134) if you can't find the file).
3. See inline comments.

hustcer:
> 
> 1. Please update the template in the OP to state the licenses of the two samples and add a reason for why you have chosen the colour you have.

I don't know what's **OP**. I have just updated the description of license for the two samples. And updated the color to `#4E9906`, that's the main theme color of nushell logo from: https://github.com/nushell

> 2. You need to add the cached license file for the grammar. This should have been created when you run `script/add-grammar` (you can manually run the command [here](https://github.com/github-linguist/linguist/blob/5a0c74277548122267d84283910abd5e0b89380e/script/add-grammar#L134) if you can't find the file).

Added

> 3. See inline comments.

I have removed `.nush` extension (It's planed to be used in the future, we can add it after that)
For the other two comments, does that mean the Nushell scripts be recognized as `Nu` language? We may want to recognized as `Nushell` to make it different from [Nu](https://github.com/programming-nu/nu)

lildude:
I'm showing my age: OP == original post 😁

> For the other two comments, does that mean the Nushell scripts be recognized as `Nu` language? We may want to recognized as `Nushell` to make it different from [Nu](https://github.com/programming-nu/nu)

No. We need to ensure that your heuristics do not incorrectly identify Nu files as Nushell.

The heurstics are read top-down. Which means we'll try the Nushell heuristic first and if it doesn't match, we move onto the next regex for that extension. If there isn't one, or there isn't a match, we move onto the classifier. The classifier should be considered the last resort as it's trained on the samples and as such may not be accurate.

By adding `- language: Nu` to the end of the heuristics, you tell Linguist that if the `.nu` file it is analysing doesn't match the Nushell regex, it will be identified as Nu and Linguist will then not try analysing the file with the classifier.

The `assert_heuristics` test you've added needs to be updated to test and confirm this behaviour for the Nu samples we already have. You can see how this is done with other tests in this file.

hustcer:
@lildude Thanks for the detailed explanation, I have just updated the PR, Please check it again

hustcer:
Should fix https://github.com/github-linguist/linguist/issues/6214

hustcer:
@lildude Can this PR be **Approved**?

lildude (approved):
LGTM. Thanks.

lildude (approved):

PR #6410, opened 2023-05-12
Fix up vendor regex for bootstrap

tclem:
The current regex for detecting if bootstrap is vendored is too broad and overly complicated (uses lookahead). I believe the intent is only to match bootstrap files (not entire directories), but I've written a few more test to demonstrate what should and should not be detected as vendored here.

The current regex looks like this:
`(^|/)bootstrap([^/.]*)(?=\.).*\.(js|css|less|scss|styl)$`

As written, this would be considered vendored:
- `bootstrap-1.4/misc/other/reset.css`

But this would not:
- `bootstrap/misc/other/reset.css`

I don't believe that's correct or intended. This vendor rule should target bootstrap artifacts specifically (not entire directories). Some history in https://github.com/github-linguist/linguist/pull/5745 and https://github.com/github-linguist/linguist/pull/5957.

The proposal is to only match on the filename (and allow filenames that include `.`).

## Checklist:
  
### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -74,7 +74,7 @@
 - ([^\s]*)import\.(css|less|scss|styl)$
 
 # Bootstrap css and js
-- (^|/)bootstrap([^/]*)\.(js|css|less|scss|styl)$
+- (^|/)bootstrap([^/.]*)(\..*)?\.(js|css|less|scss|styl)$
 - (^|/)custom\.bootstrap([^\s]*)(js|css|less|scss|styl)$
 
 # Font Awesome
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -74,7 +74,7 @@
 - ([^\s]*)import\.(css|less|scss|styl)$
 
 # Bootstrap css and js
-- (^|/)bootstrap([^/.]*)(?=\.).*\.(js|css|less|scss|styl)$
+- (^|/)bootstrap([^/]*)\.(js|css|less|scss|styl)$
 - (^|/)custom\.bootstrap([^\s]*)(js|css|less|scss|styl)$
 
 # Font Awesome
--- a/test/test_file_blob.rb
+++ b/test/test_file_blob.rb
@@ -573,9 +573,9 @@ class TestFileBlob < Minitest::Test
     # Bootstrap
     assert !sample_blob("src/bootstraps/settings.js").vendored?
     assert !sample_blob("bootstrap/misc/other/reset.css").vendored?
-    assert !sample_blob("bootstrap-1.4/misc/other/reset.css").vendored?
-    assert !sample_blob("bootstrap.10.4/misc/other/reset.css").vendored?
-    assert !sample_blob("src/bootstrap-5.4.1-beta-dist/js/bundle.js").vendored?
+    assert sample_blob("bootstrap-1.4/misc/other/reset.css").vendored?
+    assert sample_blob("bootstrap.10.4/misc/other/reset.css").vendored?
+    assert sample_blob("src/bootstrap-5.4.1-beta-dist/js/bundle.js").vendored?
     assert sample_blob("src/bootstrap-custom.js").vendored?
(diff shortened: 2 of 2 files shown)


jorendorff (approved):
regexp looks accurate.

I don't know if matching across slashes is doing useful work for us though. `bootstrap*/*.css` is a thing that exists out there, I found some in Blackbird that I guess slipped past the current regex. [css under a "bootstrap" dir](https://github.com/search?q=path%3A%2Fbootstrap-%5B0-9.-%5D*%5C%2F.*%5C.css%24%2F&type=code); [some other files](https://github.com/search?q=path%3A%2Fbootstrap.*%5C%2F%5B%5Eb%5D*%5C.%28js%7Ccss%7Cscss%29%2F&type=code)

lildude:
Pinging you for review @Alhadis as our resident regex guru 🙇‍♂️ (no rush)

Alhadis (commented):
> I don't believe that's correct or intended. This vendor rule should target bootstrap artefacts specifically (not entire directories).

IIRC, popular Bootstrap components like styled form components are often bundled alongside the core Bootstrap files themselves (so, stuff like `bootstrap/components/datepicker/**/*.{css,js}`).

I could easily be wrong, however, as I've never actually used Bootstrap before (or anything resembling it). Ergo, I trust your familiarity with Bootstrap over mine.

tclem:
That's true! Looking at the links that @jorendorff posted though, there is a real mix of original js/css content nested inside directories named `bootstrap` ([example](https://github.com/nodejs/node/blob/2dd6d76c898c930d0f7d9386bde4b6dec37b7a49/lib/internal/bootstrap/realm.js)) and  twbs source that's just copied/vendored into another repo ([example](https://github.com/mdbootstrap/mdb-ui-kit/blob/ee5cd862c83ae2f4b21901756f0927367ac8661e/src/js/bootstrap/src/modal.js)).

I ended up proposing this change in part because the lookahead in this regex isn't very portable to linguist clones that are using different regex engines (and it's the only vendor regex that uses that lookahead). In investigating and writing some tests it then also seemed like the current regex was making some arbitrary choices as documented in the main PR description. I'm open to some further discussion here on how to best detect what is actually vendored twbs vs. what is code in a subdirectory of `bootstrap`. Any other ideas?

Alhadis:
> the lookahead in this regex isn't very portable to Linguist clones that are using different regex engines

In that case, I'd replace `(?=\.).*` with `(?:(?:\.).*)?`. It effectively achieves the same effect.

> I'm open to some further discussion here on how to best detect what is actually vendored twbs vs. what is code in a subdirectory of `bootstrap`. Any other ideas?

Unfortunately, no. It's impossible to distinguish vendored components from original (or heavily-modified) ones. There's really not much we *can* do, other than encourage authors to manually unvendor files on a case-by-case basis.

tclem:
> In that case, I'd replace `(?=\.).*` with `(?:(?:\.).*)?`. It effectively achieves the same effect.

We came up with replacing `(?=\.).*` with `(\..*)?\.` which I believe is the same as well. How does that look?

Alhadis:
Ugh, did I really write `(?:\.)`? 😞 (I need sleep…) Obviously that should've been `(?:\..*)?`, which is functionally equivalent to your solution, save for the use of a non-capturing group to reduce memory footprint `(?:…)` instead of `(…)`.

> How does that look?

I think it's fine. 👍 I was about to suggest you use `(?:…)` instead of `(…)`, but since most of the patterns defined in `vendor.yml` use ordinary capturing groups, it's not worth the inconsistency.

Alhadis (approved):

tclem:
@lildude and @Alhadis, I don't have write permissions so I'm going to let you all press the merge button here. This is good to go now, right?

lildude (approved):
@tclem  yup, this looks good to me. I tend to merge PRs only close to when I make a release to keep changes to a minimum in the event there is a need for a quick patch fix release. I’m aiming for a new release early June once everything has settled down on GitHub.

lildude (approved):

PR #6425, opened 2023-05-22
Update search query in PR template

lildude:
## Description

The new search has been enabled and rolled out to all users so we can update the URL we suggest in the template to use the new syntax.

## Checklist:

[ Removed as not applicable ]

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,7 +11,7 @@
   - [ ] The new extension is used in hundreds of repositories on GitHub.com
     - Search results for each extension:
       <!-- Replace FOOBAR with the new extension, and KEYWORDS with keywords unique to the language. Repeat for each extension added. -->
-      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
+      - https://github.com/search?q=path%3A*.FOOBAR+KEYWORDS+NOT+is%3Afork&type=code
   - [ ] I have included a real-world usage sample for all extensions added in this PR:
     - Sample source(s):
       - [URL to each sample source, if applicable]
@@ -22,7 +22,7 @@
   - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
     - Search results for each extension:
       <!-- Replace FOOBAR with the new extension, and KEYWORDS with keywords unique to the language. Repeat for each extension added. -->
-      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
+      -  https://github.com/search?q=path%3A*.FOOBAR+KEYWORDS+NOT+is%3Afork&type=code
   - [ ] I have included a real-world usage sample for all extensions added in this PR:
     - Sample source(s):
       - [URL to each sample source, if applicable]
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,7 @@
   - [ ] The new extension is used in hundreds of repositories on GitHub.com
     - Search results for each extension:
       <!-- Replace FOOBAR with the new extension, and KEYWORDS with keywords unique to the language. Repeat for each extension added. -->
-      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
+      - https://github.com/search?q=path%3A*.FOOBAR+KEYWORDS+NOT+is%3Afork&type=code
```
Alhadis, line 14: 1.	I'd move `type=code` to the start of the query string, so that `q=path%3A*.FOOBAR+KEYWORDS` is less fiddly to edit (with `NOT is:fork` receiving similar treatment). So:
	```suggestion
	      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
	```
2.	To match file extensions proper (as opposed to filenames that merely begin with a full-stop), the `path:` qualifier should be `**/?*.FOOBAR` instead. I'm not sure if this is even worth clarifying for users, so I'm including this as more of an FYI.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -22,7 +22,7 @@
   - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
     - Search results for each extension:
       <!-- Replace FOOBAR with the new extension, and KEYWORDS with keywords unique to the language. Repeat for each extension added. -->
-      -  https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
+      -  https://github.com/search?q=path%3A*.FOOBAR+KEYWORDS+NOT+is%3Afork&type=code
```
Alhadis, line 25: See point <span>#</span>1 above:

```suggestion
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
```

Alhadis (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
lildude: Good idea about switching the order.

Re:

> the `path:` qualifier should be `**/?*.FOOBAR` instead.

I don't think this makes any difference when it comes to the new search. GitHub's own suggestion if you use the old `extension:foobar` is `path:*.foobar` and the [docs](https://docs.github.com/en/search-github/github-code-search/understanding-github-code-search-syntax) indicate this is sufficient too.

I've tested this with a few extensions and don't get different results. Do you have an example where you do?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,7 @@
   - [ ] The new extension is used in hundreds of repositories on GitHub.com
     - Search results for each extension:
       <!-- Replace FOOBAR with the new extension, and KEYWORDS with keywords unique to the language. Repeat for each extension added. -->
-      - https://github.com/search?utf8=%E2%9C%93&type=Code&ref=searchresults&q=extension%3AFOOBAR+KEYWORDS+NOT+nothack
+      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
```
Alhadis, line 14: Shouldn't this be `path%3A**/*.FOOBAR`? `path:*.FOOBAR` will only match `.FOOBAR` files in the root directory of a project, IIRC.

Alhadis: Oh, right. Guess I've been spending too much time in the shell recently… 😅

Alhadis (approved):

PR #6442, opened 2023-06-05
Add WGSL support

joaogabrielzo:
## Checklist:
- [x] I am adding a new language.
  - [x] The extension of the new language is used in hundreds of repositories on [GitHub.com](http://github.com/).
      - https://github.com/search?q=path%3A*.wgsl&type=code
    - Sample source(s):
      - [Bevy's Example](https://github.com/bevyengine/bevy/blob/latest/assets/shaders/animate_shader.wgsl)
    - Sample license(s):
      - MIT 
  - [x] I have included a syntax highlighting grammar: [wgsl-grammar](https://github.com/joaogabrielzo/wgsl-grammar)
  - [x] I have added a color
    - Hex value: #1a5e9a
    - Rationale: Blue is a present color in the WebGPU logo.  This shade is used in the official WGSL doc.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1320,3 +1320,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/wgsl-grammar"]
+	path = vendor/grammars/wgsl-grammar
+	url = https://github.com/joaogabrielzo/wgsl-grammar.git
--- a/grammars.yml
+++ b/grammars.yml
@@ -1179,6 +1179,8 @@ vendor/grammars/vue-syntax-highlight:
 - text.html.vue
 vendor/grammars/wdl-sublime-syntax-highlighter:
 - source.wdl
+vendor/grammars/wgsl-grammar:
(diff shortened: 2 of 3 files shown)


lildude (changes requested):
The `script/add-grammar` script should have cached a copy of the repo's license to a yaml file. We need this file as part of this PR.

If you can't find it, possibly because of an issue being hit, you can manually get the file by running: `bundle exec licensed cache -c vendor/licenses/config.yml`.

As an aside, you don't need to maintain a repo just for the grammar. We can use the wgsl-analyzer repo directly.

joaogabrielzo:
I'm running into troubles trying to install the `charlock_holmes` gem and that seems to be the reason why the script won't fully work. For the same reason I can't run the command you sent.

lildude:
`charlock_holmes` isn't needed for that command so you should find running `gem install licensed` and then `licensed cache -c vendor/licenses/config.yml` will do the trick too.

lildude (changes requested):
You need to correct the ordering in the three locations identified in the test failures. For sorting the submodule, you can use `script/sort-submodules` (it's run as part of the `script/add-grammar` script).

lildude (approved):
LGTM

PR #6460, opened 2023-06-22
Update list of tree-sitter grammars

lildude:
## Description

The syntax highlighting engine team has added a few more tree-sitter grammars to the engine and this PR updates our list to match:

- C
- C#
- Nix
- Regular Expressions

## Checklist:

Removed as it doesn't apply.

### DIFF
--- a/script/list-grammars
+++ b/script/list-grammars
@@ -6,6 +6,8 @@ require "json"
 require "yaml"
 
 TREE_SITTER_GRAMMARS = {
+  "C" => "https://github.com/tree-sitter/tree-sitter-c",
+  "C#" => "https://github.com/tree-sitter/tree-sitter-c-sharp",
   "CodeQL" => "https://github.com/tree-sitter/tree-sitter-ql",
   "CSS" => "https://github.com/tree-sitter/tree-sitter-css",
   "EJS" => "https://github.com/tree-sitter/tree-sitter-embedded-template",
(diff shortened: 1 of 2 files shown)


Alhadis:
> * Regular Expressions

Ugh, seriously? The tree-sitter [grammar](https://github.com/tree-sitter/tree-sitter-regex/blob/a74f5ef949d806fafc1da99f084a85918814d495/src/grammar.json) appears to be specific to ECMAScript regular expressions, whereas [`language-regexp`](https://github.com/Alhadis/language-regexp/) takes pains to support as many different regex flavours as possible. Argh. 😞

sambacha:
> > * Regular Expressions
> 
> Ugh, seriously? The tree-sitter [grammar](https://github.com/tree-sitter/tree-sitter-regex/blob/a74f5ef949d806fafc1da99f084a85918814d495/src/grammar.json) appears to be specific to ECMAScript regular expressions, whereas [`language-regexp`](https://github.com/Alhadis/language-regexp/) takes pains to support as many different regex flavours as possible. Argh. 😞

Is there any comparison testing done of older grammars and the newer tree sitter based ones?

lildude:
> Is there any comparison testing done of older grammars and the newer tree sitter based ones?

A quick eyeball of the appearance is done. If it looks about right, and performs well, the grammar is used. If it doesn't, as was the case when we tried to use the Svelte tree-sitter grammar as [commented here](https://github.com/github-linguist/linguist/discussions/6164#discussioncomment-4215301), we don't.

Any discrepancies over "colour choice" for various tokens are down to grammar design decisions and would need to be taken up with the upstream grammar maintainers.

sambacha:
> > Is there any comparison testing done of older grammars and the newer tree sitter based ones?
> 
> A quick eyeball of the appearance is done. If it looks about right, and performs well, the grammar is used. If it doesn't, as was the case when we tried to use the Svelte tree-sitter grammar as [commented here](https://github.com/github-linguist/linguist/discussions/6164#discussioncomment-4215301), we don't.
> 
> Any discrepancies over "colour choice" for various tokens are down to grammar design decisions and would need to be taken up with the upstream grammar maintainers.

Thanks for answering my question, cheers!

salma0303:
$ github-linguist jekyll --rev origin/gh-pages
100.00% 2568354    HTML

PR #6479, opened 2023-07-15
Create devcontainer configuration

spenserblack:
## Description

See https://github.com/github-linguist/linguist/discussions/6473

Currently, prebuilds are available from [my fork](https://github.com/spenserblack/github-linguist/tree/devcontainer), built from the `devcontainer` branch. Hopefully none of the reviewers of this PR are in East Asia, because that's the one region I didn't enable in order to save on storage usage 🙃 

This will probably need some further changes. Some things I can think of off the top of my head are
- Do we want to use a specific Ruby version, or just stick with the latest?
- Any additional desired VS Code extensions or settings?
- Right now `script/bootstrap` is a post-create command, and it can apparently hit 100% CPU usage with a 2-core codespace. Should the min requirements be specified as 4-core?

## Checklist:

(Don't think any of the checkboxes apply, sorry if one does)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,6 +15,6 @@
       "extensions": ["EditorConfig.EditorConfig", "rebornix.Ruby"]
     }
   },
-  "postCreateCommand": "script/bootstrap",
+  "postCreateCommand": ".devcontainer/postCreate.sh",
   "remoteUser": "vscode"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,6 +15,7 @@
       "extensions": ["EditorConfig.EditorConfig", "rebornix.Ruby", "redhat.vscode-yaml"]
     }
   },
+  "onCreateCommand": "./script/bootstrap",
   "postCreateCommand": ".devcontainer/postCreate.sh",
   "remoteUser": "vscode"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,14 @@
+{
+  "name": "Ruby",
+  // Available versions: https://github.com/devcontainers/images/tree/main/src/ruby
+  "image": "mcr.microsoft.com/devcontainers/ruby",
+  "customizations": {
+    "codespaces": {
+      "openFiles": ["CONTRIBUTING.md", "lib/linguist/languages.yml"]
+    },
+    "vscode": {
+      "extensions": ["EditorConfig.EditorConfig", "rebornix.Ruby"]
+    }
+  },
+  "remoteUser": "vscode"
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,10 +9,17 @@
   },
   "customizations": {
     "codespaces": {
-      "openFiles": ["CONTRIBUTING.md", "lib/linguist/languages.yml"]
+      "openFiles": [
+        "CONTRIBUTING.md",
+        "lib/linguist/languages.yml"
+      ]
     },
     "vscode": {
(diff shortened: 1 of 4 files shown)


lildude:
This is looking good. I'll see if I can find time to play in the next week or two. 

One thing we'll definitely want to do as part of this PR is update the CONTRIBUTING.md file to detail how peeps can use this for Codespaces and local docker dev.

> Do we want to use a specific Ruby version, or just stick with the latest?

Latest released version to align more closely with GitHub production.

> Any additional desired VS Code extensions or settings?

TBC. I'll need to check my original branch as I know I added some.

> Right now `script/bootstrap` is a post-create command, and it can apparently hit 100% CPU usage with a 2-core codespace. Should the min requirements be specified as 4-core?

I'm undecided on this ATM. Yes, because it's a better experience and doesn't result in the warning message. No, because it does work and is only an issue during the initial cloning of the grammars, it shouldn't be an issue with prebuilds (might need a few tweaks), and because it costs users more. I'll think about it.

spenserblack:
Oh, just found out that some tests will error out in the codespace because some refs used for testing don't exist.
Looks like copying this would roughly fix it:
https://github.com/github-linguist/linguist/blob/e855ef2b6f90c34074061a2e17acbe853e61b483/.github/workflows/ci.yml#L32C9-L32C109

Since it should work for both the main repo and forks, and forks by default only copy the main branch, perhaps this could be part of the post create command:
```shell
git remote add linguist https://github.com/github-linguist/linguist
git fetch linguist ...
```
Or we could get more advanced and detect if `upstream` exists, if so fetch from then, else fetch from `origin`.

lildude:
I've had a chance to play with this today and it's looking good. I've made a few tweaks:

- I've added the `redhat.yaml` extension as peeps will mostly be editing the YAML files so lets help make it a better experience
- I've moved the bootstrapping to `onCreateCommand` so it can be cached as part of prebuilding

Can you please also test this and know what you think @spenserblack

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
...
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {},
+    "ghcr.io/devcontainers/features/go:1": {}
+  },
+  "customizations": {
+    "codespaces": {
+      "openFiles": ["CONTRIBUTING.md", "lib/linguist/languages.yml"]
```
Alhadis, line 12: **Extremely trivial nitpick:** these arrays should be formatted one entry per line, both for consistency and ease-of-diffing.

```suggestion
      "openFiles": [
        "CONTRIBUTING.md",
        "lib/linguist/languages.yml"
      ]
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "vscode": {
+      "extensions": ["EditorConfig.EditorConfig", "rebornix.Ruby", "redhat.vscode-yaml"]
```
Alhadis, line 15:
```suggestion
      "extensions": [
        "EditorConfig.EditorConfig",
        "rebornix.Ruby",
        "redhat.vscode-yaml"
      ]
```

Alhadis (changes requested):
Minor nitpick. Don't consider it a blocker or anything.

spenserblack:
Thanks @lildude and @Alhadis!

> Can you please also test this and know what you think @spenserblack

Just tested quickly by letting a prebuild run, opening a codespace, and running `bundle exec rake test`. Works like a charm!

Unfortunately, I'm at \>90% of my codespace storage quota, so after that quick test I promptly deleted my codespace and the prebuilds :sweat_smile:

Alhadis (approved):

lildude (approved):
Thanks for this @spenserblack 🙇

PR #6500, opened 2023-08-04
Add `.resource` extension to Robot Framework

Noordsestern:
Added missing file extension for Robot Framework resource files.

## Description
Keyword-only files in Robot Framework have `.resource` extension since Robot Framework 3.1 (which is since a couple of years.). See [User Guide](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#resource-files) with the Note _The .resource extension is new in Robot Framework 3.1._

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource+Keywords
    - Sample source(s):
      - https://github.com/MarketSquare/robotframework-camunda/blob/master/tests/robot/cleanup.resource
    - Sample license(s):
  - [x] I have included a change to the heuristics to distinguish my language from others using the same extension. ( :x: I could not find another `.resource` extension in `languages.yaml`)


Nixinova:
.resource seems quite a generic extension. Are there any other languages using a .resource extension that this might be confused with? Those would need to be added in this PR too. Have a look through https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource - how many are Robot and how many aren't?

Noordsestern:
There does not seem to be another language, but as it is generic, there seem to be common files named that way in JS, CSS and XML. should i include the filenames for these languages in the PR? or better not, as it might mess something up?

Alhadis:
@Noordsestern Is there a particular pattern we can search for that unambiguously identifies `.resource` files used by the Robot Framework? If so, we could add `.resource` as a [generic extension](https://github.com/github-linguist/linguist/blob/HEAD/lib/linguist/generic.yml), which doesn't run the risk of misclassifying unrelated files, but does require an accurate and reliable heuristic be included.

Noordsestern:
Yes, there is a pattern. Content wise, they will certainly always contain at least 1 of the following lines:

`*** Settings ***`
`*** Keywords ***`

Would that be sufficient?

Alhadis:
> Due to the localization extension of Robot Framework, the lines could be translated. Maybe a regex would due, then.
>
> Would that be sufficient?

I'm afraid not; those strings could easily appear in an unrelated file that ends in `.resource`; plus, the localisation extension you speak of would diminish such a heuristic's reliability anyway.

Noordsestern:
There are thousands of files. I checked first 5 pages of all appearances of the three lines in `.resource` files and they were all Robot Framework:
- https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource+%22***+Keywords+***%22
- https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource+%22***+Settings+***%22
- https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource+%22***+Variables+***%22

It seems very Robot Framework specific.

As for the localization feature: I would not consider it. When I localize Java reserved tokens, I would not be surprised if syntax highlighting breaks. Thus, I would localisation out.

Alhadis:
> See inline comments

Where?

lildude:
I was referring to your comments @Alhadis as I can't request changes without leaving a comment 😁

Noordsestern:
I am a bit lost :cold_sweat: What is requested change exactly?

I cannot proof that there won't be a `.resource` file that 
- has a line like ``*** Settings | Variables | Keywords ***` 
- _and_ is not Robot Framework. 

The search results suggest, that it is _extremely likely_ that it is Robot Framework when above conditions are met.

I am eager to comply, but I do not understand what change is required.

F3licity:
> See inline comments

I don't see any inline comments either. One often needs to submit his review for the comments to appear. Maybe you need to double check that you completed your reviewing ?

lildude:
> I don't see any inline comments either. One often needs to submit his review for the comments to appear. Maybe you need to double check that you completed your reviewing ?

As I said in [response](https://github.com/github-linguist/linguist/pull/6500#issuecomment-1700946213) to @Alhadis's comment about this, I was referring to his comments as I couldn't mark the PR as needing changes without making a comment... so I did complete my review, just poorly worded it 😁.

I've taken a closer look at the [search results for those strings](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource+%2F%5C*%5C*%5C*+%28Keywords%7CVariables%7CSettings%29+%5C*%5C*%5C*%2F&p=3) and I think we should be OK with using them as a specific enough heuristic for this generic extension for Robot. Like you, I can't find any files containing those specific strings which don't appear to be Robot (from my naïve understanding of Robot files 😁); I can find plenty which definitely aren't Robot without them, hence we have to go the generic route.

So the next steps are to add a heuristic and add an entry to the `generic.yml` file. You can look at the PRs for one of the other extensions in that file for an example of what needs to be done.

Noordsestern:
How is the popularity being measured? What kind of popularity is meant?

[There are plenty of hits when searching for a most common combination of `.resource` files for Robot Framework](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.resource+%22***+Keywords+***%22).

Robot Framework is popular for open source test automation[ with 2,6 million downloads per month](https://pypistats.org/packages/robotframework). 

It is common topic on testing conferences, is backed by foundation, , is popular on the open source realm for 15 years now. There are[ tons of ecosystem projects ](https://pypi.org/search/?q=robotframework). And keyword files happen to have the suffix `.resource` (see [guides](https://docs.robotframework.org/docs/examples/project_structure#project-with-tests-resources-and-data-folders)).

Also take in to account, that use case for Robot Framework test automation (it also used for RPA, but that use case is less popular). Teams that use Robot Framework in big scale work for usualyl for enterprises, thus their test repositories are not public. You may find them even more often in enterprise github (that's why i opened this PR, since the syntax highlighting is missing in my enterprise repositories).

What else is needed to pass popularity threshold?

lildude:
> How is the popularity being measured? What kind of popularity is meant?

I added the label based purely on the search in the OP.

We measure popularity as detailed in https://github.com/github-linguist/linguist/issues/5756 as referenced in the CONTRIBUTING.md file.

lildude (approved):

lildude (changes requested):
Please address the issue that is causing the test failures

Noordsestern (commented):
there seems to have been a typo. fixed.

lildude (changes requested):
Needs a heuristic test still.

lildude (approved):

PR #6518, opened 2023-08-16
Harden heuristics against `Regexp::TimeoutError` errors

lildude:
## Description

Ruby 3.2 allows you to set a process global `Regexp.timeout = <time>` to limit the impact of poorly written regexes and prevent ReDoS attacks. GitHub is now enforcing a timeout which has revealed...  

a) we're not playing nicely and rescuing the `Regexp::TimeoutError` errors so GitHub may sometime respond with a 500 error or not render a file at all
b) https://github.com/github-linguist/linguist/pull/6417 introduced a [regex which suffers from catastrophic backtracking](https://github.com/github-linguist/linguist/pull/6417#pullrequestreview-1580134908) which ultimately causes the `Regexp::TimeoutError` errors. This was missed as the issue only comes to light when analysing largish files.

This PR addresses both by rescuing the `Regexp::TimeoutError` exceptions when analysing using the heuristics strategy - we return an empty result as misidentifcation is better than a 500 error -  and replaces the bad regex that brought this to light.

## Checklist:

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -686,7 +686,7 @@ disambiguations:
 - extensions: ['.stl']
   rules:
   - language: STL
-    pattern: '\A\s*solid[\s\S]*^endsolid(?:$|\s)'
+    pattern: '\A\s*solid(?:$|\s)[\s\S]*^endsolid(?:$|\s)'
 - extensions: ['.sw']
   rules:
   - language: Sway
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -686,7 +686,7 @@ disambiguations:
 - extensions: ['.stl']
   rules:
   - language: STL
-    pattern: '\A\s*solid(?=$|\s)(?:.|[\r\n])*?^endsolid(?:$|\s)'
+    pattern: '\A\s*solid[\s\S]*endsolid(?:$|\s)'
 - extensions: ['.sw']
   rules:
   - language: Sway
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -686,7 +686,7 @@ disambiguations:
 - extensions: ['.stl']
   rules:
   - language: STL
-    pattern: '\A\s*solid[\s\S]*endsolid(?:$|\s)'
+    pattern: '\A\s*solid[\s\S]*^endsolid(?:$|\s)'
 - extensions: ['.sw']
   rules:
   - language: Sway
--- a/lib/linguist/heuristics.rb
+++ b/lib/linguist/heuristics.rb
@@ -30,6 +30,8 @@ module Linguist
       end
 
       [] # No heuristics matched
+    rescue Regexp::TimeoutError
+      [] # Return nothing if we have a bad regexp which leads to a timeout enforced by Regexp.timeout in Ruby 3.2 or later
     end
 
     # Public: Get all heuristic definitions
--- a/test/test_heuristics.rb
+++ b/test/test_heuristics.rb
@@ -34,6 +34,11 @@ class TestHeuristics < Minitest::Test
     assert_equal [], Heuristics.call(file_blob("Markdown/symlink.md"), [Language["Markdown"]])
   end
 
+  def test_no_match_if_regexp_timeout
+    Regexp.any_instance.stubs(:match).raises(Regexp::TimeoutError)
+    assert_equal [], Heuristics.call(file_blob("#{fixtures_path}/Generic/stl/STL/cube1.stl"), [Language["STL"]])
+  end
+
   # alt_name is a file name that will be used instead of the file name of the
   # original sample. This is used to force a sample to go through a specific
   # heuristic even if its extension doesn't match.
--- a/test/test_heuristics.rb
+++ b/test/test_heuristics.rb
@@ -35,6 +35,8 @@ class TestHeuristics < Minitest::Test
   end
 
   def test_no_match_if_regexp_timeout
+    skip("This test requires Ruby 3.2.0 or later") if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('3.2.0')
+
     Regexp.any_instance.stubs(:match).raises(Regexp::TimeoutError)
     assert_equal [], Heuristics.call(file_blob("#{fixtures_path}/Generic/stl/STL/cube1.stl"), [Language["STL"]])
   end


DecimalTurn:
This looks good to me. On top of solving backtracking issues, `[\s\S]*` also makes things much more readable as it's clear to me that the original regex was written to match "anything including line endings" until we get to `endsolid`.

We might also want to replace other instances of `(?:.|[\r\n])*` even if they aren't causing trouble just yet.

lib/linguist/heuristics.yml
```diff
@@ -686,7 +686,7 @@ disambiguations:
 - extensions: ['.stl']
   rules:
   - language: STL
-    pattern: '\A\s*solid(?=$|\s)(?:.|[\r\n])*?^endsolid(?:$|\s)'
+    pattern: '\A\s*solid[\s\S]*^endsolid(?:$|\s)'
```
Alhadis, line 689: 1.	First of all, `solid[\s\S]` will match stuff like `solidobject` (which I presume isn't valid STL syntax), so either `(?=$|\s)` or `(?:$|\s)` is necessary here.

2.	Second, because this heuristic is anchored to the first-line of input (`\A`), it's possible to match the terminating `endsolid` directive using a separate regex:
	```suggestion
	    and:
	    - pattern: '\A\s*solid(?:$|\s)'
	    - pattern: '^\s*endsolid(?:$|\s)'
	```
	(**Nota bene:** I haven't tested these changes locally)

Alhadis (changes requested):

Alhadis:
@lildude Hope you don't mind my changing the PR's title; it exceeded the 72-character limit of well-formed Git commit-messages. 😉

lib/linguist/heuristics.yml
DecimalTurn: Is there a reason why `(?:$|\s)` is preferred over `\b` in this case?

lildude: > 2\. Second, because this heuristic is anchored to the first-line of input (`\A`), it's possible to match the terminating `endsolid` directive using a separate regex:

Splitting things makes this much more expensive. The second regex here is really expensive ([11144 steps](https://regex101.com/r/6vXA8K/1)) and quite possibly going to lead to another case of things running amok as the files grow compared my initial suggestion ([29 steps](https://regex101.com/r/VajTr7/1)), which now you point it out, needs improving for the situation you mentioned.

These all seem to match as be much more performant:

- `\A\s*solid(?:$|\s)[\s\S]*^endsolid(?:$|\s)` => [33 steps](https://regex101.com/r/HSQ919/1)
- `\A\s*solid\b[\s\S]*^endsolid(?:$|\s)` => [30 steps](https://regex101.com/r/i5LFeZ/1)
- `\A\s*solid\b[\s\S]*^endsolid\b` => [28 steps](https://regex101.com/r/aqC12C/1)

jorendorff: Sorry for the perf problems, everyone.

Does regex101 reliably approximate the cost to run these regexps in Ruby? The good original, before my bad version, was `'\A\s*solid(?=$|\s)(?m:.*?)\Rendsolid(?:$|\s)'`; the closest thing that works in regex101 is `/\A\s*solid(?=$|\s)(?:.*?)\nendsolid(?:$|\s)/gms`, and it claims that is super expensive ([107792 steps](https://regex101.com/r/HSQ919/1)).

jorendorff: OK, I tested the regexps in Ruby.

None of them perform badly on the test case, or on [a 5000 line file I found](https://github.com/prusa3d/PrusaSlicer/blob/c05542590d7c2d73eb69bbf7a82a482a075815c1/resources/profiles/Sovol/SV05_bed.stl). 

Here's how long each regex takes to run on the 500-line file, in milliseconds:

```
original: 1.9683429999859072
current:  2.581663000048138
proposed: 1.2971370000159368
```

regex101 is using a different regex engine.

jorendorff: The numbers above are for Ruby 3.2.2 on my laptop. Here's the code:

<details><summary>source code</summary>
```ruby
require 'benchmark'

ORIGINAL_RE  = /\A\s*solid(?=$|\s)(?m:.*?)\Rendsolid(?:$|\s)/
CURRENT_RE   = /\A\s*solid(?=$|\s)(?:.|[\r\n])*?^endsolid(?:$|\s)/
PROPOSED_RE  = /\A\s*solid[\s\S]*^endsolid(?:$|\s)/

text = File.read("SV05_bed.stl")

t = Benchmark.measure {
  1000.times do
    ORIGINAL_RE.match?(text)
  end
}
puts "original: #{t.real}"

t = Benchmark.measure {
  1000.times do
    CURRENT_RE.match?(text)
  end
}
puts "current:  #{t.real}"

t = Benchmark.measure {
  1000.times do
    PROPOSED_RE.match?(text)
  end
}
puts "proposed: #{t.real}"
```
<details>

Alhadis: > regex101 is using a different regex engine.

Ruby and PCRE differ in their interpretation of "multiline" modifiers (the parts scoped with `(?m:…)`). Specifically:

<table>
	<thead>
		<tr>
			<th></th>
			<th><h1><abbr title="multiline"><samp><b>(?m:</b><ins>...</ins><b>)</b></samp></abbr></h1> means…</th>
			<th><h1><abbr title="dotall"><samp><b>(?s:</b><ins>...</ins><b>)</b></samp></abbr></h1> means…</th>
		</tr>
	</thead>
	<tr>
		<th>Ruby</th>
		<td><code>.</code> matches any character, including newlines</td>
		<td>N/A (No such modifier)</td>
	</tr>
	<tr>
		<th>PCRE/Perl</th>
		<td><code>^</code> and <code>$</code> respectively match the beginning and end of each line, rather than the input.</td>
		<td><code>.</code> matches any character, including newlines; implies <code>(?m)</code>.</td>
	</tr>
</table>

See, the behaviour that `(?m)` normally enables in Perl/PCRE is instead the default in Ruby; the `\A` and `\Z`/`\z` assertions are used to match the start and end of the input string, respectively (which would normally be achieved using `^` and `$` without multiline mode).

Now, why is this relevant? Because benchmarks could be skewed in favour of whichever regex happens to fail first (if it won't match `.` because the dotall modifier is needed, then the regex engine logically has no reason to continue reading past the first newline).

Alhadis:
Linguist's heuristics are trying to
1. **Scale well** (natively-supported regex constructs like `\R` and `.` in multiline mode will likely be more performant than something like `(?:.|\r?\n)*`),
2. **Be implemented purely as regular expressions**, something that greatly restricts our available options when dealing with issues like these. Recall that Linguist's heuristics were historically implemented in Ruby (à la [`Linguist::Generated`](https://github.com/github-linguist/linguist/blob/3f6d887ecd87987ba55b88987734edf57765c1bd/lib/linguist/generated.rb)), albeit mostly with regular expressions,
3. **[Satisfy the limitations](https://github.com/github-linguist/linguist/pull/6417) of a foreign regex engine**; i.e., by removing lookarounds and replacing `\R` with `(?:.|\r?\n)*`.

To quote [Matthew 6:24-34](https://www.biblegateway.com/passage/?search=Matthew%206:24-34&version=NRSVCE): <i><q>No one can serve two masters</q></i>. If @jesuschrist were active on GitHub, I'm sure he'd agree that we're trying to serve <em>three</em> masters (i.e., conflicting objectives). That third objective is starting to become a liability, IMHO.

/cc @jorendorff

DecimalTurn:
>... by removing lookarounds and replacing \R with (?:.|\r?\n)*

I thought we were replacing `\R` by `(?:\r?\n|\r)` like discussed [here](https://github.com/github-linguist/linguist/pull/6417#issuecomment-1560201493).

In the case that this PR addresses, we tried replacing `(?m).*` with `(?:.|[\r\n])*`, but it turns out that `[\s\S]*` is much better for performance.

Based on the Regex101 debugger, it seems like `[\s\S]*` allows the engine to skip all the way to the end which is quite helpful when the file is quite large:

jorendorff:
> **Scale well** (natively-supported regex constructs like `\R` and `.` in multiline mode will likely be more performant than something like `(?:.|\r?\n)*`),

The portable regex in this PR is 50% faster than the original (which uses `\R` and `.` in multiline mode). So there aren't really conflicting priorities here.

I didn't meant to introduce one -- of course Linguist is a Ruby library above all, and you must do what is right for the project in that light.

lib/linguist/heuristics.yml
```diff
@@ -686,7 +686,7 @@ disambiguations:
 - extensions: ['.stl']
   rules:
   - language: STL
-    pattern: '\A\s*solid[\s\S]*^endsolid(?:$|\s)'
+    pattern: '\A\s*solid(?:$|\s)[\s\S]*^endsolid(?:$|\s)'
```
jorendorff, line 689: `(?:$|\s)` amounts to `\s` in this context.

lildude: I'm far from an regex expert so I'm open to suggestions. I want to match both of these cases as they're both valid (there's  no space at the end of the `solid` line in the second example): 

``` 
solid foo
  [...]
endsolid
```

```
solid
  [...]
endsolid
```

… but not `solidfoo` or `solid,`.

DecimalTurn:
@lildude are you considering a minor release for these changes (ie. v7.26.1) since this can affect other projects (such as go-enry) that are dependant on Linguist or were you planning to wait for the next major release?

Also, should there be a warning against this issue in the notes for https://github.com/github-linguist/linguist/releases/tag/v7.26.0 ?

Alhadis (approved):

lildude:
> @lildude are you considering a minor release for these changes (ie. v7.26.1) since this can affect other projects (such as go-enry) that are dependant on Linguist or were you planning to wait for the next major release?

@DecimalTurn I _was_ going to make a patch release, however it's taken so long to finish this PR (I've been distracted by higher priorities of my day job) that it's now time for a major release anyway as the freeze date for the next GitHub Enterprise Server (GHES) release is fast approaching and I like the Linguist updates to bake for a bit in prod before the freeze in case any niggles pop-up (it's a PITA to backport them to GHES 😁).

Gonna merge this and start getting things in line for a new release early next week.

PR #6530, opened 2023-08-30
Platform agnostic line endings

DecimalTurn:
## Description
This PR should make all remaining regex patterns follow our new approach of using platform-agnostic line endings as discussed here: https://github.com/github-linguist/linguist/pull/6417#issuecomment-1559966383

Here are the replacements made:
1. Standalone[^1] `\r?\n` with `(?:\r?\n|\r)` 
2. Standalone[^1] `\n` with `(?:\r?\n|\r)`
3. `(\s|\n)` with  `[\s\r\n]`
4. `[^\n]` with `[^\r\n]`

I also replaced `(^|\n)` with just `^` since matching with the previous line ending or not doesn't affect the outcome.

[^1]: In the sense that they are not part of a character set.

_(Checklist removed as it doesn't apply)._

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -74,7 +74,7 @@ disambiguations:
   - language: Public Key
     pattern: '^(----[- ]BEGIN|ssh-(rsa|dss)) '
   - language: AsciiDoc
-    pattern: '^[=-]+(\s|\n)|\{\{[A-Za-z]'
+    pattern: '^[=-]+[\s\r\n]|\{\{[A-Za-z]'
   - language: AGS Script
     pattern: '^(\/\/.+|((import|export)\s+)?(function|int|float|char)\s+((room|repeatedly|on|game)_)?([A-Za-z]+[A-Za-z_0-9]+)\s*[;\(])'
 - extensions: ['.asm']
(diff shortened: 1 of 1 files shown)


lildude (approved):

Alhadis:
Shit, sorry I missed your review request, @lildude. 😞 Been having trouble getting myself in front of a workstation lately.

> 3. `(\s|\n)` with `[\s\r\n]`

I'm not sure why or where `(\s|\n)` was used in the first place, but this should've simply been `\s`; that construct is [equivalent](https://ruby-doc.org/3.2.2/Regexp.html#class-Regexp-label-Character+Classes) to `[ \t\r\n\f\v]` in Ruby. Ergo, it should've been simplified to just `\s`.

lildude:
Ooo, I've just noticed this PR introduced a regex [warning in the test output](https://github.com/github-linguist/linguist/actions/runs/6019590751/job/16329635463#step:5:72):

```
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /^[=-]+[\s\r\n]|\{\{[A-Za-z]/
```

This is for the AsciiDoc regex.

If you've got a mo @DecimalTurn could you please address this.

Alhadis:
@lildude The `/^[=-]+[\s\r\n]` part of the offending expression is an example of the redundancy I mentioned above.

To fix:

~~~diff
-/^[=-]+[\s\r\n]|\{\{[A-Za-z]/
+/^[=-]+\s|\{\{[A-Za-z]/
~~~

DecimalTurn:
@Alhadis Thanks, for the catch. 

@lildude Should I create a new PR for this fix?

lildude:
@DecimalTurn yes please. 🙇

PR #6533, opened 2023-08-31
.vscode/settings.json as filename for jsonc

brettrowberry:
- Visual Studio Code's `.vscode/settings.json` uses `JSON with Comments`, see https://code.visualstudio.com/docs/getstarted/settings#_workspace-settingsjson-location.
- GitHub treats this file as `JSON`. 

## Description
Adds `.vscode/settings.json` as a filename for `JSON with Comments`.

## Checklist:
    - Sample source(s):
      - https://github.com/fabiorocha/vscode-settings.json/blob/master/settings.json
    - Sample license(s):
    - I don't think this applies as the filepath is quite distinct.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3176,6 +3176,7 @@ JSON with Comments:
   - ".jshintrc"
   - ".jslintrc"
   - ".swcrc"
+  - ".vscode/settings.json"
   - api-extractor.json
   - devcontainer.json
   - jsconfig.json


lib/linguist/languages.yml
```diff
@@ -3176,6 +3176,7 @@ JSON with Comments:
   - ".jshintrc"
   - ".jslintrc"
   - ".swcrc"
+  - ".vscode/settings.json"
```
Alhadis, line 3179: This won't work. Linguist doesn't consider the directory components of file paths when examining filenames.

brettrowberry: Bummer. I thought I was taking a chance when I didn't see any others like this. Would it be reasonable to just put `settings.json`? It doesn't seem like that distinctive a name, but maybe it is.

Alhadis: > Would it be reasonable to just put `settings.json`?

Nope, sorry. That filename is _waaaay_ too generic. Best solution is to add this to your project's `.gitattributes` file (create one if necessary):

~~~gitattributes
.vscode/**/*.json linguist-language=JSONC
~~~

brettrowberry: Thank you, I'll do exactly that.

brettrowberry: Looking forward, which of these possibilities seem more plausible?

- Linguist considers filepaths, not just filenames. 
- VS Code switches to more distinctively named dot files, e.g. `settings.json` -> `.vscodesettings`.

Alhadis: Neither strikes me as particularly plausible; the former would involve a complete rehaul of Linguist's codebase, while the latter would only become eligible once there's sufficient distribution of `.vscodesettings` files across GitHub (see [`CONTRIBUTING.md`](https://github.com/github-linguist/linguist/blob/master/CONTRIBUTING.md#adding-a-language) for rationale and specifics).

Instead, it'd be better if VS Code adopted `.jsonc` as a file extension (which Linguist [already supports](https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml#L3156)) instead of just `.json` (which is too easily conflated with proper JSON).

brettrowberry: Great idea!

brettrowberry: https://github.com/microsoft/vscode/issues/191988

PR #6539, opened 2023-09-08
Improve and document using Codespaces and Devcontainers

lildude:
## Description

This PR...

- tweaks the devcontainer config to move all the preconfiguration setup to the onCreateCommand step as adding the remote repo and pulling the test branch can be cached and thus can shave off a few more seconds from starting the Codespace
- updates the CONTRIBUTING.md to detail using Codespaces and Devcontainers

## Checklist:

[ Removed as it doesn't apply ]


Mk518011:
Thanks

PR #6566, opened 2023-10-05
txtpb: Adding canonical txtpb file extension to Protocol Buffer Text …

milesflo:
…Format
The [Text Format specification](https://protobuf.dev/reference/protobuf/textformat-spec/#text-format-files)  has consolidated its canonical file extension to `txtpb`, with `textproto` still supported in some legacy tooling...

3rd order legacy formats (legacy of legacy, you could say) have been removed in this change

## Description

To Protocol Buffer Text Format:

### Add Extensions

- `txtpb` canonical extension by the standard

### Remove Extensions

- `pbt` not seeing any evidence this was a canonical file extension
- `pbtxt` 3rd order legacy

## Checklist:
(Didn't see any that applied to file extension patterns)


lib/linguist/languages.yml
```diff
@@ -5410,8 +5410,7 @@ Protocol Buffer Text Format:
   - protobuf text format
   extensions:
   - ".textproto"
-  - ".pbt"
-  - ".pbtxt"
+  - ".txtpb"
```
lildude, line 5413: We never remove support for an extension (or language) as files will always exist. Please leave these and if the new extension is to be the primary extension, place it first.

lildude (changes requested):
Along with the inline comment, we will also require samples of the new extensions files, and for the popularity requirements to be met.

lib/linguist/languages.yml
milesflo: Okay, makes perfect sense to me. 

What about the `pbt` extension? I see it [elsewhere in the file](https://github.com/github-linguist/linguist/blob/017c5eaf05b6536799c5fdad4457cac127c67ee5/lib/linguist/languages.yml#L5307) for PowerBuilder

milesflo:
@lildude Comment updated 👍

lib/linguist/languages.yml
lildude: Don’t worry about it. Anything already supported should already be correctly handled… we assume things work until someone points out a problem 😁

lildude (commented):
We still need a sample for the new extension.

milesflo:
> We still need a sample for the new extension.

Ahh I see. Done: https://github.com/github-linguist/linguist/pull/6566/commits/0fc370705d76fccd898e74384635b821d352042b

lildude:
Thanks but as stated in the CONTRIBUTING.md file, the sample should be a real world sample. This appears to be a contrived "hello world" like sample. We also need to know the source of the sample and the license of the file. This was a point in the template you removed when opening this PR, which brings me to that port, please put back the template and fill it in… it's there for a reason 😉

milesflo:
@lildude Updated the sample, readded the template, and added links to everything

dougthor42:
Would it be possible to have an exception for the "pending popularity" requirement?

At least within our org (https://github.com/qh-lab), the only reason we didn't use `.txtpb` is because linguist doesn't highlight it. I'm sure other orgs/repos had the same thought.

lildude:
> Would it be possible to have an exception for the "pending popularity" requirement?

Unfortunately we can't as it sets a precedent we don't want to fight each and every time someone develops "the next greatest thing, promise on my hamster's life".

As Linguist already knows about Protobu Text formatthe language, you can use an [override](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md) to set the language until this PR is merged and deployed.

dougthor42:
I am disappointed, but completely understand and agree. But I had to try :wink:.

Also, good to know about overrides! I'll let our team know. Thanks!

lib/linguist/languages.yml
```diff
@@ -5567,6 +5567,7 @@ Protocol Buffer Text Format:
   - text proto
   - protobuf text format
   extensions:
+  - ".txtpb"
   - ".textproto"
   - ".pbt"
   - ".pbtxt"
```
lildude, line 5573:
```suggestion
  - ".txtpb"
  - ".pbt"
  - ".pbtxt"
  - ".textproto"
```

lib/linguist/languages.yml
```diff
@@ -5567,9 +5567,10 @@ Protocol Buffer Text Format:
   - text proto
   - protobuf text format
   extensions:
-  - ".textproto"
+  - ".txtpb"
   - ".pbt"
   - ".pbtxt"
+  - ".textproto"
```
lildude, line 5573: Change of plan. It looks like `textproto` is still the primary extension:

```suggestion
  - ".textproto"
  - ".pbt"
  - ".pbtxt"
  - ".txtpb"
```

stefanvanburen: I think `.txtpb` ought to be the primary extension, according to the docs: https://protobuf.dev/reference/protobuf/textformat-spec/#text-format-files.

PR #6578, opened 2023-10-12
Add Glimmer JS language

gilest:
## Description
Adds support for [Glimmer.js](https://glimmerjs.com/) which will be the [component authoring format](https://github.com/ember-template-imports/ember-template-imports) of the [next Edition of Ember.js](https://emberjs.com/editions/polaris/).

## Checklist:

      -  [.gjs search](https://github.com/search?q=%28path%3A*.gjs%29+AND+%28content%3A%3Ctemplate%3E+OR+content%3A%40glimmer+OR+content%3A%40ember%29&type=code) 1.7k files
    - Sample source(s):
	    - [class.gjs](https://github.com/discourse/discourse-ai/blob/main/assets/javascripts/discourse/components/modal/diff-modal.gjs)
	    - [class.gts](https://github.com/cardstack/cardstack/blob/main/packages/boxel/addon/components/boxel/input/time/usage.gts)
    - Sample license(s):
	    - class.gjs - (https://github.com/discourse/discourse-ai) MIT
	    - template-only.gjs (https://github.com/CrowdStrike/opensource.crowdstrike.com) MIT
	 - https://github.com/lifeart/vsc-ember-syntax – MIT
  - [x] I have added a color
    - Hex value: `#F5835F`
    - Rationale: Logo colour from https://glimmerjs.com/ at time of writing

### DIFF
--- a/grammars.yml
+++ b/grammars.yml
@@ -1057,6 +1057,11 @@ vendor/grammars/verilog.tmbundle:
 - source.verilog
 vendor/grammars/vhdl:
 - source.vhdl
+vendor/grammars/vsc-ember-syntax:
+- inline.hbs
+- inline.template
+- source.gjs
+- source.gts
 vendor/grammars/vsc-fennel:
 - source.fnl
 vendor/grammars/vscode-TalonScript:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2424,6 +2424,7 @@ Glimmer JS:
   color: "#F5835F"
   tm_scope: source.gjs
   group: JavaScript
+  language_id: 5523150
 Glyph:
   type: programming
   color: "#c1ac7f"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2416,6 +2416,14 @@ Gleam:
   - ".gleam"
   tm_scope: source.gleam
   language_id: 1054258749
+Glimmer JS:
+  type: programming
+  extensions:
+  - ".gjs"
+  ace_mode: javascript
+  color: "#F5835F"
+  tm_scope: source.gjs
+  group: JavaScript
 Glyph:
   type: programming
   color: "#c1ac7f"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1335,3 +1335,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/vsc-ember-syntax"]
+	path = vendor/grammars/vsc-ember-syntax
+	url = https://github.com/lifeart/vsc-ember-syntax.git
--- a/samples/Glimmer JS/class.gjs
+++ b/samples/Glimmer JS/class.gjs
@@ -1,0 +1,48 @@
+import Component from "@glimmer/component";
+import { action } from "@ember/object";
+import DButton from "discourse/components/d-button";
+import DModal from "discourse/components/d-modal";
+import DModalCancel from "discourse/components/d-modal-cancel";
+import I18n from "I18n";
+import { htmlSafe } from "@ember/template";
+
+const t = I18n.t.bind(I18n);
+
+export default class ModalDiffModal extends Component {
+  <template>
+    <DModal
+      class="composer-ai-helper-modal"
+      @title={{t "discourse_ai.ai_helper.context_menu.changes"}}
+      @closeModal={{@closeModal}}
+    >
+      <:body>
+        {{#if @diff}}
+          {{htmlSafe @diff}}
+        {{else}}
+          <div class="composer-ai-helper-modal__old-value">
+            {{@oldValue}}
+          </div>
+
+          <div class="composer-ai-helper-modal__new-value">
+            {{@newValue}}
+          </div>
+        {{/if}}
+      </:body>
+
+      <:footer>
+        <DButton
+          class="btn-primary confirm"
+          @action={{this.triggerConfirmChanges}}
+          @label="discourse_ai.ai_helper.context_menu.confirm"
+        />
+        <DModalCancel @close={{@closeModal}} />
+      </:footer>
+    </DModal>
+  </template>
+
+  @action
+  triggerConfirmChanges() {
+    this.args.closeModal();
+    this.args.confirm();
+  }
+}
--- a/samples/Glimmer JS/template-only.gjs
+++ b/samples/Glimmer JS/template-only.gjs
@@ -1,0 +1,55 @@
+import { ExternalLink, Link } from '@crowdstrike/ember-oss-docs';
+
+export const Footer = <template>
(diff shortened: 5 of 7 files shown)


gilest:
@lildude would you mind running the CI for me? Want to make sure I didn't make a mistake this time 😅

NullVoxPopuli:
really excited for this! looks like we're no longer "Pending Popularity"

lib/linguist/languages.yml
```diff
@@ -2416,6 +2416,15 @@ Gleam:
   - ".gleam"
   tm_scope: source.gleam
   language_id: 1054258749
+Glimmer JS:
```
lildude, line 2419: Something to ponder... what will developers expect to see in the sidebar of their repos?

As this stands they'll see JavaScript because of the `group: JavaScript` line below. Individual files will however be classified as "Glimmer JS" and correctly highlighted.

If developers expect it to be called "Glimmer JS" then great, nothing to change other than to remove the `group` line. 

If they expect it to be called "Glimmer" as was the intent of the first PR then I recommend naming this "Glimmer".

I'm thinking about the future here. When "Glimmer TS" is added, will users expect to see it reflected as "Glimmer", "Glimmer JS" or "Glimmer TS" in the sidebar?

If "Glimmer", then it's best to name this entry "Glimmer" now and "Glimmer TS" can be grouped with a `group: Glimmer` line when it's added. This will result in both variants appearing as "Glimmer" in the sidebar but files will be individually identifed as the JS/TS variants.

If "Glimmer JS", then all is good and `group: Glimmer JS` can be included in the "Glimmer TS" entry.

If "Glimmer TS", then all is good and it can be added as a completely standalone language without grouping in future.

lildude:
> really excited for this! looks like we're no longer "Pending Popularity"

Not quite. Excluding forks drops things down to ~1.8k.

lib/linguist/languages.yml
gilest: 🤔 I've definitely misconfigured the `group` key at present then.

My intuition is that Glimmer JS and Glimmer TS would be fine since it's helpful to see progress in a partially-typed repo. I'll reach out to the community for feedback, though.

gilest: Consensus from the community:

| Language  | Group (for language summary) |
| ------------- | ------------- |
| Glimmer JS  | JavaScript  |
| Glimmer TS  | TypeScript  |

So I don't think any changes will be required.

When we open the Gimmer TS PR it will add its own language entry.

NullVoxPopuli:
> Not quite. Excluding forks drops things down to ~1.8k.

can we include markdown which references gjs? (that's at 2k :partying_face: )

wagenet:
@lildude It looks like gjs without markdown and excluding forks is now at 2k.

grammars.yml
```diff
@@ -1057,6 +1057,11 @@ vendor/grammars/verilog.tmbundle:
 - source.verilog
 vendor/grammars/vhdl:
 - source.vhdl
+vendor/grammars/vsc-ember-syntax:
+- inline.hbs
+- inline.template
+- source.gjs
+- source.gts
```
bartocc, line 1064: @gilest very happy to see this PR! I hope it'll get accepted as soon as possible.

For the time being, I fallback to `JS` or `TS` when I want to have a usable syntax highlighting on github.com for my `.gjs` and `.gts` files

If this PR is only for "Glimmer JS", is this line supposed to be here?

gilest: I think so, yes. Because when Glimmer TS is added it will use the same grammar

Those entries are generated when running `script/add-grammar`

lildude:
Heads up: this language won't be using the most up-to-date version of the grammar when it is release GitHub as a duplicate scope has been added to that repo since this PR was created.

```
- [ ] repository `vendor/grammars/Handlebars` (from https://github.com/daaain/Handlebars) (1 errors)
  - Duplicate scope in repository: scope `text.html.handlebars` was already defined in repository `vendor/grammars/vsc-ember-syntax` (from https://github.com/lifeart/vsc-ember-syntax.git)
```

`vendor/grammars/Handlebars` has priority as it was here first and is dedicated to the language.

lildude:
Actually, now I look at this more closely, it appears this grammar was added _with_ that conflict already in place suggesting `script/add-grammars` wasn't used or the addition was forced and the error ignored. Reverting this PR.

Please resubmit when the issue with the upstream grammar has been fixed.

NullVoxPopuli:
Is there a way to tell `vsc-ember-syntax` to take precedence over `daaain/Handlebars`? 

For ember projects using the `.hbs` extension, "Handlebars" is _just the wrong syntax_. There are specific file differences we could query for in package/workspace boundaries to determine if a project/package/workspace is an ember project or not, but does linguist respect that sort of querying?

gilest:
> Actually, now I look at this more closely, it appears this grammar was added _with_ that conflict already in place suggesting `script/add-grammars` wasn't used or the addition was forced and the error ignored. Reverting this PR.
> 
> Please resubmit when the issue with the upstream grammar has been fixed.

Oh, damn 😓 . I used `script/add-grammars` but then manually removed `text.html.handlebars` from `vsc-ember-syntax` in `grammars.yml` to avoid conflicting with the existing Handlebars grammar. Figured since the CI passed that was ok...

We discussed this in the [public feedback PR](https://github.com/gilest/linguist/pull/1#pullrequestreview-1639016982) before this one was raised.

Is there any way to proceed without removing handlebars from the upstream grammar? It's needed in-editor to support the legacy Ember-flavoured handlebars. But ideally it's ignored in Linguist... 😅

lildude:
> Is there a way to tell `vsc-ember-syntax` to take precedence over `daaain/Handlebars`?

Not really. Linguist, more specially the highlighting engine, only has support for one grammar scope per language and it doesn't have a mechanism for overriding. Generally grammars inherit and extend upon other grammars using a different scope name if they want something to behave differently from the original grammar.

The closest option to forcing is by replacing the current Handlerbars grammar with that supplied by vsc-ember-syntax. This comes with the consequence that that project will now be responsible for the syntax highlighting of all handlebar files, not just Ember users. 

> afaict, The upstream grammar _is correct_ for ember users, and no ember user should ever be using `daaain/Handlebars` (only a file extension is shared, and nothing more)
> 
> For ember projects using the `.hbs` extension, "Handlebars" is _just the wrong syntax_. 

And there's the problem. If the file and syntax is not handlerbars, the grammar should not be using the same scope name. It should be using its own scope and either completely defining everything it supports or inheriting from the handlebars grammar and overruling the differences. This is common practice.

> There are specific file differences we could query for in package/workspace boundaries to determine if a project/package/workspace is an ember project or not, but does linguist respect that sort of querying?

Not unless there's something in the files themselves that is unique to ember. Linguist looks at files in isolation (think gists) to try and determine the language which is then fed to the highlighting engine.

If the files are definitely not handlebars and should have their own grammar, then they can be added to linguist as their own language but the grammar will still need to use a unique scope name.

lildude:
> Is there any way to proceed without removing handlebars from the upstream grammar? It's needed for in-editor support of Ember-flavoured handlebars. But ideally it's ignored in Linguist... 😅

Yes, however that's just painting over the problem and incorrect behaviour I just highlighted [above](https://github.com/github-linguist/linguist/pull/6578#issuecomment-1844787033).

PR #6588, opened 2023-10-24
Add language: Scrypto

balda-rdx:
      -  https://github.com/search?q=path%3A*.rs+%22scrypto%3A%3A%22+fork%3Atrue+path%3A*.rs+fork%3Atrue&type=Repositories&ref=advsearch&l=&l=
    - Sample source(s):
      - https://github.com/radixdlt/scrypto-examples/tree/main/core/gumball-machine
      - https://github.com/radixdlt/scrypto-examples/tree/main/core/payment-splitter
    - Sample license(s):
  - [x] I have included a syntax highlighting grammar: https://github.com/radixdlt/wg-grammar

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1335,3 +1335,6 @@
 [submodule "vendor/grammars/zephir-sublime"]
 	path = vendor/grammars/zephir-sublime
 	url = https://github.com/phalcon/zephir-sublime
+[submodule "vendor/grammars/wg-grammar"]
+	path = vendor/grammars/wg-grammar
+	url = https://github.com/radixdlt/wg-grammar.git
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6157,6 +6157,11 @@ SAS:
   codemirror_mode: sas
   codemirror_mime_type: text/x-sas
   language_id: 328
+Scrypto:
+  type: programming
+  extensions:
+  - ".rs"
+  language_id: 1051511523
 SCSS:
   type: markup
   color: "#c6538c"
--- a/samples/Scrypto/Cargo.toml
+++ b/samples/Scrypto/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "payment_splitter"
+version = "0.2.0"
+edition = "2021"
+
+[dependencies]
+sbor = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.0" }
+scrypto = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.0" }
+
+[dev-dependencies]
+transaction = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.0" }
+radix-engine = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.0" }
+scrypto-unit = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.0" }
+
+[profile.release]
+opt-level = 'z'        # Optimize for size.
+lto = true             # Enable Link Time Optimization.
+codegen-units = 1      # Reduce number of codegen units to increase optimizations.
+panic = 'abort'        # Abort on panic.
+strip = true           # Strip the symbols.
+overflow-checks = true # Panic in the case of an overflow.
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[workspace]
+# Set the package crate as its own empty workspace, to hide it from any potential ancestor workspace
+# Remove this [workspace] section if you intend the package to be part of a Cargo workspace
--- a/samples/Scrypto/README.md
+++ b/samples/Scrypto/README.md
@@ -1,0 +1,240 @@
+# PaymentSplitter
+
+> Note: the instructions here were made for a previous version of Scrypto and are now outdated.
+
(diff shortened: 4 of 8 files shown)


lildude:
🤔 I have only the teeniest of weeniest of experience with Rust, but this looks a lot like normal Rust to me (kinda like a normal library/crate) and seems to be used more like a framework like Rails than a language. The current Rust grammar is certainly not struggling at all to highlight the few files I've looked at.

How is this "language" not Rust and why does it need its own language entry with TextMate compatible grammar (the one you've added is not usable by Linguist)?

dhedey:
> 🤔 I have only the teeniest of weeniest of experience with Rust, but this looks a lot like normal Rust to me (kinda like a normal library/crate) and seems to be used more like a framework like Rails than a language. The current Rust grammar is certainly not struggling at all to highlight the few files I've looked at.
> 
> How is this "language" not Rust and why does it need its own language entry with TextMate compatible grammar (the one you've added is not usable by Linguist)?

Hi @lildude  - thanks for picking up this review so quickly. It looks like the PR needs some tweaks and improved explanation - please accept our apologies for wasting your time with this preliminary review. We'll pick up the tweaks shortly, but first some explanation!

## Justification

You're quite right that Scrypto is on one level "simply Rust", but it is quite a substantial extension to Rust. And we believe that there is sufficient justification for it to both (A) use a different grammar in theory and (B) have it classified separately to Rust.

### Rust background

For a little bit of background, Rust has a powerful macro system for custom DSLs. [Procedural macros](https://doc.rust-lang.org/beta/reference/procedural-macros.html) basically take the Rust token stream, run arbitrary code, and spit out a token stream. So macros are in theory very general, which means that the Rust textmate grammar can't assume much about the input to a macro. It therefore does its best by using an error tolerant application of the standard Rust grammar. This does a reasonable effort job highlighting the inputs to most macros - but it can't capture the nuance of the grammar of any particular macro. See for example [typed-html](https://github.com/bodil/typed-html/blob/master/examples/stdweb/src/main.rs) where it fails to understand that the input is supposed to be html (followed by the `: Stdweb` suffix).

### Scrypto as a distinct grammar

Scrypto is a set of rust libraries, procedural macros and a distinct build chain. It is designed for building packages deployed to the Radix public network. It makes heavy use of procedural macros, for example the `#[blueprint]` attribute macro, `enable_method_auth!` function-like macro, and [others in this file](https://github.com/radixdlt/radixdlt-scrypto/blob/d16b3ffa65fea417e9c6d01d76456a0b36c060fa/examples/everything/src/lib.rs). These macros capture security-critical configuration, but the syntax highlighting offered by the Rust grammar isn't great. For example, it can't tell  `=>` is a separator, nor can it distinguish patterns in `enable_method_auth!`, such as OWNER or PUBLIC being keywords in the DSL. A typical Scrypto program often has lots of security-critical configuration in these macros, and we'd like to be able to add custom syntax highlighting to these macro segments clearer, and (eg) easier to review.

### Scrypto as a distinct classification

When it comes to classification, from a developer perspective, Scrypto code is built to be deployed to the public Radix network. [It is compiled with a special CLI](https://github.com/radixdlt/radixdlt-scrypto/tree/d16b3ffa65fea417e9c6d01d76456a0b36c060fa#installation), which creates special artifacts from the code, including a deployable package and interface definitions.

For developers, there is a very clear distinction between code which is "Scrypto" (which will be deployed to be run inside the Radix network) compared with general "Rust" application code (which may also be in their repository, but would eg form their web APIs or other parts of their application), and they'd welcome such a distinction in their repository records.

## Proposed PR changes

For now, taking a look at the PR, it isn't right, so we'll go away and rework the PR, do an internal review, and share with you once ready.

As a first step, we'd like to rework the PR to:
* Start with just using the current Rust syntax highlighting grammar (ie not pull in any custom code)
* Add a couple of very specific heuristics which have to be satisfied to label a file as "Scrypto" rather than "Rust"

This will satisfy the classification part, and we can look to add a Scrypto-specific textmate grammar in future.

Obviously very happy to have a discussion / hear your thoughts on this!

Thanks,
David + the RDX team

lildude:
Thanks @dhedey I think your explanation makes sense on first read, but I'll ruminate over it. Maybe things will become clearer with time as things evolve and popularity grows. 

Current usage is a very long way away from being considered if we [make the search more specific, and exclude forks and the radixdlt org](https://github.com/search?q=path%3A*.rs+%22+scrypto%3A%3A%22+NOT+is%3Afork+NOT+user%3Aradixdlt&type=code&ref=advsearch) (as it has an undue influence on the number of files).

lildude (changes requested):
Marking this as requesting changes (I use this for easier tracking), as has already been confirmed, this is incomplete.

A few tips:

1. Follow the CONTRIBUTING.md file
2. We will need a very accurate heuristic as this has an incredibly high chance of catching legit Rust files
3. I recommend two to four real highly representative samples to better train the classifier due to how close this is to Rust
4. If the diff suppresses the content of the sample; it's too big; find a smaller sample.
5. Samples should be of only the proposed language; we don't need the other related md and toml files as we have plenty and putting them in the wrong directory will pollute the classifier.
6. Samples must be directly in the language directory and not in a sub-dir like `src` as is the case in the two `lib.rs` files in this PR

Popularity will be checked each time I make a new release which happens approximately every 3-4 months, so there's no need to post comments or updates about popularity.

lildude:
Closing as the requested changes have not been made in the 10 months since requested. Feel free to open a new PR when you have the time to complete the PR.

PR #6614, opened 2023-11-12
Add support for Expect scripts

### LINKED ISSUE #6613: Add support for expect (as tcl?)
## Language name
Expect

## URL of example file
https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/lib/compat.exp

https://github.com/textmate/tcl.tmbundle

## Most popular extensions
`.exp`

## Detected language
None

## Commentary
Expect is an extension to the Tcl scripting language, used by the dejagnu testing framework, which is in turn used to implement the testsuites for many important GNU toolchain projects, such as GCC and GDB. It can also easily be added to any project that uses `automake` by simply adding `dejagnu` to the `AM_INIT_AUTOMAKE([])` invocation in `configure.ac`, and/or the `AUTOMAKE_OPTIONS` variable in `Makefile.am`. Note that I'm not sure if it should be treated as its own separate language, or if it would make more sense to just add `.exp` to the list of supported extensions for tcl, so I'm going to hold back from submitting a PR until we can develop some consensus as to how to answer that question.

## Links
- https://en.wikipedia.org/wiki/Expect
- https://core.tcl-lang.org/expect/index
- https://sourceforge.net/projects/expect/
- https://www.nist.gov/services-resources/software/expect

cooljeanius:
Closes #6613

## Description
Expect is an extension to the Tcl scripting language, so it can be categorized as Tcl. See #6613

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.exp+proc
    - Sample source(s):
      - https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/lib/compat.exp
    - Sample license(s):
      - GPLv3+ 

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -236,6 +236,11 @@ disambiguations:
     - '^\s*def(?:exception|impl|macro|module|protocol)[(\s]'
   - language: Euphoria
     named_pattern: euphoria
+- extensions: ['.exp']
+  rules:
+  - language: Tcl
+    pattern:
+    - '^\s*(?:set|expect|proc)\s'
 - extensions: ['.f']
   rules:
   - language: Forth
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6962,9 +6962,9 @@ Tcl:
   - owh
   - starfield
   interpreters:
+  - expect
   - tclsh
   - wish
-  - expect
   tm_scope: source.tcl
   ace_mode: tcl
   codemirror_mode: tcl
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6952,8 +6952,8 @@ Tcl:
   - ".sdc"
   - ".tcl.in"
   - ".tm"
-  - ".xdc"
   - ".exp"
+  - ".xdc"
   aliases:
   - sdc
   - xdc
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6953,6 +6953,7 @@ Tcl:
   - ".tcl.in"
   - ".tm"
   - ".xdc"
+  - ".exp"
   aliases:
   - sdc
   - xdc
@@ -6962,6 +6963,7 @@ Tcl:
   interpreters:
   - tclsh
   - wish
+  - expect
   tm_scope: source.tcl
   ace_mode: tcl
   codemirror_mode: tcl
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -6949,12 +6949,13 @@ Tcl:
   extensions:
   - ".tcl"
   - ".adp"
+  - ".exp"
   - ".sdc"
   - ".tcl.in"
   - ".tm"
-  - ".exp"
   - ".xdc"
   aliases:
+  - expect
   - sdc
   - xdc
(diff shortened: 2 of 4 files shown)


cooljeanius:
OK I think this is ready for review now; I don't think a new heuristic is necessary, since I don't see anything else using the `.exp` extension that would need to be distinguished from.

Alhadis (changes requested):
> I don't think a new heuristic is necessary, since I don't see anything else using the `.exp` extension that would need to be distinguished from.

I just remembered that `.exp` is sometimes used in test-suites to contain the <samp><ins>exp</ins></samp>ected output of a test run. It's usually paired with a file ending in `.act` (to denote the <samp><ins>act</ins></samp>ual output from a test), which is compared with the `.exp` file using the <samp>diff(1)</samp> utility.

There also appear to be [roughly ~203k search results](https://github.com/search?q=path%3A**%2F%3F*.exp+NOT+%2F%5E%5Cs*%28%3F%3Aset%7Cexpect%29%5Cs%2F&type=code) for the `.exp` extension that doesn't include lines beginning with `set` or `expect`. They don't appear to be fixtures for a test-suite either, so this might warrant a more in-depth investation.

cooljeanius:
> > I don't think a new heuristic is necessary, since I don't see anything else using the `.exp` extension that would need to be distinguished from.
> 
> I just remembered that `.exp` is sometimes used in test-suites to contain the expected output of a test run. It's usually paired with a file ending in `.act` (to denote the actual output from a test), which is compared with the `.exp` file using the diff(1) utility.
> 
> There also appear to be [roughly ~203k search results](https://github.com/search?q=path%3A**%2F%3F*.exp+NOT+%2F%5E%5Cs*%28%3F%3Aset%7Cexpect%29%5Cs%2F&type=code) for the `.exp` extension that doesn't include lines beginning with `set` or `expect`. They don't appear to be fixtures for a test-suite either, so this might warrant a more in-depth investation.

Hm, so maybe just a really simple heuristic then? e.g. `'^\s*(?:set|expect|proc)\s'` or something... (I'm not very good at regexes, so please correct me if I got it wrong)

Alhadis:
> Hm, so maybe just a really simple heuristic then?

Actually, I was thinking `.exp` might need to be registered as a [generic extension](https://github.com/github-linguist/linguist/blob/9358f090deb76511b613ab024b704c1c0f13a4c2/lib/linguist/generic.yml). This is a more esoteric feature of Linguist that's called upon when an extension has widespread usage in unrelated contexts, but none of them have enough adoption individually to merit their own entries in `languages.yml`. The original use-case was to support Solidity's `.sol` extension without misclassifying the numerous repositories that used `.sol` as an abbreviated suffix for _"solution"_.

However, generic extensions have the caveat that an accurate, airtight heuristic is needed, preferably one that matches a pattern or string that's all but guaranteed to appear in the files that we're interested in supporting. The regex <code>^\s*(?:set|expect|proc)\s</code> will match any line that happens to begin with the word "set", "expect" or "proc", followed by a whitespace character (including a newline). Needless to say, that's very, *very* error-prone; even if we weren't looking to add a generic extension, I still wouldn't recommend such an open-ended heuristic.

I'm happy to help you with penning the regex, but that assumes there even _is_ a predictable and guaranteed pattern unique to Expect scripts.

PR #6665, opened 2024-01-02
Add BitBake grammar and missing extensions

ParksProjets:
## Description

The BitBake language is a metadata language created for the Yocto project. Support in linguist was added in 2014 (PR #1129), but at the time only to solve conflicts with BlitzBasic. Only the `.bb` extension was implemented without any grammar.

The Yocto project now has an [official VSCode extension](https://github.com/yoctoproject/vscode-bitbake) which includes a TextMate grammar. This PR adds the BitBake grammar from this repository (MIT License).

This PR also adds the following file extensions for the BitBake language:
 - `.bbappend`
 - `.bbclass`
 - ~`.conf`~
 - `.inc`

The two first extensions are pretty unique so no heuristics was needed. ~For `.conf`, I added it to `generic.yml` because a lot of Linux configuration files are using it (and are not BitBake). I tried to make heuristics that match BitBake specific assignments (e.g., `A += ...`) which are usually not present in generic conf files.~ Concerning `.inc`, there are many languages having this extension so making the heuristics was tricky. The current heuristics is a bit too restrictive (that's why I have added two `.inc` sample files), but still matches the majority (~ 95%) of BitBake `.inc` files in my experiments.

Finally, I fixed the `.bb` heuristics to reduce BlitzBasic misclassification. In my experiments, the error rate was less than 1%.

I made may experiments using the following repositories:
 - https://github.com/openembedded/openembedded-core
 - https://github.com/openembedded/meta-openembedded
 - https://github.com/agherzan/meta-raspberrypi

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.bbappend
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.bbclass
      - ~https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.conf+MACHINEOVERRIDES~
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.inc+SRC_URI
    - Sample source(s):
      - https://github.com/agherzan/meta-raspberrypi/blob/master/dynamic-layers/qt5-layer/recipes-qt/qt5/qtbase_%25.bbappend
      - https://github.com/openembedded/openembedded-core/blob/master/meta/classes-recipe/cmake.bbclass
      - ~https://github.com/agherzan/meta-raspberrypi/blob/master/conf/machine/raspberrypi4-64.conf~
      - https://github.com/openembedded/openembedded-core/blob/master/meta/conf/distro/include/tclibc-newlib.inc
      - https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-graphics/xorg-driver/xorg-driver-common.inc
    - Sample license(s): MIT
      - https://github.com/openembedded/openembedded-core/blob/master/LICENSE.MIT
      - https://github.com/agherzan/meta-raspberrypi/blob/master/COPYING.MIT

  - Old: None
  - New: https://github.com/yoctoproject/vscode-bitbake (MIT License)

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -103,7 +103,7 @@ disambiguations:
   - language: BlitzBasic
     pattern: '(<^\s*; |End Function)'
   - language: BitBake
-    pattern: '^\s*(# |include|require)\b'
+    pattern: '^(# |include|require|inherit)\b'
   - language: Clojure
     pattern: '\((def|defn|defmacro|let)\s'
 - extensions: ['.bf']
@@ -371,6 +371,8 @@ disambiguations:
     pattern:
     - '(?i:^\s*\{\$(?:mode|ifdef|undef|define)[ ]+[a-z0-9_]+\})'
     - '^\s*end[.;]\s*$'
+  - language: BitBake
+    pattern: '^inherit(\s+[\w.-]+)+\s*$'
 - extensions: ['.json']
   rules:
   - language: OASv2-json
--- a/.gitmodules
+++ b/.gitmodules
@@ -1205,6 +1205,9 @@
 [submodule "vendor/grammars/vscode-antlers-language-server"]
 	path = vendor/grammars/vscode-antlers-language-server
 	url = https://github.com/Stillat/vscode-antlers-language-server.git
+[submodule "vendor/grammars/vscode-bitbake"]
+	path = vendor/grammars/vscode-bitbake
+	url = https://github.com/yoctoproject/vscode-bitbake.git
 [submodule "vendor/grammars/vscode-brightscript-language"]
 	path = vendor/grammars/vscode-brightscript-language
 	url = https://github.com/rokucommunity/vscode-brightscript-language.git
--- a/grammars.yml
+++ b/grammars.yml
@@ -1081,6 +1081,8 @@ vendor/grammars/vscode-TalonScript:
 - source.talon
 vendor/grammars/vscode-antlers-language-server:
 - text.html.statamic
+vendor/grammars/vscode-bitbake:
+- source.bb
 vendor/grammars/vscode-brightscript-language:
 - source.brs
 vendor/grammars/vscode-cadence:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -598,7 +598,7 @@ Bison:
 BitBake:
   type: programming
   color: "#00bce4"
-  tm_scope: none
+  tm_scope: source.bb
   extensions:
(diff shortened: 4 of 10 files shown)


lib/linguist/heuristics.yml
```diff
@@ -164,6 +164,10 @@ disambiguations:
   rules:
   - language: Gerber Image
     pattern: '^[DGMT][0-9]{2}\*(?:\r?\n|\r)'
+- extensions: ['.conf']
+  rules:
+  - language: BitBake
+    pattern: '^[A-Z_]+(([\[:][a-z])|(\s*[?+.]=))'
```
lildude, line 170: Are you 100% sure this will only ever match BitBake files? If there's a chance this will match another language's files, it's best to not add this and the extension to `generic.yml`.

ParksProjets: I am not sure at all. On Linux people tend to use `.conf` extension for configuration files, which can be any format. I tried to make a restrictive regex, but the chance of having some false positive are great.

I will remove this extension from BitBake.

lib/linguist/heuristics.yml
```diff
@@ -371,6 +371,8 @@ disambiguations:
     pattern:
     - '(?i:^\s*\{\$(?:mode|ifdef|undef|define)[ ]+[a-z0-9_]+\})'
     - '^\s*end[.;]\s*$'
+  - language: BitBake
+    pattern: '^(include|require|inherit)\s+[a-z$]'
```
lildude, line 375: This is also waaaay too generic and definitely mis-identifies files from other languages. This is evident from [this search](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.inc+%2F%5E%28include%7Crequire%7Cinherit%29%5Cs%2B%5Ba-z%24%5D%2F) which shows many Makefiles. As with `.conf`, this is a very generic extension so we need to be incredibly precise with the regex. If in doubt, don't include it.

ParksProjets: Ok I restricted the regex to the `inherit` part only, which is likely to be present in BitBake `.inc` recipes. It will not match BitBake `.inc` files used in the configuration part but as we removed `.conf` files it's no big deal now.

`inherit` is not a Makefile keyword, this new regex seems to [only match BitBake files](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.inc+%2F%5Einherit%28%5Cs%2B%5B%5Cw.-%5D%2B%29%2B%5Cs*%24%2F).

lildude (approved):
LGTM. Thanks.

PR #6670, opened 2024-01-03
Add .txtpb extension for Protocol Buffer Text Format

dougthor42:
Add `.txtpb` extension for Protocol Buffer Text Format.

## Description

The `.txtpb` extension is the canonical and preferred extension for text protos. See https://protobuf.dev/reference/protobuf/textformat-spec/#text-format-files (emphasis mine):

> **.txtpb is the canonical text format file extension and should be preferred to the alternatives**. This suffix is preferred for its brevity and consistency with the official wire-format file extension .binpb. The legacy canonical extension .textproto still has widespread usage and tooling support. Some tooling also supports the legacy extensions .textpb and .pbtxt. All other extensions besides the above are strongly discouraged; in particular, extensions such as .protoascii wrongly imply that text format is ascii-only, and others like .pb.txt are not recognized by common tooling.

## Checklist:

      - https://github.com/search?q=path%3A*.txtpb&type=code (60 results)
      - This admittedly does not hit the "in use in at least 200 unique repositories " recommendation.
      - In our private [`qh-lab`](https://github.com/qh-lab) organization we have >250 text proto files and we _don't use the `txtpb` extension only because GitHub linguist doesn't highlight it_. I'm sure that others feel the same way. It's a chicken and egg problem :upside_down_face:.
    - Sample source(s):
      - unknown (simply copied an existing sample source and renamed the extension)
    - Sample license(s):
      - same as existing source
    - N/A - extension does not currently exist in linguist.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -5447,6 +5447,7 @@ Protocol Buffer Text Format:
   - ".textproto"
   - ".pbt"
   - ".pbtxt"
+  - ".txtpb"
   tm_scope: source.textproto
   ace_mode: text
   language_id: 436568854
--- a/samples/Protocol Buffer Text Format/schema.txtpb
+++ b/samples/Protocol Buffer Text Format/schema.txtpb
@@ -1,0 +1,271 @@
+feature {
+  name: "company"
+  value_count {
+    min: 1
+    max: 1
+  }
+  type: BYTES
+  domain: "company"
+  presence {
+    min_count: 1
+  }
+}
+feature {
+  name: "payment_type"
+  value_count {
+    min: 1
+    max: 1
+  }
+  type: BYTES
+  domain: "payment_type"
+  presence {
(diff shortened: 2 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -5447,6 +5447,7 @@ Protocol Buffer Text Format:
   - ".textproto"
   - ".pbt"
   - ".pbtxt"
+  - ".txtpb"
```
dougthor42, line 5450: > The exception is the primary extension: it should always be first.

The canonical and preferred extension is `.txtpb` (see PR message), but the _most common_ extension appears to be `.textproto`. Which should be considered the "primary" extension and thus placed first?

lildude:
Closing as duplicate of https://github.com/github-linguist/linguist/pull/6566

dougthor42:
Doh! Sorry for not doing a search beforehand...

Issue #6451, opened 2023-06-16
Modula-2 file extension '.def' is missing


trijezdci:
https://github.com/github-linguist/linguist/blob/d352058f979caab69b5c0876cd4992ce43a29132/lib/linguist/languages.yml#L4222

Modula-2 separates interface and implementation in separate files.

Interfaces (historically but incorrectly called definition modules) have file extension '.def'.
Implementation and program modules have file extension '.mod'.

This is similar to C's '.h' and '.c' files but unlike C, the modules are NOT preprocessor includes, they are separate compilation units and it is entirely impossible to write Modula-2 libraries without having '.def' files. This is not a matter of choice or style, this is an absolute requirement.

Not supporting '.def' as a file extension for Modula-2 means that the entire syntax highlighting for Modula-2 is completely broken as it is 50% incomplete.

So, can you please add recognition for the '.def' file extension.

I had reported this many years ago, but apparently nothing has happened since and I can't even find the report now. It would be nice if this wasn't just swept under the carpet again. It should be rather easy to add.

Thank you in advance.

lildude:
> So, can you please add recognition for the '.def' file extension.
>
> I had reported this many years ago, but apparently nothing has happened since and I can't even find the report now.

I can with a very easy search: https://github.com/github-linguist/linguist/issues/3657 😉 It was closed automatically back when we used that bot. That issue also went way into the weeds all outside of Linguist's scope.


And in much as my response then, the same applies now: Linguist is a community-driven project, so in short: if you want it, submit a PR to add support or implement an override and wait patiently until someone else does.

You can find details for adding support in the CONTRIBUTING.md file.

One thing to keep in mind is `.def` is a very generic extension likely to be used by many languages. We can add support but it will probably require a very precise heuristic (aka regex) for identifying the files as precisely as possible to reduce the chances of misclassifying other languages.

If you're not prepared to add support yourself, please add and fill in the "Feature request" issue template to the OP of this issue as you should have used it when opening this issue.

trijezdci:
I have forked the Sublime Text files that Linguist appears to be using for Modula-2, and I have made several corrections so that this reflects the classic Modula-2 Language originally published by Prof. Niklaus Wirth at ETH Zurich in his book "Programming in Modula-2", published by Springer Verlag. This book contains the language report and the language dialect it describes is known as PIM Modula-2, where PIM is a shorthand for the title of the book.

https://github.com/trijezdci/Sublime-Modula-2

[...]

Issue #3657, opened 2017-05-30
Modula-2 file extension recognition incomplete


trijezdci:
Modula-2 has two source file types:

* definition modules with file ending .def/.DEF
* implementation modules with file ending .mod/.MOD

Your listing only has the .mod for Modula-2, the .def is missing.

https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L2655

Github doesn't therefore syntax highlight the definition modules.

Can you add the .def wherever that needs to be added, please?

thanks a lot in advance.

lildude:
> Can you add the .def wherever that needs to be added, please?

We'd happily accept a PR with the addition from someone who knows Modula-2.  A quick cursory glance suggests adding a `.def` extension may be all we need, but I'd prefer it if someone with more experience with the language to offer up the PR.

trijezdci:
Hi and thanks for your response.

I know Modula-2 in and out, but I don't know Ruby and I don't have the time right now to get side tracked with Ruby just to teach your software to recognise .def as a Modula-2 source file ending.

Modula-2 has two ancient dialects, called PIM (from 1979) and ISO (from 1992), and there is a recent revision of the language (from ~2010) adding a third dialect. Also, most of the available PIM and ISO compilers have added their own predefined identifiers for additional built-in types.

For the Python based Pygments syntax highlighting package I had written and contributed a multi-dialect Modula-2 lexer. Modula-2 developers can control the dialect selection by dialect tags embedded within Modula-2 comments at the beginning of a Modula-2 source file which the Pygments lexer understands.

For example ...

(*!m2pim*) indicates PIM Modula-2
(*!m2iso*) indicates ISO Modula-2
(*!m2r10*) indicates Modula-2 Revision 2010
(*!m2iso+gm2*) indicates ISO Modula-2 with additions by the GNU Modula-2 compiler
(*!m2pim+mocka*) indicates PIM Modula-2 with additions by the MOCKA Modula-2 compiler

etc etc etc

The lexer also has an Algol-60 publishing mode where source code is rendered in the style of the Algol-60 report which set a standard for rendering algorithms in scientific papers. Modula-2 uses all uppercase reserved words and predefined identifiers which was actually intended to mean boldface. Thus in Algol-60 mode the Pigments package renders all Modula-2 reserved words in lowercase boldface, all predefined identifiers in lowercase boldface italic, and comments in grey. For scientific papers this is preferable to colours and an established de-facto standard used to this day.

[...]

PR #6671, opened 2024-01-04
Add missing Modula-2 *.def extension

dkearns:
Modula-2 splits the definition of a module into separate interface and implementation files having extensions `*.def` and `*.mod` respectively.

Linguist currently only detects `*.mod` files.  This PR adds support for `*.def` file detection.

See #3657 and #6451 for related discussions.

I'm not sure why `END` is in those patterns.  Assuming a top down search the `MODULE` line will obviously be found first but I may be missing something about the implementation.

## Checklist:

- [ x] **I am adding a new extension to a language.**
  - [ x] The new extension is used in hundreds of repositories on GitHub.com
      
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.def+%2F%5E%5Cs*DEFINITION+MODULE%2F
  - [ x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://github.com/gcc-mirror/gcc/blob/master/gcc/m2/gm2-libs/Environment.def
      - https://github.com/gcc-mirror/gcc/blob/master/gcc/m2/gm2-libs/Environment.mod
    - Sample license(s): GPL
  - [x ] I have included a change to the heuristics to distinguish my language from others using the same extension.


lib/linguist/heuristics.yml
```diff
@@ -195,6 +195,10 @@ disambiguations:
     #  : dependency
     # path/file.ext1 : some/path/../file.ext2
     pattern: '([\/\\].*:\s+.*\s\\$|: \\$|^[ %]:|^[\w\s\/\\.]+\w+\.\w+\s*:\s+[\w\s\/\\.]+\w+\.\w+)'
+- extensions: ['.def']
+  rules:
+  - language: Modula-2
+    pattern: '^\s*(?i:DEFINITION\s+MODULE|END)\s+[\w\.]+\s*;'
```
lildude, line 201: This heuristic will never be used as this is the only user of this extension within linguist which in itself is potentially a problem as `.def` is an incredibly generic extension used by many things.

If you're 100% certain this regex will only match Modula-2 files, then adding this extension to the `generic.yml` file will ensure this regex takes effect and is used. If there's a risk this misidentifies other languages, it's best to leave off this extension and use an override when needed.

lildude (changes requested):
See inline comment

lildude (changes requested):
Please add a heuristic test for this extension.

lildude:
Sorry. I should have been clearer... we need the heuristic test and fixtures as implemented by other generic extensions. See what has been done for the likes of `.sol` and `.cmp`.

PR #6620, opened 2023-12-03
Add Praat scripting language

JJWRoeloffs:
Adding support for the Praat scripting language. This is a domain-specific language that is quite commonly used in academia for linguistic research, but that isn't that commonly known outside that sphere.

I am not affiliated with the project - I'm just a student who had to learn the language for a BA course, and was surprised github did not support it.

## Description
If I did everything correctly, this commit adds support for the Praat scripting language in the way described by `CONTRIBUTING.md`

## Checklist:
- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension (2.5K files):
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.praat
      -  I also added `praatscript`, `praat_script` and `praat-script`, which are less frequently used, but won't likely overlap with anything. I included them because some more common grammars include them [example1](https://github.com/mauriciofigueroa/praatSublimeSyntax/blob/0077112a7435084f6281de03d16de894a8ecc1bf/Praat.tmLanguage#L4) [example2](https://github.com/massCodeIO/massCode/blob/ae9eb6b2569357eadd27e363dd403f5b24b63c2d/src/renderer/components/editor/grammars/praat.tmLanguage.json#L4)
      - I'll also list a few projects using this language, alongside their star count:
        - [FastTrack](https://github.com/santiagobarreda/FastTrack) (51) - a praat plugin
        - [SpeCT](https://github.com/lennes/spect) (49) - Speech Corpus Toolkit for Praat
        - [Praatsouce](https://github.com/kirbyj/praatsauce) (21) - Praat-based tools for spectral analysis
        - [feelins/PraatScripts](https://github.com/feelins/Praat_Scripts) (169) - A sample repository containing example scripts
        - [stylerw's praat scripts](https://github.com/stylerw/styler_praat_scripts) (87) - A repository with the praat scripts this person finds useful for research.
        - [fieldDB's praat scripts](https://github.com/FieldDB/Praat-Scripts) (50) - Another repository of scripts some researcher found useful.
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample license(s):
      - I wrote these scripts myself as a part of a Research inter-ship, and am happy to re-licence them under linguist's MIT licence (which I am allowed to do following the terms of the internship).
  - [x] I have included a syntax highlighting grammar: [URL to grammar repo](https://github.com/JJWRoeloffs/Praat-tmLanguage)
[...]

PR #6719, opened 2024-02-25
Add TextGrid data file format

JJWRoeloffs:
## Description
This adds the TextGrid data format with the `.TextGrid` file extension. This is a text-based data format used in linguistic research for aligning text (usually transcriptions) with audio. It has over 100K files checked in on GitHub. The official documentation for it can be found here: https://www.fon.hum.uva.nl/praat/manual/TextGrid_file_formats.html

As the format suggests, there are indeed more types of `ooTextFile`, several dozen in fact, but TextGrids are the only ones that are frequently checked into source control.

## Checklist:
    - Search results for each `.TextGrid`:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.TextGrid
    - Sample license(s):
    - I wrote these myself for unit tests as a part of a Research internship, and am happy to re-licence them under linguist's MIT licence (which I am allowed to do following the terms of the internship).
      - This one was already done by #6620 
  - [x] I have added a color
    - Hex value: `#c8506d`
    - Rationale: Same as Praat, as it is a data format associated with Praat. If it can "inherit" this it might be better, but I cannot find how. Also, it's a data file, so it shouldn't matter too much, as it's not displayed by default.

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7125,6 +7125,14 @@ Text:
   tm_scope: none
   ace_mode: text
   language_id: 372
+TextGrid:
+  type: data
+  color: "#c8506d"
+  tm_scope: source.textgrid
+  ace_mode: text
+  extensions:
+  - ".TextGrid"
+  language_id: 965696054
 TextMate Properties:
   type: data
   color: "#df66e4"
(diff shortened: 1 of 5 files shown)


lildude (approved):
LGTM. Thanks.

> * Rationale: Same as Praat, as it is a data format associated with Praat. If it can "inherit" this it might be better, but I cannot find how. Also, it's a data file, so it shouldn't matter too much, as it's not displayed by default.

Makes sense to me. We support grouping of languages using the `group` attribute, but this isn't really necessary as as you've said, this is a data format so ignored by default. It also doesn't transfer the colour so wouldn't help.

PR #6724, opened 2024-02-28
Remove unnecessary alias for R

InSyncWithFoo:
`R` was added as an alias of `R` 10 years ago in #783. I'm not sure what the problem it fixed back then was but it seems to be unnecessary now in 2024.


lildude (approved):
Yup, completely unnecessary now.

PR #6738, opened 2024-03-07
Switch Ruby VS Code extensions

spenserblack:
The `rebornix.Ruby` extension is deprecated in favor of `Shopify.ruby-lsp`.


lildude (approved):
I was going to do this as part of the next release, but you beat me to it 😁 Thanks 🙇 .

Merging now as peeps can gain from this immediately without needing a new release.

PR #6748, opened 2024-03-12
Add SQLx query files as generated

mercxry:
Mark `**/.sqlx/query-*.json` files as generated and added a test for it.

## Description

[SQLx is a Rust SQL library](https://github.com/launchbadge/sqlx) (11.4k stars) which generates `**/.sqlx/query-*.json` files in offline mode (enabled by default).
These are used to be able to compile a project without requiring the development database to be online.

I would like Github to automatically hide these files from diffs, since it could generate lots of files and it's a bit annoying to hide them manually one by one in each PR.

## Checklist:

      - https://github.com/search?type=code&q=path%3A**%2F.sqlx%2Fquery-*.json+NOT+is%3Afork
    - Sample source(s):
      - https://github.com/rust-lang/docs.rs/blob/master/.sqlx/query-2b8b1aae3740a05cb7179be9c7d5af30e8362c3cba0b07bc18fa32ff1a2232cc.json
    - Sample license(s): MIT
  - [ ] ~~I have included a change to the heuristics to distinguish my language from others using the same extension.~~

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -837,7 +837,7 @@ module Linguist
     #
     # Returns true or false.
     def generated_sqlx_query?
-      !!name.match(/^.*\.sqlx\/query-.+\.json$/)
+      !!name.match(/(?:^|.*\/)\.sqlx\/query-.+\.json$/)
     end
   end
 end
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -829,7 +829,7 @@ module Linguist
 
     # Internal: Is this a generated SQLx query file?
     #
-    # SQLx is a Rust SQL library which generates `**/.sqlx/queries-*.json` files
+    # SQLx is a Rust SQL library which generates `**/.sqlx/query-*.json` files
     # in offline mode (enabled by default).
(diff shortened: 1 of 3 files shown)


lib/linguist/generated.rb
```diff
@@ -825,5 +826,18 @@ def extract_html_meta(match)
...
+    #
+    # SQLx is a Rust SQL library which generates `**/.sqlx/query-*.json` files
+    # in offline mode (enabled by default).
+    #
+    # These are used to be able to compile a project without requiring
+    # the development database to be online.
+    #
+    # Returns true or false.
+    def generated_sqlx_query?
+      !!name.match(/^.*\.sqlx\/query-.+\.json$/)
```
lildude, line 840: This will match `foo.sqlx/query-blah.json`. Note that the holding directory is `foo.sqlx`. Is this intentional?

mercxry: Ah nice one, nope that's not my intention!

I wanted to match that path even if it is under one or multiple level of sub-directories, I fixed that now 👍

lildude (approved):

PR #6778, opened 2024-03-29
Detect .A86 Files As Assembly Language

Piggy63:
Added .A86 to the list of extensions for assembly language. .A86 was an extension used by 8086 assembly language source files in the late 70s and early 80s.

## Description
This pull request makes linguist detect .A86 files as x86 assembly language source code. People used to cross assemble 8086 code from 8080/Z80 machines, and .ASM was already reserved for the native assembly language of those machines, so they used .A86 instead. CP/M-86's ASM-86 and Seattle Computer Products' 8086 Cross Assembler, two of the earliest 8086 assemblers, both used the .A86 extension. GitHub currently does not detect .A86 files as assembly language.

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.A86+JNZ
    - Sample source(s):
      - https://github.com/antlr/grammars-v4/blob/master/asm/asm8086/examples/ROM.A86
      - https://github.com/erkyrath/infocom-zcode-terps/blob/master/cpm86/zip.a86
      - https://github.com/the-grue/OpenDOS/blob/master/IBMDOS/FDOS.A86
    - Sample license(s):

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -412,6 +412,7 @@ Assembly:
   - nasm
   extensions:
   - ".asm"
+  - ".a86"
   - ".a51"
   - ".i"
   - ".inc"


lildude (changes requested):
You need to add the sample you reference to the PR 😉

lildude:
> > You need to add the sample you reference to the PR 😉
> 
> With all due respect, I do not get why a sample is needed. There is no difference between a .ASM and a .A86 file, the only difference is the 3-letter extension. However, I have added a sample as requested.

Whilst this language is the only user of this extension now and there are other extensions for the same language, it may not be the case in future in which case we may need to fall back to to the classifier which is trained off these samples. So whilst it seems pointless now, we can’t say that in future so it’s best to preempt things and prepare for the future now whilst this is fresh in people’s minds.

Piggy63:
I looked around and saw that a lot of other pull requests included multiple samples, so I've added in another .A86 file.

> Whilst this language is the only user of this extension now and there are other extensions for the same language, it may not be the case in future in which case we may need to fall back to to the classifier which is trained off these samples. 

I see now.

By the way, I have a few questions:
1. What is the "Pending Popularity" tag? My understanding is that the language is not popular enough for the PR to be merged, is this correct? If so, how many repos are required before this PR can be considered? I read in the contribution guideline that the number is 200, however we have about 1.4K .A86 files on GitHub (which should be more than 200 repos).
2. Do I have to tick all boxes of the checklist? I feel like the last one is sort if irrelevant considering that this extension is currently only used by one language.

Sincerely,
Pig

lildude:
> 1\. What is the "Pending Popularity" tag? My understanding is that the language is not popular enough for the PR to be merged, is this correct? If so, how many repos are required before this PR can be considered? I read in the contribution guideline that the number is 200, however we have about 1.4K .A86 files on GitHub (which should be more than 200 repos).

Your understanding is correct, but it applies per extension too. https://github.com/github-linguist/linguist/issues/5756 as referenced in the CONTRIBUTING.md file details how we're determining popularity at the moment. We re-assess whenever we're close to making a new release which happens approximately every 3-4 months inline with the GitHub Enterprise Server release cycle.

> 2\. Do I have to tick all boxes of the checklist? I feel like the last one is sort if irrelevant considering that this extension is currently only used by one language.

Nope. Only those that apply. This last point doesn't apply as the extension is unique.

lildude (changes requested):
samples/Assembly/zip.a86 is too big (as it's suppressed in the diff by default). Please replace it with a smaller sample.

Piggy63:
Thanks for the fast reply! I have replaced the big file with a smaller one, hopefully it should be fine now and no more changes are required.

Sincerely,
Pig

lildude:
Looks good. The only thing needed now is for the license to be specified for each sample in the template. We can only accept open-source licensed files and the license needs to be specified in the template.

Piggy63:
> Looks good. The only thing needed now is for the license to be specified for each sample in the template. We can only accept open-source licensed files and the license needs to be specified in the template.

Bruh. Ok, well, how do I do that? I looked at a few other PRs and can't find where they've included the license. I'm thinking about writing my own .A86 samples at this point.

Sincerely,
Pig

lib/linguist/languages.yml
```diff
@@ -412,6 +412,7 @@ Assembly:
   - nasm
   extensions:
   - ".asm"
+  - ".a86"
   - ".a51"
```
Alhadis, line 416: Linguist requires extension lists to be sorted case-sensitively (sans the first entry, which is considered the “primary” extension), which means `.a86` should follow `.a51`, not precede it.

~~~diff
   extensions:
   - ".asm"
   - ".a51"
+  - ".a86"
~~~

Alhadis (changes requested):

lib/linguist/languages.yml
Piggy63: I didn't see that, sorry. Will change it later. However, afaik, .A51 (Intel 8051 I think) is not the same "language" as x86 assembly (8086, i286, i386 and co.). I see that Motorola 68K has its own thing, so maybe Intel 8051 should also have its own definition/whatever you call it? I can clearly see that GitHub is trying to apply the x86 syntax highlighting to the sample .A51 file. Maybe I should open an issue for that?

Again, sorry for not reading the contribution guidelines properly before opening this PR.

Alhadis: > I see that Motorola 68K has its own thing

`Motorola 68K` is [grouped](https://github.com/github-linguist/linguist/blob/4ac734c15a96f9e16fd12330d0cb8de82274f700/lib/linguist/languages.yml#L30-L31) under `Assembly`, which means M68K falls under the Assembly umbrella for the purposes of language classification.

https://github.com/github-linguist/linguist/blob/4ac734c15a96f9e16fd12330d0cb8de82274f700/lib/linguist/languages.yml#L4385-L4399

The reason for making this distinction in the first place is to employ more accurate syntax highlighting (note the `tm_scope` field above). E.g., XML property lists use a specialised grammar for highlighting the [specialised XML subset](https://en.wikipedia.org/wiki/Property_list#Format) used by [<samp>plist(5)</samp>](https://www.unix.com/man-page/OSX/5/plist), even though there's nothing wrong with XML's usual grammar.

[^1]: `tm_scope` refers to the [<samp>scopeName</samp> field](https://macromates.com/manual/en/language_grammars#example_grammar) of TextMate-compatible grammars, which acts as a GUID of sorts.

PR #5401, opened 2021-05-31
Add KoLMafia ASH language (fixes #5022)

midgleyc:
Adds the new language "KoLMafia ASH".

Closes #5022.

## Description
ASH is a currently unrecognised language used for scripting KoLMafia, a program for playing the online game "Kingdom of Loathing". There is documentation available at the [KoLMafia wiki](https://wiki.kolmafia.us/index.php?title=Main_Page).

All ASH scripts have the ".ash" extension, currently owned by AGS Scripts.

As this is a conflicting extension, I should run the "Bayesian classifier" and add a heuristic if it doesn't classify well enough. How do I do this? Compiling into docker and running that against the files in the samples directory found them as "AGS Script", which makes me think that's not the right thing to do.

## Checklist:
- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?p=1&q=extension%3A.ash+get_property&type=Code
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - https://github.com/Ezandora/Guide/blob/Release/relay/relay_Guide.ash
      - https://github.com/Ezandora/Sweet-Synthesis/blob/master/Source/synthesis.ash
      - https://github.com/Aenimus/newDG/blob/master/Release/scripts/newDG.ash
    - Sample license(s):
      - MIT (newDG)
      - Unlicense (relay_Guide, synthesis) (note that this is the public domain equivalent "unlicense", not unlicensed)
  
  - [x] I have included a syntax highlighting grammar: https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_format=json&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fmidgleyc%2Fvscode-kolmafia-ash%2Fmaster%2Fsyntaxes%2Fash.tmLanguage.json&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2FAenimus%2FnewDG%2Fmaster%2FRelease%2Fscripts%2FnewDG.ash&code=
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

- [x] **I am adding new or changing current functionality**
  
  - [ ] I have added or updated the tests for the new or changed functionality.

PR #6784, opened 2024-04-04
Add KoLmafia ASH with samples and heuristic

gausie:
## Description
I am currently missing a grammar - the closest thing that exists is here https://raw.githubusercontent.com/danielyxie/vscode-kolmafia-ash/master/syntaxes/ash.yml. How would you recommend I include that?

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ash+%2F%28%5C%24%5Ba-zA-Z%5D%2B%5C%5B%7Cvisit_url%7Csince+r%5Cd%2B%29%2F
    - Sample source(s):
      - http://svn.code.sf.net/p/zlib/code/scripts/zlib.ash
      - https://github.com/Veracity0/arrrbor-day/blob/main/scripts/Arrrbor%20Day.ash
    - Sample license(s):
      - None indicated, not typical to do so for this community
  - [x] I have included a syntax highlighting grammar
  - [x] I have added a color
    - Hex value: `#54bd32`
    - Rationale: Green picked from the lime featured in the KoLmafia icon

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1281,6 +1281,9 @@
 [submodule "vendor/grammars/vscode-just"]
 	path = vendor/grammars/vscode-just
 	url = https://github.com/skellock/vscode-just.git
+[submodule "vendor/grammars/vscode-kolmafia-ash"]
+	path = vendor/grammars/vscode-kolmafia-ash
+	url = https://github.com/midgleyc/vscode-kolmafia-ash.git
 [submodule "vendor/grammars/vscode-lean"]
 	path = vendor/grammars/vscode-lean
 	url = https://github.com/leanprover/vscode-lean
--- a/grammars.yml
+++ b/grammars.yml
@@ -1154,6 +1154,8 @@ vendor/grammars/vscode-jest:
 - source.jest.snap
 vendor/grammars/vscode-just:
 - source.just
+vendor/grammars/vscode-kolmafia-ash:
+- source.ash
 vendor/grammars/vscode-lean:
 - markdown.lean.codeblock
 - source.lean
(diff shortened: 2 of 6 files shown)


lildude:
There's already a PR for this - https://github.com/github-linguist/linguist/pull/5401. Maybe you can contribute to that PR or pull in the changes proposed there, keeping in mind all the comments on that PR.

gausie:
> There's already a PR for this - #5401. Maybe you can contribute to that PR or pull in the changes proposed there, keeping in mind all the comments on that PR.

Oh wow well done for remembering this. And sorry @midgleyc for missing your work! I think the search I've done, which is just a regex based on the heuristic I wrote, does a better job of reflecting the popularity.

I'll update this PR to use Chris's grammar.

gausie:
Ok that is done. I note that @midgleyc's PR seems to have run `script/list-grammars` but the guide for adding a new language does not include that as a step - would you like me to do so?

gausie:
> This is likely to fail the popularity test again -- I think the number of required repos has only increased since my PR several years ago.
> 
> A lot of repositories exist only on Sourceforge, and as GitHub doesn't (seem to?) list repos without recent contributions in the search, finding many of the existing scripts is going to be difficult. Even with them, the language likely isn't popular enough.

The search I have in the description is listing almost 1k files, not sure how to see individual repos but it does seem to be at least a hundred.

It would be really great to differentiate these from AGS Scripts!

lildude (changes requested):
We can't accept your samples for a few reasons:

1. they're too big; if the diff suppresses them, they're too big for our needs
2. they're not covered by an open source license.

All samples _must_ be covered by an open source license and this must be stated in the template and identifiable at the time in the source repo. We will _not_ accept samples without an indication of their license.

We're also going to need the cached license file. This should have been created when you ran `script/add-grammar` to add the grammar. Please add that file to the PR. If you can't find it, run this command:

https://github.com/github-linguist/linguist/blob/4ac734c15a96f9e16fd12330d0cb8de82274f700/script/add-grammar#L133-L134

As with the other PR, popularity still isn't sufficient for inclusion right now.

Brandylee24 (approved):

PR #6787, opened 2024-04-08
Add `Cargo.toml.orig` as TOML filename

spenserblack:
This is a file that is generated by commands like `cargo package`.

## Checklist:
  - [x] The new ~~extension~~ filename is used in hundreds of repositories on GitHub.com
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3ACargo.toml.orig
    - Sample source(s):
      - [source](https://github.com/makepad/makepad/blob/a978a7c97a56fad24f392bb899571a93f82b1011/libs/windows-core/Cargo.toml.orig)
    - Sample license(s):
      - [MIT](https://github.com/makepad/makepad/blob/a978a7c97a56fad24f392bb899571a93f82b1011/LICENSE)

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -60,6 +60,7 @@ module Linguist
       generated_net_specflow_feature_file? ||
       composer_lock? ||
       cargo_lock? ||
+      cargo_orig? ||
       flake_lock? ||
       node_modules? ||
       go_vendor? ||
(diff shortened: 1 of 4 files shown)


lildude:
@Alhadis is spot on. `teat_blob.rb` is essentially testing the helper methods used by the Blob class which in turn is doing the same things as the tests in `test_generated.rb` in this case, so there's no value in adding even more tests for generated to `test_blob.rb`, or `test_file_blob.rb` which is essentially the same, as that functionality isn't changing. Doing so only unnecessarily extends the time CI takes to run by effectively adding duplicate tests. This is one of a few places where a refactoring has left things in an incomplete and duplicating state. I've been meaning to find time to clean this all up for years 😁

lildude (approved):
LGTM. Thanks.

PR #6805, opened 2024-05-01
Allow configuring a repository's `max_tree_size` value

issyl0:
## Description

- For GitHub Code Scanning functionality, we want to be able to enable Default Setup on more repos. That relies on Linguist for the language detection.
- We found that in very large repos, the `MAX_TREE_SIZE` constant was not high enough.
- This makes the `max_tree_size` value configurable in both `initialize` and `incremental` methods, with a default value of 100,000.
- ❓ This is my first contribution to Linguist and I have no idea if this is "the way".

## Checklist:


lildude (approved):
Looks good to me especially as you've got further than I did previously 😆 

I only merge PRs when I'm getting close to making a new release, so whilst I'm approving this now, I won't merge it straight away.

issyl0:
Superseded by #6806 that's not from a fork.

Leif-Arild:
configurable-repository-max-tree-size

PR #6806, opened 2024-05-01
Make `max_tree_size` configurable per-repository

issyl0:
Supersedes #6805 since that was from my personal fork.

## Description

- For GitHub Code Scanning functionality, we want to be able to enable Default Setup on more repos. That relies on Linguist for the language detection.
- We found that in very large repos, the `MAX_TREE_SIZE` constant was not high enough.
- This makes the `max_tree_size` value configurable in both `initialize` and `incremental` methods, with a default value of 100,000.
- ❓ This is my first contribution to Linguist and I have no idea if this is "the way".

## Checklist:

### DIFF
--- a/lib/linguist/repository.rb
+++ b/lib/linguist/repository.rb
@@ -10,9 +10,11 @@ module Linguist
   class Repository
     attr_reader :repository
 
+    MAX_TREE_SIZE = 100_000
+
     # Public: Create a new Repository based on the stats of
     # an existing one
-    def self.incremental(repo, commit_oid, old_commit_oid, old_stats, max_tree_size = 100_000)
+    def self.incremental(repo, commit_oid, old_commit_oid, old_stats, max_tree_size = MAX_TREE_SIZE)
       repo = self.new(repo, commit_oid, max_tree_size)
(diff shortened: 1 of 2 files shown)


lildude (approved):

lildude:
Please leave the merging to me.

lib/linguist/repository.rb
```diff
@@ -129,10 +131,9 @@ def current_tree
     end
 
     protected
-    MAX_TREE_SIZE = 100_000
```
Alhadis, line 132: Wouldn't it be wiser to keep `MAX_TREE_SIZE` as a class constant so we can DRY up the class's `initialize` and `self.incremental` methods? Both methods hardcode `100_000` as the default value for their `max_tree_size` parameters…

issyl0: Thanks for the review. This makes a lot of sense.

Alhadis (approved):

Pea1594 (approved):
🥰

lildude:
Merging so we can start conducting testing in prod. I won't be making a new release for this as we can vendor directly off master.

PR #6814, opened 2024-05-06
Add Dune (OCaml build system)

kisaragi-hiu:
## Description
[Dune](https://dune.readthedocs.io/) is a build system for OCaml, used in a large number of OCaml projects. This PR adds classification and syntax highlighting support for it.

I specified `ace_mode: lisp` because Dune is an Sexp-based language. I'm guessing that this is appropriate.

## Checklist:

  - [x] The ~~extension~~ filename of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each file name:
      - ~~https://github.com/search?type=code&q=NOT+is%3Afork+path%3A%2F%5Edune%24%2F~~
        - This was not deemed to be enough uses, and I'm fine with leaving it out.
      - ~~https://github.com/search?type=code&q=NOT+is%3Afork+path%3A%2F%5Edune-file%24%2F~~
        - ~~This one is probably justified as it's used by Dune itself, even if just once.~~
        - This has now been removed from this PR.
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A%2F%5Edune-project%24%2F
      - ~~https://github.com/search?type=code&q=NOT+is%3Afork+path%3A%2F%5Edune-workspace%24%2F~~
        - ~~"dune-workspace" is the name that [has a special meaning in Dune](https://dune.readthedocs.io/en/latest/reference/dune-workspace/index.html), but it's more common to write it in `dune-workspace.dev` then explicitly tell Dune to use that file.~~
        - This is not popular enough and have been left out of this PR.
      - ~~https://github.com/search?type=code&q=NOT+is%3Afork+path%3A%2F%5Edune-workspace%5C.dev%24%2F~~
         - This is not popular enough and have been left out of this PR.
    - Sample source(s):
      - [dune-project](https://github.com/reasonml/reason/blob/3930092ac1cc04b7f3c62c71956592cf158ef8bb/dune-project)
    - Sample license(s):
      - [reasonml/reason](https://github.com/reasonml/reason/blob/3930092ac1cc04b7f3c62c71956592cf158ef8bb/LICENSE.txt)
  - [x] I have included a syntax highlighting grammar: https://github.com/maelvls/vscode-dune
    - [ocamllabs/vscode-ocaml-platform](https://github.com/ocamllabs/vscode-ocaml-platform/) might be more suitable in the long run, but it errored out when I tried to add it. It also provides grammars for OCaml itself within the same repository, which makes using it more complicated right now.
  - [x] I have added a color
    - Hex value: #89421e
    - Rationale: This is the main color of [the Dune logo](https://github.com/ocaml/dune/blob/main/doc/assets/imgs/dune_logo_459x116.png).


lib/linguist/languages.yml
```diff
@@ -1606,6 +1606,34 @@ Dotenv:
   tm_scope: source.dotenv
   ace_mode: text
   language_id: 111148035
+Dune:
+  type: programming
+  ace_mode: lisp
+  filenames:
+  - dune
+  - dune-file
```
lildude, line 1614: This file is nowhere near popular enough for inclusion. Its existence in this PR will significantly delay it being merged whilst we wait for popularity to be met. I recommend removing it. Users can use an override until popularity is sufficient for addition later.

lildude (changes requested):
Marking as pending popularity as the two workspace files aren't sufficient yet.

lib/linguist/languages.yml
kisaragi-hiu: I've now left out `dune-file`.

lildude (dismissed):
Reviewing this with fresh eyes, I don't think it makes sense splitting this out into different languages. They all share the same grammar and are grouped together so it makes sense to put them all into a single `Dune` language definition.

I note you've manually set each entry in this PR with a different scopes, however the upstream grammar only has a single `source.dune` so your other scopes won't work. This also further confirms these are all the same language so don't need separate entries.

Please merge these languages into a single `Dune` language.

kisaragi-hiu:
Merged the three previously declared languages into one and rebased onto master.

The split was left over from when I was trying to use [ocamllabs/vscode-ocaml-platform](https://github.com/ocamllabs/vscode-ocaml-platform/) (which has the split for different vocabulary), and I left it in by mistake when I switched to [maelvls/vscode-dune](https://github.com/maelvls/vscode-dune) due to errors when I tried the former.

lildude:
As an FYI, `dune-project` currently meets our usage requirements. `dune` is close. The other two are miles off. In short, everything but `dune-project` is preventing this PR from being merged.

If you want, you can make this PR just for `dune-project` and add support for the others as popularity grows. User will be able to use an override in the meantime.

kisaragi-hiu:
I've dropped the other three file names besides `dune-project`. Evidently I've misjudged the usage requirements to be way lower than what they actually are.

lildude (approved):

PR #6815, opened 2024-05-09
Add common AutoHotkey extensions

Banaanae:
Adds common AutoHotkey extensions

## Description

Adds the following extensions:
- ah1
- ah2

## Checklist:

      - https://github.com/search?q=NOT+is%3Afork+path%3A*.ah1+%3A%3D&type=code
      - https://github.com/search?q=NOT+is%3Afork+path%3A*.ah2+%3A%3D&type=code
    - Sample source(s):
      - Wrote my own
    - Sample license(s):
    - MIT


lildude (changes requested):
A quick look at your search links suggests most of these extensions are nowhere near popular enough for inclusion. Your samples are all exactly what we [explicitly say](https://github.com/github-linguist/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language) do not add too.

Please remove all extensions that don't meet our popularity requirements or aren't close. And replace the samples with _real_ samples.

The heuristic is also unnecessary as Linguist won't use it as it doesn't already know about this extension, which means every file on GitHub will that extension will now be associated with this language, which you might want to think twice about.

You might want to consider using an [override](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md) instead.

Banaanae:
I removed the `ahk1` and `ahk2` extensions
Made the samples better

I'm not too sure what you mean with the heuristics, there are a couple of files like [this](https://github.com/WEC-Sim/WEC-Sim/blob/main/examples/BEMIO/AQWA/RM3/RM3.AH1) which share the same extension, but isn't autohotkey
Overrides sound like a local thing?

lildude:
> I'm not too sure what you mean with the heuristics, there are a couple of files like [this](https://github.com/WEC-Sim/WEC-Sim/blob/main/examples/BEMIO/AQWA/RM3/RM3.AH1) which share the same extension, but isn't autohotkey

The heuristic is not considered at all when Linguist only knows one language for an extension. In such a case the language Linguist knows is applied to all files with that extension, unless overridden.

In order for the heuristic to be considered, Linguist needs to know of at least two languages using the extension so it can use it to differentiate between the two and make a better identification. So if you want your heuristic to take effect, you will need to identify another  other language, and add it to Linguist as part of this PR, assuming it meets the popularity requirements too.

That file in question will remain unaffected as the repo already has an override.

> Overrides sound like a local thing?

Only local to the repo. The other language file you pointed to only has a language and loose syntax highlighting because the repo owner has told Linguist to consider all files in the repo as matlab with [this line](https://github.com/WEC-Sim/WEC-Sim/blob/453af810b9f1bf5ae3b608b1e84078037dc20919/.gitattributes#L1).

Banaanae:
All requested changes should be addressed now :)

lildude (approved):

PR #6832, opened 2024-05-16
Add the Cylc language.

oliver-sanders:
## Description

Add the [Cylc](https://github.com/cylc/cylc-flow) language.

## Checklist:

      -  https://github.com/search?q=%28path%3A**%2Fsuite.rc+OR+path%3A**%2Fflow.cylc%29+-is%3Afork+-org%3Acylc&type=code&ref=advsearch
      - (note all Cylc projects must have a "root" flow.cylc or suite.rc file which is what I've searched for above, projects often contain additional "*.cylc" files)
    - Sample source(s):
      - https://github.com/MetOffice/CSET/blob/main/cset-workflow/flow.cylc
      - https://github.com/MetOffice/CSET/tree/main/cset-workflow/includes
    - Sample license(s):
      - Apache 2.0
  - [x] I have included a syntax highlighting grammar: [Cylc.tmbundle](https://github.com/cylc/Cylc.tmbundle) (BSD)
      
  - [x] I have added a color
    - Hex value: `#00b3fd`
    - Rationale: Primary brand colour used in user interfaces and the logo.
    - No other languages share the ".cylc" extension.


samples/Cylc/model.cylc
```diff
@@ -0,0 +1,138 @@
+#!Jinja2
```
oliver-sanders, line 1: I think the CI failure might be to do with this line?

```
 Cylc has a sample with an interpreter (Jinja2) that isn't explicitly defined in languages.yml
```

This shebang line is a functional part of the language (configures Cylc to run the Jinja2 preprocessor on this file before ingestion), but can remove it if necessary.

MetRonnie: (The shebang doesn't affect tokenization of the file by the textmate grammar)

oliver-sanders: Removed the Jinja2 shebangs as they are causing issues with CI.

lib/linguist/languages.yml
```diff
@@ -1399,6 +1399,17 @@ Cycript:
   codemirror_mode: javascript
   codemirror_mime_type: text/javascript
   language_id: 78
+Cylc:
+  type: programming
```
lildude, line 1403: This should be data as is INI into which you're grouping it.

oliver-sanders: Thanks:

```suggestion
  type: data
```

The syntax is INI-like, but it describes, an executable program, let me know if there's a more appropriate grouping.

lildude: My comment was only based on you grouping it with INI. Grouping isn't necessary. 

I know nothing about Cylc but from a quick view of the docs it appears these files describe workflows so it's really up to you if you'd consider this an actual programming language or data. We have other workflow-like languages which are considered data and others programming.

oliver-sanders: Yep that's right, so long as it makes sense in Linguist.

oliver-sanders:
Great, thanks for the clarification.

The extension usually only appears once per repo so I used a search query which looks for files in the project root which has ~800 results. I failed to mention it was a "once per repo" query, apologies.

I've now also included file extension queries (which will also pick up include files) which gives ~2.4k matches.

lildude (approved):

PR #6872, opened 2024-06-08
Dependabot configuration to update actions in workflows

ScottBrenner:
## Description

Noticed a few actions used in the workflows here are outdated, proposing a Dependabot configuration to update - reference https://docs.github.com/en/actions/security-guides/using-githubs-security-features-to-secure-your-use-of-github-actions#keeping-the-actions-in-your-workflows-secure-and-up-to-date

Suggest enabling https://docs.github.com/en/code-security/dependabot/working-with-dependabot/about-dependabot-on-github-actions-runners#enabling-or-disabling-for-your-repository as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,14 @@
 version: 2
 updates:
-  # Only check for Ruby gem updates
+  # Check for Ruby gem updates
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
       interval: "daily"
+  # Check for GitHub Actions updates
+  - package-ecosystem: "github-actions"
+    # Look for GitHub Actions workflows in the `root` directory
+    directory: "/"
+    # Check the for updates once a week
+    schedule:
+      interval: "weekly"


lildude:
I tried this in the past and it broke things and I didn't have time to investigate further ([revert PR](https://github.com/github-linguist/linguist/pull/5915)). It was over two years ago so maybe things are better now. I won't be merging this until I have the time to deal with the fallout if this breaks things as in the past.

ScottBrenner:
Thanks for the context - do you recall what broke? Seems checks did not run on the actions/checkout bump in https://github.com/github-linguist/linguist/pull/5911/checks, although the Actions history begins about a year _before_ that https://github.com/github-linguist/linguist/actions?page=32 so it may be lost to time

In any case, I did bump actions/checkout v4 on my fork here and all checks passed without issue https://github.com/ScottBrenner/linguist/pull/1/checks

The other action, ruby/setup-ruby, appears to keep their "v1" tag updated https://github.com/ruby/setup-ruby/tree/v1 and would not be updated by the proposed changes here

lildude:
> do you recall what broke?

I don't specifically, but from my comment in https://github.com/github-linguist/linguist/pull/5912 checkout depth was at least one problem. We need more than `master` for our tests as some need the commits in `test/attributes`.

This problem wasn't caught by the tests in the PR itself for some reason I can't recall 👴

ScottBrenner:
The commits in `test/attributes` still seem to be fetched when the version of actions/checkout is updated - https://github.com/ScottBrenner/linguist/actions/runs/9432580092/job/25982529071?pr=1#step:4:18 - via https://github.com/github-linguist/linguist/blob/master/.github/workflows/ci.yml#L32?

lildude:
That looks to be left over from when I was tatting with this last time, so maybe I've already fixed that issue 😁

I note your test PR only updates the checkout action. Do things still pass if you update all actions to their latest versions?

ScottBrenner:
Believe actions/checkout is the only action that would be updated by this, for now at least, as the only other action used ruby/setup-ruby uses v1 which they seem to keep updated under https://github.com/ruby/setup-ruby/tree/v1

lildude:
There's no need to close this.

PR #6897, opened 2024-06-21
Remove backreferences in regex patterns for `XML Property List` and `JavaScript`

DecimalTurn:
There is currently 2 discrepencies between Linguist and Enry that can be explained by the fact that RE2 doesn't support backreference:

 - [Heuristic for ".plist" extension](https://github.com/github-linguist/linguist/blob/b5432ebc7e78f25415b98d48c2fbacddbf8df317/lib/linguist/heuristics.yml#L524) in 'XML Property List', due to unsupported backreference in RE2 regexp engine.
 - [Heuristics for ".es" extension](https://github.com/github/linguist/blob/e761f9b013e5b61161481fcb898b59721ee40e3d/lib/linguist/heuristics.yml#L103) in JavaScript could not be parsed, due to unsupported backreference in RE2 regexp engine.

From: https://github.com/go-enry/go-enry#divergences-from-linguist

In both cases, this is because the regex pattern is trying to make sure that the quotation mark used is the same to start and end the string. That's something important when performing syntax highlighting, but that doesn't provide much value in terms of distinguishing one language from another. 

For that reason, the proposed changes won't make a noticable difference from Linguist's side, but it will remove the presence of backreferences for Enry.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -226,7 +226,7 @@ disambiguations:
   - language: Erlang
     pattern: '^\s*(?:%%|main\s*\(.*?\)\s*->)'
   - language: JavaScript
-    pattern: '\/\/|("|'')use strict\1|export\s+default\s|\/\*(?:.|[\r\n])*?\*\/'
+    pattern: '\/\/|["'']use strict["'']|export\s+default\s|\/\*(?:.|[\r\n])*?\*\/'
 - extensions: ['.ex']
   rules:
   - language: Elixir
@@ -552,7 +552,7 @@ disambiguations:
 - extensions: ['.plist']
   rules:
   - language: XML Property List
-    pattern: '^\s*(?:<\?xml\s|<!DOCTYPE\s+plist|<plist(?:\s+version\s*=\s*(["''])\d+(?:\.\d+)?\1)?\s*>\s*$)'
+    pattern: '^\s*(?:<\?xml\s|<!DOCTYPE\s+plist|<plist(?:\s+version\s*=\s*["'']\d+(?:\.\d+)?["''])?\s*>\s*$)'
   - language: OpenStep Property List
 - extensions: ['.plt']
   rules:


lildude (approved):
Makes sense to me. Thanks.

Alhadis (approved):
Good catch.

lib/linguist/heuristics.yml
```diff
@@ -226,7 +226,7 @@ disambiguations:
   - language: Erlang
     pattern: '^\s*(?:%%|main\s*\(.*?\)\s*->)'
   - language: JavaScript
-    pattern: '\/\/|("|'')use strict\1|export\s+default\s|\/\*(?:.|[\r\n])*?\*\/'
+    pattern: '\/\/|["'']use strict["'']|export\s+default\s|\/\*(?:.|[\r\n])*?\*\/'
```
Rudxain, line 229: If this is a char-class, then it could be simplified to `["']`

DecimalTurn: Yeah, the original regex was trying to account for the case with 2 single quotes, but in practice, that is not used (see this [search query](https://github.com/search?q=%2F%27%27use+strict%27%27%2F+path%3A*.es&type=code)). The goal of this PR wasn't to change the behavior of the regex, but it did in a small unconsequential way.

Now, we could either go all the was and make it a char-class or revert it back to `("|'')`, but it doesn't make a big difference.

DecimalTurn: However, the previous rule using `("|'')` wasn't matching for the case where a unique single quote was used, which is the [most common occurence](https://github.com/search?q=%2F%27use+strict%27%2F+path%3A*.es&type=code) of `use script` in `.es` files, so `["']` is actually simpler and more in line with was appears in `.es` files.

@Rudxain, since the PR was already merged, feel free to open a new PR and make your suggested change to simplify the rules.

DecimalTurn: Actually, you need the 2 single quotes inside a single-quoted yaml string, so what was merged is equivalent to `["']`, it's just in yaml notation.

Rudxain: @DecimalTurn thanks for explaining!

DecimalTurn: My pleasure. It was also a good reminder for me as well 😅

PR #6917, opened 2024-06-28
Split Cairo out into separate Cairo and Cairo Zero languages

mkaput:
This PR is updating the situation around the Cairo language to match the current state of things.

Disclaimer: I am involved in the development of the Cairo compiler and its dev tooling.

### Historical context

Cairo is a programming language that allows writing zero-knowledge-provable programs. Cairo is also the smart-contract language of Starknet.

In Jan 2023, Cairo 1.0 was released ([blog post](https://starkware.co/cairo-1-0-is-here/)). Cairo 1.0 (and all versions upwards) is a completely new language built from scratch. 

A couple of weeks later, we have decided a new naming strategy has been established:
- Cairo 1.0 and upwards is now called _Cairo_ proper.
- The old Cairo is now called _Cairo Zero_.
- A tiny, strict subset of Cairo Zero, the assembly language of the Cairo virtual machine became _CASM_ - Cairo Assembly.

These naming changes got community approval, but everything happened on an invite-only Telegram group, so I am unable to provide a link.

Cairo is not backwards-compatible with Cairo Zero. Cairo Zero development is frozen and the language will not receive any new features or updates.

As a side note: Cairo also brings another language - Sierra, which is an IR between Cairo and CASM. This PR is not concerning about Sierra at all.

### What this PR changes

1. Renamed existing `Cairo` language to `Cairo Zero`.
2. Added a new language `Cairo` that represents the new Cairo.
3. Updated grammar sources to the new location that provides grammars for both languages: https://github.com/software-mansion-labs/cairo-tm-grammar
    - I am maintaining this repository. It is just a repackaging of grammars that this repo pulls from upstream repositories. I made sure to have everything documented and properly attributed there.
    - The grammar for Cairo is new, and I made it for the official Visual Studio Code extension for Cairo.
    - The grammar for Cairo Zero is different from the one that was used here before.
      - The new grammar comes from the old, official Visual Studio Code extension for Cairo Zero.
      - The old grammar was written by @0xChqrles. It has bugs that make it stop highlighting on some files ([example](https://github.com/kkrt-labs/kakarot/blob/147a99602ecaebef7e44c795f0aabd713476db1c/src/kakarot/account.cairo#L474)). The new grammar does not exhibit this problem. I have talked with @0xChqrles about this, and he approves stopping using his grammar.
4. I have added heuristics for Cairo Zero that help disambiguate it from Cairo (as both languages use the same extension). With these heuristics and newly added samples, the classifier seems to identify the correct languages well, according to my testing.
5. Some samples of CASM language can appear on GitHub as `*.casm`. I have added two as `*.cairo` to help the classifier, but this PR is not doing anything more, as CASM is not significant enough.

## Checklist:

            -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.cairo+starknet
        - Sample source(s):
            - https://github.com/starkware-libs/cairo/blob/157b38ea28250bf88bdca6faf565935c1e57ff1e/corelib/src/boolean.cairo
            - https://github.com/cairo-book/cairo-book/blob/99673ae213e7ada059e3bc73251391461970b094/listings/ch11-advanced-features/no_listing_13_cust_struct_vect/src/lib.cairo
            - https://github.com/cairo-book/cairo-book/blob/99673ae213e7ada059e3bc73251391461970b094/listings/ch17-starknet-smart-contracts-security/listing_02_pizza_factory_snfoundry/src/pizza.cairo
            - https://github.com/keep-starknet-strange/alexandria/blob/0aa7cca8a1acb1f35cff32420614f7f58e38de5e/packages/merkle_tree/src/tests/merkle_tree_test.cairo
            - https://github.com/HerodotusDev/cairo-lib/blob/a0c0ee4feeb4edb61e7a35d29595ad8647c3d469/src/encoding/rlp.cairo
        - Sample license(s):
            - Apache-2.0 for `starkware-libs/cairo`
            - MIT for `cairo-book`
    - [x] I have included a syntax highlighting grammar: https://github.com/software-mansion-labs/cairo-tm-grammar
        
    - [x] I have added a color
        - Hex value: `#ff4a48`
        - Rationale: Brand color, per [Media Kit](https://www.starknet.io/media-kit/#cairo-logo-kit).
        - Not for `Cairo` language itself as it is still evolving, but added heuristics (& tests) that help to disambiguate `Cairo Zero` which is frozen.


lib/linguist/languages.yml
```diff
@@ -966,6 +966,14 @@ Cairo:
   tm_scope: source.cairo
   extensions:
   - ".cairo"
+  language_id: 620599568
```
mkaput, line 969: `script/update-ids` generated the same `language_id` as for `Cairo 0` (620599567). This is wrong, so I have just added 1, as I see it's just hashing language name so this sounds safe?

lildude: 🤔 It shouldn't.

This is where where the magic happens:

https://github.com/github-linguist/linguist/blob/39fd5e93de98de3434cb3e857c5b1972c418f8c5/script/update-ids#L8-L10

```ruby
irb(main):004:0> require 'digest'
=> true
irb(main):005:0> language = 'Cairo 0'
irb(main):006:0> Digest::SHA256.hexdigest(language).to_i(16) % (2**30 - 1)
=> 891399890
irb(main):007:0> language = 'Cairo'
irb(main):008:0> Digest::SHA256.hexdigest(language).to_i(16) % (2**30 - 1)
=> 620599567
irb(main):009:0> 
```

It however won't update an entry that already has an ID - we don't want the IDs in GitHub's database to change. As such, it's better to create a whole new `Cairo 0` entry with new ID and keep the current `Cairo` with the current ID.

I recommend adding `group: Cairo` line to the new `Cairo 0` entry too so it appears to the user that it's all Cairo.

lib/linguist/heuristics.yml
```diff
@@ -136,6 +136,10 @@ disambiguations:
   rules:
   - language: XML
     pattern: '^(\s*)(?i:<Project|<Import|<Property|<?xml|xmlns)'
+- extensions: ['.cairo']
+  rules:
+    - language: Cairo 0
+      pattern: '(^(\s*)%lang(\s+)([A-Za-z0-9_]+))|(^(\s*)%builtins(\s+)([A-Za-z0-9_]+\s*)*$)|(^(\s*)from(\s+)starkware\.(cairo|starknet)\.([A-Za-z0-9_.\s]+?)import)'
```
lildude, line 142: As these are the only two languages using this extension, we can force things to Cairo here rather than taking things further into the classifier:
```suggestion
      pattern: '(^(\s*)%lang(\s+)([A-Za-z0-9_]+))|(^(\s*)%builtins(\s+)([A-Za-z0-9_]+\s*)*$)|(^(\s*)from(\s+)starkware\.(cairo|starknet)\.([A-Za-z0-9_.\s]+?)import)'
    - language: Cairo
```

lildude (changes requested):
See inline.

Please can you replace your new samples which are not real-world files with real-world examples.

lib/linguist/heuristics.yml
mkaput: I thought about this, but I would have to extend this heuristic to match [CASM](https://github.com/github-linguist/linguist/blob/066b07b201e9ac1b287ed0560de57f406e1541a3/samples/Cairo%200/fib_u128.cairo) files, which would blow up this already complex regex. 😕 I want CASM to be counted as Cairo 0 (it's a strict subset), yet they do not exhibit such trivial and distinguishable constructs I detect here.

mkaput:
> Please can you replace your new samples which are not real-world files with real-world examples.

Do you mean the CASM ones ([1](https://github.com/github-linguist/linguist/blob/066b07b201e9ac1b287ed0560de57f406e1541a3/samples/Cairo%200/fib.cairo), [2](https://github.com/github-linguist/linguist/blob/066b07b201e9ac1b287ed0560de57f406e1541a3/samples/Cairo%200/fib_u128.cairo))? Yes, they are auto-generated, but I used them because:
1. they do appear on GitHub already, and
2. they help the classifier distinguish Cairo 0 from Cairo, as they are full of Cairo 0-specific syntax; without these, the classifier fails to differentiate remaining samples

lib/linguist/languages.yml
mkaput: > As such, it's better to create a whole new `Cairo 0` entry with new ID and keep the current `Cairo` with the current ID.

Question: Will existing files be reclassified when a linguist change is released, which reclassifies files from one language ID to another? 

> I recommend adding `group: Cairo` line to the new `Cairo 0` entry too so it appears to the user that it's all Cairo.

I'm not sure about this. For Starknet developers (majority of Cairo use) it's pretty important to know whether a codebase it Cairo 0 or Cairo 1.0+. Cairo 0 is not deployable any more on all official networks.

Given this context, would you still recommend grouping these?

lildude: > Question: Will existing files be reclassified when a linguist change is released, which reclassifies files from one language ID to another?

No. A repo will only be reclassified when reanalysed which only happens on pushes to the repo.

It's common for very old repos to show the old name until a new push is made, which is kinda right because that was what the language was called at the time of the last push. You'll see this for the variations of VBNet.

Reusing a name with a different ID really confuses things and leads to a lot of support issues because of caching. This is also complicated by Search which uses a different library which feeds off linguist, hence the suggestion of adding a new language for the old name with a new name. 

> I'm not sure about this. For Starknet developers (majority of Cairo use) it's pretty important to know whether a codebase it Cairo 0 or Cairo 1.0+. Cairo 0 is not deployable any more on all official networks.

The grouping only affects what is shown in the sidebar and as the primary language for a repo based on bytes of code. Individual files will be identified by the individual names. Additionally, given the vastly different syntax, wouldn't this be easily recognisable by users?

lildude:
> > Please can you replace your new samples which are not real-world files with real-world examples.
> 
> Do you mean the CASM ones ([1](https://github.com/github-linguist/linguist/blob/066b07b201e9ac1b287ed0560de57f406e1541a3/samples/Cairo%200/fib.cairo), [2](https://github.com/github-linguist/linguist/blob/066b07b201e9ac1b287ed0560de57f406e1541a3/samples/Cairo%200/fib_u128.cairo))? Yes, they are auto-generated, but I used them because:
> 
> 1. they do appear on GitHub already, and
> 2. they help the classifier distinguish Cairo 0 from Cairo, as they are full of Cairo 0-specific syntax; without these, the classifier fails to differentiate remaining samples
> 
> wdyt given this context?

Yes, those files. The fact you can only get the classifier to match our current samples by fudging things kinda speaks for itself… either the current samples aren't good illustrations of the language or the new ones aren't. Or you're trying to match things which look completely different which normally indicates a whole different syntax and thus commonly a whole different language.

BaljeettheBro (approved):

lib/linguist/heuristics.yml
mkaput: I removed this heuristic. With new samples, the classifier now works great.

lib/linguist/languages.yml
mkaput: > Additionally, given the vastly different syntax, wouldn't this be easily recognisable by users?

Yes, they can.

---

Done

mkaput:
> Yes, those files. The fact you can only get the classifier to match our current samples by fudging things kinda speaks for itself… either the current samples aren't good illustrations of the language or the new ones aren't. Or you're trying to match things which look completely different which normally indicates a whole different syntax and thus commonly a whole different language.

I removed the bigger file, but left the smaller one as it's a good representation and actually hand-writable for some example purposes. I added more real-world Cairo and Cairo 0 samples to provide more data for the classifier. PR description is updated with source links.

lib/linguist/heuristics.yml
lildude: I'm happy with this, but Cairo user may not be 😁  

The classifier is not very reliable and is really a "last resort" - it's literally the last thing we try if we can't find a result any other way. You might want to consider a simple heuristic that would match more often than not to help things and leave the classifier to those occasions where the heuristic doesn't match. There's no need to be 100% perfect; close enough is good enough. 

Maybe try matching for most of Cairo, like your search result, and leave the classifier to handle Cairo 0 and CASM and the more obscure uses?

mkaput: Done. Went with your original suggestion to try to match most of `Cairo 0` and fall back to `Cairo` because I realised there is one more very common pattern in `Cairo 0` code. Now, I believe one would have to handcraft Cairo 0 code deliberately to be identified as Cairo.

mkaput:
@lildude is there anything I have to add here? :)

lildude (approved):
LGTM. Thanks.

mkaput:
Awesome! 😍 Thank you for review!

I saw this section in docs; thought you'd merge this now and will wait, just releasing from `main`. But it's fine whatever you do.

Are you able to give an order of magnitude, when do you plan to do a next release? I bet my community will love to hear about it, but it's also fine if you can't tell it here.

lildude:
I make a release approx every 3-4 months. The last one was on 7 Jun so it won't be for a few months... probably late August.

lildude (approved):

PR #6936, opened 2024-07-11
Add Octave language with its own language distinct from MATLAB

pr0m1th3as:
## Description
This PR is adding a distinction between MATLAB and GNU Octave languages, which both use `.m` files for their code files. Although GNU Octave is often described as the open source MATLAB alternative due to their similarities, the reality is that it is a separate programming language. Most importantly, although MATLAB syntax can be interpreted by GNU Octave, the opposite is not true. Octave syntax includes `#` for comment lines, and `endif`, `endfor`, `endfunction`, `endclassdef` etc as block's closing statements, which are explicitly recognized by the Octave interpreter, but cannot be parsed/run in MATLAB. As a result of these differences, GitHub falsely recognizes Octave code as MATLAB and by using MATLAB's syntax highlighting grammar results into mis-displaying code, which is written explicitly for the Octave language.

Besides the mis-highlighted code syntax of files containing Octave code, this is also confusing for users who might consider that code written for Octave may work in MATLAB, since GitHub reports all Octave code as MATLAB. GNU Octave is a community driven open source project, which has been in constant development for more than 30 years now, and with a reasonably large user base across academic and industrial sectors, especially in engineering fields. Thus, there is also an ethical aspect towards our community of developers and users that the Octave language should not be misclassified as another proprietary language by the worlds largest developer platform, GitHub, which gratefully provides free hosting for a myriad of open source projects.

## Checklist:
      
      -  The following search query:
[`NOT is:fork path:*.m AND (/^\s*endfor/ OR /^\s*endfunction/ OR /^\s*endclassdef/ OR /^\s*endif/)`](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.m+AND+%28%2F%5E%5Cs*endfor%2F+OR+%2F%5E%5Cs*endfunction%2F+OR+%2F%5E%5Cs*endclassdef%2F+OR+%2F%5E%5Cs*endif%2F%29)
      returns 44.5k files in more than 200 repositories.
    - Sample source(s):
      The following files have been included as real-world usage samples
      - [accumarray.m](https://github.com/gnu-octave/octave/blob/default/scripts/general/accumarray.m)
      - [BetaDistribution.m](https://github.com/gnu-octave/statistics/blob/main/inst/dist_obj/BetaDistribution.m)
      - [ClusterCriterion.m](https://github.com/gnu-octave/statistics/blob/main/inst/Clustering/ClusterCriterion.m)
      - [anova1.m](https://github.com/gnu-octave/statistics/blob/main/inst/anova1.m)
      - [silhouette.m](https://github.com/gnu-octave/statistics/blob/main/inst/silhouette.m)
      - [var.m](https://github.com/gnu-octave/octave/blob/default/scripts/statistics/var.m)
    - Sample license(s): All samples are licensed under the GNU GPLv3.
  - [x] I have included a syntax highlighting grammar: [Octave-language-grammar](https://github.com/gnu-octave/Octave-language-grammar)
      
  - [x] I have added a color
    - Hex value: `#2299C4`
    - Rationale: The color choice is based on the most prominent color of the GNU Octave logo


lib/linguist/languages.yml
```diff
@@ -4884,6 +4880,14 @@ Objective-J:
...
   ace_mode: text
   language_id: 259
+Octave:
+  type: programming
+  color: "#2299c4"
+  extensions:
+  - ".m"
+  tm_scope: source.octave
+  ace_mode: octave
+  language_id: 312
```
lildude, line 4890: Please follow the CONTRIBUTING.md details for adding a new language and generating this ID.

test/test_heuristics.rb
```diff
@@ -633,6 +633,7 @@ def test_m_by_heuristics
       "M" => all_fixtures("M", "MDB.m"),
       "Mathematica" => all_fixtures("Mathematica", "*.m") - all_fixtures("Mathematica", "Problem12.m"),
       "MATLAB" => all_fixtures("MATLAB", "create_ieee_paper_plots.m"),
+      "Octave" => all_fixtures("Octave", "accumarray.m"),
```
lildude, line 636: As there are quite a few samples, we want this heuristic to correctly identify all of them correctly, not just the one file.

```suggestion
      "Octave" => all_fixtures("Octave", "*.m"),
```

lildude (changes requested):
See inline comments.

Additionally, several of your samples are too big for our needs (the ones suppressed in the diff). Please replace them with smaller samples.

You also need to add the cached license file the `add-grammar` script should have created when you ran it.

If you can't find it or it didn't get create it, you can do it manually by running:

https://github.com/github-linguist/linguist/blob/39fd5e93de98de3434cb3e857c5b1972c418f8c5/script/add-grammar#L134

tpapastylianou:
Spotted this PR and just wanted to add that, while octave and matlab do indeed share an odd relationship, matlab is not the only language with which github language detection clashes with, particularly when ".m" files are involved, which are also used by objective-c, wolfram mathematica, and mercury-lang as well.

So I would say this is less a case of "making octave distinct from matlab" specifically per se, but more about adding octave as a proper language, and disambiguating octave files among all the above languages that also use the ".m" extension, including matlab  (whose syntax is indeed different, despite the many similarities).

E.g. if one searches github for "language:Objective-C octave" (or indeed "language:Objective-C matlab") one will witness this confusion, and that it's not necessarily octave/matlab-exclusive.

lib/linguist/languages.yml
pr0m1th3as: A proper language_id has been generated and added to the `languages.yml` file.

test/test_heuristics.rb
pr0m1th3as: The very large sample files have been removed. The remaining ones express Octave syntax in a suitable manner. Corrected the heuristic to correctly identify all available sample files of the Octave language.

pr0m1th3as:
@lildude I have made all necessary changes to the best of my understanding of the errors produced from the automated tests. Please, run once again the testing worklfow and give me some guidance about the necessary changes that might still be required. Thank you very much for your help.

lib/linguist/languages.yml
```diff
@@ -4020,15 +4020,11 @@ M4Sugar:
...
   color: "#e16737"
-  aliases:
-  - octave
   extensions:
   - ".matlab"
   - ".m"
   tm_scope: source.matlab
   ace_mode: matlab
-  codemirror_mode: octave
-  codemirror_mime_type: text/x-octave
```
lildude, line 4031: Whilst Octave isn't Matlab, I think we should keep this for current Matlab users unless this is really is horribly wrong.

lib/linguist/languages.yml
```diff
@@ -4884,6 +4880,14 @@ Objective-J:
...
   ace_mode: text
   language_id: 259
+Octave:
+  type: programming
+  color: "#2299c4"
+  extensions:
+  - ".m"
+  tm_scope: source.octave
+  ace_mode: text
+  language_id: 854759768
```
lildude, line 4890: As there is a codemirror mode for Octave, you may as well add it here too.

lildude (changes requested):
Looking better. Just a few more suggestions.

lib/linguist/languages.yml
pr0m1th3as: Apart from providing a correct syntax highlight for Octave code, the other reason behind this PR is to segregate the two languages in order to prevent people from falsely assuming that MATLAB and Octave are bidirectionally compatible. GNU Octave can run MATLAB code, but MATLAB will not run Octave code. IMHO, keeping the `codemirror_mode` in MATLAB's definition is not a good idea.

lib/linguist/languages.yml
pr0m1th3as: See my previous comment on avoiding codemirror more.

lib/linguist/languages.yml
lildude: This isn't about running the code or giving an impression is would work, it's about getting syntax highlighting in the code editor box. Without this, users get nothing. With it, they'll get some syntax highlighting.

lib/linguist/languages.yml
lildude: Ditto 😉

lib/linguist/languages.yml
pr0m1th3as: I am not sure I follow your comment. Isn't the case that MATLAB code gets highlighted according to MATLAB's syntax tmLanguage file and Octave code (once merged and deployed in linguist) will get the syntax highlight from the new grammar tmLanguage file published in gnu-octave/Octave-language-grammar repo? In what case, the code editor will not apply any syntax highlighting? In any case, if you consider that the presence of codemirror code is a necessity for linguist's proper (or safe) operation, then I assume we can add it.

lildude: > Isn't the case that MATLAB code gets highlighted according to MATLAB's syntax tmLanguage file and Octave code (once merged and deployed in linguist) will get the syntax highlight from the new grammar tmLanguage file published in gnu-octave/Octave-language-grammar repo?

Yes, when rendered once you've commited and pushed the files to GitHub. 

From the very top of `languages.yml`:

https://github.com/github-linguist/linguist/blob/39fd5e93de98de3434cb3e857c5b1972c418f8c5/lib/linguist/languages.yml#L12-L15

To extend on this, the codemirror options are used for the editor when you edit files directly on the GitHub website. Find a file in a repo on GitHub, open it and click the pencil icon - that syntax highlighting is coming from Codemirror, not the grammar included in Linguist.

We've not been consistent with ensuring we set these options in the past as not all languages have Codemirror support but your changes reminded me that this is a situation where it does.

If you remove this here, anyone who uses the GitHub UI to edit Matlab files will lose whatever syntax highlighting they get now, whether it's 100% accurate or not. 

This is also why I suggested adding it to the Octave language... that way peeps editing Octave files directly on GitHub will get syntax highlighting in the code editor. It seems silly not to offer it when it is available.

pr0m1th3as: Ah, ok, it totally makes sense. Thanks for clarifying what codemirror actually does. I will push these changes shortly.

lib/linguist/languages.yml
```diff
@@ -4020,14 +4020,12 @@ M4Sugar:
...
   color: "#e16737"
-  aliases:
-  - octave
   extensions:
   - ".matlab"
   - ".m"
   tm_scope: source.matlab
   ace_mode: matlab
-  codemirror_mode: octave
+  codemirror_code: octave
```
lildude, line 4028: Whoops

```suggestion
  codemirror_mode: octave
```

pr0m1th3as:
@lildude Are there any other changes required before this PR can be merged?

lildude:
> @lildude Are there any other changes required before this PR can be merged?

Yup. See the test failure.

Alhadis (commented):
Is the distinction between MATLAB and Octave that great that Octave projects don't contribute to the usage statistics of MATLAB, and vice versa? If not, then the new `Octave` entry should be grouped under `MATLAB` (using a `group: MATLAB` field).

Otherwise, this will make a mess of classification: I can already see projects being classified as a "mixture" of MATLAB and Octave files, purely because each file was analysed in isolation, as per Linguist's implementation.

Grouping won't stop you from selecting a more accurate grammar to highlight Octave source (and by extension, CodeMirror modes). In fact, this is a large part of why this language "grouping" feature exists.

PR #6968, opened 2024-07-25
Update languages.yml

JW0914:
Added INI syntax highlighting for config file extensions *.conf and *.ovpn:
  - Some programs (like OpenVPN) use .conf instead of .cnf for config files
  - OpenVPN client .ovpn files are INI config files

## Checklist:

      - [.conf+OpenVPN](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.CONF+OPENVPN)
      - [.ovpn+OpenVPN](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.OVPN+OPENVPN)
    - Sample source(s):
      - [`OpenVPN-Server.conf`](https://github.com/JW0914/Wikis/blob/master/Scripts%2BConfigs/OpenVPN/OpenVPN-Server.conf) _(Compare to: [`openvpn-server.cnf`](https://github.com/JW0914/Wikis/blob/master/Scripts%2BConfigs/OpenVPN/openvpn-server.cnf))_
      - [`Client.ovpn`](https://github.com/JW0914/Wikis/blob/master/Scripts%2BConfigs/OpenVPN/Client.ovpn)
    - Sample license(s):

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2998,9 +2998,11 @@ INI:
   extensions:
   - ".ini"
   - ".cfg"
+  - ".conf"
   - ".cnf"
   - ".dof"
   - ".lektorproject"
+  - ".ovpn"
   - ".prefs"
   - ".pro"
   - ".properties"


DecimalTurn:
It seems like the syntax of a lot of OpenVPN's files in the search queries (and the samples) you provided aren't compatible with the current syntax highlighting for INI files. 

The current syntax expects an equal sign (=) between the key and the value, but most of the OpenVPN's .conf files seem to use a space instead, hence the syntax highlighting doesn't really work:

```ini

server {{OVPN_NETWORK}} {{OVPN_SUBNET}}
verb 3

pkcs12 {{OVPN_CERTS}}
dh {{OVPN_DH}}

key-direction 0
keepalive 10 60
persist-key
persist-tun

proto {{OVPN_PROTO}}
port 1194
dev tun0
status /tmp/openvpn-status.log

user nobody
group nogroup

{{MORE_OPTS}}

push "route {{OVPN_K8S_SERVICE_NETWORK}} {{OVPN_K8S_SERVICE_SUBNET}}"
{{POD_CONF}}
push "dhcp-option DOMAIN {{OVPN_K8S_DOMAIN}}"
push "dhcp-option DNS {{OVPN_K8S_DNS}}"

```
([sample source](https://github.com/offlinehacker/openvpn-k8s/blob/56050fb1cde50fc86918e834faf90d41f841ebdf/openvpn.conf))

However a lot of `.conf` file seem to be using the INI format that linguist uses, so maybe I'd suggest to add a heuristic taking into account that the file needs to have at least one well specified key followed by an equal sign to be considered an INI file (some thing like `^[\w_]*\s?=`).

This still gives [5K files](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.CONF+OPENVPN+%2F%5E%5B%5Cw_%5D*%5Cs%3F%3D%2F), but if we drop the "OPENVPN" search term we even get [782K files](https://github.com/search?q=path%3A*.conf+%2F%5E%5B%5Cw_%5D*%5Cs%3F%3D%2F&type=code).

lildude (changes requested):
This PR is not complete. Please follow the CONTRIBUTING.md file for adding extensions and update the title of the PR.

You're going to need to add `.conf` as a generic extension (also add to `generic.yml`) with a very precise heuristic (there is no scope for error) as it is such a generic and common extension used by many languages which don't use the INI format. Some use their own language, some use JSON, some use XML etc. This is a potential minefield 😁

JW0914:
@DecimalTurn and @lildude will do

PR #6976, opened 2024-07-31
Fix heuristic for extension `.yy` (JSON vs Yacc)

DecimalTurn:
## Description

Closes https://github.com/github-linguist/linguist/issues/6517
[496k+ files](https://github.com/search?q=lang%3Ayacc+path%3A*.yy+%2F%5C%22resourceType%5C%22%5C%3A%5Cs*%5C%22GM%2F&type=code) affected

~~Implements fix suggested by @lildude and backed by @RobQuistNL in https://github.com/github-linguist/linguist/issues/6517#issuecomment-1680403065.~~

See this comment for the approach used: https://github.com/github-linguist/linguist/pull/6976#issuecomment-2264230000

## Checklist:

    - Sample source: https://github.com/Ttanasart-pt/Pixel-Composer/blob/363a1dec53f6785ee58a7b3f3b8407a48dcdd2c2/scripts/VCT/VCT.yy
    - Sample License: [MIT license](https://github.com/Ttanasart-pt/Pixel-Composer#)

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -698,13 +698,13 @@ module Linguist
     # Internal: Is this a generated Game Maker Studio (2) metadata file?
     #
     # All Game Maker Studio 2 generated files will be JSON, .yy or .yyp, and have
-    # a part that looks like "modelName: GMname" on the 3rd line
+    # a line that looks like "modelName: GMname" or "resourceType: GMname"
     #
     # Return true or false
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines[2].match(/\"(modelName|resourceType)\"\:\s*\"GM/) ||
+      return lines.any? { |line| line.match(/\"(modelName|resourceType)\"\:\s*\"GM/) } ||
              lines[0] =~ /^\d\.\d\.\d.+\|\{/
     end
 
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -704,7 +704,7 @@ module Linguist
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines[2].match(/\"modelName\"\:\s*\"GM/) ||
+      return lines[2].match(/\"(modelName|resourceType)\"\:\s*\"GM/) ||
              lines[0] =~ /^\d\.\d\.\d.+\|\{/
     end
 
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -704,7 +704,7 @@ module Linguist
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines.any? { |line| line.match(/\"(modelName|resourceType)\"\:\s*\"GM/) } ||
+      return lines.first(3).join('').match?(/^\s*[\{\[]/) ||
              lines[0] =~ /^\d\.\d\.\d.+\|\{/
     end
 
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -697,9 +697,6 @@ module Linguist
 
     # Internal: Is this a generated Game Maker Studio (2) metadata file?
     #
-    # All Game Maker Studio 2 generated files will be JSON, .yy or .yyp, and have
-    # a line that looks like "modelName: GMname" or "resourceType: GMname"
-    #
     # Return true or false
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
--- a/samples/JSON/VCT.yy
+++ b/samples/JSON/VCT.yy
@@ -1,0 +1,13 @@
+{
+  "$GMScript":"v1",
+  "%Name":"VCT",
+  "isCompatibility":false,
+  "isDnD":false,
+  "name":"VCT",
+  "parent":{
+    "name":"VCT",
+    "path":"folders/VCT.yy",
+  },
+  "resourceType":"GMScript",
+  "resourceVersion":"2.0",
+}
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -351,7 +351,7 @@ disambiguations:
 - extensions: ['.ice']
   rules:
   - language: JSON
-    pattern: '\A\s*[{\[]'
+    named_pattern: json
   - language: Slice
 - extensions: ['.inc']
   rules:
@@ -848,7 +848,7 @@ disambiguations:
 - extensions: ['.yy']
   rules:
   - language: JSON
-    pattern: '\"(modelName|resourceType)\"\:\s*\"GM'
+    named_pattern: json
   - language: Yacc
 named_patterns:
   cpp:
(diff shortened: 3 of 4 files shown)


lildude (changes requested):
Can we have a new sample and updated heuristic test, if necessary, for this too please.

DecimalTurn:
> Can we have a new sample and updated heuristic test, if necessary, for this too please.

Sure. I've added a sample JSON file with the .yy extension.

The test doesn't need any changes:

https://github.com/DecimalTurn/linguist/blob/4cbec096a484b0fdb756f598769a142c9d828762/test/test_heuristics.rb#L1087-L1093

lildude:
I was referring to adding a new GMStudio sample that matches the heuristic change.

DecimalTurn:
`.yy` files are produced by GMStudio, but they are classified as JSON by Linguist. That's why I've added the sample in the JSON folder.

The file added matches the heuristic added with the following line:
`"resourceType":"GMScript",`

lildude:
🤦 Whoops. I missed that.

lildude (approved):
LGTM. Thanks.

lildude:
Yeah, probably best to fix that too. Lets do it in this PR.

DecimalTurn:
Should be all done now.
Note that I've had to relax the constraint that the property has to be on the 3rd line. There seems to be [50k+ files](https://github.com/search?q=path%3A*.yy+%2F%5CA%28%3F%3A.*%28%3F%3A%5Cr%3F%5Cn%7C%5Cr%29%29%7B3%2C%7D%5Cs*%5C%22modelName%5C%22%5C%3A%5Cs*%5C%22GM%2F+&type=code) of GMStudio generated JSON where the old `modelName` property appears after the 3rd line, so even for GMStudio before v2.3, that constraint was too restrictive.

Sec32fun32 (approved):

lib/linguist/generated.rb
```diff
@@ -698,13 +698,13 @@ def generated_graphql_relay?
...
     # All Game Maker Studio 2 generated files will be JSON, .yy or .yyp, and have
-    # a part that looks like "modelName: GMname" on the 3rd line
+    # a line that looks like "modelName: GMname" or "resourceType: GMname"
     #
     # Return true or false
     def generated_gamemakerstudio?
       return false unless ['.yy', '.yyp'].include? extname
       return false unless lines.count > 3
-      return lines[2].match(/\"modelName\"\:\s*\"GM/) ||
+      return lines.any? { |line| line.match(/\"(modelName|resourceType)\"\:\s*\"GM/) } ||
```
lildude, line 707: Can we please keep this limited to a number of lines to protect against a massive rogue file crippling things.

samples/JSON/VCT.yy
```diff
@@ -0,0 +1,13 @@
+{
+  "$GMScript":"v1",
```
RobQuistNL, line 2: Maybe this `$GMScript` is a better one to match on given that it sounds pretty exclusive to GM? And looks like its a fixed line too.

Not sure if this is true for all newer files.. Haven't touched GM in a while :(

lib/linguist/generated.rb
RobQuistNL: Agreed - this previous one on a fixed line worked perfect for the older GM files. Lets keep that line in there. Maybe add an extra line that scans just a single fixed line (like I mentioned) if that works.

RobQuistNL (changes requested):

samples/JSON/VCT.yy
DecimalTurn: Hum, based on [this search](https://github.com/search?q=path%3A*.yy+%2F%5CA%28%3F%3A.*%28%3F%3A%5Cr%3F%5Cn%7C%5Cr%29%29%7B0%2C%7D%5Cs*%5C%22%28%5C%24GMScript%29%5C%22%5C%3A%5Cs*%5C%22v%2F++NOT+%2F%5CA%28%3F%3A.*%28%3F%3A%5Cr%3F%5Cn%7C%5Cr%29%29%7B0%2C%7D%5Cs*%5C%22%28modelName%7CresourceType%29%5C%22%5C%3A%5Cs*%5C%22GM%2F+&type=code), there is no files that has `$GMScript` and not `modelName` or `resourceType`, so adding `$GMScript` to the matching rules wouldn't help match more files at least.

lib/linguist/generated.rb
RobQuistNL: I would opt for keeping the lines lower (3 lines per file max) and just extend the amount of checks;
`$GM` as the first 3 letters of the first key
`resourceType` key with value starting with `GM`
`isDnD` in the key

things like that :thinking:

samples/JSON/VCT.yy
RobQuistNL: $GMScript and things like that seem to appear in the first 2 lines from what I've seen - `modelName` and `resourceType` might be all the way at the bottom (or in the middle if we're talking "room" files) :thinking:

lib/linguist/generated.rb
DecimalTurn: >  just extend the amount of checks

@RobQuistNL I personally think having a check that looks at the first 200 lines is OK since the regex used is really simple and should still perform well. However, if you have any example of a GMStudio generated .yy file that doesn't match with the current rule, but matches one of your suggested checks, I'd be happy to make changes to fix that.

DecimalTurn:
After looking a little more at the [syntax for Yacc](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/yacc.html), I realize that a Yacc file cannot start with a `{` or `[`. It has to start with a declaration which beggins with a `%`. This means that we could simply use the heuristic that we used for JSON from here (checking that the first non-whitespace character is either `{` or `[`):

https://github.com/github-linguist/linguist/blob/39fd5e93de98de3434cb3e857c5b1972c418f8c5/lib/linguist/heuristics.yml#L353-L354

And for the generated condition inside generated.rb , we could simply have something like:
```ruby
return lines.first(3).join('').match?(/^\s*[{\[]/) ||
```

Thoughts?

<hr>

[Corresponding search query](https://github.com/search?q=path%3A*.yy+%2F%5CA%5Cs*%5B%7B%5C%5B%5D%2F&type=code): Still 600K files.

RobQuistNL:
That is actually a much better approach in my opinion :+) Good find! I've [seen one other](https://github.com/fuzzdebugplatform/fuzz_debug_platform/blob/8f828483ea8dc5d37c60268cbc92c958adb63c87/exampleyy/functions.yy) `.yy` syntax that I don't think is either GM or Yacc, but since this setup only has these 2 included I think this might be a very clean and good solution.

JSON + .yy extension = GM
No JSON + .yy extension = Yacc

DecimalTurn:
We should be all good now with the latest changes [`bb34e54` (#6976)](https://github.com/github-linguist/linguist/pull/6976/commits/bb34e5422ef05c59bce59c81f33b0e03a593f086).

> I've [seen one other](https://github.com/fuzzdebugplatform/fuzz_debug_platform/blob/8f828483ea8dc5d37c60268cbc92c958adb63c87/exampleyy/functions.yy) .yy syntax that I don't think is either GM or Yacc

Good catch! This actually seems like an unusual Yacc file that skips the directives section going against the convention/documentation, so it would still be correct to categorize them as Yacc techniquely.

However, for this to be a problem, those unconventional Yacc files would need to start with a `{`. GitHub Search tells me the number of files where they have one of these is likely in the single digits as it's struggling to find more than 3 files.

RobQuistNL (approved):
LGTM!

lildude (approved):
LGTM. Thanks.

PR #6996, opened 2024-08-16
Add go.mod heuristics

goto1134:
## Checklist:

    - Sample source(s):
    - Sample license(s):
  
  ---
  
  As for the samples I've provided, they are modified copies of existing samples in the repository.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -471,6 +471,8 @@ disambiguations:
     pattern: '<!ENTITY '
   - language: NMODL
     pattern: '\b(NEURON|INITIAL|UNITS)\b'
+  - language: Go Module
+    pattern: '^\s*module\s+(\(|[\w\-./]+|"[^"]*")\s*$'
   - language: Modula-2
     pattern: '^\s*(?i:MODULE|END) [\w\.]+;'
   - language: [Linux Kernel Module, AMPL]
--- a/test/fixtures/Go Module/multiline/go.mod
+++ b/test/fixtures/Go Module/multiline/go.mod
@@ -1,0 +1,12 @@
+module (
+	golang.org/x/oauth2
+)
+
+go 1.11
+
+require (
+	cloud.google.com/go v0.34.0
+	golang.org/x/net v0.0.0-20190108225652-1e06a53dbb7e
+	golang.org/x/sync v0.0.0-20181221193216-37e7f081c4d4 // indirect
+	google.golang.org/appengine v1.4.0
+)
--- a/test/fixtures/Go Module/quoted/go.mod
+++ b/test/fixtures/Go Module/quoted/go.mod
@@ -1,0 +1,10 @@
+module "golang.org/x/oauth2"
+
+go 1.11
+
+require (
+	cloud.google.com/go v0.34.0
+	golang.org/x/net v0.0.0-20190108225652-1e06a53dbb7e
+	golang.org/x/sync v0.0.0-20181221193216-37e7f081c4d4 // indirect
+	google.golang.org/appengine v1.4.0
+)
--- a/test/fixtures/Go Module/unquoted/go.mod
+++ b/test/fixtures/Go Module/unquoted/go.mod
@@ -1,0 +1,10 @@
+module golang.org/x/oauth2
+
+go 1.11
+
+require (
+	cloud.google.com/go v0.34.0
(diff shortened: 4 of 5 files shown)


lildude:
This won't work as `.mod` is not associated with the "Go Module" language, and it probably shouldn't be.

A better solution would be to add `.go.mod` as an extension to the current language definition:

https://github.com/github-linguist/linguist/blob/d190ad88af555bb2ee213aa64205f2b9a97a077d/lib/linguist/languages.yml#L2547-L2557

... though usage isn't anywhere near popular enough for inclusion right now.

Until then, peeps will need to use an override to force the langauge.

goto1134:
> though usage isn't anywhere near popular enough for inclusion right now.

What are the criteria for being popular enough?

For the record, almost every Golang repo uses the `go.mod` files. >900k results in the following search: https://github.com/search?q=path:go.mod&type=code

> This won't work as .mod is not associated with the "Go Module" language, and it probably shouldn't be.

> A better solution would be to add .go.mod as an extension to the current language definition:

@lildude , thank you for your suggestion!

Most of the current `go.mod` files are misclassified for AMPL in the [search](https://github.com/search?q=path:go.mod&type=code), unless I query [`path:go.mod language:"Go Module"`](https://github.com/search?q=path:go.mod+language:%22Go+Module%22&type=code). 

As far as I understand, the problem is in the following code:
https://github.com/github-linguist/linguist/blob/a0a6d59bdcb7e4f4b3f549a7256cd8ff685a1d08/lib/linguist/heuristics.rb#L123

... heuristics are applied based on a filename suffix, and they do not depend on the extensions listed in `languages.yml`.

Could you please elaborate more on how are `languages.yml` connected with the `heuristics.yml` and how modifying the `languages.yml` will fix the problem I see?

lildude:
> What are the criteria for being popular enough?

See https://github.com/github-linguist/linguist/issues/5756, also referenced from the CONTRIBUTING.md file.

> For the record, almost every Golang repo uses the `go.mod` files. >900k results in the following search

Indeed. And Linguist already has support for `go.mod` files. I was referring to `*.go.mod` files. There are only 61 files indexed with the [`.go.mod` extension](https://github.com/search?q=path%3A*.go.mod&type=code).

> Most of the current `go.mod` files are misclassified for AMPL in the [search](https://github.com/search?q=path:go.mod&type=code), unless I query [`path:go.mod language:"Go Module"`](https://github.com/search?q=path:go.mod+language:%22Go+Module%22&type=code).

That's only in the search results, which uses a different internal library to detect the language. This internal library happens to feed off Linguist, but not perfectly as it's not a 1:1 match in terms of functionality. The actual files are already correctly identified by Linguist and the syntax highlighter applies the correct syntax highlighting too.

We can confirm this from the first result in your search results: https://github.com/jmoiron/sqlx. This is what the search results show - wrong language and wrong syntax highlighting:

Screenshot 2024-08-16 at 11 47 05

However if you view the file, you get the correct syntax highlighting:

Screenshot 2024-08-16 at 11 48 21

And if we use Linguist to analyse it, we get the correct language:

```console
➜ github-linguist sqlx/
99.74%  220442     Go
0.26%   583        Makefile
➜ github-linguist sqlx/go.mod
sqlx/go.mod: 9 lines (7 sloc)
  type:      Text
  mime type: text/plain
  language:  Go Module
➜ 
```

Which aligns with the language stats shown in the sidebar of the repo:

Screenshot 2024-08-16 at 11 54 19

AMPL is considered a detectable language by default in Linguist so if Linguist had incorrectly determined the language as AMPL, it would appear in the results above.

Unfortunately, no amount of tweaking Linguist is going to resolve this issue with what Search is showing. The team that handles Search will need to address this incorrect behaviour.

Please use the "Contact" link at the bottom of the page showing the wrong search results and log a support ticket. The issue will get routed to the correct team to address.

goto1134:
@lildude , it seems like my changes are not required, and the problem is on the GitHub side. Should I report the problem somewhere?

lildude:
> @lildude , thank you for clarifying that! It looks like my changes are not required, and the problem is on the GitHub side. Should I report the problem somewhere?

Yes. From my previous comment 😉:

> Please use the "Contact" link at the bottom of the page showing the wrong search results and log a support ticket. The issue will get routed to the correct team to address.

goto1134:
For the record, here is a bug report: https://github.com/orgs/community/discussions/136097

PR #7003, opened 2024-08-20
Optimization: Use `match?` instead of `match`

DecimalTurn:
## Description

This PR replaces instances of `.match` with `.match?` in heuristics.rb.
This change should have a small performance improvement as discussed [here](https://stackoverflow.com/a/42834177/5958842).

I also removed the double-negation inside `And.match?` to improve readability.

## Checklist:
N/A


Alhadis (approved):
I vaguely recall using this method some time ago, but it triggered test failures in an outdated version of Ruby that I assume we've dropped support for.

lildude (approved):
Makes sense to me. Thanks.

PR #7006, opened 2024-08-22
Add `uv.lock` as a TOML file

edgarrmondragon:
## Description

> `uv.lock` is a human-readable TOML file but is managed by uv and should not be edited manually.

https://docs.astral.sh/uv/concepts/projects/#lockfile

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3Auv.lock
    - Sample source(s):
      - https://github.com/edgarrmondragon/uv-click-example/blob/ce8dbff4f7b16e50b159b0cb46b00bd2ee8357b3/uv.lock
    - Sample license(s): MIT


lildude (changes requested):
Your sample is being surpressed by the diff because of its size which means it's too big for us. Please replace it with a smaller real-world sample.

edgarrmondragon:
> Your sample is being surpressed by the diff because of its size which means it's too big for us. Please replace it with a smaller real-world sample.

Done! 99b49baf54ebcc95c2b130a051305c839bc385a6

edgarrmondragon:
FWIW regarding https://github.com/github-linguist/linguist/labels/Pending%20Popularity, there were ~150 repos with a `uv.lock` file in search results this morning and we're at 183 now, so it's safe to assume the ecosystem will quickly reach the "hundreds of repositories" threshold.

lildude (approved):
Wow! A jump of several hundred in a week. Nice.

PR #7007, opened 2024-08-22
Add .ccm, .cxxm, .c++m extensions for C++

rdong8:
Adding the C++ module file extensions  `.c++m` , `.ccm`, `.cxxm` to [follow CMake's lead](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8308). Also made the extension list for C++ in ASCII alphabetical order (moved `.cpp` extension). 

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.c%2B%2Bm+module
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ccm+module
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.cxxm+module
    - Sample source(s):
      - All samples are my own
    - Sample license(s):
      - MIT
    - N/A

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -772,12 +772,15 @@ C++:
   aliases:
   - cpp
   extensions:
-  - ".cpp"
   - ".c++"
+  - ".c++m"
   - ".cc"
+  - ".ccm"
   - ".cp"
+  - ".cpp"
   - ".cppm"
   - ".cxx"
+  - ".cxxm"
   - ".h"
   - ".h++"
   - ".hh"
--- a/samples/C++/mathematics.ccm
+++ b/samples/C++/mathematics.ccm
@@ -1,0 +1,143 @@
+/// @module mathematics
+/// A sample library containing math functions.
+/// A documentation comment for the file like this one must be in any file you wish to be tracked by Doxygen.
+
+module;
+
+#include <cmath>
+#include <compare>
+#include <concepts>
+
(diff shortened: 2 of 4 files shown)


lib/linguist/languages.yml
```diff
@@ -772,12 +772,15 @@ C++:
   aliases:
   - cpp
   extensions:
-  - ".cpp"
```
lildude, line 775: Looks like you missed the comment right at the top of this file:

https://github.com/github-linguist/linguist/blob/d190ad88af555bb2ee213aa64205f2b9a97a077d/lib/linguist/languages.yml#L17-L19

If `.c++` is now considered the primary extension for the language, then this change is fine, but if it's not (which I suspect is the case) `.cpp` needs to remain the first extension in the list.

lildude (changes requested):
See inline comment. Additionally, I think this PR is very very premature with none of the extensions being anywhere near meeting our [usage requirements](https://github.com/github-linguist/linguist/issues/5756) (referenced in the CONTRIBUTING.md file too).

As such, I'm going to mark this as needing changes and close this PR. 

Feel free to open a new PR when usage has grown to close to meeting our requirements. You'll probably want to do this in multiple PRs so one extension doesn't hold up others. For example, all the indexed `.c++m` files are owned by two users so that alone would block this PR assuming all the other extensions met our requirements.

Until then, users can use an override to mark the files as C++.

PR #7009, opened 2024-08-22
Added The Device Modeling Language (DML)

TSonono:
- Add The Device Modeling Language (DML) to the Linguist project.
- Add the `.dml` as an extension for the DML language

## Description
The Device Modeling Language (DML) is an open source domain-specific language for writing fast functional or transaction-level device models for virtual platforms. DML provides high-level abstractions suitable for functional device models, including constructs like register banks, registers, bit fields, event posting, interfaces between models, and logging. DML code is compiled by the DML Compiler (DMLC), producing C code with API calls tailored for a particular simulator.

Currently, the compiler supports building models for the Intel® [Simics®](https://www.intel.com/content/www/us/en/developer/articles/tool/simics-simulator.html) simulator, but other back-ends may be added in the future.

DMLC is maintained by Intel.

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml+%22dml+1.4%22
      - For some reason I see 0 repositories. While it's unclear to me if there are 200 unique public repositories. Intel has a lot of code internally written in DML but they are unfortunately in private repositories. The same is the case with customer's using Intel Simics to write their device models using DML.
    - Sample source(s):
      - [sample-device.dml](https://github.com/TSonono/linguist/blob/dml/samples/Device%20Modeling%20Language/sample-device.dml)
      - [sample-interface.dml](https://github.com/TSonono/linguist/blob/dml/samples/Device%20Modeling%20Language/sample-interface.dml)
    - Sample license(s): Same as for Linguist project (MIT)
  - [x] I have included a syntax highlighting grammar: https://github.com/intel/device-modeling-language
  - [x] I have added a color
    - Hex value: `#0068B5`
    - Rationale: Intel logo and Simics logo are blue.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -479,6 +479,9 @@
 [submodule "vendor/grammars/desktop.tmbundle"]
 	path = vendor/grammars/desktop.tmbundle
 	url = https://github.com/Mailaender/desktop.tmbundle.git
+[submodule "vendor/grammars/device-modeling-language"]
+	path = vendor/grammars/device-modeling-language
+	url = https://github.com/intel/device-modeling-language.git
 [submodule "vendor/grammars/diff.tmbundle"]
 	path = vendor/grammars/diff.tmbundle
 	url = https://github.com/textmate/diff.tmbundle
--- a/grammars.yml
+++ b/grammars.yml
@@ -381,6 +381,8 @@ vendor/grammars/denizenscript-grammar:
 - source.denizenscript
 vendor/grammars/desktop.tmbundle:
 - source.desktop
+vendor/grammars/device-modeling-language:
+- source.dml
 vendor/grammars/diff.tmbundle:
 - source.diff
 vendor/grammars/dm-syntax:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -1762,6 +1762,16 @@ DenizenScript:
   - ".dsc"
   tm_scope: source.denizenscript
   language_id: 435000929
+Device Modeling Language:
+  type: programming
+  aliases:
+  - DML
+  extensions:
+  - ".dml"
+  ace_mode: text
+  color: "#0068B5"
+  tm_scope: source.dml
+  language_id: 806438883
 Dhall:
   type: programming
   color: "#dfafff"
--- a/samples/Device Modeling Language/sample-device.dml
+++ b/samples/Device Modeling Language/sample-device.dml
@@ -1,0 +1,73 @@
+dml 1.4;
+
+device sample_device;
+
+param desc = "A sample DML device";
+
+param documentation = "This is a sample DML device intended as a sample for"
+    + " the Linguist project";
+
+import "sample-interface.dml";
+
+attribute n_regs_written is (pseudo_attr, uint64_attr) "" {
+    method get() -> (attr_value_t) {
+        local uint8 n_written_regs = 0;
+        foreach reg in (each register_written in (regs)) {
+            if (reg.has_been_written())
+                n_written_regs += 1;
+        }
+        return SIM_make_attr_uint64(n_written_regs);
+    }
+
+    method set(attr_value_t val) throws {
+        if (SIM_attr_integer(val) != 0) {
+            log error: "n_regs_written can be only set to 0 to reset written"
+                + " status on all registers in the 'regs' bank";
+            throw;
+        }
+
+        foreach reg in (each register_written in (regs)) {
+            reg.reset_written();
+        }
+    }
+}
+
+implement sample {
+    saved int arg_sum = 0;
+    method simple_method(int arg) {
+        arg_sum += arg;
+    }
+}
+
+template register_written is (register, write) {
+    saved bool written;
+
+    method write(uint64 value) default {
+        this.written = true;
+        default(value);
+    }
+
+    shared method has_been_written() -> (bool) {
+        return this.written;
+    }
+
+    shared method reset_written() {
+        this.written = false;
+    }
+}
+
+bank regs {
+    param desc = dev.desc + "custom desc";
+    register r1 size 4 @ 0x0000 is (read, register_written) {
+        method read() -> (uint64) {
+            log info, 3: "read from r1";
+            return 42 + sample.arg_sum;
+        }
+    }
+    register r2 size 4 @ 0x0004 is (write, register_written) {
+        method write(uint64 value) {
+            log info, 3: "wrote %d to r2", value;
+            default(value);
+        }
+    }
+}
--- a/samples/Device Modeling Language/sample-interface.dml
+++ b/samples/Device Modeling Language/sample-interface.dml
@@ -1,0 +1,11 @@
+dml 1.4;
+
+header %{
+typedef struct {
+    void (*simple_method)(conf_object_t *obj, int arg);
+} sample_interface_t;
+%}
+
+extern typedef struct {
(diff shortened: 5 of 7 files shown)


lib/linguist/heuristics.yml
```diff
@@ -195,6 +195,10 @@ disambiguations:
     #  : dependency
     # path/file.ext1 : some/path/../file.ext2
     pattern: '([\/\\].*:\s+.*\s\\$|: \\$|^[ %]:|^[\w\s\/\\.]+\w+\.\w+\s*:\s+[\w\s\/\\.]+\w+\.\w+)'
+- extensions: ['.dml']
+  rules:
+  - language: Device Modeling Language
+    pattern: '(^dml\s1.4;)'
```
lildude, line 201: This will have no effect as Linguist doesn't know of any other language that uses the `.dml` extension. See my primary comment for more details.

lildude (changes requested):
> * For some reason I see 0 repositories. While it's unclear to me if there are 200 unique public repositories.

This will be because that only returns the count of repos with "dml 1.4" in their name.

> * Intel has a lot of code internally written in DML but they are unfortunately in private repositories. The same is the case with customer's using Intel Simics to write their device models using DML.

We rely on public usage only as there is no way to query private usage... cos it's private, even to GitHub 😁 

Regarding your search query:

> * https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml+%22dml+1.4%22

This search is quite precise. I assume `1.4` is the version. What about earlier versions? A [quick search](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml+%22dml+1%22) shows there are some files with `dml 1.0`, `dml 1.2`, `dml 1.4`. Are these the same language? 

Your specific search means there are not enough files for this PR to meet our [usage requirements](https://github.com/github-linguist/linguist/issues/5756) (also referenced in the CONTRIBUTING.md file).

If we [remove your qualifier](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml), we get a lot more files and things look a bit more promising, however a lot of the files returned don't appear to be the same language as you're proposing with this PR as we can see if we [exclude the `dml 1` string](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml+NOT+%22dml+1%22). As a result, these will all be misidentified as this language if this PR is merged as Linguist doesn't know of any other languages with the `.dml` extension.

There are far more of those other languages than this language if `dml 1.?` always has to be in the file, so at least one of those other languages needs to be identified and added at the same time in this PR. Your (relaxed?) heuristic will then be used to differentiate the two.

lib/linguist/heuristics.yml
TSonono: Ah, I was under the impression that if this regex didn't match, and given that there are no other languages in linguist that has a `.dml` extension, the language of the file would simply not be recognized.

TSonono:
> This search is quite precise. I assume 1.4 is the version. What about earlier versions? A [quick search](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml+%22dml+1%22) shows there are some files with dml 1.0, dml 1.2, dml 1.4. Are these the same language?

Yes, they are the same language. However, the syntax is quite different between 1.2 and 1.4, and the grammar file only covers 1.4. Do you have any suggestion regarding how this can be handled?


> If we [remove your qualifier](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml), we get a lot more files and things look a bit more promising, however a lot of the files returned don't appear to be the same language as you're proposing with this PR as we can see if we [exclude the dml 1 string](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.dml+NOT+%22dml+1%22). As a result, these will all be misidentified as this language if this PR is merged as Linguist doesn't know of any other languages with the .dml extension.

> There are far more of those other languages than this language if dml 1.? always has to be in the file, so at least one of those other languages needs to be identified and added at the same time in this PR. Your (relaxed?) heuristic will then be used to differentiate the two.

Would you say that if we include 1.0, 1.2 and 1.4 and add also add another language with a `.dml` extension, there is a chance that support for the DML (as in device modeling language) to be added to Github Linguist? Or is it not popular enough anyway?

lildude:
> Would you say that if we include 1.0, 1.2 and 1.4 and add also add another language with a `.dml` extension, there is a chance that support for the DML (as in device modeling language) to be added to Github Linguist? Or is it not popular enough anyway?

Even if you add a single entry for all `.dml` files that contain `dml 1`, you're still a long way off. When adding support for a language, all languages, filenames and extensions being added in the PR need to meet requirements.

kbrunham-intel:
We (Altera) have plans to public release repos with Simics code in the near future and would greatly benefit from having DML support in linguist. The repos we plan would be FPGA designs with their associated Simics model, in addition to specific training examples on using Simics and modeling.

I am aware of at least 4 major engineering companies that use both GitHub and Simics, where this linguist support would benefit them too. 

Please consider this a very strong endorsement of this PR.

lib/linguist/heuristics.yml
lildude: Nope. Linguist works like a funnel... everything goes in at the top and then works through each of the strategies in decreasing order of specificity until a single language is identified:

https://github.com/github-linguist/linguist/blob/5fad8d57605a914026a65b0e3ff6815d739944de/lib/linguist.rb#L51-L71

As Linguist would only know of the `.dml` extension with this language, it would stop the process there which as as you can see is before the heuristics strategy. It would also however apply this to all `.dml` files found on GitHub.

There is an exception to this: very generic extensions which we add to the [`generic.yml`](https://github.com/github-linguist/linguist/blob/main/lib/linguist/generic.yml) file. These are checked against the heuristics really early in the extension strategy.

Thinking about this further, I think `.dml` is a good candidate to add to `generic.yml` too which would then mean this heuristic _is_ used and will leave the other languages unaffected.

Please can you add this extension to the `generic.yml` file. This heuristic can also be improved (we don't need the capturing group and the tiny overhead it brings with it):

```suggestion
    pattern: '^dml\s1.4;'
```

lildude (changes requested):
I've checked the usage again for this precise version and the more lax `dml 1` and we're still a long way off from meeting out usage requirements.

I still recommend considering loosening the version you're targeting, at the expense of some possible syntax highlighting anomolies, as I see a lot of "older" versions associated with Simics based on the paths in the search results. This will speed up meeting our usage requirements.

I know this introduces a chicken-and-egg situation, but we have these requirements because we get a lot of requests to add support for hobby and very new languages that don't amount to anything over time. Making an exception for one means an exception must be made for all which defeats the purpose of having this requirement. Removing this requirement will lead to a very bloated and hard to manage and maintain project with no real benefit to the wider GitHub user base. 

I review usage of all PRs pending popularity each quarter when I prepare the Linguist release so this won't be forgotten. I will not comment each time as there are waaay too many PRs to do this for each.

lib/linguist/heuristics.yml
```diff
@@ -248,6 +248,10 @@ disambiguations:
...
```
lildude, line 254: Heuristics have no effect if Linguist only knows of one language with an extension. Removing.

```suggestion
```

lildude (changes requested):
Popularity looks good now, however you've got failing tests that need resolving, along with removing the unused heuristic.

lib/linguist/heuristics.yml
TSonono: Fixed

PR #7020, opened 2024-08-28
Add vimscript alias to Vim Script

jmdevin:
Add two aliases for vimscript

## Description
Using `vimscript` as a language tag does not cause it to be recognized and highlighted. This adds `vimscript` as an alias to allow that.

## Checklist:
[ Removed as it doensn't apply]

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7583,6 +7583,8 @@ Vim Script:
   - vim
   - viml
   - nvim
+  - vim script
+  - vimscript
   extensions:
   - ".vim"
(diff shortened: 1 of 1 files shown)


lib/linguist/languages.yml
```diff
@@ -7583,6 +7583,8 @@ Vim Script:
   - vim
   - viml
   - nvim
+  - vim script
```
lildude, line 7586: This is not needed as this is just the lowercase version of the current language.

```suggestion
```

lildude (approved):

PR #7026, opened 2024-09-02
Update number of acceptable classification errors.

DecimalTurn:
## Description

With the introduciton of the new classifier (https://github.com/github-linguist/linguist/pull/5103), the number of classification errors went from 19 to 14. 🎉

`Number of errors (14) is within the acceptable threshold (19)`
https://github.com/github-linguist/linguist/actions/runs/10663397934/job/29552498132?pr=7021

This means we have to change the threshold number to 14, otherwise it won't catch new errors that could be introduced in new PRs.

## Checklist:
N/A

### DIFF
--- a/script/cross-validation
+++ b/script/cross-validation
@@ -6,7 +6,7 @@ ACCEPTABLE_ERRORS = 14
 
 # Number of acceptable classification errors when using --all.
 # It should only be decreased.
-ACCEPTABLE_ERRORS_ALL = 777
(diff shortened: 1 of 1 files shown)


DecimalTurn:
Regarding the number of errors when using the `--all` option, it gave me 744 errors instead of 777, so I've updated it as well, but not sure if this is really used.

lildude (approved):
Whoops. I forgot about those. Thanks.

PR #7049, opened 2024-09-19
Update heuristics to fallback to TSX language if .tsx file is not XML

### LINKED ISSUE #7039: Syntax highlighting doesn't work well on files with `.tsx` extension
## Describe the bug

When opening a pull request or when the code is merged into the main branch, the syntax highlighting for `.tsx` TypeScript React files doesn't work properly, leaving part of the file in plain text.

### Expected behaviour

The expected behavior should be to always display the syntax highlighting that corresponds to the `.tsx` extension.

### Additional notes

As additional information, I have managed to correct this by importing React at the beginning of each file.

```tsx
import * as React from 'react';
```

But this should not be necessary today, since it is not an obligation to have to import React.

Captura de pantalla 2024-09-15 a la(s) 11 33 13

vicasas:
## Description

This PR fixes the syntax highlighting issue for .tsx files in the TSX language. Previously, the heuristic required detecting a React import pattern in the code. However, in recent versions of React, importing React is no longer necessary, which often resulted in the file being treated as plain text.

The behavior has now been updated to first check for matches with XML languages. If no match is found, it will assume the file is a TSX.

Closes #7039

## Checklist:

    - Sample source(s):
    - Sample license(s):

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -780,10 +780,9 @@ disambiguations:
   - language: Scilab
 - extensions: ['.tsx']
   rules:
-  - language: TSX
-    pattern: '^\s*(import.+(from\s+|require\()[''"]react|\/\/\/\s*<reference\s)'
   - language: XML
     pattern: '(?i:^\s*<\?xml\s+version)'
+  - language: TSX
 - extensions: ['.txt']
   rules:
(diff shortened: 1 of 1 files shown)


lildude (approved):
LGTM. Thanks.

PR #7052, opened 2024-09-23
Add C3 support, again.

RealPacket-sys:
I am honestly tired of doing this just to know that randomly at some point, for no predictable reason, GitHub will suspend my account, whether at 6 PM or 10 PM. I probably should go change my main email to something that isn't outlook before they delete my account again.
Also, the accounts with the default profile is because it can't find the profile of my suspended account.

## Checklist:
      
    - Sample source(s):
      - The C3 standard library, all of it.
    - Sample license(s):
       - MIT (the compiler is licensed under LGPL, however)
  - [X] I have included a syntax highlighting grammar: [URL to grammar repo]
  - [X] I have added a color
    - Hex value: `#7C3AED`
    - Rationale: It's the end of the gradient of the C3 logo from the C3 website. See [c3lang/c3-web/public/logo.svg#17](https://github.com/c3lang/c3-web/blob/cb671a8cf5925e2319b6ced92a7a353f7fe8627f/public/logo.svg#L17)
  - [X] I have updated the heuristics to distinguish my language from others using the same extension.


lib/linguist/heuristics.yml
```diff
@@ -340,6 +340,10 @@ disambiguations:
   - language: C++
     named_pattern: cpp
   - language: C
+- extensions: ['.c3', '.c3i', '.c3t']
+  rules:
+  - language: C3
+    named_pattern: c3
```
lildude, line 346: A heuristic is not necessary when there is only one language using an extension. You can remove this.

lib/linguist/languages.yml
```diff
@@ -824,6 +824,18 @@ C2hs Haskell:
   codemirror_mode: haskell
   codemirror_mime_type: text/x-haskell
   language_id: 45
+C3:
+  type: programming
+  color: "#7C3AED"
+  extensions:
+  - ".c3"
+  - ".c3i"
+  - ".c3t"
```
lildude, line 833: We need search result links in the template for all extensions individually. Three extensions means three links, and they each need to meet our usage requirements individually.

I recommend removing any that are no where near meeting our requirements.

lildude (changes requested):
> The C3 standard library, all of it.

We do not need this many samples. 2-3 of the most representative of each extension is enough.

lib/linguist/heuristics.yml
RealPacket-sys: There are [1] [2] [3] some that have the C3 extension (see https://github.com/search?q=%28path%3A*.c3+OR+path%3A*.c3i+OR+path%3A*.c3t%29+language%3AC+AND+NOT+owner%3Ac3lang+NOT+is%3Afork&type=code)
[1]: https://github.com/windelbouwman/ppci/blob/915c069e0667042c085ec42c78e9e3c9a5295324/librt/io.c3#L4
[2]: https://github.com/arcanon/ipadflash/blob/6e3f35fe4b36156d79d3b09ac9b13f36a9802346/src/SDL/src/main/dummy/SDL_dummy_main.c3#L4
[3]: https://github.com/as112/STM32F103-Boiler-Controller/blob/7d84ac6992df5cbe436a73450e2d2a10eb6fb8c7/ds18b20-termostat-v2/CMSIS/src/core_cm3.c3#L4

RealPacket-sys: The heuristics should help.
Also, should help more in cases where there are no file extension, if linguist even considers them or uses the heuristics / or samples.

lib/linguist/languages.yml
RealPacket-sys: > We need search result links in the template for all extensions individually. Three extensions means three links, and they each need to meet our usage requirements individually.
> 
> I recommend removing any that are no where near meeting our requirements.

I'll add it though.

RealPacket-sys:
image
"Hey, we don't support this query when searching in X mode, would you like to search in X mode instead?"

lib/linguist/heuristics.yml
lildude: The heuristic won't help at all as it won't be used. If linguist only knows of one language with a certain extension, it'll identify all files it finds with that extension as that language

RealPacket-sys: Oh, ok.

lib/linguist/languages.yml
lildude: Your `.c3t` search returns zero results. This will hold back this PR. We look at extensions in isolation so each needs to meet our requirements before the PR is merged. Once Linguist knows about a language, users can use an override for less popular extensions until they meet usage requirements and are added.

RealPacket-sys:
Some files seem to be using `.c3t` as a JSON file (in the context of 3D / 2D models?). I think maybe later JSON should have C3T as a file, but that does mean that this would need heuristics.

RealPacket-sys:
This is the search that should get most of the C3T files being used as JSON Models:
https://github.com/search?q=%28path%3A*.c3t%29+AND+%7B+AND+%28id+OR+version+OR+%28meshes+AND+attributes%29%29+NOT+is%3Afork&type=code

PR #7062, opened 2024-09-26
Add varlink

fossdd:
## Description

https://varlink.org/

Adds varlink grammar and language

[systemd](https://github.com/systemd/systemd) started using it (see https://media.ccc.de/v/all-systems-go-2024-276-varlink-now-), so I think it's getting popular

## Checklist:
      
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.varlink
    - Sample source(s):
      - https://github.com/varlink/com.redhat.system/blob/master/src/com.redhat.system.varlink
    - Sample license(s):
      - MIT and Apache
  - [x] I have included a syntax highlighting grammar: https://github.com/varlink/language-varlink
      
  - [x] I have added a color
    - Hex value: `#42f1f4`
    - Rationale: It's the prefered color of varlink

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -8590,6 +8590,17 @@ sed:
 templ:
   type: markup
   color: "#66D0DD"
+varlink:
+  type: markup
+  color: "#5882FA"
+  extensions:
+  - ".varlink"
+  ace_mode: text
+  tm_scope: source.varlink
+  language_id: 396563549
+wdl:
+  type: programming
+  color: "#42f1f4"
   extensions:
   - ".templ"
   ace_mode: text
--- a/vendor/README.md
+++ b/vendor/README.md
@@ -598,6 +598,7 @@ This is a list of grammars that Linguist selects to provide syntax highlighting 
 - **UnrealScript:** [textmate/java.tmbundle](https://github.com/textmate/java.tmbundle)
 - **UrWeb:** [gwalborn/UrWeb-Language-Definition](https://github.com/gwalborn/UrWeb-Language-Definition)
 - **V:** [0x9ef/vscode-vlang](https://github.com/0x9ef/vscode-vlang)
+- **varlink:** [varlink/language-varlink](https://github.com/varlink/language-varlink)
 - **VBA:** [serkonda7/vscode-vba](https://github.com/serkonda7/vscode-vba)
 - **VBScript:** [peters-ben-0007/VBDotNetSyntax](https://github.com/peters-ben-0007/VBDotNetSyntax)
 - **VCL:** [brandonwamboldt/sublime-varnish](https://github.com/brandonwamboldt/sublime-varnish)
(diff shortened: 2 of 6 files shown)


lildude:
> systemd started using it (see https://media.ccc.de/v/all-systems-go-2024-276-varlink-now-), so I think it's getting popular

With only 51 indexed files, it's got a very long way to go to meet our usage requirements.

vendor/README.md
```diff
@@ -597,9 +597,14 @@ This is a list of grammars that Linguist selects to provide syntax highlighting
...
 - **UnrealScript:** [textmate/java.tmbundle](https://github.com/textmate/java.tmbundle)
 - **UrWeb:** [gwalborn/UrWeb-Language-Definition](https://github.com/gwalborn/UrWeb-Language-Definition)
+<<<<<<< HEAD
 - **V:** [0x9ef/vscode-vlang](https://github.com/0x9ef/vscode-vlang)
 - **VBA:** [serkonda7/vscode-vba](https://github.com/serkonda7/vscode-vba)
 - **VBScript:** [peters-ben-0007/VBDotNetSyntax](https://github.com/peters-ben-0007/VBDotNetSyntax)
+=======
+- **Vala:** [technosophos/Vala-TMBundle](https://github.com/technosophos/Vala-TMBundle)
+- **varlink:** [varlink/language-varlink](https://github.com/varlink/language-varlink)
+>>>>>>> Add varlink grammar
```
lildude, line 607: You need to fix this

lildude (changes requested):
Looks like you've still got remnants of a merge conflict that need cleaning up.

lib/linguist/languages.yml
```diff
@@ -8590,6 +8590,17 @@ sed:
 templ:
   type: markup
   color: "#66D0DD"
+varlink:
+  type: programming
```
lildude, line 8594: Is this really a programming language. Feels more like configuration so would fall under data

fossdd: I think it's a markup language as a varlink file describes interfaces (see https://en.wikipedia.org/wiki/Interface_description_language). What do you think?

vendor/README.md
fossdd: Oh yeah, thanks fixed it!

fossdd:
Thanks for the review! I addressed the issues. Yeah, varlink still isn't used that much (even though it's a few years old now), but I think with systemd adopting varlink it has a pretty good starting point.

lildude (changes requested):
The cached licence file needs to be the one generated by the `script/add-grammar` script you should have used to add the grammar. If you did not use this script, revert your changes and use it and then push your changes. Don't worry about keeping the history clean as we squash all commits when merging.

lib/linguist/languages.yml
lildude: Markup is for things that result in visual effects, think html. From what I've read, the only varlink files that will be committed are the interface definition files which are data by the fact they're definition files… they're defining the methods the interface implements.

PR #7084, opened 2024-10-09
Add `.slnx` to XML extensions

kasperk81:
## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.slnx+solution
    - Sample source(s):
      - https://github.com/visose/Robots/blob/aa2be4ba40947b950b79d68fae3e22f32a0184c2/Robots.slnx
    - Sample license(s):

Fix https://github.com/github-linguist/linguist/issues/7083


rainersigwald (commented):
Ah yeah, getting the default XML highlighting is a great low-cost way to do this!

lildude (changes requested):
Your sample should be in the `samples/XML` directory.

lildude:
> @lildude ok, but can you tell me why?

Sure... it's what you want 😉 

Screenshot 2024-10-14 at 17 52 58

As you're adding the extension to the XML language entry, you're saying these files are XML and should be highlighted as XML and as such, the sample needs to go into the XML language directory.

lildude:
You need to add this to the list of no-root-tag entries in the test too:

https://github.com/github-linguist/linguist/blob/f164d13fa618023ecf2d8f2ed9a6ce5fae731346/test/test_strategies.rb#L176-L194

This will exclude the sample from the XML strategy testing which is what is failing here.

lildude:
> Hello @lildude just wondering if you and your team had had the chance of taking a look into this.

No. I only review pending popularity PRs when I make a new release which happens approximately every 3-4 months.

PR #7098, opened 2024-10-21
Add Cargo.nix generated by crate2nix

flashgnash:
Cargo.nix is the name of a file generated by crate2nix from cargo.toml and cargo.lock which dwarfs all other files in smaller projects, heavily skewing language percentage readouts

(it is effectively just a nix-ified version of cargo.lock)

Add crate2nix generated file "Cargo.nix" 

## Checklist:

I am unsure what test changes would be neccesary, as I have added a single glob


lildude (changes requested):
If you're wanting the file to be marked as generated, you should add it to the [`generated.rb`](https://github.com/github-linguist/linguist/blob/f164d13fa618023ecf2d8f2ed9a6ce5fae731346/lib/linguist/generated.rb) file, not the `vendor.yml` file.

You need to add a test too.

PR #7121, opened 2024-11-03
Add WORKSPACE.bzlmod to Starlark

petemounce:
Bazel's bzlmod dependency manager involves Starlark files with `.bzlmod` file extensions.

https://bazel.build/external/migration#workspace.bzlmod

## Checklist:

      Please note: bazel-y people tend toward monorepos, so high contributor count vs smaller repo count

      - 8k repos: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.bzlmod+bazel
      - 108 repos: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.bzlmod+bzlmod
    - Sample source(s):
      - https://github.com/bazel-contrib/rules_dotnet/pull/460 -> https://github.com/bazel-contrib/rules_dotnet/blob/master/examples/WORKSPACE.bzlmod
    - Sample license(s):
  - > [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.
  
      Sorry, I don't follow what this one is asking?

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7004,6 +7004,7 @@ Starlark:
   color: "#76d275"
   extensions:
   - ".bzl"
+  - ".bzlmod"
   - ".star"
   filenames:
   - BUCK
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7004,7 +7004,6 @@ Starlark:
   color: "#76d275"
   extensions:
   - ".bzl"
-  - ".bzlmod"
   - ".star"
   filenames:
   - BUCK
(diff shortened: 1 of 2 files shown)


lildude (changes requested):
The sample needs to be added to this PR.

The docs you referenced also only mentions `WORKSPACE.bzlmod` and excluding this file from search returns only 7 files, six of which are owned by the same repo so I think it's probably better to add this as a new named file for now and not as a new extension.

petemounce:
@lildude many thanks for the fast turnaround.

I haven't understood what to do about the sample license(s) part; how to proceed there?

I've adjusted to only aim for `WORKSPACE.bzlmod`.

lildude:
> I've adjusted to only aim for `WORKSPACE.bzlmod`.

Thanks. We still need a sample.

petemounce:
@lildude is something like ln 9650c7c what you have in mind?

PR #7144, opened 2024-12-02
Add URScript language

EbbeFuglsang:
URScript is the custom programming language, developed by Universal Robots, that controls the robot arm. Any interactions with PolyScope get converted into URScript commands, and are subsequently sent to the robot to be executed. This high level language is easy to learn, and allows you to program the robot without the PolyScope graphical interface. The URScript Manual contains an overview of the language structure, information on data types, and a complete reference of the standard functions.

## Checklist:

    - Search results for each extension: https://github.com/search?utf8=%E2%9C%93&q=urscript&type=repositories&ref=searchresults
      
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.script+URScript
    - Sample source(s):
    - Sample license(s): https://github.com/UniversalRobots/URScript_Examples/tree/main/admittance-control BSD 3-Clause License
  - [x] I have included a syntax highlighting grammar: [URL to grammar repo]
      
  - [x] I have added a color
    - Hex value: `#RRGGBB`
    - Rationale: This is Universal Robots brand color
  - [X] I have updated the heuristics to distinguish my language from others using the same extension.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '^\s*end\b'
+    pattern: '^\s*def\s+[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\s+end+\s'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -709,6 +709,10 @@ disambiguations:
   - language: Markdown
     # Markdown syntax for scdoc
     pattern: '^#+\s+(NAME|SYNOPSIS|DESCRIPTION)'
+- extensions: ['.script']
+  rules:
+  - language: URScript
+    pattern: '^\s*end\b'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '^\s*def\s+[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\s+end+\s'
+    pattern: '^\s*def\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*\(([^)]*)\)\s*:(((?:(?!^\send($|\s))[\s\S])*)\send($|\s))'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '^\s*def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
+    pattern: '(^\s*|\s+)def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '^\s*def\s*[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\send($|\s)'
+    pattern: '^\s*def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '^\s*def\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*\(([^)]*)\)\s*:(((?:(?!^\send($|\s))[\s\S])*)\send($|\s))'
+    pattern: '^\s*def\s*[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\send($|\s)'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '(^\s*|\s+)def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
+    pattern: '^\s*def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7612,7 +7612,7 @@ URScript:
   color: "#56A2D6"
   extensions:
   - ".script"
-  tm_scope: source.python
+  tm_scope: source.urscript
   ace_mode: text
   language_id: 431829457
 UrWeb:
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7607,6 +7607,14 @@ UnrealScript:
   codemirror_mode: clike
   codemirror_mime_type: text/x-java
   language_id: 382
+URScript:
+  type: programming
+  color: "#56A2D6"
+  extensions:
+  - ".script"
+  tm_scope: source.python
+  ace_mode: text
+  language_id: 431829457
 UrWeb:
   type: programming
   color: "#ccccee"
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7547,6 +7547,14 @@ Typst:
   tm_scope: source.typst
   ace_mode: text
   language_id: 704730682
+URScript:
+  type: programming
+  color: "#56A2D6"
+  extensions:
+  - ".script"
+  tm_scope: source.urscript
+  ace_mode: text
+  language_id: 431829457
 Unified Parallel C:
   type: programming
   color: "#4e3617"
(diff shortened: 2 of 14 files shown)


EbbeFuglsang (commented):
.

lib/linguist/languages.yml
```diff
@@ -7607,6 +7607,14 @@ UnrealScript:
   codemirror_mode: clike
   codemirror_mime_type: text/x-java
   language_id: 382
+URScript:
+  type: programming
+  color: "#56A2D6"
+  extensions:
+  - ".script"
+  tm_scope: source.python
```
lildude, line 7615: This should be the scope from the grammar:

```suggestion
  tm_scope: source.urscript
```

vendor/README.md
```diff
@@ -595,6 +595,7 @@ This is a list of grammars that Linguist selects to provide syntax highlighting
 - **TypeScript:** [tree-sitter/tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript) 🐌
 - **TypeSpec:** [microsoft/typespec](https://github.com/microsoft/typespec)
 - **Typst:** [michidk/typst-grammar](https://github.com/michidk/typst-grammar)
+- **URScript:** [MagicStack/MagicPython](https://github.com/MagicStack/MagicPython)
```
lildude, line 598: Huh!? This file is auto-updated so this shouldn't be pointing to the MagicPython grammar.

Remove the line and run `script/list-grammars` - this should fix things.

vendor/licenses/git_submodule/urscript-extension.dep.yml
```diff
@@ -1,22 +1,26 @@
 ---
-name: typst-grammar
-version: 180d59e6f4e09941d778b5fa6fc48d52211294a1
+name: urscript-extension
+version: 51d5c4021fccc1fd9549c01f80156d3e001ee5ed
```
lildude, line 3: Why are you renaming and updating this file? The `script/add-grammar` script you should have run to add the grammar should have created a new file for the new grammar.

Please revert this change and add the correct license file.

lildude (changes requested):
In addition to the inline comments, you're also going to need to add this to `generic.yml` and come up with a 100% precise matching regex and add it to the heuristics. This is because `.script` is incredibly generic so the only way we'll be able to correctly identify these files is via a heuristc.

Popularity is no where near ready for merging so you might want to consider holding off on the suggested changes, close this PR and wait for popularity to meet our requirements.

vendor/licenses/git_submodule/urscript-extension.dep.yml
lildude: Ooof. I know why this happened. https://github.com/github-linguist/linguist/issues/7145 You'll need to redo this one that issue has been resolved.

EbbeFuglsang:
Regarding popularity this search:
[urscript OR UR5 AND ROS OR URCap](https://github.com/search?type=repositories&q=urscript+OR+UR5+AND+ROS+OR+URCap) gives 1.6k repos. They might not all have the URScript in seperate file. But with this search:
https://github.com/search?type=code&q=%28urscript+AND+path%3A*.script%29+OR+%28UR5+AND+ROS+AND+path%3A*.script%29+OR+%28URCap+AND+path%3A*.script%29+ gives 542 different .script files that holds URScript.
What is the threshold?

Regarding the heuristc. That makes a lot of sense. I will look into that. Thank you for you feedback so far🙏

lildude:
> What is the threshold?

https://github.com/github-linguist/linguist/blob/5fad8d57605a914026a65b0e3ff6815d739944de/CONTRIBUTING.md?plain=1#L76-L80

EbbeFuglsang:
I just added some heuristics I think that should reduce the risk of false positives.

lib/linguist/heuristics.yml
```diff
@@ -709,6 +709,10 @@ disambiguations:
   - language: Markdown
     # Markdown syntax for scdoc
     pattern: '^#+\s+(NAME|SYNOPSIS|DESCRIPTION)'
+- extensions: ['.script']
+  rules:
+  - language: URScript
+    pattern: '^\s*end\b'
```
lildude, line 715: This is way too vague and will result in a lot of misclassification. This heuristic needs to be 100% unlikely to match anything other than this language due to the nature of the generic script.

We also need a test for the heuristic.

EbbeFuglsang: That makes a lot of sense! Thank you for the feedback🙏

I updated it for being more specific and added a test.

lib/linguist/heuristics.yml
```diff
@@ -709,6 +709,10 @@ disambiguations:
   - language: Markdown
     # Markdown syntax for scdoc
     pattern: '^#+\s+(NAME|SYNOPSIS|DESCRIPTION)'
+- extensions: ['.script']
+  rules:
+  - language: URScript
+    pattern: '^\s*def\s+[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\s+end+\s'
```
lildude, line 715: This regex is vulnerable to ReDoS attack.

EbbeFuglsang: I see. I will try to find something that is not suffering from that and is still able to capture the URScript. 

Again thank you for your feedback🙏

EbbeFuglsang: I updated the regex and tested it with https://devina.io/redos-checker 

Regarding the specificity of the regex. I try to match against all other samples in the project, and got the following results:  
```
@EbbeFuglsang ➜ /workspaces $ ruby search.rb ./linguist/samples '^\s*def\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*\(([^)]*)\)\s*:(((?:(?!^\send($|\s))[\s\S])*)\send($|\s))'
Could not read file ./linguist/samples/Altium Designer/Sample Schematic Sheet.SchDoc: invalid byte sequence in UTF-8
Could not read file ./linguist/samples/Oberon/beNull.ob2: invalid byte sequence in UTF-8
Could not read file ./linguist/samples/Roff/roff.1in: invalid byte sequence in UTF-8
Match found in: ./linguist/samples/URScript/admittance_control.script
Match found in: ./linguist/samples/URScript/math.script
Match found in: ./linguist/samples/URScript/payload_get_struct.script
Match found in: ./linguist/samples/URScript/payload_manipulators.script
Match found in: ./linguist/samples/URScript/payload_set.script
Match found in: ./linguist/samples/URScript/payload_tests.script
Match found in: ./linguist/samples/jq/builtin.jq
Match count: 7
Files count: 3047
```
If the samples is representative for all other code, the amount of false positive is not too bad.

lib/linguist/heuristics.yml
```diff
@@ -709,6 +709,10 @@ disambiguations:
   - language: Markdown
     # Markdown syntax for scdoc
     pattern: '^#+\s+(NAME|SYNOPSIS|DESCRIPTION)'
+- extensions: ['.script']
+  rules:
+  - language: URScript
+    pattern: '^\s*def\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*\(([^)]*)\)\s*:(((?:(?!^\send($|\s))[\s\S])*)\send($|\s))'
```
lildude, line 715: Thanks for the improvement, however this now has a problem in that it's no longer run linearly which means it'll likely have performance issues:

```ruby
irb(main):002> Regexp.linear_time?(/^\s*def\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*\(([^)]*)\)\s*:(((?:(?!^\send($|\s))[\s\S])*)\send($|\s))/)
=> false
```

We're trying to improve the regexes in this file to improve performance and prevent timing out (we catch these ourselves but really shouldn't) so we're aiming for all regexes to have linear execution whilst also being immune from ReDos. [^1]

/cc @Alhadis

[^1]: I'm slowly doing this in a separate branch but ultimately this will be enforced so we may as well get it right now rather than down the road when it's enforced.

vendor/licenses/git_submodule/urscript-extension.dep.yml
```diff
@@ -1,22 +1,26 @@
 ---
-name: typst-grammar
-version: 180d59e6f4e09941d778b5fa6fc48d52211294a1
+name: urscript-extension
```
lildude, line 2: You need to merge main into your fork. Your PR should not be renaming and changing the `typst-grammar.dep.yml` file; it should only be adding new file.

lildude (changes requested):
See inline comments.

lib/linguist/heuristics.yml
EbbeFuglsang: It makes a lot of sense. Thank you for insight!

vendor/licenses/git_submodule/urscript-extension.dep.yml
EbbeFuglsang: I am not sure how that happened. I hope it is in order now🤞

lildude: Yup. Looking good now.

lib/linguist/heuristics.yml
```diff
@@ -709,6 +709,10 @@ disambiguations:
   - language: Markdown
     # Markdown syntax for scdoc
     pattern: '^#+\s+(NAME|SYNOPSIS|DESCRIPTION)'
+- extensions: ['.script']
+  rules:
+  - language: URScript
+    pattern: '^\s*def\s*[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\send($|\s)'
```
lildude, line 715: This looks much better, thanks. I think there's still room for improvement:

```suggestion
    pattern: '^\s*def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*\send\b'
```

- `^\s*def\s*` replaced with `^\s*def\s+` as it appears at least one space is required
- `[a-zA-Z0-9_]*` is the same as `\w*`
- `\s*end($|\s)` is simplified to `\s*end\b` as `\b` matches a word boundary, which is equivalent to the end of the word `end`.

@Alhadis may have some further suggestions.

Alhadis: 1.	`\send\b` will match a (presumably illegal) construction like `end.`, `end-`, `end(`, and many, _many_ others. 
2.	For performance's sake, `[\s\S]*` should match conservatively (`[\s\S]*?`) so that anything past the first `\send\b` isn't searched and included as part of the match:

	~~~ruby
	#!/usr/bin/env ruby
	greedy  = /^\s*def\s*[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*\send($|\s)/
	thrifty = /^\s*def\s*[a-zA-Z_][a-zA-Z0-9_]*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)/
	input   = File.read("samples/URScript/admittance_control.script")

	greedy  =~ input; $&.length # => 10019
	thrifty =~ input; $&.length # => 126
	~~~

lib/linguist/heuristics.yml
```diff
@@ -712,7 +712,7 @@ disambiguations:
 - extensions: ['.script']
   rules:
   - language: URScript
-    pattern: '^\s*def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
+    pattern: '(^\s*|\s+)def\s+[a-zA-Z_]\w*\s*\([^)]*\)\s*:[\s\S]*?\send($|\s)'
```
lildude, line 715: This is unnecessary (`\s*` is zero or more which includes `\s+` which is one or more) and introduces a ReDoS.

EbbeFuglsang: I see. I reverted the commit.

PR #7158, opened 2024-12-14
docs: Add link to Docker build instructions further down in the README to the How To Run

vorburger:
This is a very minor addition to the README.

Because I was (initially) too dumb to see this, and without a Ruby background, did not immediately know how to build this, in order to contribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ cd /path-to-repository
 github-linguist
 ```
 
+You can locally build the `github-linguist` binary either [using Docker](#docker) or via the [contributing documention](#contributing).
+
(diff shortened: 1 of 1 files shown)


lildude:
I'm not sure we need this. The point where you've added this is based on the fact you've installed the gem which is detailed right at the beginning of the file.

vorburger:
> I'm not sure we need this. The point where you've added this is based on the fact you've installed the gem which is detailed right at the beginning of the file.

All right, no worries; I'm happy to close this. (I'll update my [unrelated] other #7163 some time over the coming days or so.)

PR #7163, opened 2024-12-14
Add `.timer` & various other systemd extensions (fixes #7161)

### LINKED ISSUE #7161: Missing additional extensions for systemd INI format style files
## Describe the bug

I noticed on my https://github.com/vorburger/vorburger-dotfiles-bin-etc/commit/d6eb521e0e43de7c198b167d05126946c9b56a72 that a (systemd, INI format style) `.service` had syntax highlighting, but `.timer` did not:

``` 
$ github-linguist systemd/flatpak-update.service
systemd/flatpak-update.service: 6 lines (5 sloc)
  type:      Text
  mime type: text/plain
  language:  desktop

$ github-linguist systemd/flatpak-update.timer
systemd/flatpak-update.timer: 10 lines (8 sloc)
  type:      Text
  mime type: text/plain
  language:
```

I'll see if it's easy enough to contribute a PR with a proposed fix for this...

vorburger:
## Description

Fixes #7161.

## Checklist:

  - TODO [ ] The new extension is used in hundreds of repositories on GitHub.com
      
  - TODO [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
    - Sample license(s):
  - [X] I have verified that there is no need to include a change to the heuristics to distinguish my language from others using the same extension (because all these new systemd specific file name extensions are not yet used by other other language).

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -8468,10 +8468,38 @@ crontab:
   language_id: 705203557
 desktop:
   type: data
+  # https://en.wikipedia.org/wiki/INI_file
+  filenames:
+  - "systemd-system.conf"
+  - "systemd-user.conf"
+  - "logind.conf"
+  - "journald.conf"
+  - "journal-remote.conf"
+  - "journal-upload.conf"
+  - "systemd-sleep.conf"
+  - "timesyncd.conf"
   extensions:
   - ".desktop"
   - ".desktop.in"
+  # systemd unit-file types; extensions from
+  # https://www.freedesktop.org/software/systemd/man/latest/systemd.syntax.html
+  - ".automount"
+  - ".device"
+  - ".link"
+  - ".mount"
+  - ".netdev"
+  - ".network"
+  - ".nspawn"
+  - ".path"
+  - ".scope"
   - ".service"
+  - ".slice"
+  - ".snapshot"
+  - ".socket"
+  - ".swap"
+  - ".target"
+  - ".timer"
+  - ".unit"
   tm_scope: source.desktop
   ace_mode: text
   language_id: 412


lib/linguist/languages.yml
```diff
@@ -8468,10 +8468,28 @@ crontab:
   language_id: 705203557
 desktop:
   type: data
+  # https://en.wikipedia.org/wiki/INI_file
```
lildude, line 8471: These are all in the wrong place. If they're INI files, we already have a [section](https://github.com/github-linguist/linguist/blob/1a61b82c42754ff84f1a3e108150a2a7461c10bc/lib/linguist/languages.yml#L3056) for that.

lildude (commented):
> How strictly does this project require "including a real-world usage sample for all extensions added in this PR"? I sure could copy/paste a random example from those "Search results for each extension" linked above, but... what's the point of that? (I'm happy to do so, if it's required to merge this, of course; just wasn't sure if really was a hard-requirement.)

It is. They're used to train the classifier.

Also every filename and extension needs to meet our usage requirements. One lesser used file or extension will hold up the PR until it meets the usage requirements.

vorburger:
> > How strictly does this project require "including a real-world usage sample for all extensions added in this PR"? I sure could copy/paste a random example from those "Search results for each extension" linked above, but... what's the point of that? (I'm happy to do so, if it's required to merge this, of course; just wasn't sure if really was a hard-requirement.)
> 
> It is. They're used to train the classifier.

ACK, thanks; I'll add samples.

> Also every filename and extension needs to meet our usage requirements. One lesser used file or extension will hold up the PR until it meets the usage requirements.

Do you prefer separate small PRs - one for each?

lib/linguist/languages.yml
vorburger: > These are all in the wrong place. If they're INI files, we already have a [section](https://github.com/github-linguist/linguist/blob/1a61b82c42754ff84f1a3e108150a2a7461c10bc/lib/linguist/languages.yml#L3056) for that.

But then the already existing (systemd) ".service" is in the wrong place as well? Do you want me to move that one too? Otherwise it's inconsistent. Full disclaimer: I don't yet fully understand what difference "desktop" (which are also INI) VS "ini" language makes either way. What I could do is raise a separate small PR, just to first move . service?

lildude:
> Do you prefer separate small PRs - one for each?

One PR is fine, and be sure to only include extensions that meet our usage requirements as the PR will be held up waiting for any that don't meet our requirements.

lib/linguist/languages.yml
lildude: 🤔 good point. Leave these here. I'd forgotten @Alhadis added `.service` in https://github.com/github-linguist/linguist/pull/5768 and makes the point the files are based on desktop which is essentially INI.

We def don't need the comments and only really want files and extensions which meet our usage requirements, along with a sample of each.

vorburger:
> We def don't need the comments

Inline comments now removed.

> and only really want files and extensions which meet our usage requirements

These do, do the best of my understanding of your criteria, and knowledge. If I misunderstood, let me know, and I'll remove.

See https://github.com/github-linguist/linguist/issues/7161#issuecomment-2543120512 for which extensions I have opted NOT to include.

> along with a sample of each

Samples now added.

lib/linguist/languages.yml
Alhadis: @lildude Yeah, something I didn't address in #5768 was the observation that the "Desktop" language ought to be rolled into INI (which appears to be the correct course of action, AFAICT), because apart from having distinct syntax highlighting grammars, each entry refers to essentially the same format.

I only added `.service` to Desktop's file extensions because its usage appeared to be affiliated with <samp>systemd(1)</samp> and I didn't have time to look into whether or not it was worth merging Desktop into INI.

lildude (changes requested):
Please check the usage for each of these new files and extensions and add searches for the filenames.

A quick look at several of your results indicates a lot of these don't meet our usage requirements as repos can reasonably be expected to have more than one of them, which means the number of files criteria needs to be met. Some of the filenames you're adding only have single digit results.

Please address the test failures.

> 
>     * Sample source(s):
>       * [URL to each sample source, if applicable]
>     * Sample license(s):

This needs to be filled in for every sample.

PR #7175, opened 2024-12-22
add Jule language

adamperkowski:
## Checklist
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.jule
  - [x] I have included a real-world usage sample for all extensions added in this PR
  - [x] I have included a syntax highlighting grammar: https://github.com/julelang/vscode-jule/tree/master/syntaxes
  - [x] I have added a color
    - Hex value: `#5f7489`
    - Rationale: Jule's logo color

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1338,6 +1338,9 @@
 [submodule "vendor/grammars/vscode-jest"]
 	path = vendor/grammars/vscode-jest
 	url = https://github.com/jest-community/vscode-jest
+[submodule "vendor/grammars/vscode-jule"]
+	path = vendor/grammars/vscode-jule
+	url = https://github.com/julelang/vscode-jule.git
 [submodule "vendor/grammars/vscode-just"]
 	path = vendor/grammars/vscode-just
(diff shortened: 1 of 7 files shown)


lildude (changes requested):
> I have included a real-world usage sample for all extensions added in this PR

Why have you removed the two lines below this rather than filling them in? We need to know where the files come from and the license covering each for legal reasons: eg we can't include proprietary code. Please add back those lines and fill them in.

As for the samples:

https://github.com/github-linguist/linguist/blob/2409807814a3ff386294b1f217b886a1594642cd/CONTRIBUTING.md?plain=1#L130

Usage will be reviewed again when the next release is made.

koibtw:
> Why have you removed the two lines below this rather than filling them in? We need to know where the files come from and the license covering each for legal reasons: eg we can't include proprietary code. Please add back those lines and fill them in.

not sure because it was almost a year ago, but probably because as of opening this PR the samples didn't have sources nor licenses

Issue #7178, opened 2024-12-23
New language: LPC


jlchmura:
## Language name
LPC

## URL of example repository
https://github.com/jlchmura/lpc-language-server

## URL of syntax highlighting grammar
https://github.com/jlchmura/lpc-language-server/tree/main/syntaxes

## Most popular extensions
.c
.h

## Detected language

lpc

jlchmura:
I am happy to contribute a PR, but the file extensions will conflict with `c`.  Is there a way to distinguish between the two languages?

Mehran33SP:
Deleted my account

PR #7179, opened 2024-12-24
Add language: LPC

jlchmura:
## Description
This PR adds syntax highlighting for the LPC language, a C-like language used in MUD drivers such as [LDMud](https://www.ldmud.eu/) and [FluffOS](https://www.fluffos.info/).

## Checklist:
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.c+%28%22this_player%28%29%22+OR+%22this_object%28%29-%3E%22%29
    - Sample source(s):
      - https://github.com/jlchmura/lpc-language-server/blob/main/server/src/tests/cases/compiler/
    - Sample license(s):
      - MIT 
  - [x] I have included a syntax highlighting grammar: https://github.com/jlchmura/lpc-language-server/tree/main/syntaxes
  - [x] I have added a color
    - Hex value: `#ffb100`
    - Rationale: It is reminiscent of the amber screens that were in use when many MUDs got their start

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -956,7 +956,7 @@ named_patterns:
   - '^[ \t]*#[ \t]*(?:precache|using_animtree)[ \t]*\('
   json: '\A\s*[{\[]'
   key_equals_value: '^[^#!;][^=]*='
-  lpc: '(mixed|nosave|clone_object)|(^inherit [\"A-Za-z][A-Za-z0-9_]*)|(((public|private|protected)\s+)?((varargs|nomask|noshadow)+\s+)?((string|int|mixed|object|mapping)\s+)?[A-Za-z][A-Za-z0-9_]*\()'
+  lpc: '(mixed|nosave|clone_object)|(this_player\(\))|(^inherit [\"A-Za-z][A-Za-z0-9_]*)|(((public|private|protected)\s+)?((varargs|nomask|noshadow)+\s+)?((string|int|mixed|object|mapping)\s+)?[A-Za-z][A-Za-z0-9_]*\()'
   m68k:
   - '(?im)\bmoveq(?:\.l)?\s+#(?:\$-?[0-9a-f]{1,3}|%[0-1]{1,8}|-?[0-9]{1,3}),\s*d[0-7]\b'
   - '(?im)^\s*move(?:\.[bwl])?\s+(?:sr|usp),\s*[^\s]+'
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -146,6 +146,12 @@ disambiguations:
   rules:
   - language: XML
     pattern: '^(\s*)(?i:<Project|<Import|<Property|<?xml|xmlns)'
+- extensions: ['.c']
+  rules:
+    - language: LPC
+      named_pattern: lpc
+      negative_pattern: '(?i)([ \t]*(unsigned|uint|enum)|^\s*\*[A-Za-z]|([A-Za-z][A-Za-z0-9_]*\[\]))'
+    - language: C
 - extensions: ['.cairo']
   rules:
     - language: Cairo Zero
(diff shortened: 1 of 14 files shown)


jlchmura:
Note to the maintainers: I was not able to write a heuristic that could handle `samples/c/nightmare.h` since C and LPC are so similar in that regard.  Do you recommend removing LPC from the `.h` extension, and let the user override via gitattributes `linguist-language`?

jlchmura:
This PR will close #7178

jlchmura:
@Alhadis @lildude Can you add this PR to your to-do list to review?

lildude (changes requested):
Please review your regexes to ensure they run linearly and do not suffer from ReDoS... your `lpc` pattern definitely suffers from ReDoS.

Please also update your sample to only include real-world examples. Several of your samples are contrived just to validate the syntax highlighter. Additionally, and sample that is suppressed in the diff indicates it's too big for our needs so will need removing or reducing.

jlchmura:
ReDos has been addressed.  Also replaced samples as requested.

@lildude I still need advice on `samples/c/nightmare.h`.  Remove LPC from the `.h` extension?

lib/linguist/heuristics.yml
```diff
@@ -174,6 +174,12 @@ disambiguations:
   rules:
   - language: XML
     pattern: '^(\s*)(?i:<Project|<Import|<Property|<?xml|xmlns)'
+- extensions: ['.c']
+  rules:
+    - language: LPC
+      named_pattern: lpc
+      negative_pattern: '(?i:(?:[ \t]*(?:unsigned|uint|enum)|^\s*\*[A-Za-z]|[A-Za-z][A-Za-z0-9_]*\[\]))'
```
lildude: This is vulnerable to ReDos. Making it case-insensitive also defeats the point of then trying to match case-sensitive characters.

lib/linguist/heuristics.yml
```diff
@@ -379,6 +385,9 @@ disambiguations:
     named_pattern: objectivec
   - language: C++
     named_pattern: cpp
+  - language: LPC
+    named_pattern: lpc
+    negative_pattern: '(?i:(?:^[ \t]*(?:unsigned|uint|enum|extern|pragma[ \t]+mark|__inline__)\b|[ \t]*\*\(|^[ \t]*\*[A-Za-z]|[A-Za-z][A-Za-z0-9_]*\[\]))'
```
lildude: This is also vulnerable to ReDos.

lildude (changes requested):
Your regexes are still vulnerable to ReDoS.

> @lildude I still need advice on `samples/c/nightmare.h`. Remove LPC from the `.h` extension?

Whooops. Sorry I missed that. If you can't write a heuristic that only matches LPC, then it's best to leave it off. We fought for years trying to find the compromise we currently have, and even have it [documented in our troubleshooting guide](https://github.com/github-linguist/linguist/blob/main/docs/troubleshooting.md#my-ccobjective-c-h-header-file-is-detected-as-the-wrong-language). 

The same applies for your `.c` heuristic - it need to 100% only match LPC as I don't want to get into a similar battle with C.

The regexes also need to be safe from ReDos and RE2 compatible.

lib/linguist/heuristics.yml
jlchmura: This is no longer needed after revising the `lpc` named pattern.

lib/linguist/heuristics.yml
jlchmura: This is no longer needed after revising the `lpc` named pattern.  I removed it in the latest commit.

jlchmura:
Thanks @lildude.
I reworked the LPC named pattern to make it even more LPC-specific, and removed the ReDoS-vulnerable negative patterns in the heuristic, which I don't believe are needed anymore.  

Let me know...

lildude (changes requested):
We need tests for your heuristics and a search query for the `.h` extension in the PR template too.

PR #7193, opened 2025-01-07
Register `bun.lock` as JSON 

RiskyMH:
## Description

[Bun](https://github.com/oven-sh/bun) recently added a text lockfile called `bun.lock` (https://bun.sh/blog/bun-lock-text-lockfile) and it will soon become default. 

## Checklist:

      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*bun.lock+lockfileVersion+NOT+org%3Aoven-sh
    - Sample source(s):
      - https://github.com/oven-sh/bun/blob/main/packages/bun-vscode/example/bun.lock
    - Sample license(s): **MIT**

---

Although `bun.lock` officially uses JSONC, this PR sets it to standard JSON. Bun only relies on the trailing comma feature, which JSON happens to support without errors. The main reason to use standard JSON is that it provides better syntax highlighting on GitHub (https://github.com/github-linguist/linguist/discussions/6197)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7516,6 +7516,7 @@ TypeScript:
   aliases:
   - ts
   interpreters:
+  - bun
   - deno
   - ts-node
   - tsx
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3307,8 +3307,8 @@ JSON:
   - ".watchmanconfig"
   - MODULE.bazel.lock
   - Pipfile.lock
-  - composer.lock
   - bun.lock
+  - composer.lock
   - deno.lock
   - flake.lock
   - mcmod.info
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -3308,6 +3308,7 @@ JSON:
   - MODULE.bazel.lock
   - Pipfile.lock
   - composer.lock
+  - bun.lock
   - deno.lock
   - flake.lock
   - mcmod.info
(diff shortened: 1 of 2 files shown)


RiskyMH:
Hi @lildude, thanks for reviewing this PR 🙏.  

GitHub search seems to be inconsistent with the result count—it’s fluctuated a lot recently. From what I can tell, there are at least **220 repositories** using `bun.lock`, which meets the minimum requirements. Each repo typically has only one lockfile, even in monorepo setups with `bun.lock` at the root.  

The fluctuations seem to be a search or caching issue rather than a decline in usage. Adoption is expected to grow significantly soon, as `bun.lock` will become the default for new projects in Bun 1.2. However we would love it if JSON highlighting could be default by then!

The search is currently showing the correct number again, so I’d appreciate it if you could take another look:  
**👉 https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*bun.lock+lockfileVersion**

Thank you! 🙌

Jarred-Sumner:
@lildude currently the search shows 1.7k files matching. Does that meet the popularity criteria?

image

DecimalTurn:
> The main reason to use standard JSON is that it provides better syntax highlighting on GitHub (https://github.com/github-linguist/linguist/discussions/6197)

JSONC's syntax highlighting should be much better once https://github.com/github-linguist/linguist/pull/7255 is live on GitHub.

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7198, opened 2025-01-10
Update references to go-enry in documentation

lildude:
## Description

GitHub's search no longer users go-enry and instead uses an internally developed library for language detection. This still feeds off Linguist so the same delays and limitations apply.

This PR updates the docs to reflect we no longer use go-enry.

## Checklist:

N/A

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -22,8 +22,8 @@ If you have not made any changes to your repository in a while, you may find pus
 
 There are a few reasons that could lead to this outcome:
 
-1. If the repository implements an override via the `linguist-language` attribute, it won't be taken into account in GitHub's search results since GitHub Search relies on [go-enry](https://github.com/go-enry/go-enry) which doesn't support overrides at the moment ([More info](https://github.com/src-d/enry/issues/18)).
-2. go-enry might not be using the latest version of Linguist which means that files could be detected differently in search compared to Linguist (see also the note at the end of [this section](#my-linguist-pr-has-been-merged-but-github-doesnt-reflect-my-changes)).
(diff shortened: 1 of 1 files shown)


DecimalTurn:
Can we know what language is used for the internal library? It would help make sure that the regex syntax we use in Linguist is compatible.

lildude:
It written in Rust.

DecimalTurn:
Is it the project mentioned [here](https://github.com/github-linguist/linguist/pull/6417)?

lildude:
> Is it the project mentioned [here](https://github.com/github-linguist/linguist/pull/6417)?

Yup.

> I'm asking because I would like to confirm what engine is used for regex patterns. For instance, if they use [this implementation](https://github.com/rust-lang/regex) of regex, they would have no support for possessive qualifiers (and most non-Re2 regex patterns) as discussed [here](https://github.com/rust-lang/regex/issues/765). Which means we should probably try to avoid or remove them from the heuristics in Linguist.

That's the implementation that is used and you raise a good point. Thanks and thanks for https://github.com/github-linguist/linguist/pull/7238.

If you've got an urge to fix more regexes, we have several regexes that need fixing as they don't run linearly or are vulnerable to ReDoS. I've started https://github.com/github-linguist/linguist/compare/main...lildude/linear-regex-redos to add a test and clean them up as and when I have the time, so feel free to continue cleaning up our regexes.

DecimalTurn:
@lildude Thanks for confirming and yes, I do have the intention to work on some more regex adjustments. I actually have some more changes that I was going to submit PRs for, so I'll get on with it.

PR #7229, opened 2025-02-01
Refactor assembly language detection

donno2048:
Pretty much refactor the entire assembly language detection

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.s+%22org%20100h%22
    - Sample source(s):
      -  https://github.com/xorvoid/sectorc/blob/main/sectorc.s
    - Sample license(s): CC0


lib/linguist/heuristics.yml
```diff
@@ -699,6 +699,10 @@ disambiguations:
     pattern: '^\s*<\?xml'
 - extensions: ['.s']
   rules:
+  - language: Assembly
+    and:
+      - negative_pattern: 'mov\s+%'
+      - pattern: 'mov\s+[^%]{2,},'
```
lildude, line 705: This regex is vulnerable to ReDoS attack. Please find a better regex which isn't that also runs linearly.

lildude (changes requested):
See inline comment.

lib/linguist/heuristics.yml
donno2048: fixed

donno2048:
@lildude fixed

donno2048:
I'll delete `Assembly/audio.i` and `Assembly/macros.inc` as they both are **not** nasm assembly so will create false-negative

donno2048:
I've never ever seen anyone uses `.asm` for unix assembly so going to remove it

DecimalTurn:
> I'll delete `Assembly/audio.i` and `Assembly/macros.inc` as they both are **not** nasm assembly so will create false-negative 

If they're not nasm, what are they? Could the samples be moved to the folder of another Assembly dialect?

lildude:
What @DecimalTurn said. We don't remove samples unless they are blatantly complete rubbish which doesn't appear to be the case here.

The failing test is going to be because the Unix Assembly `.s` samples are not unique enough for the classifier to successfully differentiate the two different languages. Add one or more real world Unix Assembly samples that contain syntax and tokens that are unique to this language to improve the classifier training. This may take a few attempts so I recommend you test this locally rather than pushing and waiting each time.

donno2048:
> > I'll delete `Assembly/audio.i` and `Assembly/macros.inc` as they both are **not** nasm assembly so will create false-negative 
> 
> If they're not nasm, what are they? Could the samples be moved to the folder of another Assembly dialect?

`audio.i` looks like asm but line 62 has the instruction `deek` and 107 has `doke` which are BASIC commands, maybe there was a way to embed BASIC in asm I don't know about, I'm not old enough to know anyways...

donno2048:
> What @DecimalTurn said. We don't remove samples unless they are blatantly complete rubbish which doesn't appear to be the case here.

Those files aren't NASM nor UNIX assembly so for the specific case of this repository they **are** complete rubbish

> The failing test is going to be because the Unix Assembly `.s` samples are not unique enough for the classifier to successfully differentiate the two different languages. 
> ```
> Unix Assembly/gemm_kernel_1x4.S BAD (Assembly)
> ```

The sample has a `movl %` so should be picked up, figured it's probably a caching issue or something, that's why I didn't replace it...

Also just to make sure I tested it locally and it passed.


> Add one or more **real world** Unix Assembly samples (we don't want contrived samples) that contain syntax and tokens that are unique to this language to improve the classifier training. This may take a few attempts so I recommend you test this locally ([run these commands](https://github.com/github-linguist/linguist/blob/1647b5be1658fb15a5e09bf89067f79b392301e2/.github/workflows/ci.yml#L62-L65)) rather than pushing and waiting each time.

    
No problem

lildude:
> `audio.i` looks like asm but line 62 has the instruction `deek` and 107 has `doke` which are BASIC commands, maybe there was a way to embed BASIC in asm I don't know about, I'm not old enough to know anyways...

Is it possibly the whole file is actually BASIC or one of the variants?

> `macros.inc` isn't unix as it uses `.endmacros` and not `.endm`.

A quick search suggests it _is_ some form of Assembly as detailed [here](https://onlinedocs.microchip.com/oxy/GUID-E06F3258-483F-4A7B-B1F8-69933E029363-en-US-2/GUID-FD9F4042-970A-41D3-ABAC-7639908B8B01.html). \

I have no clue about Assembly but Linguist recognises many forms... see all the entries in the `language.yml` file with Assembly in the name or grouping... and doesn't attempt to be 100% perfect as people make mistakes and languages evolve. The Assembly language has evolved over years from a catch-all to be more refined as more variants are broken out into their own languages.

If in doubt, leave it be. These samples have been included in Linguist for over 10 years and are only used to train the classifier so provided your heurstics are sound and there are sufficient other samples for the classifier, the chances of things slipping through due to these two tokens are low and tolerable.

> Those files aren't NASM nor UNIX assembly so for the specific case of this repository they **are** complete rubbish

By "complete rubbish" I mean something that is clearly not code. The content is legit code so the correct language should be identified rather than taking the easy way out and removing it.

donno2048:
> What @DecimalTurn said. We don't remove samples unless they are blatantly complete rubbish which doesn't appear to be the case here.

Those files aren't NASM nor UNIX assembly so for the specific case of this repository they **are** complete rubbish

> The failing test is going to be because the Unix Assembly `.s` samples are not unique enough for the classifier to successfully differentiate the two different languages. 
> ```
> Unix Assembly/gemm_kernel_1x4.S BAD (Assembly)
> ```

The sample has a `movl %` so should be picked up, figured it's probably a caching issue or something, that's why I didn't replace it...

Also just to make sure I tested it locally and it passed.


> Add one or more **real world** Unix Assembly samples (we don't want contrived samples) that contain syntax and tokens that are unique to this language to improve the classifier training. This may take a few attempts so I recommend you test this locally ([run these commands](https://github.com/github-linguist/linguist/blob/1647b5be1658fb15a5e09bf89067f79b392301e2/.github/workflows/ci.yml#L62-L65)) rather than pushing and waiting each time.

    
No problem

DecimalTurn:
Regarding `audio.i`, it was introduced in [this PR](https://github.com/github-linguist/linguist/pull/4637) and was originally taken from [this location](https://github.com/kervinck/gigatron-rom/blob//b0471f594ac33017c3c5af023bfc1bcca1a3c9f8/Contrib/at67/gasm/tetronis/audio.i). Looking at the repo where it's located seems to indicate that the version of Assembly used is for the Neo6502 Processor instead of NASM (x86). If so, it is still Assembly with some additional BASIC instructions that are supported.

Neo6502's Assembly might be too niche to have it's own language entry, but grouping it with the general Assembly language might cause issues, so I'm not sure what is best here.

donno2048:
> > `audio.i` looks like asm but line 62 has the instruction `deek` and 107 has `doke` which are BASIC commands, maybe there was a way to embed BASIC in asm I don't know about, I'm not old enough to know anyways...
> 
> Is it possibly the whole file is actually BASIC or one of the variants?

The whole file is written like asm so I doubt it's BASIC, maybe some very old assembly dialects have support for these BASIC commands but I don't think `linguist` should support it...

> > `macros.inc` isn't unix as it uses `.endmacros` and not `.endm`.
> 
> A quick search suggests it _is_ some form of Assembly as detailed [here](https://onlinedocs.microchip.com/oxy/GUID-E06F3258-483F-4A7B-B1F8-69933E029363-en-US-2/GUID-FD9F4042-970A-41D3-ABAC-7639908B8B01.html). \

The problem with assembly is people call it just "Assembly" so people miss the fact that assembly isn't a language but a category of languages, hence the difference between, say, x86 NASM assembly and UNIX gas assembly.

Think about it like "markup", markup isn't a language it's a category of languages (e.g. HTML, markdown, etc.).

Another point to think of is that even when we say "NASM assembly" it's not a single "language" as NASM is assembling the syntax of many different possible ASM languages, among them the one relevant to this discussion `x86` this drill-down from assembly category to NASM assembler to x86 language gives us the the "x86 NASM assembly" syntax.

The reason I'm writing all this is to say, that there are thousands of assembly languages and dozens upon dozens of assemblers, we don't need to (and probably can't) implement syntax for every assembler that exists.

The page linked is documentation about the AVR assembler which assembles code for the AVR which is a subclass of RISC (which by the way GAS supports) which isn't even that popular.

NASM and GAS (and maybe also MASM and TASM but they both also use intel syntax) are the most popular assemblers so considering their syntax makes sense.

> I have no clue about Assembly but Linguist recognises many forms... see all the entries in the `language.yml` file with Assembly in the name or grouping... and doesn't attempt to be 100% perfect as people make mistakes and languages evolve. The Assembly language has evolved over years from a catch-all to be more refined as more variants are broken out into their own languages.

It's not about making mistakes, it just doesn't make sense trying to support all assembly dialects exactly because there are so many variants. 

> If in doubt, leave it be. These samples have been included in Linguist for over 10 years and are only used to train the classifier so provided your heurstics are sound and there are sufficient other samples for the classifier, the chances of things slipping through due to these two tokens are low and tolerable.
> 
> > Those files aren't NASM nor UNIX assembly so for the specific case of this repository they **are** complete rubbish
> 
> By "complete rubbish" I mean something that is clearly not code. The content is legit code so the correct language should be identified rather than taking the easy way out and removing it.
> 
> 


But what would be considered correct? This isn't intel nor unix syntax...

donno2048:
> Regarding `audio.i`, it was introduced in [this PR](https://github.com/github-linguist/linguist/pull/4637) and was originally taken from [this location](https://github.com/kervinck/gigatron-rom/blob//b0471f594ac33017c3c5af023bfc1bcca1a3c9f8/Contrib/at67/gasm/tetronis/audio.i). Looking at the repo where it's located seems to indicate that the version of Assembly used is for the Neo6502 Processor instead of NASM (x86). If so, it is still Assembly with some additional BASIC commands that are supported.
> 
> Neo6502's Assembly might be too niche to have it's own language entry, but grouping it with the "general" Assembly language might cause issues, so I'm not sure what is best here.

From the GitHub page of Neo6502:

> Neo6502 is small board with dimensions 80x55mm.
> it consist of real W65C02S processor and RP2040 which emulates RAM memory, video, keyboard, SPI, I2C, UART, GPIOs, Sound


In essence, what this means is this person made a board not in common use, made an assembler custom for it that essentially tailored around his board-specific implementation.


I think not only we shouldn't use it to check for assembly syntax, this plainly isn't even assembly, he's implementing memory operations that are architecture-dependent into his assembler, calling it an assembler doesn't change the fact it's a high-level programming language...

donno2048:
From the Wikipedia article of `PEEK and POKE`:

> In computing, PEEK and POKE are commands used in some high-level programming languages for accessing the contents of a specific memory cell referenced by its memory address.

DecimalTurn:
Regarding the Classifier cross-validation error, it seems like there is still an issue with the `Unix Assembly/gemm_kernel_1x4.S` sample that is identified as `Assembly` instead of `Unix Assembly`. I'd suggest adding 1 or 2 samples with the `.S` extension to `Unix Assembly` that are similar to help the classifier group them together.

Also note that you can use a filter to test the classifer for specific extensions to save some time:
ie.:
```
bundle exec script/cross-validation --extensions=".s, .S"
```

donno2048:
@DecimalTurn the problem is that `gemm_kernel_1X4.S` uses `gcc` syntax while the heuristics probably refer to `as` synatx

DecimalTurn:
The classifier doesn't rely on the heuristics however, it uses a statistical approach based on file similarities. It is a fallback identification strategy in case the heuristics aren't enough.

donno2048:
But I mean, the question is if the syntax definition used is actually relevant for the case of `gcc` macros syntax

DecimalTurn:
Even if it's not perfect, having some syntax highlighting is better than none at all.

donno2048:
ig

donno2048:
@lildude

lildude (changes requested):
You still need to restore the two deleted files or identify and correctly categorise them. The `audio.i` file was added in https://github.com/github-linguist/linguist/pull/4637 and comes from https://github.com/kervinck/gigatron-rom/blob//b0471f594ac33017c3c5af023bfc1bcca1a3c9f8/Contrib/at67/gasm/tetronis/audio.i

donno2048:
@lildude regarding `audio.i` I already explained it's irrelevant as not only it's not one of the recognized assembly languages it's not an assembly language at all...

It's a custom language as explained in my and @DecimalTurn 's comments.

Also, this custom language only appears in 49 files (including the copy in this repository): https://github.com/search?q=path%3A**%2F*.i+%22doke%22+NOT+is%3Afork&type=code

So overall, it doesn't fit the criteria to even be included here.

donno2048:
As for `macros.inc`, would you rather I add the AVR assembly language too, then move `macros.inc` into `samples/AVR Assembly`?

lildude:
Ah, I missed that… too many words and not enough time to read them all 🤣

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

lildude:
@donno2048 I completely missed this before, this PR has introduced warnings about the regexes you've changed as can be [seen here](https://github.com/github-linguist/linguist/actions/runs/13481437201/job/37667140113#step:5:76) in the CI output:

```
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /^\s+(i?)db\s+[\d\w]/
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /(?i-mx:mov\s+[^\s]+,)|(?-mix:^\s+(i?)db\s+[\d\w])/
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /^\s+(?i)db\s+[\d\w]/
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:129: warning: character class has duplicated range
```

Please can you address these in a new PR.

lildude:
> From a quick look, I think this is going to be due to `[\d\w]`.

Yup. I think all of these cases can be simplified to `\w`

I think this should do the trick:

```diff
@@ -368,7 +368,7 @@ disambiguations:
   - language: Assembly
     pattern:
     - '(?i)mov\s+[^\s]+,'
-    - '^\s+(i?)db\s+[\d\w]'
+    - '^\s+(i?)db\s+\w+'
 - extensions: ['.ice']
   rules:
   - language: JSON
@@ -403,7 +403,7 @@ disambiguations:
   - language: Assembly
     pattern:
     - '^(?i)[^"m]*mov\s+[^\s]+,'
-    - '^\s+(?i)db\s+[\d\w]'
+    - '^\s+(?i)db\s+\w'
 - extensions: ['.json']
   rules:
   - language: OASv2-json
```

PR #7235, opened 2025-02-14
Hide .snap snapshot files by insta.rs framework

nyurik:
## Description
`*.snap` are snapshot files stored by the popular Rust [Insta](https://insta.rs/docs/snapshot-files/)  framework, and are auto-generated when the test results are accepted. Subsequent test runs just match results against these files.

By default .snap files are stored in the `.../snapshots/*.snap`, but user may modify the location. For now, I marked all `.snap` files as internal, but this may need to be revisited if other common extension is also .snap

Examples of [/snapshots/*.snap](https://github.com/search?q=path%3A%2F%5C%2Fsnapshots%5C%2F%5B%5E%5C%2F%5D%2B%5C.snap%24%2F&type=code) - 103k instances.

Note that it appears `.snap` is often used by other repositories for similar purpose, so I think it might be OK to hide it for them as well.  [Code search without /snapshots/ prefix](https://github.com/search?q=path%3A%2F%5C.snap%24%2F&type=code)

## Checklist:
  
### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -74,6 +74,7 @@ module Linguist
       npm_shrinkwrap_or_package_lock? ||
       pnpm_lock? ||
       terraform_lock? ||
+      insta_snap? ||
       generated_yarn_plugnplay? ||
       godeps? ||
       generated_by_zephir? ||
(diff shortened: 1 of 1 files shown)


lildude (changes requested):
This is too blunt a weapon. Linguist only associates this extension with the Jest Snapshot language. This change will mark all of files as generated too which is not correct else they would have been marked as generated at the time language support was added.

nyurik:
> This is too blunt a weapon. Linguist only associates this extension with the Jest Snapshot language. This change will mark all `.snap` files as generated too which is not correct else they would have been marked as generated at the time language support was added.

@lildude I was concerned about it too tbh... can you suggest a way to limit these files somehow, e.g. maybe by its content?

lildude:
> @lildude I was concerned about it too tbh... can you suggest a way to limit these files somehow, e.g. maybe by its content? Or only if this is a "rust language" repo?

There are several things you can do in combination:

- check the path and extension (you can start with the most common/default path... users can always override if they change it)
- check the content for something unique to these files (according to the page you linked to, you've got `---`, `expression` and `source` occurring in the first 3-5 lines of every file you can match on).

You can see examples of all of these in the the [`generated.rb`](https://github.com/github-linguist/linguist/blob/main/lib/linguist/generated.rb) file and you can combine them. Something like done for .Net Doc files:

https://github.com/github-linguist/linguist/blob/6f9715869791ffc235df50171dcb097af8f292c4/lib/linguist/generated.rb#L252-L261

PR #7238, opened 2025-02-16
Remove remaining backreferences

DecimalTurn:
## Description

This PR removes remaining backreferences that were not already removed in https://github.com/github-linguist/linguist/pull/6897. It turns out the list was incomplete in the Go-Enry README and even if Go-Enry is no longer used for search on GitHub, rust-lang/regex also [doesn't support backreferences](https://github.com/rust-lang/regex/issues/302), so we should still remove them.

Affected heuristics:
* Bikeshed
* FreeMarker
* StringTemplate

Like in the previous PR, the presence of the backreference in those regex was to make sure that open and close tags were matching which is important for syntax checking, but not language identification. 

## Checklist:
NA


lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7240, opened 2025-02-17
Remove atomic groups

DecimalTurn:
## Description

Atomic groups (aka. Non-backtracking subexpression or possessive match) aren't supported in re2. I've removed them to increase portability accross libraries that use Linguist's heuristics.

Note that atomic groups can be useful to increase performance, but the risk of excessive backtracking in the cases here is minimal or I've added other countermesures when needed.

***Win32 Message File***

There was no real use of atomic grouping there since `\/\*\s*` can't really cause any branching in this context.

***INI***

The original use of an atomic group in `(?>[^\s\[][^\r\n]*(?:\r?\n|\r))*` can prevent some backtracking in case a file contains "InternetShortcut]", but no "URL" property afterwards. However, even without the atomic grouping, the execution time remains linear since the capture group doesn't have multiple ways of matching a whole line. I've also added a limit of 20 lines for the "URL" property to appear since that should be more than enough.

***GSC***

First pattern: The atomic grouping was only preventing minor backtracking due to the presence of 2 options.

Second pattern: That was probably the most useful case of atomic grouping among the one presented here. However, simply combining `(?>\w+\.)*\w+` into `(\w+\.)+` like proposed in this PR will already prevent the vast majority of potential backtracking.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -952,7 +952,7 @@ named_patterns:
   gsc:
   - '^\s*#\s*(?:using|insert|include|define|namespace)[ \t]+\w'
   - '^\s*((?:autoexec|private)\s+){0,2}function\s+((?:autoexec|private)\s+){0,2}\w+\s*\('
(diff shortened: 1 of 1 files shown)


lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7242, opened 2025-02-17
Remove some lookaheads

DecimalTurn:
## Description

Removing lookaheads in those regexes will help with portability across librairies using Re2-style regexes and should help ensure that our regexes run linearly.

Note that to demonstrate that this PR doesn't break ActionScript's heuristics, https://github.com/github-linguist/linguist/pull/7241
 should be merged before changing this PR from a draft to a real PR.


DecimalTurn:
Brief explanation of the logic behind the changes:

**ActionScript:**
The lookahead was there to check if at least one of the 2 keywords (intrinsic or extends) was present. By splitting the expression into 2 expressions seperated by `|`, we avoid the need for this check.

**BrighterScript:**
Lookaheads where pretty much unneeded here.

**FreeMarker:**
The lookahead checked that the next character was either `\s` or `]`, but then it checks that the next character is anything but `]`, so the only thing that could have matched was a `\s` anyways. 

**RUNOFF:**
Because the lookahead is right before a `|`, we can easily replace it by matching the content directly.

**Solidity:**
We can avoid the negative lookahead by using a character class for the next character that excludes numbers.

**Hosts File:**
Because the lookahead is at the end, we can easily replace it by matching the content directly.

DecimalTurn:
@Alhadis, I think I'll let you handle the ` Vim Help File` case, I'm not ready to deal with with nested lookaheads at the moment 😅.

Alhadis:
@DecimalTurn The heuristic used by `Vim Help Text` was lifted from Linguist's `Modeline` strategy, which presents a much more readable and self-documenting version of the regex:

https://github.com/github-linguist/linguist/blob/c959cb53f48552db0e42cded66f3e5a64749b9ec/lib/linguist/strategy/modeline.rb#L52-L126

Any changes to <samp>Linguist::Strategy::Modeline::VIM_MODELINE</samp> will logically need to be reflected in the heuristic, and I strongly advise _against_ making the regex even less readable by replacing nested lookaheads with… whatever works with `re2`.

In other words, you'll probably have to make an exception upstream for this particular heuristic. 🤷‍♂️

lildude (approved):

PR #7265, opened 2025-03-04
Fix Assembly heuristics to remove char class duplication 

donno2048:
https://github.com/github-linguist/linguist/pull/7229 introduced changes to the Assembly heuristics that included character class duplication which were reported as warnings in CI but missed at the time:

```
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /^\s+(i?)db\s+[\d\w]/
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /(?i-mx:mov\s+[^\s]+,)|(?-mix:^\s+(i?)db\s+[\d\w])/
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range: /^\s+(?i)db\s+[\d\w]/
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:91: warning: character class has duplicated range
/home/runner/work/linguist/linguist/lib/linguist/heuristics.rb:129: warning: character class has duplicated range
```

For future reference, @donno2048, useful PR bodies like this are preferred over simple pings as context is needed for reviewing by those unfamiliar at the time or who question the change in future.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -368,7 +368,7 @@ disambiguations:
   - language: Assembly
     pattern:
     - '(?i)mov\s+[^\s]+,'
-    - '^\s+(i?)db\s+[\d\w]'
+    - '^\s+(i?)db\s+[a-z\d]'
 - extensions: ['.ice']
   rules:
   - language: JSON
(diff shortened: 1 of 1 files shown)


lib/linguist/heuristics.yml
```diff
@@ -368,7 +368,7 @@ disambiguations:
   - language: Assembly
     pattern:
     - '(?i)mov\s+[^\s]+,'
-    - '^\s+(i?)db\s+[\d\w]'
+    - '^\s+(i?)db\s+[a-z\d]'
```
lildude, line 371: So just to confirm, this should definitely not match any upper case letters?

donno2048: The `i` modifier handles it

lildude (approved):

PR #7320, opened 2025-04-06
Add Koka

toh995:
## Description
Add support for [koka](https://koka-lang.github.io/koka/doc/index.html).

## Checklist:
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.kk
      
      
    - Sample source(s):
      - For the file `samples/Koka/community-docs.kk`, I consolidated various examples from https://github.com/koka-community/koka-docs/tree/0bf7262 into a single file.
     
      - The other files are copied from the main koka repo, https://github.com/koka-lang/koka.
        
        - `samples/Koka/official-tutorial1.kk` and `samples/Koka/official-tutorial2.kk` - this code was consolidated/extracted from the official "language tutorial" docs https://github.com/koka-lang/koka/blob/8b18c48a32398c7cc98a9e33e7d25832d418e9ec/doc/spec/tour.kk.md
        
    - Sample license(s):
      - In the file `samples/Koka/community-docs.kk`, the code is licensed under MIT (included in a comment at the top of the file). This license was copied from https://github.com/koka-community/koka-docs/blob/6be3e57/LICENSE.
      
      - The other files are licensed under Apache v2.0. See original license: https://github.com/koka-lang/koka/blob/c607f91/LICENSE.
  - [x] I have included a syntax highlighting grammar: https://github.com/koka-community/koka-textmate-grammar/
    -  The decision was made to maintain the textmate grammar file separately from the main repo, for linguist's consumption. See [related discussion](https://github.com/koka-lang/koka/issues/686).
  - [x] I have added a color
    - Hex value: `#215166`
    - Rationale: This color is featured prominently on [koka's official website](https://koka-lang.github.io/koka/doc/index.html) (see the screenshot). Also, see this [color-picking discussion](https://github.com/koka-lang/koka/issues/685). 429766380-24e1f2c5-1fb7-466b-8e3d-a3dfa6a693e1
  - ~[ ] I have updated the heuristics to distinguish my language from others using the same extension.~ N/A, there are no other languages with the same extensions


TimWhiting:
@toh995 I don't think `.core` is a Koka extension. `.kkc` is probably what you are thinking of. 

`.kk` is source
`.kkc` is core (a subset of source), highlights pretty well with the textmate grammar, but might also have some features less supported - it is a generated intermediate file representation, probably not ever checked into git.
`.kki` is interface files, always generated as an interface for distributing binary libraries, and linking to them. Also just a subset of the source syntax, and highlights fine with that.

toh995:
Hi @TimWhiting. Thanks for the advice.

I was going off of this part of the textmate grammar: https://github.com/koka-lang/koka/blob/e55036d/support/vscode/koka.language-koka/syntaxes/koka.json#L3

I had thought that there was a one-to-one mapping between file extensions vs. file types. But looks like that is not actually the case?

Which file extensions would you recommend adding linguist support for, amongst the three? Obviously we'd like `.kk` support, but what about the others?

TimWhiting:
I've made a change in the original textmate grammar in the Koka repo to reflect the actual file type.

I'd say add all of the extensions unless there is any concern from the linguist side. They all should be highlighted using this grammar if they show up in a repository or on github issues. We probably should add examples of those files though.

lildude:
> I'd say add all of the extensions unless there is any concern from the linguist side.

All extensions need to meet our usage requirements. Any that don't will hold back the PR from being merged. Once merged and deployed, users can use an override to set the language for the less poor extensions.

samples/Koka/LICENSE
```diff
@@ -0,0 +1,210 @@
+Copyright 2012-2023, Microsoft Research, Daan Leijen.
```
lildude, line 1: We don't want this file. All samples are used to train the classifier and this is clearly not written in the language you are adding 😉

lildude (changes requested):
Several of your samples are too big… if the diff suppresses them, they're too big for our needs. Please remove or replace the large samples. We don't need many, just a few most representative of real usage of the language.

samples/Koka/LICENSE
toh995: Thanks. How would you suggest including the entire license then?

I want to ensure we comply with copyright law :sweat_smile:

toh995:
Thanks @lildude for weighing RE: file extensions!

Looks like `.kki` and `.kkc` clearly do not meet the usage requirements. I'll remove those file extensions.

samples/Koka/LICENSE
lildude: We don't need it. Were only need the license stated in the PR template.

vendor/licenses/git_submodule/koka-textmate-grammar.dep.yml
```diff
@@ -0,0 +1,139 @@
+---
+name: koka-textmate-grammar
+version: 66402e23e04f4e8b8abdf67093593644c110f0fe
+type: git_submodule
+homepage: https://github.com/koka-community/koka-textmate-grammar.git
+license: other
```
lildude, line 6: The license detection hasn't succeeded to automatically detect the license so we need to set it manually.

```suggestion
license: apache-2.0
```

lildude (changes requested):
> Since no other language uses `.kk` I think we only need one sample.

As suggested, we don't need this many samples. Adding two or three now will ensure we have sufficient samples if a new language is added with a matching extension.

toh995:
Thanks @lildude for the review!

For the code samples, is it okay if it doesn't contain all possible language features? I guess that's why I added so many samples, I wanted to ensure we get reasonable coverage of all language features 😅

TimWhiting:
@toh995 I don't think we need all features, just enough to distinguish from another language using the `.kk` extension. 

From looking at the other languages in the samples folder it seems like there is a wide variety of lengths and numbers of samples, and many have only one or two samples per extension type. For example, Kotlin, Dart, ReScript and Scala have very few samples (1-2). There are others that have a lot (Swift). I feel like the ones that have a lot are going overboard. What other programming language is going to use the `.swift` extension? Also the samples don't necessarily need to be a full module with tests and main functions. Maybe some different examples of omitting `{} and ;` versus using `{} and ;` could be useful. I don't think it has to be comprehensive. If another language decides to use `.kk`, then they will have to add enough samples to distinguish from Koka's syntax, which is unlikely to be hard - since Koka has some unique features and keywords.

toh995:
@lildude I believe this is ready for re-review

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7384, opened 2025-05-17
Add support for MiniScript

ayecue:
**From the MiniScript manual:**

> MiniScript is a modern scripting language designed to be clean, simple, and easy to learn. It
was designed from the ground up by borrowing only the best ideas from other languages
such as Python, Lua, Basic, and C#. If you know pretty much any other programming
language, you’ll pick up MiniScript almost immediately.
> 
> And if you’ve never written a line of code in your life, don’t panic! MiniScript is the
friendliest and most fun way to get started. It’s much easier than you probably expect.
> 
> Important: MiniScript is designed as an embedded programming language. That means
you will usually use it inside some other program, such as a video game. You should find
another document that describes how to access and use MiniScript within that other
program. This document only describes the MiniScript language itself, and the intrinsic
functions that are common to most MiniScript applications.

**More informations about MiniScript can be found here:**
- https://miniscript.org/
- https://github.com/JoeStrout/miniscript
- https://miniscript.org/wiki/Main_Page

**Following changes are performed in this PR:**
- add MiniScript entry to languages.yml
- add MiniScript samples from https://github.com/JoeStrout/miniscript repo
- add MiniScript grammar from https://github.com/ayecue/miniscript-textmate-linguist
- add heuristics related to the file extension `.ms`

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ms+%2F%5E%5Cs*%5Cbend+%28%3F%3Afunction%7Cwhile%7Cif%7Cfor%29%5Cb%7C%5Cs%5Cb%28%3F%3Aisa%29%5Cb%5Cs%2F (2.6k files)
    - Sample source(s):
      - https://github.com/JoeStrout/miniscript/tree/master/MiniScript-cpp/demo
    - Sample license(s):
      - https://github.com/JoeStrout/miniscript/blob/master/LICENSE
  - [x] I have included a syntax highlighting grammar: https://github.com/ayecue/miniscript-textmate-linguist
  - [x] I have added a color
    - Hex value: `#494247`
    - Rationale: Color of the fur of the MiniScript mascot.


lildude (changes requested):
`samples/MiniScript/superstartrek.ms` and `samples/MiniScript/textAdventure.ms` are way too big for our needs - if the diff suppresses the content, it's too big. Please remove them.

Please also update the search query in the PR template to use your heuristics so we can assess popularity and your heuristics.

ayecue:
> `samples/MiniScript/superstartrek.ms` and `samples/MiniScript/textAdventure.ms` are way too big for our needs - if the diff suppresses the content, it's too big. Please remove them.

Done. Both large MiniScript samples have been removed.

> Please also update the search query in the PR template to use your heuristics so we can assess popularity and your heuristics.

I've updated the query to include some of the keywords and special variables. While it might not catch every single MiniScript related file, it should cover the majority.

lib/linguist/heuristics.yml
```diff
@@ -523,6 +523,10 @@ disambiguations:
     and:
       - negative_pattern: '/\*'
       - pattern: '^\s*\.(?:include\s|globa?l\s|[A-Za-z][_A-Za-z0-9]*:)'
+  - language: MiniScript
+    and:
+      - negative_pattern: '^\s*(?:--|/\*|(?:local|global)\s+[a-zA-Z_][a-zA-Z0-9_]*\s*=|fn\s+[a-zA-Z_][a-zA-Z0-9_]*\s+[a-zA-Z_][a-zA-Z0-9_]*)'
+      - pattern: '^\s*(?:end (?:function|while|if|for)|for\s+[a-zA-Z_][a-zA-Z0-9_]*\s+in|else|else if)|\b(?:then|return|locals|globals|outer|self|super|isa|and|or|not)\b|//'
```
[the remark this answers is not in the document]
ayecue: > I've not validated this, but from a quick look, this looks like it might match more than just MiniScript files.

I've modified the positive pattern to be a bit more specific and rather look for comments, specific keywords and variable names at the beginning of the line and not just somewhere in the code. Additionally I've added another negative pattern related to function not being allowed to be at the beginning of the line.

> We also need a test for the heuristics.

I've added MiniScript to the existing `.ms` test case in the heuristics test file.

lib/linguist/heuristics.yml
```diff
@@ -523,6 +523,10 @@ disambiguations:
     and:
       - negative_pattern: '/\*'
       - pattern: '^\s*\.(?:include\s|globa?l\s|[A-Za-z][_A-Za-z0-9]*:)'
+  - language: MiniScript
+    and:
+      - negative_pattern: '^\s*(?:--|\.|/\*|function|(?:local|global)\s+[a-zA-Z_][a-zA-Z0-9_]*\s*=|fn\s+[a-zA-Z_][a-zA-Z0-9_]*\s+[a-zA-Z_])'
+      - pattern: '^\s*\b(?:end (?:function|while|if|for)|else|else if|return|locals|globals|outer|self|super|print|import)\b|^\s*//|\s\b(?:isa|and|or)\b\s'
```
lildude, line 529: These still look too generic as you're matching commonly used syntax keywords. This needs to try and match something that is unique to the language and likely to occur in most files. `if`, `else` etc are very common so are likely to match more incorrectly than correctly as demonstrated by your search results.

ayecue: Makes sense. I initially based the pattern on how other `.ms` languages are handled, since MaxScript and MiniScript share some common syntax. I tried to separate them using negative patterns, but I agree that matching on common keywords like if and else isn't reliable.

I've now trimmed the keyword list significantly. Detection now requires at least one block, which reduces false positives. I kept import since it’s less common and might still help identify MiniScript.

ayecue:
I've reduced the complexity of the heuristics pattern significantly now. Mainly relying on the end block keyword. Additionally I added two more extensions since those should also be relevant to MiniScript.

lib/linguist/languages.yml
```diff
@@ -4486,6 +4486,16 @@ MiniD:
   tm_scope: none
   ace_mode: text
   language_id: 231
+MiniScript:
+  type: programming
+  color: "#4B4A56"
+  extensions:
+  - ".gs"
```
lildude, line 4493: This is no where near meeting our usage requirements and will block this PR even longer than the lack of popularity of the `.ms` extension indicates.

It's probably best to remove this and users can use an override once this PR is merged.

lib/linguist/languages.yml
```diff
@@ -4486,6 +4486,16 @@ MiniD:
...
+  - ".ms"
+  - ".src"
```
lildude, line 4495: This is a very generic extension and as this is the only instance of it, the heuristic will not be used and every `.src` file will be classified as this language.

You could add it to the `generic.yml` file which would then use the heuristic, but then your heuristic needs to be 100% bomb proof with zero scope for error.

If in doubt, I recommend _not_ adding it.

lildude (changes requested):
See inline comments

JoeStrout:
Commenting mainly to get notifications on this issue (I can't find a Subscribe or Watch button).  But just to be sure I understand:

With the `Pending Popularity` tag, this issue is on hold until we cross some threshold?  Is it for [this search](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ms+%2F%5E%5Cs*%5Cbend+(%3F%3Afunction%7Cwhile%7Cif%7Cfor)%5Cb%7C%5Cs%5Cb(%3F%3Aisa)%5Cb%5Cs%2F) to return 2000 files, or something else?

lildude:
> With the `Pending Popularity` tag, this issue is on hold until we cross some threshold? Is it for [this search](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ms+%2F%5E%5Cs*%5Cbend+(%3F%3Afunction%7Cwhile%7Cif%7Cfor)%5Cb%7C%5Cs%5Cb(%3F%3Aisa)%5Cb%5Cs%2F) to return 2000 files, or something else?

Essentially, yes. More deets in https://github.com/github-linguist/linguist/issues/5756

EntitySeaker:
Is it possible for this pr to get accepted before the 2000 files mark, or is that a hard requirement?

lildude:
Hard requirement. If we make an exception for one there's no point having the restriction and we'd have no choice but to accept every submission, even for hobby languages that no one uses.

JoeStrout:
It looks like the guidelines have changed; the global search now quits counting files at ~100, so the PR template now says we need "hundreds of repositories", as evidenced by a search which, in our case, would be something like:

https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ms+%22end%20function%22

And that's showing 269 repositories today.  🥳

So @ayecue, could I trouble you to update your PR with the latest template?  I think we should be good to go now!

lildude:
> And that's showing 269 repositories today. 🥳

Which means nothing in the context of this PR 😁 That's showing 269 repositories which contain "end function" in their name or description and has nothing to do with the file extensions or content of the files in the repo. Poke around in the results and see for yourself.

> It looks like the guidelines have changed; the global search now quits counting files at ~100,

~~It hasn't/doesn't. See [`NOT is:fork path:*.rb puts`](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.rb+puts). The search only returns what is currently indexed. I don't know how the Search index is managed but I know the results are also affected by Search's [documented limitations](https://docs.github.com/en/enterprise-cloud@latest/search-github/github-code-search/about-github-code-search#limitations).~~

Scrap that. That's indeed new to me. I'll need to investigate that more.

lildude:
> Scrap that. That's indeed new to me. I'll need to investigate that more.

Looks like this might be a bug from a recent change. Still investigating.

lildude:
> Looks like this might be a bug from a recent change. Still investigating.

Confirmed. This is a bug with the number shown when using the `path` qualifier. The number at the top of the page shows the correct behaviour if you don't use that qualifier, eg use `language`. The search team are looking into it.

JoeStrout:
Hmm, I see you're right, that's not a useful search result (for our purposes) at all.  I was confused by the PR template asking for "hundreds of repos" and then suggesting a search of that form — kind of implying that such a search would be helpful for answering how many repos you have.

I know the old search found 1500+ files back in November, and I'd expect it to be substantially higher now, so hopefully the search team can track down the bug.  But even if they do, that only gives us a number of files, not repos.  I found another source (which I can't seem to find again now) claiming that the criterion these days was 200 repos, not 2000 files.  But I don't know how we would measure that.

JoeStrout:
Actually, since MiniScript files are currently identified as MAXScript, I guess we can just do:

https://github.com/search?type=code&q=NOT+is%3Afork+language%3AMAXScript++%22end+function%22

which currently shows "2k files".  Perhaps that's good enough for our purposes?

lildude:
> I found another source (which I can't seem to find again now) claiming that the criterion these days was 200 repos, not 2000 files. But I don't know how we would measure that.

That's the old guidance for all files but now only applies to files/extensions that are expected to occur only _once_ per repo. The newer guidance, which has been in place for quite some time in https://github.com/github-linguist/linguist/issues/5756 is now finalised in the CONTRIBUTING.md file [here](https://github.com/github-linguist/linguist/blob/main/CONTRIBUTING.md#language-extension-and-filename-usage-requirements).

> which currently shows "2k files". Perhaps that's good enough for our purposes?

Possibly. Reassessment will take place when I next make a release which happens approximately every 3-4 months.

lildude:
> Confirmed. This is a bug with the number shown when using the path qualifier. The number at the top of the page shows the correct behaviour if you don't use that qualifier, eg use language. The search team are looking into it.

The fix rolled out yesterday.

lildude (approved):

PR #7401, opened 2025-05-22
Add 'gel' alias for 'esdl' (existing grammar update!)

1st1:
**This PR updates the already existing and supported grammar.**

EdgeDB got recently renamed to Gel [1]. The schema language extension changed from `.esdl` to `.gel`. This PR adds `.gel` as a recognized file extension and `gel` as a recognized code-block language selector.

[1] https://www.geldata.com/blog/edgedb-is-now-gel-and-postgres-is-the-future


lildude (changes requested):
Please follow the instructions in CONTRIBUTING.md for adding a new extension. Please also use and correctly complete the PR template… it's there for a reason.

lib/linguist/heuristics.yml
```diff
@@ -317,6 +317,28 @@ disambiguations:
...
+    negative_pattern:
+    # Check for common comments that are invalid in EdgeQL.
+    #
+    # The caveat is that they technically would be legal inside string
+    # literals, but it is extremely unlikely to have such string literals in
+    # the Gel schema (i.e. appearing as defaults or annotations). However, to
+    # reduce the possibility of false positives here we also restrict illegal
+    # comment detection to the beginning of a line.
+    - '^\/\/.+$'
+    - '^\/\*.*?\*\/'
```
lildude, line 341: This is unnecessary as this is the only language associated with the `.gel` extension so we'll never get to the heuristics.

```suggestion
```

lildude (changes requested):
I recommend replacing your one-liner samples with more representative examples as they're used to train the classifier and very small samples are known to cause incorrect results.

vpetrovykh:
I've added a more representative example. It's still synthetic, but it's based much more closely on actual real schema files.

lib/linguist/heuristics.yml
vpetrovykh: We've included the heuristics largely because we've found a lot of `.gel` files that have nothing to do with our database schema (https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.gel has ~1.8K code hits). So we didn't think it would be appropriate to target all `.gel` files indiscriminately.

If you're saying that regardless of that intent, this is not how the heuristics work and would not help there, we can remove them, of course.

lildude: > If you're saying that regardless of that intent, this is not how the heuristics work and would not help there, we can remove them, of course.

Yes. The heuristics won't be used at all as we don't get to the heuristics strategy thanks to the extension being unique. Heuristics can be added in future if another language with the same extension is added.

PR #7431, opened 2025-06-07
Tweaked tree-sitter query classification

mkatychev:
Updated Tree-sitter Query filename detection and heuristics such that all previously classified as Scheme languages are now properly identified as Tree-sitter Query in https://github.com/nvim-treesitter/nvim-treesitter:
image

## Checklist:

    - Sample source(s):
      - https://github.com/nvim-treesitter/nvim-treesitter/tree/master/queries
    - Sample license(s):
      - Apache-2.0 license 

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -754,7 +754,7 @@ disambiguations:
     - '(?:[\)\]]\s*[\*\+\?](?:\s|$))'
     - '(?:^\s*\w+:\s*[\(\[\"])'
     - '\(#(?:set!|(?:not-)?(?:any-of|match)\?)'
-    - '@[\w.-]+(?:\s|$)'
+    - '@[\w.-]+(?:\)\s|$)'
     negative_pattern:
     - '\([+\-:<>\/=~\)]'
   - language: Scheme
@@ -764,7 +764,7 @@ disambiguations:
     negative_pattern:
     - '\(#[\w-]+[!\?]'
     - '(?:[\)\]]\s*[\*\+\?](?:\s|$))'
-    - '@[\w.-]+(?:\s|$)'
+    - '@[\w.-]+(?:\)\s|$)'
 - extensions: ['.sol']
   rules:
   - language: Solidity
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7712,12 +7712,6 @@ Tree-sitter Query:
   - tsq
   extensions:
   - ".scm"
-  filenames:
-  - folds.scm
-  - highlights.scm
-  - indents.scm
-  - injections.scm
-  - locals.scm
   tm_scope: source.scm
   ace_mode: text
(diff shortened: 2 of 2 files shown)


lib/linguist/heuristics.yml
```diff
@@ -737,21 +737,23 @@ disambiguations:
...
   - language: Tree-sitter Query
     pattern:
     - '\(#[\w-]+[!\?]'
     - '[\)\]"_]\s*(?:[\*\+\?]|@\w)'
     - '(?:^\s*\w+:\s*[\(\[\"])'
     - '\(#(?:set!|(?:not-)?(?:any-of|match)\?)'
+    - '@[\w.-]+(?:\s|$)'
     negative_pattern:
     - '\([+\-:<>\/=~\)]'
+  - language: Scheme
```
mkatychev, line 760: Moving TSQ above Scheme somehow increased classification hit-rate 🤷

lib/linguist/languages.yml
```diff
@@ -7704,6 +7704,12 @@ Tree-sitter Query:
   - tsq
   extensions:
   - ".scm"
+  filenames:
+  - folds.scm
+  - highlights.scm
+  - indents.scm
+  - injections.scm
+  - locals.scm
```
lildude: You shouldn't need these as they all share the `.scm` extension which should be sufficient.

```suggestion
  filenames:
  - folds.scm
  - highlights.scm
  - indents.scm
  - injections.scm
  - locals.scm
```

mkatychev: I tested it out, it seems those lines change classification, before deletion for [nvim-tresitter](https://github.com/nvim-treesitter/nvim-treesitter):

```io
$ ./bin/github-linguist $doc/nvim-treesitter | rg '(Tree-sitter|Scheme)'
69.59%  1128396    Tree-sitter Query
```

After deleting those lines:

```io
$ ./bin/github-linguist $doc/nvim-treesitter | rg '(Tree-sitter|Scheme)'
69.52%  1127317    Tree-sitter Query
0.07%   1079       Scheme
```

lildude: That would then indicate the heuristics are missing a scenario specifically in your repo or the samples are not sufficiently teaching the classifier. Keep in mind we’re aiming for good enough, not perfection as every repo is different. It’s also why we have overrides.

The use of specific files is for cases where a specific filename is always used by that language, and generally doesn’t have an extension or has a generic or very common extension associated with another language. Using a filename will mean every instance of that filename will be classified as that language, even if the file is legitimately written in another language. So if a Scala developer creates a file called `locals.scm`, they’ll be forced to use an override because you decided all files with this name should be Tree-sitter query.

mkatychev: Thanks for the explanation, the changed turned out to be a rather small tweak.

mkatychev:
@lildude are there further changes needed?

PR #7475, opened 2025-06-30
Rename JavaScript to WebScript

alanhamlett:
Renames `JavaScript` to `WebScript` because Oracle won't release the trademark on `JavaScript`.
Reference: https://news.ycombinator.com/item?id=44412408


lildude:
Sorry, this is very premature. We're not renaming a language based off a HN discussion around a suggestion made less than 24 hours ago… the outrage is just not worth it. We'll certainly entertain the idea when it is accepted by the wider community and accepted as "the new way to refer to the language".

Issue #7171, opened 2024-12-19
DataFlex - add support for this language


NilsSve:
## Language name
DataFlex

## URL of example repository
https://github.com/DataFlexCode
https://github.com/NilsSve 

## URL of syntax highlighting grammar
https://docs.dataaccess.com/dataflexhelp/mergedProjects/LanguageGuide/Language_Guide.htm

## Most popular extensions
*.src,*.pkg,*.vw,*.dg,*.sl,*.bp,*.wo,*.inc

## Detected language
DataFlex

sonnyfalk:
DataFlex uses a lot of different file extensions, some of which will likely have conflicts. Modern DataFlex uses a common workspace structure, such as `AppSrc/cMyClass.pkg`. Searching for [NOT is:fork path:Appsrc/**.pkg](https://github.com/search?type=code&q=NOT+is%3Afork+path%3AAppsrc%2F**.pkg
) narrows it down to effectively DataFlex source code files and provides 900+ results. Many of the other file extensions are nowhere near that many, so it won't meet the minimum usage requirements.

Given that there are so many file extensions in use, maybe a better initial approach could be to add support for DataFlex without specifying `extensions` or `filenames` to avoid conflicts. And DataFlex source repositories can use `.gitattributes` overrides for the relevant extensions instead. Obviously not ideal, but it would at least provide recognition for the DataFlex language, and that's a great starting point.

I see there is some precedence for languages without specifying file extensions, such as for Elvish transcripts from PR #6302, which is mainly used inside markdown files. @lildude and @NilsSve, I can help and put up a PR for adding the DataFlex language without specifying any file extensions, would that work?

NilsSve:
Hi Sonny,

If you are asking me, yes, that would help! The current repositories at @NilsSve are using .gitattributes exactly as suggested.

lildude:
> I see there is some precedence for languages without specifying file extensions, such as for Elvish transcripts from PR [#6302](https://github.com/github-linguist/linguist/pull/6302), which is mainly used inside markdown files. [@lildude](https://github.com/lildude) and [@NilsSve](https://github.com/NilsSve), I can help and put up a PR for adding the DataFlex language without specifying any file extensions, would that work?

It would "work" but it won't be accepted. The Elvish exception is because there isn't a common extension as it's predominantly used in a markdown files. This is not the case with DataFlex.

[...]

PR #7479, opened 2025-07-03
Add support for DataFlex language

sonnyfalk:
This adds DataFlex language support for `.pkg` file extension with strict heuristics requiring both a Use statement and a class or function/procedure definition. This will resolve issue #7171.

The main DataFlex language website is https://www.dataflex.dev.

## Description
As discussed in issue #7171, the most common DataFlex source file extension is `.pkg`. Since it's a pretty generic extension, this adds it to `generic.yml` with some strict rules in `heuristics.yml` and corresponding tests.

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.pkg+End_Procedure+OR+External_Function
    - Sample source(s):
      - https://github.com/DataFlex-dev/DFUnit/blob/release/DFUnit/AppSrc/DFUnit/Testing/TestRunner.pkg
      - https://github.com/DataFlex-dev/JsonWebToken/blob/production/stable/Library/AppSrc/JWT/UnixTime.pkg
    - Sample license(s): MIT License and BSD 3-Clause License.
  - [x] I have included a syntax highlighting grammar: https://github.com/sonnyfalk/vscode-dataflex
  - [x] I have added a color
    - Hex value: `#ff5f1f`
    - Rationale: This is the DataFlex logo color.
      - This adds new heuristics for `.pkg` file extension.


lib/linguist/heuristics.yml
```diff
@@ -589,6 +589,18 @@ disambiguations:
...
   - language: PHP
     pattern: '<\?[^h]'
+- extensions: ['.pkg']
+  rules:
+  - language: DataFlex
+    and:
+    - pattern: '^\s*Use [a-zA-Z0-9_$]+'
+    - pattern: '^\s*Class [a-zA-Z0-9_$]+ is a|is an [a-zA-Z0-9_$]+'
+    - pattern: '^\s*End_Class'
+  - language: DataFlex
```
lildude: A language should only appear once per extension so this needs to be merged with the heuristic above.

test/test_heuristics.rb
```diff
@@ -802,6 +802,13 @@ def test_php_by_heuristics
     })
   end
 
+  def test_pkg_by_heuristics
+    assert_heuristics({
+      "DataFlex" => Dir.glob("#{fixtures_path}/Generic/pkg/DataFlex/*.pkg"),
```
lildude: This needs to test your samples too, not just the fixtures contrived to match the regex.

```suggestion
      "DataFlex" => all_fixtures("DataFlex") + Dir.glob("#{fixtures_path}/Generic/pkg/DataFlex/*.pkg"),
```

lildude (changes requested):
See inline comments. 

`samples/DataFlex/FileType.pkg` is also too big; if the diff suppresses it, it’s too big for our needs. Please replace it.

lib/linguist/heuristics.yml
sonnyfalk: Thanks! I I merged the two rule sets so it's just one for DataFlex now.

test/test_heuristics.rb
sonnyfalk: Thanks! I updated it as suggested so it's testing the samples too.

sonnyfalk: Thanks for the review. I've made all the changes to address the comments. I replaced the sample file that was too big with another one and updated the PR description to point to the new source file. Let me know if you want me to make any further changes.

lib/linguist/languages.yml
```diff
@@ -1611,6 +1611,14 @@ Daslang:
   tm_scope: source.daslang
   ace_mode: text
   language_id: 648759486
+DataFlex:
+  type: programming
+  color: "#ff5f1f"
```
sonnyfalk, line 1732: @HarmWibier FYI, I got this color from the dataflex.dev website.

HarmWibier: @sonnyfalk : Nice! Seems like the best pick!

sonnyfalk:
I've made all the changes to address the comments. Please let me know if there's something else that I need to do for this to get merged.

PR #7493, opened 2025-07-14
New Verilog samples / update to heuristics to distinguish from Coq

r-hensley:
## Description

This pull request fixes misclassification of certain `.v` Verilog files that were previously identified as Rocq Prover or Coq due to syntax edge cases. These cases include:

- Module stubs where the `module` keyword is followed directly by the port list with no space (e.g., `module foo(...)`)
- Files consisting mainly of `` `pragma protect `` directives (common in encrypted IP cores)
- Compact syntax using `always@` or `initial@` without space before the sensitivity list

The updated `heuristics.yml` pattern broadens the existing Verilog match to allow for:

- Optional spacing before `#` and `(` in module declarations
- `` `pragma `` as a valid Verilog preprocessor directive
- Optional spacing before `@` in `always` and `initial` blocks

Two new `.v` samples are added to demonstrate these cases, and an existing sample (`button_debounce.v`) is updated to include a real-world use of `always@`.

---

## Checklist:

    - Sample source(s):
      - [`module_stub.v`](https://github.com/r-hensley/linguist/blob/main/samples/Verilog/module_stub.v): Vivado-generated module stub
      - [`encrypted_module.v`](https://github.com/r-hensley/linguist/blob/main/samples/Verilog/encrypted_module.v): encrypted IP file
    - Sample license(s):
      - Vivado-generated code: compatible with example/sample use under Xilinx’s tools
    - Updated source(s):
      - [`button_debounce.v`](https://github.com/r-hensley/linguist/blob/main/samples/Verilog/button_debounce.v): Change `always @ (` to `always@(` to provide an example of the possibility of no whitespace between these operators.

---

## Supporting Evidence

- GitHub code search shows both `always @` and `always@` are widely used:
  - [`always @`](https://github.com/search?q=always+%40+language%3AVerilog&type=code): ~618,000 results
  - [`always@`](https://github.com/search?q=always%40+language%3AVerilog&type=code): ~60,900 results
- Example of real-world use of `always@`:  
  [1bitSDR/NCO.v#L36](https://github.com/alberto-grl/1bitSDR/blob/3e2e01da0357b38a9badff832c65a185b2fdcaf0/NCO.v#L36)


samples/Verilog/button_debounce.v
```diff
@@ -61,7 +61,7 @@ module button_debounce
   always @ (posedge clk or negedge reset_n)
     state <= (!reset_n) ? WAIT : next_state;
 
-  always @ (posedge clk or negedge reset_n) begin
+  always@(posedge clk or negedge reset_n) begin
```
lildude: We generally don't change samples. It's better to add a new sample.

lildude:
```suggestion
  always @ (posedge clk or negedge reset_n) begin
```

PR #7495, opened 2025-07-18
Rewrite and reformat documentation of `languages.yml`

Alhadis:
# Description
This pull-request is a rewrite of the comment-header at the start of `languages.yml`, which currently serves as the file's only canonical documentation. Aside from being ordered arbitrarily, it also contains some falsehoods sitting in plain sight:

1.	> type - Either data, programming, markup, prose, <b><i><ins>or nil</ins></i></b>

	As we all know, `nil` isn't a valid language type. This field is mandatory.

2.	> color - CSS hex color to represent the language. <b><i><ins>Only used if type is "programming" or "markup".</ins></i></b>

	This hasn't been true for a while. These colours are used to identify the languages of search results, and are also visible if [`linguist-detectable`](https://github.com/github-linguist/linguist/blob/c1c34e5260797b4d598f5ec76f19723bfc5a1894/docs/overrides.md#detectable) is used to include `data` and `prose` languages in language breakdown graphs.

Other improvements include:

1. Eliminating references to Ruby jargon (e.g., `name.downcase` may be obvious to a Ruby programmer, but less so to a novice coder who's only ever seen `toLowerCase()` used, for example).
2. Ordering fields alphabetically[^1], and grouping them by their requirement status.
3. Referring to field values in more generic terms, rather than the RubyDoc convention of capitalising class-names that aren't proper nouns (e.g., `a String name`, `an Array of`, etc).
4. Clarifying what certain fields mean for those less familiar with the sacred ways of Unix (a Windows user might misinterpret the, uh, `interpreters` field if they've only ever worked in an IDE). This holds especially true for anybody whose English may not be perfect.

Please let me know if anything sticks out as strangely-worded or whatever. I spent so much time juggling words and resmithing paragraphs, I developed tunnel vision. It may need a second pair of eyes that aren't glazing over like mine are right now.

[^1]: Yes, I'm aware that the `type` field is ordered before the rest of the required fields. That's because it's the most important field of the group, so I'm abusing the same arbitrary logic we employ for primary extensions sorted before an alphabetised list. Deal with it.


lib/linguist/languages.yml
```diff
@@ -1,34 +1,35 @@
...
-#                         changed for existing languages without the explicit permission of GitHub staff.
-# color                 - CSS hex color to represent the language. Only used if type is "programming" or "markup".
-# tm_scope              - The TextMate scope that represents this programming
-#                         language. This should match one of the scopes listed in
-#                         the grammars.yml file. Use "none" if there is no grammar
-#                         for this language.
-# group                 - Name of the parent language. Languages in a group are counted
-#                         in the statistics as the parent language.
+# Required fields:
+#    type                  - Either "data", "programming", "markup" or "prose".
```
Alhadis, line 4: I was tempted to elaborate on what kinds of languages and formats each type refers to, but there was no concise, skimmable way of explaining that semantic quagmire. So I omitted it.

lib/linguist/languages.yml
```diff
@@ -1,34 +1,35 @@
...
+#    ace_mode              - Ace mode used when editing files which use the language. This must match
+#                            one of the names in https://gh.io/acemodes. Use "text" if no mode exists.
+#    extensions            - List of associated file extensions, sorted in ascending ASCII order,
+#                            except for the language's primary extension, which is always listed first.
+#    filenames             - List of associated filenames, sorted in ascending ASCII order.
+#                            May be omitted if the extensions field is present (and vice versa).
+#    language_id           - Integer used as a language-name-independent indexed field so that we can rename
+#                            languages in Linguist without reindexing all the code on GitHub. Must not be
+#                            changed for existing languages without the explicit permission of GitHub staff.
```
lildude: Should we also note that this is generated by `script/update-ids`?

Alhadis: Good idea. In fact, we shouldn't even make this field out as something the user needs to specify themselves. I'll amend the description.

Alhadis: @lildude I've amended the field description, aiming for something short-and-sweet rather than too elaborate (since users don't need to know/care about the semantics of the `language_id` field).

Alhadis:
@lildude Don't forget about this. 😉

PR #7513, opened 2025-08-02
Add BIRD2 language

Alice39s:
## Summary
  
This PR introduces syntax highlighting support for the **BIRD2 configuration language** via `.conf` files used by the [BIRD Internet Routing Daemon](https://bird.network.cz/). 

Unlike simpler configuration models (e.g., INI or Nginx), BIRD employs a uniquely intricate paradigm with distinct implementation challenges. We address this by adding a `textmate` grammar with precise scoping for:

- Protocol definitions  
- Filter functions  
- Route maps  
- Other core BIRD2 constructs  

## Motivation

BIRD2 powers critical Internet infrastructure at scale, including:  
- **AMS-IX** – 870+ ASNs, 14Tb/s traffic ([ref](https://www.ams-ix.net/ams/documentation/ams-ix-route-servers))  
- **LINX** – 1,000+ BGP sessions on v2.13 ([ref](https://www.linx.net/wp-content/uploads/2024/05/Day-1-P4-LINX_Technology-Presentation_v3.0.pdf))  
- **Cloudflare** – Anycast routing across all edge PoPs ([ref](https://blog.cloudflare.com/cloudflares-architecture-eliminating-single-p/))  

Despite its critical role in major Internet infrastructure, editors like VSCode and Shiki lack native grammar support for BIRD2 configurations. This PR aims to fill that gap.

## References

- **Grammar Source**: [bird-chinese-community/BIRD-tm-language-grammar](https://github.com/bird-chinese-community/BIRD-tm-language-grammar)  
- **Main Contributors**: [Alice39s](https://github.com/Alice39s), [pppwaw](https://github.com/pppwaw)

## License

- Syntax files are distributed under **[Mozilla Public License 2.0](https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/main/LICENSE.syntax)**
- Sample configuration files are distributed under **[MIT License](https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/main/LICENSE.sample)**

## Checklist:

      - [Public code rearch results](https://github.com/search?q=%22protocol+bgp%22+OR+%22neighbor%22+OR+%22local+as%22+path%3A*.conf+NOT+is%3Afork&type=code&ref=advsearch) -- **27k+** BIRD2 configuration snippets found in public repositories
      - [Public repo rearch results](https://github.com/search?q=bird+config&type=repositories&ref=advsearch) -- **883+** active repositories using BIRD configurations
      - `bird.conf`: 1.3k+ https://github.com/search?q=path%3Abird.conf&type=code
      - `bird6.conf`: 275+ https://github.com/search?q=path%3Abird6.conf&type=code
      - `bird2.conf`: 149+ https://github.com/search?q=path%3Abird2.conf&type=code

    - Sample source:
      - https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/main/sample/
    - Sample license: **[MIT License](https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/main/LICENSE.sample)**
  - [x] I have included a syntax highlighting grammar: https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/main/Syntaxes/bird2.tmLanguage.json
  - [x] I have added a color
    - Hex value: `#b6d7e4`
    - Rationale: This color comes from the [Official Documentation](https://bird.network.cz/) logo and accent colors.


lib/linguist/languages.yml
```diff
@@ -634,6 +634,23 @@ Bikeshed:
   codemirror_mode: htmlmixed
   codemirror_mime_type: text/html
   language_id: 1055528081
+BIRD2:
+  type: data
+  color: "#b6d7e4"
+  extensions:
+  - ".conf"
```
lildude: This is a very generic extension and adding this would mean _every_ `.conf` file on GitHub would be considered BIRD2 as no other language is associated with this extension, because it is so generic. The heuristic you've added won't take effect because this would be the only language using this extension. 

If you are 100% sure the regex will never match anything other than BIRD2 files - it doesn't even match your bogon.conf sample - you can add the extension to the `generic.yml` file which will then cause the heuristic to be used.

If there's a risk of matching a file that isn't BIRD2, it's best to remove the extension and rely on the filenames entries. User can implement an override if they use another filename.

lildude (changes requested):
See inline comment. Please also follow the instructions for adding the grammar in CONTRIBUTING.md.

lib/linguist/languages.yml
Alice39s: Thanks for the thorough review 🙏

You're absolutely right — using a generic .conf extension without cross-language sharing risks generating false positives. 

As suggested, I have now removed .conf from the BIRD2 language definition and am relying solely on specific file names and dedicated extensions (e.g. .bird, .bird2.conf).

Let me know if there's anything else that needs improving and thank you again for your guidance!

lib/linguist/languages.yml
```diff
@@ -634,6 +634,29 @@ Bikeshed:
...
+  type: data
+  color: "#b6d7e4"
+  extensions:
+  - ".bird"
+  - ".bird2"
+  - ".bird3"
+  - ".bird.conf"
+  - ".bird2.conf"
+  - ".bird3.conf"
+  - ".birdconf"
```
lildude: We need search results and samples for every extension being proposed. Any extension that doesn't meet our usage requirements will prevent the entire PR from being merged.

lildude (changes requested):
Don't remove the two missing grammar licences. I'm trying to get the author to restore the repos.

Alice39s:
You're right — thanks for pointing that out.

I ran `script/add-grammar` and didn’t realize it would remove those entries automatically. That wasn’t my intention, and I’d be happy to fix it — though I’m not entirely sure what the best way is to restore them manually.

If you could point me in the right direction, I’d really appreciate it 🙏!

Apologies again for the unintended change.

lib/linguist/languages.yml
Alice39s: Thanks for the review!

You're right — `.bird.conf` is in use today (e.g. [examples here](https://github.com/search?q=.bird.conf&type=code)).

The others (`.bird2.conf`, `.bird3.conf`, `.birdconf`) were added in anticipation of future usage.

I'm happy to drop them if needed and revisit later as adoption grows.

lildude:
The easiest would be to revert your previous commit and then re-apply just your changes.

lib/linguist/languages.yml
Alice39s: Hi, I've removed `.bird2.conf`, `.bird3.conf`, and `.birdconf` since they don't currently meet our usage requirements.

These were originally added for forward compatibility, but I'm happy to revisit them in a future PR once real-world adoption increases.

Alice39s:
Hi @lildude  👋

Gentle ping on this — updates done per your feedback (extensions & licenses) - Could you take a look when you have a moment?

Thanks for your time!

lib/linguist/languages.yml
```diff
@@ -634,6 +634,26 @@ Bikeshed:
...
   codemirror_mime_type: text/html
   language_id: 1055528081
+BIRD2:
+  type: data
+  color: "#b6d7e4"
+  extensions:
+  - ".bird"
+  - ".bird2"
+  - ".bird3"
+  - ".bird.conf"
```
lildude: We need search results and samples for all of these extensions too. If any don't meet our usage requirements, they will prevent this PR from being merged so it's best to keep things to a minimum and users can use an override until popularity is sufficient to add the other extensions.

lib/linguist/languages.yml
```diff
@@ -634,6 +634,26 @@ Bikeshed:
...
+  filenames:
+  - BIRD.conf
+  - BIRD2.conf
+  - BIRD3.conf
+  - bird.conf
+  - bird2.conf
+  - bird3.conf
```
lildude: We need search results to confirm popularity of all of these too. See my comment about extension popularity.

vendor/licenses/git_submodule/BIRD-tm-language-grammar.dep.yml
```diff
@@ -0,0 +1,417 @@
+---
+name: BIRD-tm-language-grammar
+version: da32221ba3b21e4fb6bba98a88ff567b9594a472
+type: git_submodule
+homepage: https://github.com/bird-chinese-community/BIRD-tm-language-grammar.git
+license: other
```
lildude: You can manually set this as the script wasn't able to correctly determine the grammar license itself.

```suggestion
license: mpl-2.0
```

lildude (changes requested):
See inline comments. Additionally, as you're adding support only for filenames, the samples need to match those filenames. None of your samples do so they need to be replaced with real examples of those you're adding support for.

Alice39s (commented):
Resolved https://github.com/github-linguist/linguist/pull/7513#pullrequestreview-3110504561

@lildude

lildude (changes requested):
Samples can not be in subdirectories. Please move them to a level.

Please also update the PR template to provide individual links to the search results for each of the filenames you're adding and fix the information about the samples' sources and licences.

Alice39s:
> Samples can not be in subdirectories. Please move them to a level.
> 
> Please also update the PR template to provide individual links to the search results for each of the filenames you're adding and fix the information about the samples' sources and licences.

@lildude Thanks for the feedback and detailed explanation!

I've made the following changes based on your suggestions:

1. I've moved all the sample configuration files from subdirectories to the top level. They are now all located in the `samples/BIRD2` directory.
2. I've updated the PR template to provide individual search result links for each new filename.
3. I've corrected the information regarding the samples' sources and licenses.

Please take another look, and let me know if there are any further changes required.

lib/linguist/languages.yml
```diff
@@ -634,6 +634,18 @@ Bikeshed:
   codemirror_mode: htmlmixed
   codemirror_mime_type: text/html
   language_id: 1055528081
+BIRD2:
+  type: data
+  color: "#b6d7e4"
+  filenames:
+  - bird.conf
+  - bird6.conf
+  - bird2.conf
```
lildude: These two files are not sufficiently popular for inclusion right now and will prevent this PR from being merged until they're both popular enough. If you want this PR merged, I suggest removing these two. Users can use an override if they want.

Alice39s (commented):
Resolved https://github.com/github-linguist/linguist/pull/7513#discussion_r2281682779

@lildude

Alice39s:
Hi @lildude, could you take a look at this when you have a chance? 🙂

Updates are in place, and I really appreciate your help!

Alice39s:
Hi @lildude 👋

Hope you’re doing well! Just wanted to gently follow up on this PR when you have a moment. I believe all the requested changes have been addressed, and it’s currently awaiting your review.

Please let me know if there’s anything else you’d like me to adjust. Really appreciate your time and guidance on this!

Thanks in advance 🙏

lildude (changes requested):
The test failure is because `samples/BIRD2/bird.conf` is in the wrong place. It needs to be at `samples/BIRD2/filenames/bird.conf`

Alice39s:
> The test failure is because `samples/BIRD2/bird.conf` is in the wrong place. It needs to be at `samples/BIRD2/filenames/bird.conf`

Thanks for catching that! 🙏 I've moved the sample config to `samples/BIRD2/filenames/bird.conf` as suggested.  

Please let me know if there’s anything else you'd like me to adjust.

RedCMD:
a bunch of the regexes have extra erroneous backslashes
https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/c39653d00f6337f9fe4ba9c75f403c29dbec3907/grammars/bird2.tmLanguage.json#L558

image

pppwaw:
> a bunch of the regexes have extra erroneous backslashes
> 
> https://github.com/bird-chinese-community/BIRD-tm-language-grammar/blob/c39653d00f6337f9fe4ba9c75f403c29dbec3907/grammars/bird2.tmLanguage.json#L558
> 
> 
> 
> image
> 
> 

Hi, I I’ve just fixed the issue in the repo.

lildude (approved):

PR #7524, opened 2025-08-12
Replace Vanish and Nginx grammars

lildude:
## Description

The original maintainer of these two grammars has decided to stop using GitHub and has removed access to all their repos so this PR replaces the Varnish and Nginx grammars with others.

Both projects appear to be more actively maintained too so we should get improved syntax highlighting now too.

## Checklist:

  Varnish/VCL: 
  - Old: https://github.com/brandonwamboldt/sublime-varnish
  - New: https://github.com/fastly/vscode-fastly-vcl
  Nginx:
  - Old: https://github.com/brandonwamboldt/sublime-nginx
  - New: https://github.com/william-voyek/vscode-nginx


lildude:
Whoops. Missed nginx was covered by GPL. Will find another later

Alhadis (approved):

PR #7532, opened 2025-08-17
Add XVBA dependencies as vendored

DecimalTurn:
## Description

At the moment, [XVBA](https://marketplace.visualstudio.com/items?itemName=local-smart.excel-live-server) is the most popular VS Code extension for VBA. One feature it offers is the inclusion of VBA "modules" and other dependencies inside the `xvba_modules` directory. The files included there should be considered vendored files in the same way we do for `node_modules`.

Example repo that has this directory: https://github.com/jimbrig/VBASuite
Example of a "module" included in the xvba_modules folder: https://github.com/Aeraphe/excel-types

## Checklist:

  ~~- [ ] I have added or updated the tests for the new or changed functionality.~~ (There's no tests to update for vendored files at the moment it seems)

### DIFF
--- a/lib/linguist/vendor.yml
+++ b/lib/linguist/vendor.yml
@@ -391,3 +391,6 @@
 
 # teamcity CI configuration
 - (^|/)\.teamcity/
+
+# VBAX dependencies
+- (^|/)xvba_modules/


lildude:
> * There's no tests to update for vendored files at the moment it seems)

👇 😉 

https://github.com/github-linguist/linguist/blob/1040da7115035a67fd0b8bfeb12f38b9719b16af/test/test_file_blob.rb#L270

lildude:
This doesn't appear to be particularly popular with search only returning 51 files within this path.

DecimalTurn:
> This doesn't appear to be particularly popular with search only returning 51 files within this path.

I thought so too at first, but there seems to be a bug with the number displayed in Search and the actual number of files that match. It's almost like Search has discovered that all these files are identical, so it applied some kind of single-counting optimization that reduce the number of results appearing in Search. One way to see it clearly is with a search for a specific file like worksheet.d.vb: https://github.com/search?q=path%3Axvba_modules%2F*%2Fworksheet.d.vb&type=code

It says "2 files", but I've counted around 46 results and since there is usually one of these per repo, that would mean there is at least 46 repos that use the xvba_modules.

image

DecimalTurn:
Also, I've been also looking to see if I can find projects that are not indexed and I could easily find 4 with a [quick search](https://github.com/search?q=xvba&type=repositories&p=1) and looking only at the first 2 pages of results. So there are definitely more than 50 repos using xvba_modules.

Screenshot 2025-08-18 133726

image

image


image

lildude:
That's still incredibly low. The lack of indexing is generally a sign of a lack of activity. We switched the search backend a few years back and IIRC only index based on activity. (Yes, this makes checking usage hard and not very accurate but we don't really have any other options).

> It says "2 files", but I've counted around 46 results and since there is usually one of these per repo, that would mean there is at least 46 repos that use the xvba_modules.

I have no idea about this. Very odd. If you really want to know what's up, I'd recommend opening a support ticket.

DecimalTurn:
Ok, I've submitted a support ticket for this.

DecimalTurn:
I've received a response from the support team a few days ago and it was confirmed that there is no double-counting in Search:  if a file has identical copies across GitHub, it will only count for one result it the counter that appears at the top of search results. For the file `worksheet.d.vb` that I linked above, the only reason why it says 2 results is because there is one version with CRLF and another with LF.

The support team acknowledged that there is a display bug in the sense that there should be a window asking to display identical files with the total like this one:

image

However, there is no indication that there is more than 50 or so repos that contain one of these modules at the moment.
@lildude would 200 repos be considered enough?

lildude:
> @lildude would 200 repos be considered enough?

Yeah, that seems reasonable to me.

PR #7533, opened 2025-08-17
Add ALGOL

DecimalTurn:
## Description

Adding ALGOL as a "new" language.
Note that ALGOL 60 and ALGOL 68 don't have the necessary popularity to be added seperatly at the moment, but we can add them later.

Closes https://github.com/github-linguist/linguist/issues/7196

## Checklist:

      -  1.2k files: https://github.com/search?q=path%3A*.alg+AND+%28%2F%5Ecomment%5Cb.*%28%3B%7Ccomment%29%2F+OR+%2F%5Ebegin%5B%5Cs%5CS%5D*%5Eend%2F+OR++%2F%5Cb%28integer%7Creal%7Cboolean%7Cstring%29%5Cb%5Cs%2B%5Cw%2B%5Cs*%3A%3D%5Cs*%2F+OR+%2F%5E%23%5B%5E%23%5Cn%5D%2B%23%24%2F%29+NOT+is%3Afork+NOT+user%3Aacmeism&type=code
    - Sample source(s):
      - https://github.com/ailrk/blgol60/blob/04398762ef256906dc2be69a9ec8b43d9ddd19d2/data/examples/example15.alg
    - Sample license(s): [MIT](https://github.com/ailrk/blgol60/blob/master/LICENSE)
  - [x] I have included a syntax highlighting grammar: https://github.com/PolariTOON/language-algol60?rgh-link-date=2025-01-09T00%3A10%3A49.000Z
      
  - [x] I have added a color
    - Hex value: `#D1E0DB` (very light green)
    - Rationale: Color of the publication shown [here](https://en.wikipedia.org/wiki/ALGOL_68#/media/File:Algol68RevisedReportCover.jpg)

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -75,7 +75,7 @@ disambiguations:
     # ALGOL 60 comments
     - '(?i)^comment\b.*(;|comment)'
     # ALGOL 68 comments
-    - '^#[^#\n]+#$'
+    - '^#[^#\r\n]+#$'
     # ALGOL 60/68 keywords
     - '(?i)\b(integer|real|boolean|string)\b\s+\w+\s*:=\s*'
     - '(?i)^begin[\s\S]*^end'


lib/linguist/heuristics.yml
```diff
@@ -68,6 +68,17 @@ disambiguations:
...
   # Open-ended fallback to Perl AutoLoader
   - language: Perl
+- extensions: ['.alg']
+  rules:
+  - language: ALGOL
+    pattern:
+    # ALGOL 60 comments
+    - '(?i)^comment\b.*(;|comment)'
+    # ALGOL 68 comments
+    - '^#.*?#'
```
lildude: This is a bit loose in it'll match any line that starts with `##` as can be seen by the search returning [this file](https://github.com/apache/lucene/blob/772f63c0a4ba507fc0b26307c0d8cdac22ff3957/lucene/benchmark/conf/spatial.alg#L23) which doesn't look like ALGOL.

DecimalTurn: I've replace the rule `^#.*?#` with `^#[^#\n]+#$` which is much more targeted at the `# TEXT #` type of comments by needing other characters between the 2 `#` and making sure we match the end of the line.

DecimalTurn:
I've updated the search query at the top of the PR with the new regex for ALGOL 68 comments and we are now at 2.1k results.

DecimalTurn:
If we exclude the user acmeism from search, the total falls back to 1.2k files, so it seems that the popularity threshold has not been reached actually. ALGOL will have to wait a little longer, but it has been around for more than 60 years, so what's a few more years of wait, right? 😅

Issue #7550, opened 2025-08-29
Use Freemarker syntax highlighting for *.ftlh and *.ftlx files


tajakobsen:
Freemarker has two alternative file extensions:
 1. *.ftlh _(correct escaping for HTML-output)_
 2. *.ftlx _(correct escaping for XML-output)_

See also the documentation:
https://freemarker.apache.org/docs/dgui_misc_autoescaping.html#:~:text=ftlh,extensions

I would like the **same syntax highlighting** used for *.ftl (Freemarker) files to also be used for *.ftlh and *.ftlx files.

lildude:
Both extensions are popular enough for inclusion so go ahead, open a PR 😉 See CONTRIBUTING.md for details on adding a new extension.

tajakobsen:
Support for `.ftlh` has been merged.

PR #7551, opened 2025-08-29
Add `.ftlh` extension to Freemarker

tajakobsen:
- Add `.ftlh` as an extension for Freemarker
 - Add samples for `.ftlh`

## Description

Hi,

I propose adding the `.ftlh` extension for Freemarker as discussed in #7550 .

The alternative file extension indicate to the Freemarker-library if the output should be HTML-escaped. See https://freemarker.apache.org/docs/dgui_misc_autoescaping.html#:~:text=ftlh,extensions

## Checklist:

      - `.ftlh` (12.9k files) https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ftlh
    - Sample source(s):
      - https://github.com/tajakobsen/linguist/blob/support-for-ftlh-and-ftlx/samples/FreeMarker/blocks-card.ftlh
      - https://github.com/tajakobsen/linguist/blob/support-for-ftlh-and-ftlx/samples/FreeMarker/header.ftlh
    - Sample license(s):
      - https://github.com/ItemConsulting/lib-xp-item-blocks/blob/main/src/main/resources/site/mixins/blocks-card/blocks-card.ftlh (MIT License)
      - https://github.com/ItemConsulting/xp-feature-toggles/blob/main/src/main/resources/admin/views/header/header.ftlh (MIT License)


lildude (changes requested):
> The sample files added in this PR reflects this:
> 
>     * The `.ftlh` files are copies of the `.ftl` files, which will render HTML-documents.

Please don't do this. These are very old and terrible samples from before we started requiring real-world usage examples. Please find real examples and add those. Update the PR template to reflect the license and source of each file.

tajakobsen:
I replaced the example files with two new ones from real-world projects. 

I also wanted to ensure that [both "syntax flavors"](https://freemarker.apache.org/docs/dgui_misc_alternativesyntax.html) are represented in the samples:

 1. `blocks-card.ftlh` uses the angle-bracket-syntax
 2. `header.ftlh` uses the square-bracket-syntax

lib/linguist/languages.yml
```diff
@@ -2253,6 +2253,8 @@ FreeMarker:
   - ftl
   extensions:
   - ".ftl"
+  - ".ftlh"
+  - ".ftlx"
```
lildude: Usage of this extension is a long way from meeting out usage requirements and as such including this extension in this PR will delay it from being merged until it meets our requirements.

I recommend removing it and add it later when usage meets our requirements. Users can use an override in the meantime.

tajakobsen: I agree that removing the `ftlx`-extension makes sense, so I have now removed it and removed the ftlx-sample-file.

Only the functionality around `ftlh` should be left in the PR now.

lildude:
> Sample source(s):
> 
>     * https://github.com/tajakobsen/linguist/blob/support-for-ftlh-and-ftlx/samples/FreeMarker/blocks-card.ftlh
> 
>     * https://github.com/tajakobsen/linguist/blob/support-for-ftlh-and-ftlx/samples/FreeMarker/header.ftlh

Please update these links to point to the actual source, not your fork of Linguist.

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7556, opened 2025-08-30
Add bpftrace language

amiremohamadi:
## Description
Add bpftrace language.

## Checklist:
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.bt+bpftrace
    - Sample source(s):
      - https://github.com/bpftrace/bpftrace/blob/master/tools/syscount.bt
    - Sample license(s): GPL v2 ([commit](https://github.com/bpftrace/bpftrace/commit/1d2e9bb59c1d45795dca4b2fb82ea04c2f75e0f7))
  - [x] I have included a syntax highlighting grammar: https://github.com/amiremohamadi/btls/
  - [x] I have added a color
    - Hex value: `#ebbe1d`
    - Rationale: based on [bpftrace's website theme](https://github.com/bpftrace/website/blob/master/src/css/custom.css)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -732,6 +732,16 @@ Boogie:
   tm_scope: source.boogie
   ace_mode: text
   language_id: 955017407
+BPFTrace:
+  type: programming
+  color: "#ebbe1d"
+  extensions:
+  - ".bt"
+  interpreters:
+  - bpftrace
+  tm_scope: source.bpftrace
+  ace_mode: text
+  language_id: 607227601
 Brainfuck:
   type: programming
   color: "#2F2530"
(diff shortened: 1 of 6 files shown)


lib/linguist/languages.yml
```diff
@@ -732,6 +732,16 @@ Boogie:
   tm_scope: source.boogie
   ace_mode: text
   language_id: 955017407
+BPFTrace:
+  type: programming
+  color: "#ebbe1d"
+  extensions:
+  - ".bt"
```
lildude, line 739: This is a very popular extension which appears to be use more by languages that are _not_ BPFTrace than BPFTrace according to your search. As such all of those files will now become incorrectly classified.

To prevent this, you will need to find at least another language that uses this extension and meets our usage requirements and add it, along with a heuristic to differentiate the two, as part of this PR.

lildude (changes requested):

PR #7562, opened 2025-09-05
Add `.bxl` to Starlark

lf-:
https://buck2.build/docs/bxl/

## Description

Buck2 has PACKAGE files which do directory specific configuration. I am not familiar enough with Bazel to draw a comparison to WORKSPACE, but I think the Buck2 equivalent is a combo of buckconfig and PACKAGE. BXL is the Buck eXtension language and it lets you write programs in Starlark which query your build graph.

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.bxl
    - Sample source(s):
      - https://github.com/facebook/buck2/blob/c779e6f75b80779a4951f99774040a324ff32f2e/examples/bxl_tutorial/part2.bxl
    - Sample license(s): Apache 2 & MIT

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7313,6 +7313,7 @@ Starlark:
   color: "#76d275"
   extensions:
   - ".bzl"
+  - ".bxl"
   - ".star"
   filenames:
   - BUCK
--- a/samples/Starlark/part2.bxl
+++ b/samples/Starlark/part2.bxl
@@ -1,0 +1,26 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is dual-licensed under either the MIT license found in the
+# LICENSE-MIT file in the root directory of this source tree or the Apache
+# License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+# of this source tree. You may select, at your option, one of the
+# above-listed licenses.
+
+def _main(ctx: bxl.Context):
+    universe = ctx.target_universe(ctx.cli_args.target)
+    all_targets = universe.universe_target_set()
+    nodes = ctx.cquery().kind("^(binary|library)$", all_targets)
+    analysis_res_dict = ctx.analysis(nodes)
+    index_outputs = []
+    for _, analysis_res in analysis_res_dict.items():
+        default_info = analysis_res.as_dependency()[DefaultInfo]
+        index_sub_target_provider = default_info.sub_targets["index"]
+        index_outputs.extend(index_sub_target_provider[DefaultInfo].default_outputs)
+    ctx.output.print(pstr(index_outputs))
+
+main = bxl_main(
+    impl = _main,
+    cli_args = {
+        "target": cli_args.target_label(),
+    },
+)


lib/linguist/languages.yml
```diff
@@ -7312,13 +7312,15 @@ Starlark:
   codemirror_mime_type: text/x-python
   color: "#76d275"
   extensions:
+  - ".bxl"
```
lildude: The first extension in this list should be the languages primary extension (see the comment at the top of this file). Usage is no where near meeting our usage requirements so I doubt this is the primary extension for the language.

lib/linguist/languages.yml
```diff
@@ -7312,13 +7312,15 @@ Starlark:
...
   - ".bzl"
   - ".star"
   filenames:
   - BUCK
   - BUILD
   - BUILD.bazel
   - MODULE.bazel
+  - PACKAGE
```
lildude: This is a very generic filename. Are you 100% sure this is the only language using this filename? If not, I recommend removing it and users can use an override as Linguist doesn't support heuristics for named files as they're meant to be unique to a language.

lildude (changes requested):

lib/linguist/languages.yml
lf-: I can't get GitHub code search to tell me, so I have no idea. Do you know how to make it tell me? 

Specifically it being all uppercase makes it likely much more unique. See also WORKSPACE.

lib/linguist/languages.yml
lf-: You're definitely correct, I was having issues with the sorting test but maybe I was doing it wrong.

lib/linguist/languages.yml
lildude: > Do you know how to make it tell me?

Use a regex like this https://github.com/search?q=NOT+is%3Afork+path%3A%2F%28%3F-i%29%28%5E%7C%5C%2F%29PACKAGE%24%2F&type=code

lf-: Yeah, that's not at all unique, ty.

lib/linguist/languages.yml
```diff
@@ -7314,11 +7314,13 @@ Starlark:
   extensions:
   - ".bzl"
   - ".star"
+  - ".bxl"
```
lildude, line 7316: This needs to move one up.

lib/linguist/languages.yml
```diff
@@ -7314,11 +7314,13 @@ Starlark:
...
   filenames:
   - BUCK
   - BUILD
   - BUILD.bazel
   - MODULE.bazel
+  # N.B. PACKAGE files in buck2 are not sufficiently uniquely named so we cannot set them here.
```
lildude: Please don't do this.

lf-:
I have looked at the statistics for popularity. There seem to be no real false positives; some HTML temp files people mysteriously checked in, some seemingly randomized file names. I can't really determine whether there are hundreds of repositories with the UI, but there *are* hundreds of true positive matches for the search.

Issue #5598, opened 2021-10-15
Add `gitcommit` and `gitrebase` language?


spenserblack:
## Language name

- Git Commit
- Git Rebase

## URL of example repository

I think this is N/A, since a normal repository wouldn't commit any files that are in the `.git` directory.

## URL of syntax highlighting grammar

- https://github.com/textmate/git.tmbundle/blob/master/Syntaxes/Git%20Commit%20Message.tmLanguage
- https://github.com/textmate/git.tmbundle/blob/master/Syntaxes/Git%20Rebase%20Message.tmLanguage

## Most popular extensions

Likely N/A, but the filenames are 

- `COMMIT_EDITMSG`
- `git-rebase-todo`

## Detected language

None/Plain text

***

While the files themselves are unlikely to be present in a repository (there are [a lot of results for `COMMIT_EDITMSG`](https://github.com/search?q=filename%3ACOMMIT_EDITMSG&type=Code), though), GitHub also uses linguist to highlight code blocks in Markdown, per my understanding.

Considering that these are common git use cases, I believe that the option to highlight their syntax in Markdown should be available. I think that ***Git***Hub should support syntax highlighting for git filetypes :laughing: 

If it is reasonable to support syntax highlighting for these types of files, could you please consider it?

spenserblack:
Thanks for looking that up! I'm glad that it might not an exception to the rule after all :smiley:

eric-wieser:
> GitHub also uses linguist to highlight code blocks in Markdown, per my understanding.

What is the mapping between ```` ```foo ```` and the language chosen by linguist?

spenserblack:
Wow, this issue is a blast from the past :laughing:
Looking it up again, [here are the relevant docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting). I believe that the language's name or one of its aliases can be `foo` in this case.

Nixinova:
Git-rebase-todo doesnt have any usages. https://github.com/search?q=path%3Agit-rebase-todo+-path%3A*.*&type=code

commit_editmsg does, i will open pr for that one shortly

PR #7596, opened 2025-09-21
Add Git Commit language

Nixinova:
- Implements half of #5598

## Description

Adds the git commit file as its own language. This shouldnt be committed, but is found on github.
Name: "Git Commit", with alias "Commit" (for usage like ````` ```commit `````)

## Checklist:

      - [COMMIT_EDITMSG](https://github.com/search?q=path%3ACOMMIT_EDITMSG+-path%3A*.*&type=code&p=2) 4.9K+ files
    - [Sample](https://github.com/JazzDev26/JazzDev26.github.io/blob/a65f0b815271a59073d43b32e518bab5c0304db0/COMMIT_EDITMSG#L4), under [MIT](https://github.com/JazzDev26/JazzDev26.github.io/blob/main/LICENSE)
  - [x] I have included a syntax highlighting grammar
    - Using `text.git-commit` from my NovaGrammars repo. [NovaLightshow preview](https://nixinova.github.io/NovaLightshow/?grammar-type=url&grammar=https://raw.githubusercontent.com/Nixinova/NovaGrammars/main/grammars/git-commit.yaml-tmLanguage&sample-type=url&sample=https://raw.githubusercontent.com/Nixinova/NovaGrammars/main/samples/COMMIT_EDITMSG)
  - [x] I have added a color
    - Hex value: `#F44D27`
    - Rationale: From the other Git languages

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2588,7 +2588,8 @@ Git Commit:
   filenames:
   - COMMIT_EDITMSG
   ace_mode: text
-  tm_scope: text.hash-commented
+  tm_scope: text.git-commit
+  wrap: true
   language_id: 131750475
 Git Config:
   type: data
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2580,6 +2580,16 @@ Git Attributes:
   codemirror_mode: shell
   codemirror_mime_type: text/x-sh
   language_id: 956324166
+Git Commit:
+  type: data
+  color: "#F44D27"
+  aliases:
+  - commit
+  filenames:
+  - COMMIT_EDITMSG
+  ace_mode: text
+  tm_scope: text.hash-commented
+  language_id: 131750475
 Git Config:
   type: data
   color: "#F44D27"
--- a/grammars.yml
+++ b/grammars.yml
@@ -88,6 +88,7 @@ vendor/grammars/NovaGrammars:
 - source.solution
 - source.win32-messages
 - source.ws
+- text.git-commit
 - text.hash-commented
 - text.robots-txt
 vendor/grammars/ObjectScript.tmBundle:
--- a/samples/Git Commit/filenames/COMMIT_EDITMSG
+++ b/samples/Git Commit/filenames/COMMIT_EDITMSG
@@ -1,0 +1,15 @@
+updated css for all pages 
+# Please enter the commit message for your changes. Lines starting
+# with '#' will be ignored, and an empty message aborts the commit.
+#
+# On branch master
+# Changes to be committed:
+#	new file:   cv.css
+#	modified:   cv.html
+#	new file:   cvdraft.html
+#	modified:   index.css
+#	modified:   index.html
+#
+# Changes not staged for commit:
+#	deleted:    home.css
+#
--- a/vendor/README.md
+++ b/vendor/README.md
@@ -220,6 +220,7 @@ This is a list of grammars that Linguist selects to provide syntax highlighting 
 - **Gettext Catalog:** [textmate/gettext.tmbundle](https://github.com/textmate/gettext.tmbundle)
 - **Gherkin:** [cucumber/cucumber-tmbundle](https://github.com/cucumber/cucumber-tmbundle)
 - **Git Attributes:** [Alhadis/language-etc](https://github.com/Alhadis/language-etc)
+- **Git Commit:** [Nixinova/NovaGrammars](https://github.com/Nixinova/NovaGrammars)
 - **Git Config:** [Alhadis/language-etc](https://github.com/Alhadis/language-etc)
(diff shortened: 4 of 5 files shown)


lib/linguist/languages.yml
```diff
@@ -2580,6 +2580,16 @@ Git Attributes:
...
   language_id: 956324166
+Git Commit:
+  type: data
+  color: "#F44D27"
+  aliases:
+  - commit
+  filenames:
+  - COMMIT_EDITMSG
+  ace_mode: text
+  tm_scope: text.hash-commented
```
Alhadis: Atom's [`language-git` package](https://github.com/atom/language-git) does a great job of highlighting commit messages.

```suggestion
  tm_scope: text.git-commit
```

Alhadis (changes requested):

lib/linguist/languages.yml
Nixinova: Not sure I like the error highlighting for just recommendation.

image

<del>Also, do the special tokens actually work with github styling? They're of the form `markup.deleted`. They dont work in my lightshow but ive forgotten what limitations it has.</del>


I also note that `<space>#` isnt considered a comment, so my hash commented grammar wont work exactly.

Might just make a grammar myself then.

Nixinova: Okay it does look like the `markup.*` tokens work.

image

Might just cope with long line length being error

Alhadis: > Not sure I like the error highlighting for just recommendation.

That's a well-established and widely-accepted convention in Git commit messages. 72 characters on an 80-character wide terminal leaves 8 columns free to indent by one level. This convention even has a name: the 50/72 rule (referring to the line-length limits for commit summary lines and bodies, respectively).

> I also note that `<space>#` isnt considered a comment, so my hash commented grammar wont work exactly.

Yes, that's the expected behaviour; a Git commit-message comment requires a `#` character in the first column.

Nixinova: breaks convention != disallowed though, which is what "invalid.illegal" implies.

Also,  these seem like they would be a problem:

image
image

Making a less opinionated grammar seems better

Alhadis: In that case, I suggest adding `wrap: true` so that sloppily-formatted commit messages are still readable on GitHub.

Alhadis: Also, the [VS Code grammar for commit-messagess](https://github.com/walles/git-commit-message-plus) doesn't enforce line-length limits. Maybe use that instead.

lib/linguist/languages.yml
```diff
@@ -2580,6 +2580,17 @@ Git Attributes:
   codemirror_mode: shell
   codemirror_mime_type: text/x-sh
   language_id: 956324166
+Git Commit:
+  type: data
```
Nixinova, line 2584: data, or prose?

Alhadis: @Nixinova That's a good question. I'm tentatively leaning towards `data`, for reasons similar to why [`E-mail`](https://github.com/github-linguist/linguist/blob/1d913bbdff8e31f1dcf0ba6ce482f4edb5715fc7/lib/linguist/languages.yml#L1777-L1791) is categorised as `data`: it's a transient data format that, while user-readable, is ultimately destined for processing by a program instead of presented as-is.

That's an unbelievably shaky justification, but classifying Git commit-messages as `prose` feels… wrong, somehow.

Nixinova: Yeah I agree

Alhadis (dismissed):

Nixinova:
what kind of keyboard shortcut did i hit to dismiss that review (???)
or does it auto dismiss if i update a dep?

RedCMD:
simply just updating the PR code iirc

Nixinova:
weird, never seen that before and use GitHub every day. don't like the UI saying I'm doing things I'm not...

Alhadis:
> what kind of keyboard shortcut did i hit to dismiss that review (???) or does it auto dismiss if i update a dep?

The thread update says that [my review](https://github.com/github-linguist/linguist/pull/7596#pullrequestreview-3252134150) was dismissed <i><q>via [2c2286c](https://github.com/github-linguist/linguist/commit/2c2286c1aa9d55e4d1aa7b9dca723cdc3b488c69)</q></i>, indicating that GitHub's UI thought to dismiss it automatically out of irrelevancy.

Which, like @Nixinova, leaves me with mixed feelings. I'd dislike software that tries doing more than what it's asked/expected to ([“Worse is Better”](https://en.wikipedia.org/wiki/Worse_is_better)), especially since the commit that rendered my review "stale" was unrelated to the changes I suggested.

Alhadis (approved):

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7600, opened 2025-09-23
Add .gitattributes override mention when returning the strategy

DecimalTurn:
This pull request introduces a note mentioning `.gitattributes` overrides when using the `--strategies` option. It also adds small integration tests to verify that CLI behavior reflects these changes while preserving the normal behavior when  `--strategies` is not specified.

**Language detection improvements:**

* When a `linguist-language` attribute is present in the `.gitattributes` file for the analyzed file, a note is added to mention the override.

**Testing and instrumentation enhancements:**

* Added tests to `test_basic_instrumenter.rb` to verify that the detection strategy and language are correctly tracked when `.gitattributes` overrides are present and the strategy information is displayed correctly.

**CLI integration and coverage:**

* Added a new `test_cli_integration.rb` suite with tests for CLI flags (`--strategies`, `--breakdown`, `--json`) to ensure that `.gitattributes` overrides are detected, the correct strategy info is reported, and JSON output remains accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,14 @@ project.languages      #=> { "Ruby" => 119387 }
 
 ### Command line usage
 
+The `github-linguist` executable operates in two distinct modes:
+
+1. **[Git Repository mode](#git-repository)** - Analyzes an entire Git repository (when given a directory path or no path)
+2. **[Single file mode](#single-file)** - Analyzes a specific file (when given a file path)
+
 #### Git Repository
 
-A repository's languages stats can also be assessed from the command line using the `github-linguist` executable.
+A repository's languages stats can be assessed from the command line using the `github-linguist` executable.
 Without any options, `github-linguist` will output the language breakdown by percentage and file size.
 
 ```bash
@@ -151,6 +156,41 @@ lib/linguist.rb
 …
 ```
 
+##### `--strategies`
+
+The `--strategies` or `-s` flag will show the language detection strategy used for each file. This is useful for understanding how Linguist determined the language of specific files. Note that unless the `--json` flag is specified, this flag will set the `--breakdown` flag implicitly.
+
+You can try running `github-linguist` on the root directory in this repository itself with the strategies flag:
+
+```console
+$ github-linguist --breakdown --strategies
+66.84%  264519     Ruby
+24.68%  97685      C
+6.57%   25999      Go
+1.29%   5098       Lex
+0.32%   1257       Shell
+0.31%   1212       Dockerfile
+
+Ruby:
+  Gemfile [Filename]
+  Rakefile [Filename]
+  bin/git-linguist [Extension]
+  bin/github-linguist [Extension]
+  lib/linguist.rb [Extension]
+  …
+```
+
+If a file's language was overridden using `.gitattributes`, the strategy will show the original detection method along with an override note:
+
+```console
+$ github-linguist --strategies .devcontainer/devcontainer.json
+.devcontainer/devcontainer.json: 27 lines (27 sloc)
+  type:      Text
+  mime type: application/json
+  language:  JSON with Comments
+  strategy:  Filename (overridden by .gitattributes)
+```
+
 ##### `--json`
 
 The `--json` or `-j` flag output the data into JSON format.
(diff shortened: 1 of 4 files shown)


Alhadis:
I think it would be clearer to users if we reported _"(overridden by <samp>[<var>path/to/.gitattributes</var>]</samp>)"_ instead of merely "GitAttributes" (as there's really only one that affects Linguist's classification).

So instead of treating `linguist-language` overrides as a faux-strategy, perhaps tailor the output to include an [overridden] flag, possibly prepended to the strategy that would otherwise have matched in the absence of the override.

DecimalTurn:
Sounds good, I've made the change to display the information like so:

`<FilePath> [<StrategyName> (overridden by .gitattributes)]`

This change means that detection has to take place even when there is an override in order to determine what strategy would be used. However, the changes in [e879628](https://github.com/github-linguist/linguist/commit/e879628af216f951f5c2f4eed820e5ea53a1e77e) also introduce a check on `Linguist.instrumenter` to see if its defined before running the full detection (this is equivalent to check if `--strategies` was specified). This way, we keep the lazy approach if the strategy isn't needed.

Adding the `.gitattributes` path seems to be too complicated to implement for what it adds, but I'm not against the idea in principle.

lildude (changes requested):
Seems like a good idea to me. Can you please also update the README.md to document and show the output this change introduces.

DecimalTurn:
> Seems like a good idea to me. Can you please also update the README.md to document and show the output this change introduces.

In the latest commits, I've added:

* An intro section about the different modes (_Git Repository_ vs _Single file_)
* Documentation about `--strategies` in _Git Repository_ mode
* Documentation about all relevant flags in _Single file_ mode

README.md
```diff
@@ -151,6 +156,41 @@ lib/linguist.rb
...
+Ruby:
+  Gemfile [Filename]
+  Rakefile [Filename]
+  bin/git-linguist [Extension]
+  bin/github-linguist [Extension]
+  lib/linguist.rb [Extension]
+  …
+```
+
+If a file's language was overridden using `.gitattributes`, the strategy will show the original detection method along with an override note:
```
lildude: "strategy will show the original detection method along with an override note"

🤔 this is confusing with the output below. `devcontainer.json` is already JSONC but the output suggests it's overridden to JSONC by an override. Why would you override to the same language?

I think it would be better to use an example where there is a clear difference in languages.

DecimalTurn: > Why would you override to the same language?

I don't know why, but it's the only "override" in the linguist repo 🤷

image

DecimalTurn: > I think it would be better to use an example where there is a clear difference in languages.

Ok, I'll change it for a better example... I'm thinking of using .bas file override in https://github.com/tannerhelland/PhotoDemon

lildude: > I don't know why, but it's the only "override" in the linguist repo 🤷

Whoops. I'm not sure how that slipped past. We shouldn't need that override.

As an aside, we use the `test/attributes` branch for testing overrides. The `.gitattributes` in that branch has a lot more overrides.

DecimalTurn: This discussion makes me think that maybe the implementation of the `--strategies` flag should make the note appears different if the detection gives the same language as the override.

DecimalTurn: Something like:

```ruby
          # Get the original strategy by calling super (which calls Linguist.detect)
          original_language = super
          original_strategy_info = Linguist.instrumenter.detected_info[self.name]
          original_strategy = original_strategy_info ? original_strategy_info[:strategy] : "Unknown"

          # Determine if gitattributes actually changed the result
          if original_language == detected_language
            strategy_name = "#{original_strategy} (confirmed by .gitattributes)"
          else
            strategy_name = "#{original_strategy} (overridden by .gitattributes)"
          end
```

spenserblack: > I don't know why, but it's the only "override" in the linguist repo 🤷

IIRC I made that override because `devcontainer.json` wasn't yet registered as a JSONC filename at the time.

DecimalTurn: @spenserblack would you like to have the honor of removing it?

spenserblack: @DecimalTurn sure! https://github.com/github-linguist/linguist/pull/7711#issue-3740056914

spenserblack: Oh, and regarding the IIRC... I was not remembering correctly at all :sweat_smile:
https://github.com/github-linguist/linguist/pull/5081
https://github.com/github-linguist/linguist/pull/6479

DecimalTurn: Well, at least it's now properly addressed. I'll mark this thread as resolved now.

lildude (approved):

PR #7622, opened 2025-09-30
Mark Gradle and Maven wrappers as generated

### LINKED ISSUE #7621: Gradle wrapper scripts should be marked as generated
## Describe the enhancement

The [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) scripts, `gradlew` and `gradlew.bat` should be marked as generated by default, as they are purely to bootstrap gradle locally, are auto-generated and never to be touched.

Same may be the case for the [maven wrapper](https://maven.apache.org/tools/wrapper/), which has the scripts `mvnw` and `mvnw.cmd`.

megla-tlanghorst:
## Description

Fixes #7621 

Adds the `gradlew`, `gradlew.bat`, `mvnw` and `mvnw.cmd` scripts to the generated files.

## Checklist:

- [x] **I am adding a new extension to a language.** _Kinda, it's less of a language, more of a tool, but in the java context_
      
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3Agradlew+Gradle+start+up+script+for+POSIX+generated+by+Gradle
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3Amvnw+Apache+Maven+Wrapper+startup+batch+script
      - Both of these searches don't seem to be properly finding the right files, even if the found repos definitely have them  if you check
    - Sample source(s):
      - Gradle Wrapper: Just pasted it from one of my projects where it has been generated, but it provides a license header
      - Maven Wrapper: Just pasted it from one of my projects where it has been generated, but it provides a license header
    - Sample license(s):
      - Apache-2.0


megla-tlanghorst:
Test are still failing, but I'm not completely sure why

lib/linguist/generated.rb
```diff
@@ -558,6 +560,20 @@ def bazel_lock?
       !!name.match(/(^|\/)MODULE\.bazel\.lock$/)
     end
 
+    # Internal: Is the blob a generated gradle wrapper file?
+    #
+    # Returns true or false.
+    def gradle_wrapper?
+      !!name.match(/gradlew(\.bat)?/)
```
Alhadis: This expression will match partial filenames like `create-gradlew` or `docs.gradlew.bat.html`. So it needs to be:

```suggestion
      !!name.match(/(?:^|\/)gradlew(?:\.bat)?$/i)
```

(Since we're expected to match batch files, it's expected that this will need to match files on case-insensitive filesystems, such as those used on Windows).

lib/linguist/generated.rb
```diff
@@ -558,6 +560,20 @@ def bazel_lock?
...
+    end
+
+    # Internal: Is the blob a generated maven wrapper file?
+    #
+    # Returns true or false.
+    def maven_wrapper?
+      !!name.match(/mvnw(\.cmd)?/)
```
Alhadis: See above.
```suggestion
      !!name.match(/(?:^|\/)mvnw(?:\.cmd)?$/i)
```

megla-tlanghorst:
> Wouldn't this be a better thing to check for than assuming all files named `gradlew` are generated? BTW, I found a few that _didn't_ contain `generated by Gradle` but _did_ contain the `Gradle start up script` part. As I know nothing about Gradle, I have to ask: are some of these `gradlew` files hand-written (or expected to be hand-written)?

No they are never handwritten (or at least never expected to be manually written or modified, I mean in practice no one can prevent you from doing that, but I think that's the case with all generated files), but they are frequently automatically replaced with newer versions as Gradle is being upgraded. That's why I'm not a big fan of matching by content: I have no idea if that sentence might be changed in the future some day. It might be changed or it might not be, who knows. As far as I know there's no real pattern to the script, only a changeable template from which it is generated and the filenames, which are _always_ the same.

The best other indicators are that it should always be next to a `settings.gradle`, `settings.gradle.kts` or whatever the settings file name of that new format which is coming up was (but as you can see that format those can be written in may expand over the years) and basically always next to a `gradle` folder (though I'm not sure whether that was changeable).

Alhadis:
> That's why I'm not a big fan of matching by content: I have no idea if that sentence might be changed in the future some day.

Good point. Let's stick to regular filename matching then. You certainly know this software a hell of a lot better than I do.

Alhadis (approved):

megla-tlanghorst:
> You're adding explicitly named files so the samples need to be in the `filename` sub-directory of each language's samples.

Do you have any example for that? I can't find any for a `Cargo.lock` for example.

Only [here](https://github.com/megla-tlanghorst/linguist/blob/0230fb9413307edc901922d07a989badd4530bd7/test/test_blob.rb#L161) and I'm not sure whether that's correct?

lildude:
> Do you have any example for that? I can't find any for a `Cargo.lock` for example.

It's [here](https://github.com/github-linguist/linguist/blob/main/samples/TOML/filenames/Cargo.lock) because the file is written in TOML, not Rust.

Accordingly, you should put the files in `filenames` directories in the samples directories of the languages they're actually written in - `Shell` and `Batchfile`.

megla-tlanghorst:
Okay, should be better now I think

megla-tlanghorst:
Okay, I just though I'd update it to the latest version that currently available to align with the bat version (Gradle 9.1.0).

megla-tlanghorst:
Did a revert on that

megla-tlanghorst:
```
TestSamples#test_filename_listed [test/test_samples.rb:53]:
/home/runner/work/linguist/linguist/samples/Batchfile/filenames/mvnw.cmd isn't listed as a filename for Batchfile in languages.yml.
Expected [] to include "mvnw.cmd".
```

So should the CMD files not be in the filenames folder?

lildude:
> ```
> TestSamples#test_filename_listed [test/test_samples.rb:53]:
> /home/runner/work/linguist/linguist/samples/Batchfile/filenames/mvnw.cmd isn't listed as a filename for Batchfile in languages.yml.
> Expected [] to include "mvnw.cmd".
> ```
> 
> So should the CMD files not be in the filenames folder?

You need to add it to the filenames list under the Batchfile section in the `languages.yml` file like you did `mvnw` when you added it to the Shell section (you'll need to add the list). You'll need to do the same thing for the `gradlew.bat` file.

megla-tlanghorst:
Okay, seems a bit redundant, but should work now, at least the tests are green now.

megla-tlanghorst:

I need them in GitLab, so it'll probably be even longer...

PR #7651, opened 2025-10-29
Add B (Formal Method) language

JJWRoeloffs:
This pull request adds the detection of files used for the B formal method.

The B-method is a formal method for writing code that satisfies some formal mathematical description. You start with a mathematical machine (`.mch`), refine it in refinement steps (`.ref`), and finally end with an implementation (`.imp`), which can be transpiled into an executable programming language (such as C, ADA, or Rust.)

These files are not commonly checked into public github repositories, as the projects are generally either too academic or too corporate. This means the latter two extensions are on the border of acceptable of being added to linguist, with the most questionable being `.ref`, which has a little over two hundred files, most of which are in separate repositories as far as I can tell.

For the syntax highlighting, I had to write my own tmGrammar file, as none exist ([though Vim does supports B without plugins](https://github.com/vim/vim/blob/110656ba607d22bbd6b1b25f0c05a1f7bad205c0/runtime/syntax/b.vim),) which I did basing myself on the open [B Language Reference](https://www.atelierb.eu/wp-content/uploads/2023/10/b-language-reference-manual.pdf), making a quick [VSCode extension](https://github.com/JJWRoeloffs/b-vscode). I did this in one grammar file instead of three because the B Language Reference describes it like that.

I used the name `B (Formal Method)` to not confuse anyone about it being the predecessor of C, which is a different language called B. This B is a successor of Z.

Because machine, refinement (or reference) and implementation are not terms unique to B, the extensions are not either, even while no other users are in linguist currently. Because of this, I added some heuristics that should allow these extensions to only be classified as B when they are B, leaving them unclassified otherwise.

## Checklist:
      - `.mch`, Machine file, starting point of the B-method: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.mch+MACHINE+END
      - `.imp`, Implementation file, end point of the B-method: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.imp+IMPLEMENTATION+OPERATIONS+END
      - `.ref`, Refinement file, containing the intermediate steps from machine to implementation: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.ref+REFINEMENT+REFINES+END
    - Some I wrote my own and am happy to license under linguist's MIT, the others are from [Let's prove Leftpad](https://github.com/hwayne/lets-prove-leftpad), which is licensed under CC0 and so can be licensed under MIT. 
     - Wrote my own: https://github.com/JJWRoeloffs/b-vscode/blob/08b8a94567efed815c301c112037a731fe54f58b/syntaxes/b.tmLanguage.json
  - [x] I have added a color
    - Hex value: `#8aa8c5`
    - Rationale: Nearly randomly chosen. It is the color of the logo of a conference where research on the B method is frequently presented, the [abz-conference](https://abz-conf.org/).
    - There are no other languages in linguist using `.mch`, `.imp`, or `.ref`, but a search of the files finds that all of these extensions are more commonly used outside of the B method, so I added heuristics. For each of those heuristics, my intent is to have the file only be classified as B if the regex pattern matches, else having it remain an unclassified file, but I did not find a way to test this working.


JJWRoeloffs:
Fixing all my stuff to make the tests pass took me way longer than it should have. The documentation is way too good for me to make this many mistakes.

---

> There are no other languages in linguist using `.mch`, `.imp`, or `.ref`, but a search of the files finds that all of these extensions are more commonly used outside of the B method, so I added heuristics. For each of those heuristics, my intent is to have the file only be classified as B if the regex pattern matches, else having it remain an unclassified file, but I did not find a way to test this working.

Looking at the way in which the other tests are written, I am _really_ not sure if I did this correctly. In another pull request, you ask for the user to find another popular language with the extension to differentiate, but, in this case, `.ref` and `.imp` just mean "reference" and "implementation" and are used for all sorts of files. I am presuming Linguist has to have a feature for this if the way I did it is not correct, though I cannot find it.

lib/linguist/languages.yml
```diff
@@ -528,6 +528,16 @@ Awk:
   tm_scope: source.awk
   ace_mode: text
   language_id: 28
+B (Formal Method):
+  type: programming
+  color: "#8aa8c5"
+  extensions:
+  - ".mch"
+  - ".imp"
+  - ".ref"
```
lildude: These two extensions are no where need meeting our usage requirements from your search results and will hold back this PR. I recommend removing them and the samples for now. Users can always use a override once this PR is merged and deployed whilst we wait for the popularity to grow.

JJWRoeloffs:
Removed the extensions and heuristics as requested.

Removing the (attempt at) heuritsics means a few `.mch` files that are assembly will be wrongly classified as B method files. These appear to be around 1/20 of all `.mch` files. I guess anyone wanting to fix that could add this assembly themselves. :P

PR #7657, opened 2025-11-02
Add .tsconfig.json to JSONC

ArnaudBarre:
## Description

In a TypeScript codebase, it's frequent to require multiple configuration files for a project. The convention is to use `<prefix>.tsconfig.json` like `test.tsconfig.json` or `node.tsconfig.json`.

## Checklist:

  - [x] The new extension is used in hundreds of repositories on github.com
    - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.tsconfig.json
    - https://github.com/DevExpress/devextreme-vue/blob/9340a269a4a145418b5571fcc2009805455d03a0/jest.tsconfig.json
    - Sample license(s): MIT


DecimalTurn:
The tests are failing because the filename of the sample you submitted is of the form: `tsconfig.<suffix>.json` instead of the extension `.tsconfig.json`.

If the goal of your PR was simply to support this file extension, I guess you can simply replace with another sample from the search results in your original post.

Unfortunately Linguist doesn't support wildcards, so we can't have `.tsconfig.*.json` as an extension. We would need to add every possible tsconfig extension types to add support to them assuming they meet the popularity threshold.

However, looking at [search results ](https://github.com/search?q=path%3A*.tsconfig.*.json&type=code) for files using a file extension of the form `.tsconfig.*.json`, there's only a few results. This tells me it would be better to target direct filenames such as the ones seen [here](https://github.com/search?q=path%3Atsconfig.*.json&type=code):
- tsconfig.app.json
- tsconfig.build.json
- tsconfig.editor.json
- tsconfig.spec.json
- etc.

EDIT: Fixed some errors in original message

ArnaudBarre:
Thanks for looking into it and sorry for doing this too quickly and not seing that it's actually way more common to do `tsconfig.*.json` than `*.tsconfig.json` and that's what the Vite templates are using 🙈

I don't think listing `tsconfig.app.json`, `tsconfig.build.json`, ... is a good direction because it will be more maintenance to decide what is going in the list or not. Instead I think it's better to push for a unique `.tsconfig.json` extension and spread the word that naming it this way get's you better syntax highlighting. I will update the Vite template accordingly and try to spread the word.

DecimalTurn:
> Instead I think it's better to push for a unique .tsconfig.json extension and spread the word that naming it this way get's you better syntax highlighting.

It would be much simpler indeed, but if there's enough people that use those filenames like `tsconfig.app.json`, I don't see a reason not to support it. That can be done seperatly from this PR however.

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #7687, opened 2025-11-27
Add Go Template and differentiate from Smarty

trewallier:
## Description
Add Go template as a new language.
This PR is intended to solve https://github.com/github-linguist/linguist/issues/7639 and https://github.com/github-linguist/linguist/issues/6196 so that Helm chart repositories are classified correctly.
Note that Go template use braces similarly to Jinja `{{ }}` and the file extension might not differentiate template languages in general. So this PR focus only on Smarty - Go template differences as they usually share the same extension in Helm charts.

Rationale for added extensions:
- `.gohtml`: IntelliJ IDEA has association with this extension. ([link to source](https://www.jetbrains.com/help/idea/integration-with-go-templates.html))
- `.html.tmpl`: GitHub search verifies that is widely used. Mentioned in a [StackOverflow answer](https://stackoverflow.com/a/43295290) and also default in a [Prettier plugin](https://github.com/NiklasPor/prettier-plugin-go-template).
- `.gotmpl`, `.tmpl`: Default extensions of Go template files the Go language server (gopls) ([link to documentation](https://go.dev/gopls/features/templates#enabling-template-support)). Also the VS Code Go Plugin use this default ([link to docs](https://github.com/golang/vscode-go/wiki/features#go-template-syntax-highlighting)).
- `.tpl`: Convention used in Helm charts as stated by the [Helm documentation](https://helm.sh/docs/chart_template_guide/named_templates) ("Conventionally, Helm charts put [...] templates inside of a partials file, usually _helpers.tpl")

Since none of the extensions are defined by the [Go template documentation](https://pkg.go.dev/text/template), and none of the supporting tools stick to one particular default, I left the list in alphabetical order in the `languages.yml` config.

## Checklist:
      - `_helpers.tpl` https://github.com/search?type=code&q=NOT+is%3Afork+path%3A**%2F_helpers.tpl+%7B%7B-
      - `.gohtml` https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.gohtml
      - `.gotmpl` https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.gotmpl
      - `.html.tmpl` https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.html.tmpl
      - `.tmpl` https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.tmpl+%7B%7B-
      - `.tpl` https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.tpl+%7B%7B-
    - Sample source(s):
      - _helpers.tpl [source](https://helm.sh/docs/chart_template_guide/named_templates/)
      - base.tpl [source](https://github.com/go-echarts/go-echarts/blob/master/templates/base.tpl)
      - batch.gohtml [source](https://github.com/ncruces/RethinkRAW/blob/master/assets/batch.gohtml)
      - _templates.gotmpl [source](https://github.com/redhat-cop/helm-charts/blob/main/_templates.gotmpl)
      - crd.tmpl [source](https://github.com/grafana/k6-operator/blob/main/docs/crd.tmpl)
      - letter.tmpl [source](https://pkg.go.dev/text/template#example-Template)
      - prettier.html.tmpl [source](https://github.com/NiklasPor/prettier-plugin-go-template/blob/master/README.md)
    - Sample license(s):
      - _helpers.tpl [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)
      - base.tpl [MIT](https://github.com/go-echarts/go-echarts/blob/master/LICENSE)
      - batch.gohtml [MIT](https://github.com/ncruces/RethinkRAW/blob/master/LICENSE)
      - _templates.gotmpl [Apache License 2.0](https://github.com/redhat-cop/helm-charts/blob/main/LICENSE)
      - crd.tmpl [Apache License 2.0](https://github.com/grafana/k6-operator/blob/main/LICENSE)
      - letter.tmpl [BSD-3-Clause](https://pkg.go.dev/text/template?tab=licenses)
      - prettier.html.tmpl [MIT License](https://github.com/NiklasPor/prettier-plugin-go-template/blob/master/LICENSE)
  - [x] I have included a syntax highlighting grammar: https://github.com/jinliming2/vscode-go-template
  - [x] I have added a color
    - Hex value: `#00ADD8`
    - Rationale: Used same color as all other Go languages

    - Sample source(s):
      - header.tpl [source](https://github.com/smarty-php/smarty/blob/master/demo/templates/header.tpl)
      - index.tpl [source](https://github.com/smarty-php/smarty/blob/master/demo/templates/index.tpl)
    - Sample license(s):
      - header.tpl [LGPL LICENSE](https://github.com/smarty-php/smarty/blob/master/LICENSE)
      - index.tpl [LGPL LICENSE](https://github.com/smarty-php/smarty/blob/master/LICENSE)

- [x] **I am updating a grammar submodule**

### DIFF
--- a/samples/Go Template/_helpers.tpl
+++ b/samples/Go Template/_helpers.tpl
@@ -1,4 +1,4 @@
-{{/* Generate basic labels */}}
+{{- /* Generate basic labels */ -}}
 {{- define "mychart.labels" }}
   labels:
     generator: helm
--- a/samples/Go Template/_templates.gotmpl
+++ b/samples/Go Template/_templates.gotmpl
@@ -1,0 +1,71 @@
+{{- /* https://github.com/redhat-cop/helm-charts/blob/main/_templates.gotmpl */ -}}
+
+{{ define "doc.header" -}}
+
+{{ template "chart.header" . }}
+  [![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+
+  {{ template "chart.versionBadge" . }}
+
+  {{ template "chart.deprecationWarning" . }}
+
+  ## Description
+
+  {{ template "chart.description" . }}
+
+{{- end }}
+
+{{ define "doc.maintainer_and_sources" -}}
+
+{{ template "chart.maintainersSection" . }}
+
+## Sources
+Source: {{ template "chart.sourcesList" . }}
+
+Source code: {{ template "chart.homepage" . }}
+
+{{- end }}
+
+
+{{ define "doc.deps" -}}
+
+## Dependencies
+
+This chart has the following dependencies:
+
+{{ template "chart.requirementsTable" . }}
+
+{{- end }}
+
+{{ define "doc.values" -}}
+
+{{ template "chart.valuesSection" . }}
+
+{{- end }}
+
+
+{{ define "doc.footer" -}}
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```console
+helm install my-release repo/<chart-name>>
+```
+
+The command deploys the chart on the Kubernetes cluster in the default configuration.
+
+## Uninstalling the Chart
+
+To uninstall/delete the my-release deployment:
+
+```console
+helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+{{ template "helm-docs.versionFooter" . }}
+
+{{- end }}
--- a/samples/Go Template/base.tpl
+++ b/samples/Go Template/base.tpl
@@ -1,0 +1,36 @@
+{{ /* Example from https://github.com/go-echarts/go-echarts/blob/master/templates/base.tpl */ }}
+{{- define "base_element" -}}
+<div class="container">
+    <div class="item" id="{{ .ChartID }}" style="width:{{ .Initialization.Width }};height:{{ .Initialization.Height }};"></div>
+</div>
+{{- end -}}
+
+{{- define "base_script" -}}
+<script type="text/javascript">
+    "use strict";
+    let goecharts_{{ .ChartID | safeJS }} = echarts.init(document.getElementById('{{ .ChartID | safeJS }}'), "{{ .Theme }}", { renderer: "{{  .Initialization.Renderer }}" });
+    let option_{{ .ChartID | safeJS }} = {{ template "base_option" . }}
+    goecharts_{{ .ChartID | safeJS }}.setOption(option_{{ .ChartID | safeJS }});
+
+  {{- range  $listener := .EventListeners }}
+    {{if .Query  }}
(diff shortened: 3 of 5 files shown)


lib/linguist/languages.yml
```diff
@@ -2736,6 +2736,20 @@ Go Module:
   tm_scope: go.mod
   ace_mode: text
   language_id: 947461016
+Go Template:
+  type: markup
+  color: "#00ADD8"
+  extensions:
+  - ".gohtml"
```
lildude, line 2745: See the comment at the top of this file. The first extension should be the language's primary extension. Your comments and the fact you've not provided a sample or supporting search result suggests this is not the primary extension for the language.

lib/linguist/languages.yml
```diff
@@ -2736,6 +2736,20 @@ Go Module:
...
+  - ".gotmpl"
+  - ".html.tmpl"
+  - ".tmpl"
+  - ".tpl"
+  filenames:
+  - _helpers.tpl
+  tm_scope: text.html.go-template
```
lildude: You're guessing and haven't run the tests 😉 

This needs a supporting grammar. Linguist doesn't currently have a grammar that provides this scope so this either needs to be set to `text` (which kinda defeats the point in differentiating when you're not going to get any syntax highlighting) or you need to add a grammar. 

I notice you conveniently removed that section of the template too 😉

lib/linguist/languages.yml
trewallier: Thanks for the reminder, added my reason to the PR description why I did not choose default. If you suggest one particular extension as default based on the description I can update the PR happily.

lib/linguist/languages.yml
trewallier: Sure I missed the test, and tried to fix them in https://github.com/github-linguist/linguist/pull/7687/commits/f0ffa723a9bc6a3f15f2522fb1969a8a23482fd0. Still I get some repo reference errors on my fork. Do you have any suggestion where should I start with errors like this? Or is it just happening because on a fork?
`Rugged::OdbError: object not found - no match for id (7dbcffcf982e766fc711e633322de848f2b60ba5)`

Added supporting grammar. Feel free to suggest anything else which might fit better.

Also sorry for the sloppy PR description, I was not expecting such a quick review, and wanted to save my progress.

trewallier:
Thanks for the quick review, tried to fix every issue from the first commit, and added more info to the PR description.

Regarding the extensions I researched the topic from every aspect I could, and included search strings. Still if you feel I should remove some, just let me know. For me it looked like every search string has at least 10k results, but I may interpret the search result incorrecty.

lib/linguist/languages.yml
lildude: > Or is it just happening because on a fork?
Rugged::OdbError: object not found - no match for id (7dbcffcf982e766fc711e633322de848f2b60ba5)

This normally happens if you've only included the `main` branch when forking. We have tests that rely on objects in the `test/attributes` brach.

trewallier:
@lildude If you have some time, please let me know what I need to update so that this PR can progress. I am eager to settle this topic and provide any improvement.

PR #7696, opened 2025-12-07
Replacement of COBOL sample

AntonioGibas:
removing samples/COBOL/hello_world.cbl, replacing it with example that covers more language usage
## Description
Reason for removal:
The repository rules have been updated and explicitly state that generic 'Hello World' examples are not allowed within the samples directory. This commit ensures the COBOL section is compliant with the repository's current guidelines.

Relevant commit:
* Delete samples/COBOL/hello_world.cbl
* Add Fibonacci sequence generator and analyzer program

## Checklist:
    - Sample source(s):
    - Sample license(s):

  - [x] I have included a new sample for the misclassified language

### DIFF
--- a/samples/COBOL/ADVFIBB.cbl
+++ b/samples/COBOL/ADVFIBB.cbl
@@ -1,0 +1,384 @@
+      ******************************************************************
+      * This COBOL program generates and analyzes the Fibonacci sequence 
+      * with comprehensive statistical calculations. 
+      * It classifies each  number by parity (even/odd), primality, and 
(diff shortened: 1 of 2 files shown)


lildude (changes requested):
> Add Fibonacci sequence generator and analyzer program

If you're going to remove samples in the name of adhering to the guidelines, you should be adding a real-world sample. Fibonacci examples are just like "Hello world" samples in that no one really uses these outside of learning the language.

The PR template should also point to the source of the real-world sample and state the license of the sample.

PR #7704, opened 2025-12-12
Update the heuristics and examples for Mojo language to include the "comptime" keyword

dgurchenkov:
I am updating a definition of an existing language. 

## Description

This PR contains updated heuristics and new examples for Mojo language. The language was added to Linguist more than a year ago and since then the language syntax has changed. In particular, we added a new keyword ("comptime") that is unique to the language and thus it makes sense to include it in the detection heuristics. I also added new examples that better reflect the current syntax of the language.

## Checklist:

    - Sample source(s):
      - https://github.com/modular/modular/blob/main/max/kernels/src/linalg/matmul/cpu/impl.mojo (as samples/Mojo/Matmul_25_7.mojo)
      - https://github.com/modular/modular/blob/main/mojo/stdlib/stdlib/collections/list.mojo (as samples/Mojo/List_25_7.mojo)
    - Sample license(s):
      - Apache License v2.0 with LLVM Exceptions: https://llvm.org/LICENSE.txt 
      - License included in both sample files


dgurchenkov:
> We do not change samples. These samples are used to train the classifier and whilst languages evolve, the content on GitHub may not, and as such we prefer to keep samples as they were when added and prefer to add more samples.
> 
> Please revert the changes to the samples and instead add more real-world samples representing the evolution of the language. Be sure to update the PR template to link to the original source of the new samples and state their license.

Thank you Colin for looking at this so quickly!
I reverted changes to the samples, and instead added two more sample files. Links to the sample files are in the updated description. License is stated and included in the files as well.

Please let me know if the changes are in line with your expectations, happy to change it further.

Also, if you don't mind me asking -- I have updated the language grammar in the repo linked from the linguist/blob/main/vendor/README.md file (https://github.com/modular/mojo-syntax) so I don't need to do anything in this PR, the grammar change will be pulled from the linked repo automatically on the next release. Is that right?

dgurchenkov:
I just realized that there are several more syntax changes coming to Mojo in the coming weeks. There is no value in adding examples now, if the language syntax is unstable. 
I'm going to abandon this PR and will reopen a new one once the syntax stabilizes.

dgurchenkov:
> `samples/Mojo/List_25_7.mojo` is waaay too big. If the sample is suppressed in the diff, it's too big. Please replace this with a smaller sample.
> 
> On the topic: I'm not sure why you're extending the regex other than because this is a new keyword. Both of your samples would be matched by the current regex so extending it is unnecessary. We generally only need to modify the heuristics when they're not matching as expected and the samples would then be a case where the old regex didn't match but the new does.

Thanks Colin, this is helpful.

dgurchenkov:
Abandoning this PR, will re-open or create a new one once the @parameter decorator is renamed and the assert keyword is merged. Otherwise we'd be adding examples that would not represent the language.

PR #7184, opened 2024-12-30
Add Answer Set Programming and Linear Programming

nickswalker:
This PR adds language entries for the two types of programs that use the `.lp` extension; answer set programs and linear programs. They're both common, and they shouldn't be added independently to avoid widespread misclassification.

## Description 

Adding Answer Set Programming was first suggested in #3867, but it was difficult to assess popularity at the time because GitHub code search didn't support special characters. https://github.com/github-linguist/linguist/issues/3867#issuecomment-338144722 points out that `.lp` isn't necessarily exclusive to ASP, but fortunately Prolog (the most similar commonly used logic programming language) has coalesced on `.pl`.

#4860 discussed adding a language group for ASP (the language varies slightly across different solvers). The included grammar is quite generic, so I don't believe there's a need for specific sub-grammars/languages. 

#4860 also points at the `.lp` is also used for "linear programs" (common [human-readable input format](https://docs.gurobi.com/projects/optimizer/en/current/reference/fileformats/modelformats.html#formatlp) for LP solvers). These can be told apart by looking for the required objective statement. [Gurobi docs](https://docs.gurobi.com/projects/optimizer/en/current/reference/fileformats/modelformats.html#objective-section):

> The first section in an LP file is the objective section. This section begins with one of the following six keywords: minimize, maximize, minimum, maximum, min, or max. Capitalization is ignored. This keyword may appear alone, or it may be immediately followed by multi-objectives, which indicates that the model contains multiple objective functions.

## Answer Set Programming
[Answer Set Programming](https://en.wikipedia.org/wiki/Answer_set_programming) is a logic programming paradigm which dates back to the 1990s.  The `.lp` (logic program) file extension has been used for ASP programs since at least 1998, when it was used by [lparse](http://www.tcs.hut.fi/Software/smodels/) and the smodels solver.

Today, there are many ASP-focused repos on GitHub. Here are [more than a hundred that opted into the `#answer-set-programming` topic](https://github.com/topics/answer-set-programming).

### Checklist:

- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      - `.lp` with `:-`: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.lp+%3A-
        - Returns 36k files as of December 2024. I would conservatively estimate >1000 repos, assuming 36 .lp files on average. 
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
[...]

PR #7716, opened 2025-12-24
Improve detection of ASP and LP files

nickswalker:
Add a pattern for Answer Set Programming.

## Description

Answer Set Programming (#7184) was patternless, leading to false positives on LibrePCB .lp files (which don't appear popular enough for inclusion themselves). Now, it now matches files containing specific ASP directives (`#include`, `#const`, etc.). Because some ASP files won't have directives, it also looks for rules defined with `:-`, or lines simply containing "facts" (`foo(x).` or `atom.` or `edge(2, 4).`).

Added an additional ASP sample `tohB.lp` from Potassco's clingo repo (MIT licensed) to exercise the patterns.

## Checklist:
    - Sample source(s):
      - [tohB.lp](https://github.com/potassco/clingo/blob/920d06bcda7dd420814ce50953feff260a60fd8b/examples/clingo/multishot/tohB.lp)
    - Sample license(s): MIT

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -458,7 +458,7 @@ disambiguations:
 - extensions: ['.lp']
   rules:
   - language: Linear Programming
-    pattern: '^\s*(?i:minimize|minimum|min|maximize|maximum|max)(?i:\s+multi-objectives)?(?:\s|$)'
+    pattern: '(?i)^\s*(?:minimize|minimum|min|maximize|maximum|max)(?:\s+multi-objectives)?\s*$'
   - language: Answer Set Programming
     pattern:
     - '^\s*#(?:include|const|program|script|show|minimize|maximize|heuristic|project|edge|theory|defined)\b'


lib/linguist/heuristics.yml
```diff
@@ -458,8 +458,12 @@ disambiguations:
 - extensions: ['.lp']
   rules:
   - language: Linear Programming
-    pattern: '^(?i:minimize|minimum|min|maximize|maximum|max)(?i:\s+multi-objectives)?$'
+    pattern: '^\s*(?i:minimize|minimum|min|maximize|maximum|max)(?i:\s+multi-objectives)?(?:\s|$)'
```
lildude: This regex is vulnerable to ReDos attack. Please fix it, and also ensure all regexes you're adding are RE2 compatible.

lildude (changes requested):
See inline comment.

Additionally, we need tests for all heuristics.

Issue #3835, opened 2017-09-23
Language of repo incorrectly identified as OpenEdge ABL


tmadhuri:
I had a repo with majorly python files and some data files in `.txt` and `.p`. The `.p` files are incorrectly identified as OpenEdge ABL source files.

Edit: Repo link: https://github.com/tmadhuri/Community-Detection

lildude:
This is happening because of the `.p` extension. This is only associated with OpenEdge. You'll need to implement a manual override to force these to Python or use one of the more commonly used Python extensions for those files.

pchaigno:
See the [README.md](https://github.com/github/linguist#overrides) for manual overrides.

Issue #3619, opened 2017-05-11
Python pickles are not OpenEdge ABL


pratheeksh:
It mistakes .p files to be something strange and since they're large, my repo is considered to be ABL. That makes no sense

pchaigno:
`.p` is currently not recognized by Linguist as a Python extension. If you think it should, you can take a look at the [guidelines to add support for a new extension](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language). In particular, notice the requirement on in-the-wild usage.

PR #7724, opened 2025-12-31
Add heuristics for OpenEdge ABL

nickswalker:
## Description

`.p` is a common file extension which is getting mapped to OpenEdge ABL in most cases (except when matching a pattern for GNUplot files). This has been raised several times ( #4216, #4074, #3835, #3619, #1277). This PR adds patterns to prevent this. I also added a pattern for the language under the `.cls` extension, as it was similar and there were already samples in the repo.

There were already two OpenEdge .p samples, but one was a single-line hello world. I have replaced this sample.

## Checklist:

    - Sample source(s): 
      - https://github.com/chriscamicas/vscode-abl/blob/a3e1fe7445fb87ee16271e1527b3da5ca99b573d/abl-src/run.p#L6
    - Sample license(s): MIT

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -192,7 +192,7 @@ disambiguations:
   - language: ObjectScript
     pattern: '^Class\s'
   - language: OpenEdge ABL
-    pattern: '(?i)^\s*(end\s+(class|interface)\.|(?:class|interface)\s+[a-z_][a-z0-9_#$%&-]*:)'
+    pattern: '(?i)^\s*(?:end\s+(?:class|interface)\.|(?:class|interface)\s+[a-z_][a-z0-9_#$%&-]*:)'
 - extensions: ['.cmp']
   rules:
   - language: Gerber Image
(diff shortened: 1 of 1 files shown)


lib/linguist/heuristics.yml
```diff
@@ -608,6 +610,7 @@ disambiguations:
     - '^s?plot\b'
     - '^set\s+(term|terminal|out|output|[xy]tics|[xy]label|[xy]range|style)\b'
   - language: OpenEdge ABL
+    pattern: '^\s*(?i)(end\s+procedure\.|procedure\s+[a-z_][a-z0-9_#$%&-]*:|define\s+variable\s+[a-z_][a-z0-9_#$%&-]*\s+as\s+[^.]+\.)'
```
lildude: This regex is vulnerable to redos attack. Please fix it, and ensure your new regexes are also RE2 compatible.

nickswalker: It looks like the expensive part is `[^.]+\.` which consumes characters until the statement-ending period. I see that this could cause $n$ backtracking steps if there's no period at the end of a string.

I swapped to a couple simpler patterns that just look for keywords to avoid this, and put the case insensitivity flag at the beginning for RE2.

PR #6680, opened 2024-01-17
Add Glimmer TS language

gilest:
## Description
Adds support for TypeScript-flavoured [Glimmer.js](https://glimmerjs.com/) which will be the [component authoring format](https://github.com/ember-template-imports/ember-template-imports) of the [next Edition of Ember.js](https://emberjs.com/editions/polaris/).

## Checklist:
- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      - [.gts search](https://github.com/search?q=%28path%3A*.gts%29+AND+%28content%3A%3Ctemplate%3E+OR+content%3A%40glimmer+OR+content%3A%40ember%29+NOT+is%3Afork&type=code) 2k files excluding forks
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
	    - [class.gts](https://github.com/cardstack/cardstack/blob/main/packages/boxel/addon/components/boxel/input/time/usage.gts)
	    - [template-only.gts](https://github.com/EmberCrate/website/blob/main/app/components/resource-card.gts)
    - Sample license(s):
	    - class.gts -(https://github.com/cardstack/cardstack) MIT
	    - template-only.gjs (https://github.com/CrowdStrike/opensource.crowdstrike.com) MIT
  - [x] I have added a color
    - Hex value: `#3178c6`
    - Rationale: TypeScript colour which is also used by TypeScript and TSX
  - [x] I have updated the heuristics to distinguish my language from others using the same extension.

PR #7732, opened 2026-01-05
Add aliases to glimmer-js and glimmer-ts

ijlee2:
## Description

Patches #6630 and #6680.

While the terms "Glimmer JS" and "Glimmer TS" do currently appear in public repos in the context of parsers, I think these terms are technical and not what Ember developers would colloquially refer to files with `<template>` tags. Rather, we refer to `gjs` and `gts`, respectively.

Example: [Ember's official tutorial](https://guides.emberjs.com/release/tutorial/part-1/orientation/) doesn't mention "Glimmer JS" or "Glimmer TS."

Examples of more technical documentations:

- https://github.com/ember-cli/ember-template-imports/tree/v4.4.0-ember-template-imports?tab=readme-ov-file#using-template-tags-and-gjsgts-files
- https://rfcs.emberjs.com/id/1046-template-tag-in-routes/

I added the `aliases` key to these 2 languages, in particular, to ease writing code blocks on GitHub. Once this pull request is approved and released, developers would be able to write the following:

````diff
- ```glimmer-js
+ ```gjs
const Hello = <template>
  <div>Hello, {{@name}}!</div>
</template>;

export default Hello;
```
````

````diff
- ```glimmer-ts
+ ```gts
interface HelloSignature {
  Args: {
    name: string;
  }
}

const Hello: HelloSignature = <template>
  <div>Hello, {{@name}}!</div>
</template>;

export default Hello;
```
````

## Checklist:

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -2662,6 +2662,8 @@ Glimmer JS:
   language_id: 5523150
 Glimmer TS:
   type: programming
+  aliases:
+  - gts
   extensions:
   - ".gts"
   ace_mode: typescript

### CODE REVIEW

File: vendor/README.md [hunk 1]
```diff
@@ -226,8 +226,8 @@ This is a list of grammars that Linguist selects to provide syntax highlighting
 - **Git Config:** [Alhadis/language-etc](https://github.com/Alhadis/language-etc)
 - **Git Revision List:** [Nixinova/NovaGrammars](https://github.com/Nixinova/NovaGrammars)
 - **Gleam:** [gleam-lang/tree-sitter-gleam](https://github.com/gleam-lang/tree-sitter-gleam) 🐌
-- **Glimmer JS:** [lifeart/vsc-ember-syntax](https://github.com/lifeart/vsc-ember-syntax)
-- **Glimmer TS:** [lifeart/vsc-ember-syntax](https://github.com/lifeart/vsc-ember-syntax)
+- **Glimmer JS (GJS):** [lifeart/vsc-ember-syntax](https://github.com/lifeart/vsc-ember-syntax)
+- **Glimmer TS (GTS):** [lifeart/vsc-ember-syntax](https://github.com/lifeart/vsc-ember-syntax)
```
lildude: This file is automatically generated with these values coming from the language name. Please revert this change.

lildude (changes requested):
See inline comment.

vendor/README.md
ijlee2: Ah, I didn't know. Will do!

ijlee2:
I have a question. In the `main` branch, [`lib/linguist/languages.yml#L3684`](https://github.com/github-linguist/linguist/blob/a7e40d31e271f6747fa1234df37b5fff3e6e2406/lib/linguist/languages.yml#L3684) lists `gjs` as a possible interpreter for JavaScript (introduced in #4783). Would this create a conflict with the alias `gjs` for `*.gjs` files?

(Currently on GitHub, code blocks with `gjs` language do get syntax highlighting, but not `gts`. I wonder if the discrepancy in behavior is due to `gjs` having been listed in `interpreters`.)

lildude:
> Would this create a conflict with the alias `gjs` for `*.gjs` files?

No. The `gjs` codeblocks are being correctly rendered due to the Glimmer JS entry as this is the only language with this extension.

The `gts` blocks are being detected as Gerber because there are two languages with this extension but it appears the heuristic isn't being used to differentiate so it's going on the first result for some reason:

Screenshot 2026-01-08 at 10 01 31

Codeblocks are rendered using the [markup library](https://github.com/github/markup) which feeds off linguist, but isn't really maintained much these days. I'm not very familiar with this how markup works, but this should improve things.

lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

ijlee2:
@lildude Thank you for your prompt response and help with looking into what may be happening differently between `gjs` and `gts`. 🎉

PR #7702, opened 2025-12-10
Update `vscode-motoko` GitHub organization

rvanasa:
This PR replaces the `dfinity/vscode-motoko` submodule with [`caffeinelabs/vscode-motoko`](https://github.com/caffeinelabs/vscode-motoko) to account for the repository being transferred between organizations.

## Checklist:

- [ ] **I am adding a new extension to a language.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am adding a new language.**
  - [ ] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a syntax highlighting grammar: [URL to grammar repo]
      
  - [ ] I have added a color
    - Hex value: `#RRGGBB`
    - Rationale: 
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.

- [ ] **I am fixing a misclassified language**
  - [ ] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [x] **I am changing the source of a syntax highlighting grammar**
  - Old: https://github.com/dfinity/vscode-motoko
  - New: https://github.com/caffeinelabs/vscode-motoko

- [ ] **I am updating a grammar submodule**
  

- [ ] **I am adding new or changing current functionality**
  
  - [ ] I have added or updated the tests for the new or changed functionality.

- [ ] **I am changing the color associated with a language**
  
  - [ ] I have obtained agreement from the wider language community on this color change.
    - [URL to public discussion]
    - [Optional: URL to official branding guidelines for the language]

PR #7734, opened 2026-01-06
Replace Motoko grammar

rvanasa:
This PR replaces the `dfinity/vscode-motoko` submodule with [`caffeinelabs/vscode-motoko`](https://github.com/caffeinelabs/vscode-motoko). Note that this is the same repository/grammar under a new organization.

Supersedes #7702.

## Checklist:

  - Old: https://github.com/dfinity/vscode-motoko
  - New: https://github.com/caffeinelabs/vscode-motoko

### DIFF
--- a/vendor/licenses/git_submodule/ArkTS.dep.yml
+++ b/vendor/licenses/git_submodule/ArkTS.dep.yml
@@ -1,0 +1,10 @@
+---
+name: ArkTS
+version: 107a21c78819540405549c9e65d9447d313c165a
+type: git_submodule
+homepage: https://github.com/ohosvscode/arkTS
+license: mit
+licenses:
+- sources: LICENSE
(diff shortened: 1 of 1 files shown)


vendor/README.md
```diff
@@ -40,7 +40,6 @@ This is a list of grammars that Linguist selects to provide syntax highlighting
 - **Apex:** [forcedotcom/apex-tmLanguage](https://github.com/forcedotcom/apex-tmLanguage)
 - **Apollo Guidance Computer:** [Alhadis/language-agc](https://github.com/Alhadis/language-agc)
 - **AppleScript:** [textmate/applescript.tmbundle](https://github.com/textmate/applescript.tmbundle)
-- **ArkTS:** [ohosvscode/arkTS](https://github.com/ohosvscode/arkTS)
```
lildude, line 43: Whoops. You should not be removing grammars.

rvanasa: I followed the instructions of running `script/add-grammar --replace vscode-motoko https://github.com/caffeinelabs/vscode-motoko`, which caused that other grammar to be removed. I could manually remove that from the diff.

lildude: I'm not sure. It normally happens if the grammar fails to be cloned, but the source looks good now, however there does appear to have been some changes recently, so it's possible it broke something. I can investigate that later.

You should find that reverting your last commit and then running that command again will do the same thing. If it does, you can reset that submodule before committing by running `git submodule update --init -- vendor/grammars/ArkTS` and then `script/list-grammars`.

If you check the changes at this point, it should only be the files related to the new grammar which you can then commit and push. No need to force push as we squash all commits when merging.

rvanasa: Ok, thanks! I reran the script (same result) and then manually removed the changes related to the ArkTS grammar.

lildude: > I can investigate that later.

Looks like the grammar has problems and wasn't added correctly in https://github.com/github-linguist/linguist/pull/7593#issuecomment-3723261971.

We've not made a release since that was added so I'm reverting the PR: https://github.com/github-linguist/linguist/pull/7737

PR #7759, opened 2026-01-13
Add Eask file as Emacs Lisp

jcs090218:
## Description

Add Eask file as Emacs Lisp.

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A**/Eask
    - Sample source(s):
      - https://github.com/jcs090218/setup-emacs-windows/blob/master/Eask
    - Sample license(s):
      - MIT

### DIFF
--- a/samples/Emacs Lisp/filenames/Eask
+++ b/samples/Emacs Lisp/filenames/Eask
@@ -1,21 +1,22 @@
 ;; -*- mode: eask; lexical-binding: t -*-
 
-(package "composer"
-         "0.2.0"
-         "Interface to PHP Composer")
+(package "setup-emacs-windows"
+         "4.0.0"
+         "A Github Action that installs a specific emacs version")
 
-(website-url "https://github.com/emacs-php/composer.el")
-(keywords "tools" "php" "dependency" "manager")
+(website-url "https://github.com/jcs090218/setup-emacs-windows")
+(keywords "ci" "gha")
 
-(package-file "composer.el")
+(package-file "setup-windows.el")
 
 (script "test" "echo \"Error: no test specified\" && exit 1")
 
 (source 'gnu)
 (source 'melpa)
+(source 'jcs-elpa)
 
-(depends-on "emacs" "25.1")
-(depends-on "seq")
-(depends-on "php-runtime")
+(depends-on "emacs" "26.1")
+(depends-on "request")
+(depends-on "s")
 
 (setq network-security-level 'low)  ; see https://github.com/jcs090218/setup-emacs-windows/issues/156#issuecomment-932956432


lildude:
> 
>   * Sample source(s):
>     * https://github.com/emacs-php/composer.el/blob/master/Eask
>   * Sample license(s):

It doesn't look like you can use this file. The repo doesn't have a license so it's assumed it's proprietary. Please find a sample with an open source license and update this point to point to the source and state the license.

Popularity isn't sufficient to merge this PR yet either.

jcs090218:
> It doesn't look like you can use this file. The repo doesn't have a license so it's assumed it's proprietary. Please find a sample with an open source license and update this point to point to the source and state the license.

I have updated the link to https://github.com/jcs090218/setup-emacs-windows/blob/master/Eask. Does this one work? 🤔 

> Popularity isn't sufficient to merge this PR yet either.

Oh, sorry—what level of popularity is required to open a PR for this repository? Thanks!

lildude:
That new sample is good. I've updated the template to state it's covered by MIT.

As for popularity… it's detailed in the CONTRIBUTING.md file and issue it references.

jcs090218:
I assumed is this:

> We try only to add new extensions once they have some usage on GitHub. In most cases we prefer that each new file extension be in use in at least 200 unique :user/:repo repositories before supporting them in Linguist (but see https://github.com/github/linguist/issues/5756 for a temporary change in the criteria).

Consider how many people are interested in Emacs Lisp, so I think the bar is a bit too high. 🤔 And, I am not sure how accurate the measurement is since the search doesn't list out all repos (not even my own repos...). If you look at the graph from https://github.com/emacs-eask/setup-eask/network/dependents, there is a total of `515` repos that use this action (including forks). Of course, the graph isn't accurate as well since there are people who prefer not to use GHA or simply use bash to webinstall, etc. Therefore, my question is, what do you use to measure this? 😅 Thanks!

lildude:
> Consider how many people are interested in Emacs Lisp, so I think the bar is a bit too high.

It's impossible to measure interest. We rely on usage as it can be measured, to a certain extent.

> And, I am not sure how accurate the measurement is since the search doesn't list out all repos (not even my own repos...).

Unfortunately it won't. Search only returns repos that have been recently indexed. At some point untouched repos drop out of the index which does affect our measurements, but unfortunately we have nothing else to quantify usage. If a repo is in active use, it should be indexed and appear in the search results.

> If you look at the graph from https://github.com/emacs-eask/setup-eask/network/dependents, there is a total of `515` repos that use this action (including forks).

They may use the workflow but that doesn't mean their repo contains the file and thus aren't counted.  A quick look at a few of those indicates quite a few don't contain an Eask file.

> Therefore, my question is, what do you use to measure this?

Your search query, and then I apply a bit of filtering as detailed in the issue linked in the CONTRIBUTING.md file.

jcs090218:
Okay, sounds reasonable for me to close this! Thank you for your explanation! :)

PR #7766, opened 2026-01-15
Make `get_tree_size` robust against git bombs

jorendorff:
## Description

For pathological repositories like git bombs with deeply nested tree structures, Rugged's `count_recursive` method can hang indefinitely even with a file limit, because it has to visit every tree object to find and count the blobs. See libgit2/rugged#995.

That causes `RuggedRepository#get_tree_size` to hang, so `git-linguist stats` can hang.

This PR works around that issue by using a custom count method that also applies a tree limit, so that it terminates promptly regardless of repository structure. The return value is still the blob count for normal repos, preserving existing behavior.

## Checklist:

### DIFF
--- a/lib/linguist/source/rugged.rb
+++ b/lib/linguist/source/rugged.rb
@@ -44,7 +44,31 @@ module Linguist
       end
 
       def get_tree_size(commit_id, limit)
-        get_tree(commit_id).count_recursive(limit)
+        tree = get_tree(commit_id)
+        tree_count = 0
+        count = 0
+        stack = [tree.each]
+
+        while !stack.empty?
+          begin
+            entry = stack.last.next
+          rescue StopIteration
+            stack.pop
+            next
+          end
+
+          if entry[:type] == :tree
+            tree_count += 1
+            return limit if tree_count >= limit
+            subtree = @rugged.lookup(entry[:oid])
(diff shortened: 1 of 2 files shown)


carlosmn:
This reimplements `count_recursive` but with the tree limit and the trick in the implementation to return the limit if we have too many trees. I think ideally the tree limit would be different and we could also just return 0 or some value that means unknown if we do see too many trees, but I  don't know where all this ends up getting used.

```ruby
        def get_tree_size(commit_id, limit)
          tree_count = 0
          blob_count = 0
          get_tree(commit_id).walk(:preorder) do |root, entry|
            if blob_count >= limit || tree_count >= limit
              blob_count = limit # If we have too many trees we return the limit
              raise StopIteration
            end

            case entry[:type]
            when :blob
              blob_count += 1
            when :tree
              tree_count += 1
            end

            true # go into a tree if that's what we were given
          end

          blob_count
        rescue StopIteration
          blob_count
        end
```

jorendorff:
I ended up with [this](https://github.com/github-linguist/linguist/pull/7766/changes#diff-31f1d3299206746a961dd94f6fff5ecb724a4a1fe7618477547cedeb67af17d7R46-R63). This is now ready for review.

lildude (approved):

PR #7798, opened 2026-02-11
Add `AGENTS.md`

lildude:
## Description

People are going to use AI to create PRs so lets make their life easier, and ours too, by adding an AGENTS.md file. This will hopefully reduce the amount of slop we're getting and also ensure things like our PR template are used and properly filled in.

This is agent-agnostic so should work for all agents.

## Checklist:

N/A

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,131 @@
+# AGENTS.md
+
+This file provides guidance for AI agents working with the Linguist repository.
+
+## Project Overview
+
+Linguist is a Ruby library used on GitHub.com to:
+- Detect blob languages
+- Ignore binary and vendored files
+- Suppress generated files in diffs
+- Generate language breakdown graphs
+
+It's packaged as the `github-linguist` gem and includes CLI tools: `github-linguist` and `git-linguist`.
+
+## Key Files and Directories
+
+| Path | Purpose |
+|------|---------|
+| `lib/linguist/languages.yml` | Master list of all languages, extensions, filenames, and their properties |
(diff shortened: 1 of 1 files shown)


Alhadis:
I'd love to offer my input on this, but the number of times I've used AI-powered tools can be boiled down to just two brief, pointless excursions.[^1][^2] I assume `AGENTS.md` is consumed by tooling like Copilot, yes?

While I can't comment on the efficacy of these changes as they pertain to AI, I will say that `AGENTS.md` is a neatly bulleted, concise list of of stuff most first-time submitters who TL;DR over `CONTRIBUTING.md` ought to read through instead. 😁

[^1]: The first was an attempt to pick a fight with ChatGPT by asking it if it preferred tabs or spaces (naturally expecting the latter). It gave an obsequious response explaining the pros and cons of either style, so I quickly lost interest (I was in a bad mood and wanted to beat up a robot to feel better).

[^2]: An attempt to create a stock image using Adobe's whatever-the-hell-their-GenAI-product-is-called, which—after about 6–7 different prompts paraphrasing the same thing—failed to generate an image that matched what I had in my head. It was supposed to be a VHS tape dunked in a jug of orange juice, and every angle it generated looked wrong.

lildude:
> While I can't comment on the efficacy of these changes as they pertain to AI, I will say that `AGENTS.md` is a neatly bulleted, concise list of of stuff most first-time submitters who TL;DR over `CONTRIBUTING.md` ought to read through instead. 😁

Yup, with the advantage agents are likely to do what humons don't 😁 

My next trick is to see if I can get Copilot to do an early review and possibly auto-close the cases where peeps/bots clearly don't follow instruction and don't use our template.

PR #7870, opened 2026-03-23
Detect `zz_generated.*.go` as generated

pirothat:
## Description

https://github.com/github-linguist/linguist/issues/7784

## Checklist:

    - Sample source(s):
      - https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/pkg/apis/cr/v1/zz_generated.deepcopy.go

    - Sample source(s):
    - Sample license(s):
      
  - [ ] I have added a color
    - Hex value: `#RRGGBB`
    - Rationale: 

    - Sample source(s):
    - Sample license(s):

  - Old: [URL to grammar repo]
  - New: [URL to grammar repo]

    - [URL to public discussion]


lildude:
Please state in the PR template where you got the sample from and it's license.

lib/linguist/generated.rb
```diff
@@ -337,6 +337,7 @@ def generated_postscript?
 
     def generated_go?
       return false unless extname == '.go'
+      return true if !!name.match(/(^|\/)[^\/]*_generated\..+\.go$/)
```
Alhadis: There's a much cleaner way to write this:

```suggestion
      return false unless name =~ %r{(?:^|/)[^/]*_generated\..+\.go$}
```

Alhadis (changes requested):

lib/linguist/generated.rb
pirothat: Thanks! I agree the regexp can be written more cleanly.

I updated the regexp style, but kept `return true if` rather than changing it to `return false unless`, because `generated_go?` currently detects Go generated files based on the `// Code generated ...` marker regardless of the filename.

Changing this to `return false unless` would make the filename pattern mandatory and could stop detecting existing generated Go files whose names do not match `*_generated.*.go`, for example protobuf-generated files like `api.pb.go`.

pirothat:
> Please state in the PR template where you got the sample from and it's license.

 I updated the PR template to include the sample source and license.

The sample is derived from Kubernetes' `zz_generated.deepcopy.go`, and the sample license is listed as Apache-2.0.

pirothat:
I realized almost all go file of real world project contains // Code generated lines. 

I commented on issue https://github.com/github-linguist/linguist/issues/7784#issuecomment-4427801361

lildude:
> I realized almost all go file of real world project contains // Code generated lines.

That would mean we don't need this PR then, because if the comment isn't there, it's not likely to have been generated?

pirothat:
>That would mean we don't need this PR then, because if the comment isn't there, it's not likely to have been generated?

Yes, I will close this PR.

PR #7878, opened 2026-03-26
Add `bun` and `deno` to interpreters in languages.yml for `JavaScript`

oopsio:
Added the bun and deno interpreters to languages.yml.

## Description
This PR adds 4 examples, each with the MIT license, and adds the deno and bun interpreters to languages.yml for the `JavaScript` language. The license(s) are included as comments in the code. 

## Checklist:

      - https://github.com/search?q=NOT+is%3Afork+path%3A*.js+const+function+async+export+return+await&type=code
      - https://github.com/search?q=NOT+is%3Afork+path%3A*.ts+interface+type+enum+namespace+readonly+private&type=code
    - Sample source(s):
      - https://github.com/oopsio/linguist/blob/main/samples/JavaScript/bun.js
      - https://github.com/oopsio/linguist/blob/main/samples/JavaScript/deno.js
      - https://github.com/oopsio/linguist/blob/main/samples/TypeScript/deno.ts
      - https://github.com/oopsio/linguist/blob/main/samples/TypeScript/bun.ts
    - Sample license(s):
    - (note: all are included in the same file as comments)


test/test_language.rb
```diff
@@ -207,6 +207,8 @@ def test_find_by_filename
 
   def test_find_by_interpreter
     {
+      "bun" => "JavaScript",
+      "deno" => "JavaScript",
```
lildude: 🤔 this is ignoring these two are already associated with TypeScript:

https://github.com/github-linguist/linguist/blob/537297cdae3ab05f8d5dd1c03627a5bd73707b19/lib/linguist/languages.yml#L8029-L8039

oopsio: Fixed, please review it

lildude: I wasn't thinking about removing it, I was thinking about changing it into an array with both languages. 

If you'd run the tests you'd have seen it fail with this:

```
TestLanguage#test_find_by_interpreter [test/test_language.rb:222]:
--- expected
+++ actual
@@ -1 +1 @@
-[#<Linguist::Language name=JavaScript>]
+[#<Linguist::Language name=JavaScript>, #<Linguist::Language name=TypeScript>]
```

However now I look at the test code, this would also require an update to the test as it isn't expecting more than one language with the same interpreter.

This change would be required too:

```diff
--- test/test_language.rb
+++ test/test_language.rb
@@ -217,7 +217,8 @@ def test_find_by_interpreter
       "sbcl" => "Common Lisp",
       "sclang" => "SuperCollider"
     }.each do |interpreter, language|
-      assert_equal [Language[language]], Language.find_by_interpreter(interpreter)
+      languages = Array(language).map { |lang| Language[lang] }
+      assert_equal languages, Language.find_by_interpreter(interpreter)
     end
 
     assert_equal [], Language.find_by_interpreter(nil)
```

So if we put both together, please make this change:

```diff
--- test/test_language.rb
+++ test/test_language.rb
@@ -207,6 +207,8 @@ def test_find_by_filename
 
   def test_find_by_interpreter
     {
+      "bun" => ["JavaScript", "TypeScript"],
+      "deno" => ["JavaScript", "TypeScript"],
       "ruby" => "Ruby",
       "Rscript" => "R",
       "sh" => "Shell",
@@ -217,7 +219,8 @@ def test_find_by_interpreter
       "sbcl" => "Common Lisp",
       "sclang" => "SuperCollider"
     }.each do |interpreter, language|
-      assert_equal [Language[language]], Language.find_by_interpreter(interpreter)
+      languages = Array(language).map { |lang| Language[lang] }
+      assert_equal languages, Language.find_by_interpreter(interpreter)
     end
 
     assert_equal [], Language.find_by_interpreter(nil)
```

lildude (changes requested):
Along with my inline comment, you've got test failures due to the extensionless `test/fixtures/TypeScript/main` test file. I've searched GitHub and there are only 2 other users of this that are not Linguist and it's forks.

With that in mind, please:

- remove that file
- add real-world samples for both JavaScript and TypeScript which have bun and deno shebangs into their respective `samples/` directories.

test/test_language.rb
oopsio: I have added the changes, it adds the changes you required me to and added samples of Deno and Bun with the shebangs. Also, a question, can linguist handle multiple arguments of shebang's? (example: if i have #!/usr/bin/env -S deno run, does it detect the `deno` interpreter?.) Let me know!

lildude: > Also, a question, can linguist handle multiple arguments of shebang's? (example: if i have #!/usr/bin/env -S deno run, does it detect the `deno` interpreter?.) Let me know!

Yup:

https://github.com/github-linguist/linguist/blob/537297cdae3ab05f8d5dd1c03627a5bd73707b19/lib/linguist/shebang.rb#L41-L49

lildude (changes requested):
Please update the PR body to use the "I am adding a new extension to a language." part of the [template](https://github.com/github-linguist/linguist/blob/main/.github/PULL_REQUEST_TEMPLATE.md) and fill it in as if you're adding an extension.

The important part is we need real-world samples with their original filenames, and a link to the original source location and the license for legal reasons.

Please also add equivalent samples for TypeScript too.

Alhadis:
@lildude Hold on, I thought Linguist had troubles disambiguating between multiple languages that share the same interpreter? (Or am I misremembering something?)

I ask because `osascript`, normally used for executing AppleScript programs, can also be used to execute JavaScript that runs atop macOS's Open Scripting Architecture (which is the "OSA" in `osascript`). Unlike TypeScript/JavaScript, it'd be relatively trivial to write a heuristic to disambiguate JavaScript and AppleScript.

lildude:
> @lildude Hold on, I thought Linguist had troubles disambiguating between multiple languages that share the same interpreter? (Or am I misremembering something?)

I thought the same thing but this PR prompted me to look at the code again and I found it works like all the other strategies (makes sense) and the shebang strategy will return multiple languages down to the next strategy. The only issue seems to have been that the tests didn't expect this, but they do now thanks to the changes in this PR.

lildude (approved):

PR #7880, opened 2026-03-27
Add Vespa Schema Definition language

thomasht86:
## Description

Add [Vespa Schema Definition](https://docs.vespa.ai/en/schemas.html) as a new language. Vespa `.sd` files define document schemas, ranking profiles, and indexing pipelines for the [Vespa](https://vespa.ai) search engine. The language has programming characteristics — ranking expressions, tensor operations, conditional logic, and ONNX model integration.

The grammar lives in the [vespa-engine/vespa](https://github.com/vespa-engine/vespa/) monorepo at `integration/tmgrammar/grammars/vespa-schema.tmLanguage.json`. Following the review feedback, the directory was renamed from `grammar/` to the conventional `grammars/` upstream in vespa-engine/vespa#37518, so the grammar is now picked up from a standard path and **no change to `GrammarsInNonStdPath` is required**.

The `.sd` extension is not currently claimed by any other language in linguist.

## Checklist:

    - Search results for .sd extension:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.sd+schema
    - Sample source(s):
      - `music.sd`: https://github.com/vespa-engine/vespa/blob/master/config-model/src/test/derived/music/music.sd
      - `msmarco.sd`: https://github.com/vespa-engine/vespa/blob/master/client/go/internal/cli/cmd/testdata/applications/withSource/src/main/application/schemas/msmarco.sd
      - `ranking.sd`: https://github.com/vespa-engine/vespa/blob/master/config-model/src/test/derived/globalphase_token_functions/test.sd
  - [x] I have included a syntax highlighting grammar: https://github.com/vespa-engine/vespa (`integration/tmgrammar/grammars/vespa-schema.tmLanguage.json`)
  - [x] I have added a color
    - Hex value: `#61D790`
    - Rationale: Derived from the green in the Vespa logo/branding


lib/linguist/heuristics.yml
```diff
@@ -819,6 +819,10 @@ disambiguations:
     - '\(#[\w-]+[!\?]'
     - '(?:[\)\]]\s*[\*\+\?](?:\s|$))'
     - '@[\w.-]+(?:\)\s|$)'
+- extensions: ['.sd']
+  rules:
+  - language: Vespa Schema Definition
+    pattern: '^\s*(schema|document|field|rank-profile)\b'
```
lildude: This isn't needed when there is only one language with an extension. Please remove.

lib/linguist/languages.yml
```diff
@@ -8259,6 +8259,16 @@ Verilog:
   codemirror_mode: verilog
   codemirror_mime_type: text/x-verilog
   language_id: 387
+Vespa Schema Definition:
+  type: programming
```
lildude: Schemas are generally not programming languages and are generally considered data.

 
```suggestion
  type: data
```

test/test_heuristics.rb
```diff
@@ -991,6 +991,12 @@ def test_scm_by_heuristics
     })
   end
 
+  def test_sd_by_heuristics
+    assert_heuristics({
+      "Vespa Schema Definition" => all_fixtures("Vespa Schema Definition", "*.sd")
+    })
+  end
+
```
lildude: As the heuristic isn't needed, this isn't needed either.

lildude (changes requested):
See inline comments.

lib/linguist/heuristics.yml
thomasht86: Done

thomasht86:
Thanks a lot for the quick feedback @lildude! 
I see you added "pending popularity" label, but afaict there are a lot more than the requirements?
Vespa Schemas are also most commonly in the "1 per application" - category, althought it is possible to have multiple. 
Any chance you could reconsider? 🙏

thomasht86:
ping @lildude - any pointer on what needs to be done to get this through? ☺️

lildude:
Popularity is reassessed when preparing for a new release, which happens approximately every 3 months.

thomasht86:
Hey @lildude - i assume you are aiming for a release next month? 
Could you indicate whether current popularity for this is sufficient (based on the fact that a schema typically exists only one per vespa application), and thus will be merged in time for next release? Many in the community are eager to see this. :)

thomasht86:
Ping @lildude - would be great to know if we pass the reqs here so we have time to address if not

lildude:
> Ping @lildude - would be great to know if we pass the reqs here so we have time to address if not

Popularity is still not there. I will review with each release, but I will _not_ comment each time.

lildude:
Popularity is good now, however there are now conflicts that need resolving. Please resolve the conflicts.

thomasht86:
Thanks @lildude. Conflicts should be resolved now!

lildude (changes requested):
> The grammar lives in the [vespa-engine/vespa](https://github.com/vespa-engine/vespa/) monorepo at `integration/tmgrammar/grammar/vespa-schema.tmLanguage.json`. Since the grammar directory is named `grammar/` (singular) rather than the conventional `grammars/`, `"grammar"` was added to `GrammarsInNonStdPath` in `tools/grammars/compiler/data.go`. No other submodules use a `grammar/` directory with `.json` files, so there are no false positives.

I'm not comfortable with the changes required for this. The original intent of this workaround was to handle unmaintained grammars that were problematic at the time.

All new grammars should adhere to the layout expected by the text editors they're written for, and which we look for, else there's a risk of silent breakage in future with a very real risk of someone not being around to resolve it. We've had it in the past (see the growing list at the end of the OP in https://github.com/github-linguist/linguist/issues/3924) and I want to prevent it from happening again in future by pushing for all new grammars to adhere to the requirements enforced by the compiler from the start.

thomasht86:
@lildude - changed path upstream, and updated to use standard path. Please take another look.

lildude:
Now comes the next challenge. The `.sd` extension is way more popular [_excluding_ the `schema` keyword](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.sd+NOT+schema) than including it suggesting this isn't the most popular language for this extension. Merging this PR as-is will mean all those files will be incorrectly classified as Vespa Schema Definition.

To prevent this, this PR will need to be updated to add the next most popular language for this extension with a heuristic to differentiate the two.

thomasht86:
Thanks for the quick feedback again @lildude 

After doing some digging, it seems like the next most popular language is a [Flexcat Source definition file](https://github.com/search?q=repo%3Aadtools%2Fflexcat+.sd&type=code).

IIUC, having another language defined for the extension is needed to have the heuristics run at all, instead of the extension be classified as Vespa schema straight away, if it was the only one with `.sd` extension.

So, by adding the Flexcat file heuristics, and making the Vespa `.sd` heuristics stricter, we will be able to correctly classify all Flexcat and Vespa `.sd`-files, right? 

The next question is maybe the fallback if none of the heuristics match. AFAICT there is no option to fallback to unclassified, so adding Flexcat may be the least bad option? At least we are sure that `.sd`-files not matching the vespa heuristics are not Vespa schema files. 

Adding a commit with this as proposal. 

Flexcat heuristic added from their [parser]([`src/createcatsrc.c#L499-L561`](https://github.com/adtools/flexcat/blob/38ad07cf0d2bb789826a929f4e6f7d26556414ff/src/createcatsrc.c#L499-L561))

lib/linguist/heuristics.yml
```diff
@@ -870,6 +870,13 @@ disambiguations:
     - '\(#[\w-]+[!\?]'
     - '(?:[\)\]]\s*[\*\+\?](?:\s|$))'
     - '@[\w.-]+(?:\)\s|$)'
+- extensions: ['.sd']
+  rules:
+  - language: FlexCat Source Description
+    pattern: '^##\s*(?i:rem|stringtype|shortstrings)\b'
+  - language: Vespa Schema Definition
+    pattern: '^\s*(?:schema|search|document)\s+[\w./-]+[^{\n]*\{'
```
lildude: This regex is vulnerable to ReDoS attack.

thomasht86: Nice catch - changed regex so that it is linear now.

lib/linguist/heuristics.yml
```diff
@@ -880,6 +880,13 @@ disambiguations:
     - '\(#[\w-]+[!\?]'
     - '(?:[\)\]]\s*[\*\+\?](?:\s|$))'
     - '@[\w.-]+(?:\)\s|$)'
+- extensions: ['.sd']
+  rules:
+  - language: FlexCat Source Description
+    pattern: '^##[ \t]*(?i:rem|stringtype|shortstrings)\b'
+  - language: Vespa Schema Definition
+    pattern: '^[ \t]*(?:schema|search|document)[ \t]+[\w./-]+[ \t]*(?:inherits[ \t][^{\n]*)?\{'
```
lildude: I think you've got an error here. I think the `/` in `[\w./-]+` should be escaped.

thomasht86: Thanks! Added the escape

lildude:
> After doing some digging, it seems like the next most popular language is a [Flexcat Source definition file](https://github.com/search?q=repo%3Aadtools%2Fflexcat+.sd&type=code).

🤔 I don't think it is. A [search using the heuristic you've added](https://github.com/search?q=path%3A*.sd+%2F%5E%23%23%5B+%5Ct%5D*%28%3Fi%3Arem%7Cstringtype%7Cshortstrings%29%5Cb%2F&type=code) returns ~236 files which is a long way from the ~60k from my previous comment. This also doesn't meet our usage requirements.

thomasht86:
It's not supposed to be easy, is it? 😅 

Here is a summary my findings: 
I sampled 3,599 .sd files stratified across 12 size buckets. 
Most of them are generated output, not a language: 
- ~5.4k GNU binutils ld/gas testsuite section dumps
- ANSYS AEDT solver output (numbers only)
- MPEG indexes, PDB ligand lists, git-lfs pointers and fstab dumps.

The only thing that could be a "competitor" is a proprietary 4GL screen-definition format. ~2.8k files, but all 300 of 300 sampled hits belong to one user, davidlpotter59, across 11 repos. Filtering out this user, I found none other of this type.

So Vespa at ~2.8k is definitely the largest language-like user of .sd. 
A fallback language accounting for the bulk of the 57k can't be met by any language, because the 57k isn't source code.
The Flexcat Source definition files seem to be the next most popular language-like format, if you filter out generated output files and files that only exists from a single user. 

What would the best solution to this look like to you @lildude ?

lildude:
> It's not supposed to be easy, is it? 😅

I was hoping it would be easier than it's turned out 😬 

Given your findings, I think it might be best to remove the Flexcat changes add `.sd` to the `generic.yml` and keep just the Vespa heuristic. This will allow the identification purely based on the heuristic and leave all other `.sd` files unclassified.

thomasht86:
Thank you - sounds reasonable!
Done in latest commit.

lildude (approved):

PR #7911, opened 2026-04-10
Add PowerPC Assembly language

RoyalAce22:
Add PowerPC Assembly as a recognized language.

## Description

PowerPC assembly shows up in the Linux kernel (`arch/powerpc/`), FreeBSD, GCC, glibc, LLVM test suites, and the PS3/Wii/GC homebrew scene. Currently there's no way to distinguish it from other assembly on GitHub.

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.S+mflr+blr
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.s+lis+stw+bdnz
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.S+path%3Apowerpc
    - Sample source(s):
      - `cpu_idle_power.S` -- from FreeBSD [`sys/powerpc/powerpc/cpu_subr64.S`](https://github.com/freebsd/freebsd-src/blob/main/sys/powerpc/powerpc/cpu_subr64.S)
      - `exception_handler.S`, `spinlock.s`, `strcpy.s`, `altivec_memset.ppc` -- written for this PR
    - Sample license(s): BSD-2-Clause (FreeBSD), MIT (rest)
  - [x] I have included a syntax highlighting grammar: https://github.com/RoyalAce22/VSCode-PowerPC-Syntax
  - [x] I have added a color
    - Hex value: `#0f62fe`
    - Rationale: IBM Blue 60 from the Carbon Design System

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -4871,7 +4871,7 @@ Motoko:
   language_id: 202937027
 Motorola 68K Assembly:
   type: programming
-  color: "#005daa"
+  color: "#0f62fe"
   group: Assembly
   aliases:
   - m68k
@@ -5948,6 +5948,21 @@ PowerBuilder:
   tm_scope: source.powerbuilder
   ace_mode: text
   language_id: 292
+PowerPC Assembly:
+  type: programming
+  color: "#0f62fe"
+  group: Assembly
+  aliases:
+  - ppc
+  - ppc-asm
+  - powerpc
+  extensions:
+  - ".ppc"
+  - ".S"
+  - ".s"
+  tm_scope: source.asm
+  ace_mode: text
+  language_id: 798592640
 PowerShell:
   type: programming
   color: "#012456"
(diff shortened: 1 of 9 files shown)


lib/linguist/languages.yml
```diff
@@ -4871,7 +4871,7 @@ Motoko:
   language_id: 202937027
 Motorola 68K Assembly:
   type: programming
-  color: "#005daa"
+  color: "#0f62fe"
```
lildude, line 4874: Why are you changing this colour? We don't change language colours without full public support from the language's community. See the CONTRIBUTING.md file.

Please revert this change.

lildude (changes requested):
See inline comment.

Additionally, when you ran `script/add-grammar` a copy of the grammar's licence should have been saved to a yaml file. We need that file too.

RoyalAce22:
Closing this. After further research, .ppc  on its own doesn't have enough real-world usage on GitHub to meet the threshold. Most .ppc files are Makefiles and configs, not ASM. The extension people actually use for PowerPC assembly is .S/.s, on which I'm having issues w/ heuristics against the existing Assembly/Unix Assembly/Motorola 68K rules, and that's a bit more complicated than I wish to deal with right now. The Motorola color change wasn't intentional. #005daa was my original color choice and a simple find/replace #005daa -> #0f62fe collided with Motorola's color. My mistake.

PR #7929, opened 2026-04-24
Improve semantic correctness and consistent presentation in the `ReadMe`.

beedell-roke:
## Description

The WCAG AA requires adherence to the chosen markup language. That's GHFM. GHFM necessarily requires adherence to CommonMark, which is essentially 1:1 with the WHATWG HTML LS for the purposes of this patch.

Consequently:

- I added a `\n` between `li`s, such that they become `li > p`s. This matters with GHFM's CSS3, because they don't wrap well otherwise.

- I added shebangs for `pre > code`s compliant with IEEE Standard 1003.1-2017 Shell Command Language, and/or its superordinate implementations. This matters because `pwsh` (and similar) shan't act as `bash` does.

- I added periods at the end of the CLI usage `li`s' sentences, *because* they are sentences.

- I removed erroneously-added `\n`s at the start end end of some `pre > code`s, because they don't render on GitHub. Having the documentation render differently when local and not is undesirable.

This is solely *partially* so that I can acquire collaborator status, thereby allowing me to ask for [`issues/7637`](https://github.com/github-linguist/linguist/issues/7637#event-24566660360) to be re-closed as a duplicate of [`github/markup/issues/2000`](https://github.com/github/markup/issues/2000#issuecomment-4311404928)! 😆

## Checklist:

Ironically, I've avoided incurring [`github/markup/issues/1857`](https://github.com/github/markup/issues/1857#issue-2456610906) for the undermentioned:

<blockquote>

- [ ] **I am adding a new extension to a language.**

</blockquote>

I have actually read all of these, but due to what [`orgs/community/discussions/19199#discussioncomment-10371423`](https://github.com/orgs/community/discussions/19199#discussioncomment-10371423) describes, cannot indicate that they are inapplicable. Consequently, I have removed their dependent subordinates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,11 @@ This library is used on GitHub.com to detect blob languages, ignore binary or ve
 ## Documentation
 
 - [How Linguist works](/docs/how-linguist-works.md)
+
 - [Change Linguist's behaviour with overrides](/docs/overrides.md)
+
 - [Troubleshooting](/docs/troubleshooting.md)
+
 - [Contributing guidelines](CONTRIBUTING.md)
 
 ## Installation
@@ -18,6 +21,7 @@ This library is used on GitHub.com to detect blob languages, ignore binary or ve
 Install the gem:
 
 ```bash
+#!/usr/bin/env bash
 gem install github-linguist
 ```
 
@@ -27,28 +31,37 @@ Linguist is a Ruby library so you will need a recent version of Ruby installed.
 There are known problems with the macOS/Xcode supplied version of Ruby that causes problems installing some of the dependencies.
 Accordingly, we highly recommend you install a version of Ruby using Homebrew, `rbenv`, `rvm`, `ruby-build`, `asdf` or other packaging system, before attempting to install Linguist and the dependencies.
 
-Linguist uses [`charlock_holmes`](https://github.com/brianmario/charlock_holmes) for character encoding and [`rugged`](https://github.com/libgit2/rugged) for libgit2 bindings for Ruby.
-These components have their own dependencies.
+Linguist uses [`charlock_holmes`](https://github.com/brianmario/charlock_holmes) for character encoding and [`rugged`](https://github.com/libgit2/rugged) for `libgit2` bindings for Ruby.
+These components have their own dependencies:
 
-1. charlock_holmes
-    * cmake
-    * pkg-config
+1. `charlock_holmes`
+
+    * `cmake`
+
+    * `pkg-config`
+
     * [ICU](http://site.icu-project.org/)
-    * [zlib](https://zlib.net/)
-2. rugged
-    * [libcurl](https://curl.haxx.se/libcurl/)
+
+    * [`zlib`](https://zlib.net/)
+
+3. `rugged`
+
+    * [`libcurl`](https://curl.haxx.se/libcurl/)
+
     * [OpenSSL](https://www.openssl.org)
 
 You may need to install missing dependencies before you can install Linguist.
 For example, on macOS with [Homebrew](http://brew.sh/):
 
 ```bash
+#!/usr/bin/env bash
 brew install cmake pkg-config icu4c
 ```
 
 On Ubuntu:
 
 ```bash
+#!/usr/bin/env bash
 sudo apt-get install build-essential cmake pkg-config libicu-dev zlib1g-dev libcurl4-openssl-dev libssl-dev ruby-dev
 ```
 
@@ -72,8 +85,9 @@ project.languages      #=> { "Ruby" => 119387 }
 
 The `github-linguist` executable operates in two distinct modes:
 
-1. **[Git Repository mode](#git-repository)** - Analyzes an entire Git repository (when given a directory path or no path)
-2. **[Single file mode](#single-file)** - Analyzes a specific file (when given a file path)
+1. **[Git Repository mode](#git-repository)** – Analyzes an entire Git repository (when given a directory path or no path).
+
+2. **[Single file mode](#single-file)** – Analyzes a specific file (when given a file path).
 
 #### Git Repository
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,28 +31,37 @@ Linguist is a Ruby library so you will need a recent version of Ruby installed.
...
-Linguist uses [`charlock_holmes`](https://github.com/brianmario/charlock_holmes) for character encoding and [`rugged`](https://github.com/libgit2/rugged) for libgit2 bindings for Ruby.
-These components have their own dependencies.
+Linguist uses [`charlock_holmes`](https://github.com/brianmario/charlock_holmes) for character encoding and [`rugged`](https://github.com/libgit2/rugged) for `libgit2` bindings for Ruby.
+These components have their own dependencies:
+
+1. `charlock_holmes`
+
+    * `cmake`
+
+    * `pkg-config`
```
Alhadis, line 42: The backticks you've added to the library names can stay, but the extraneous newlines you've inserted need to be removed.

```suggestion
Linguist uses [`charlock_holmes`](https://github.com/brianmario/charlock_holmes) for character encoding and [`rugged`](https://github.com/libgit2/rugged) for `libgit2` bindings for Ruby.
These components have their own dependencies:

1. `charlock_holmes`
    * `cmake`
    * `pkg-config`
```

README.md
```diff
@@ -27,28 +31,37 @@ Linguist is a Ruby library so you will need a recent version of Ruby installed.
...
-    * [zlib](https://zlib.net/)
-2. rugged
-    * [libcurl](https://curl.haxx.se/libcurl/)
+
+    * [`zlib`](https://zlib.net/)
+
+3. `rugged`
+
+    * [`libcurl`](https://curl.haxx.se/libcurl/)
+
```
Alhadis, line 50: See above.

```suggestion
    * [`zlib`](https://zlib.net/)
3. `rugged`
    * [`libcurl`](https://curl.haxx.se/libcurl/)
```

README.md
```diff
@@ -27,28 +31,37 @@ Linguist is a Ruby library so you will need a recent version of Ruby installed.
...
     * [OpenSSL](https://www.openssl.org)
 
 You may need to install missing dependencies before you can install Linguist.
 For example, on macOS with [Homebrew](http://brew.sh/):
 
 ```bash
+#!/usr/bin/env bash
```
Alhadis, line 57: Remove this.

README.md
```diff
@@ -27,28 +31,37 @@ Linguist is a Ruby library so you will need a recent version of Ruby installed.
...
 brew install cmake pkg-config icu4c
 ```
 
 On Ubuntu:
 
 ```bash
+#!/usr/bin/env bash
```
Alhadis, line 64: Remove this.

README.md
```diff
@@ -72,8 +85,9 @@ project.languages      #=> { "Ruby" => 119387 }
 
 The `github-linguist` executable operates in two distinct modes:
 
-1. **[Git Repository mode](#git-repository)** - Analyzes an entire Git repository (when given a directory path or no path)
-2. **[Single file mode](#single-file)** - Analyzes a specific file (when given a file path)
+1. **[Git Repository mode](#git-repository)** – Analyzes an entire Git repository (when given a directory path or no path).
+
+2. **[Single file mode](#single-file)** – Analyzes a specific file (when given a file path).
```
Alhadis, line 90: The full-stops you've added can stay, but the superfluous line-break needs to go.

```suggestion
1. **[Git Repository mode](#git-repository)** – Analyzes an entire Git repository (when given a directory path or no path).
2. **[Single file mode](#single-file)** – Analyzes a specific file (when given a file path).
```

Alhadis (changes requested):
> The WCAG AA requires adherence to the chosen markup language. That's GHFM.

No, that's _HTML_. GHFM (the correct acronym of which is “GFM”, BTW) is a superset[^1] of a standardised form of Markdown, a lightweight markup language chiefly used to generate HTML output. The WCAG has nothing to do with Markdown.

[^1]: https://github.github.com/gfm/#what-is-github-flavored-markdown%2D

> GHFM necessarily requires adherence to CommonMark, which is essentially 1:1 with the WHATWG HTML LS for the purposes of this patch.

Erh, no it's not. CommonMark uses an [explicitly versioned](https://spec.commonmark.org/current/) specification, while the WHATWG adopts a "rolling updates" approach (hence the term, Living Standard). In any case, the comparison is an apples-to-oranges scenario.

> I added a `\n` between `li`s, such that they become `li > p`s. This matters with GHFM's CSS3, because they don't wrap well otherwise.

I've no idea what you're referring to here. The existing lists soft-wrap perfectly fine, and the only thing you've done by converting `li > text` to `li > p > text` is add unnecessary vertical space to an already overly-long page.

> I added shebangs for `pre > code`s compliant with IEEE Standard 1003.1-2017 Shell Command Language, and/or its superordinate implementations. This matters because `pwsh` (and similar) shan't act as `bash` does.

Interpreter directives (otherwise known as "hashbangs" or "shebangs") aren't specified by POSIX (and thus, IEEE 1003.1-2017), but even if they were, adding them makes no sense in the context of the code-blocks you've modified. These are intended to be single-line, copy+pastable terminal commands for the reader's benefit, not isolated executables. Moreover, they're simple enough that both Bourne-style shells (like <samp>bash(1)</samp> and <samp>zsh(1)</samp>) _and_ non POSIX-compliant shell languages (like PowerShell/`pwsh`) will interpret them correctly.

(It's also odd to bring up other command interpreters like PowerShell when the hashbangs you've added specify `bash`, but whatever).

> I added periods at the end of the CLI usage `li`s' sentences, _because_ they are sentences.

In this case, you're right, but list-items themselves don't always merit a full-stop (such as when they contain fragments instead of sentences).

> I removed erroneously-added `\n`s at the start end end of some `pre > codes`, because they don't render on GitHub. Having the documentation render differently when local and not is undesirable.

Well, it's going to render differently anyway. Markdown footnotes are supported on GitHub,[^2] but they aren't part of the GFM specification itself. So concerns over how Linguist's readme gets rendered elsewhere is a moot point.

Instead, you could've simply said you <i><q>removed extraneous empty lines in code-blocks</q></i>, which by itself would've been adequate.

[^2]: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes

PR #7930, opened 2026-04-25
Add .MODULE.bazel extension to Starlark

auscompgeek:
## Description
Bazel, in some circumstances, supports splitting a `MODULE.bazel` file into [multiple `*.MODULE.bazel` files](https://bazel.build/external/faq#why-does-module-bazel-not-support-loads) with an `include()` directive.

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.MODULE.bazel
    - Sample source(s):
      - https://github.com/GoogleContainerTools/distroless/blob/517026aa21cf5ce3a6e27f224cc93265070e3a75/private/repos/deb/deb.MODULE.bazel
    - Sample license(s):
      - [Apache-2.0](https://github.com/GoogleContainerTools/distroless/blob/517026aa21cf5ce3a6e27f224cc93265070e3a75/LICENSE)

### DIFF
--- a/lib/linguist/languages.yml
+++ b/lib/linguist/languages.yml
@@ -7500,6 +7500,7 @@ Starlark:
   extensions:
   - ".bzl"
   - ".star"
+  - ".MODULE.bazel"
   filenames:
   - BUCK
   - BUILD
--- a/samples/Starlark/deb.MODULE.bazel
+++ b/samples/Starlark/deb.MODULE.bazel
@@ -1,0 +1,40 @@
+"debian dependencies"
+
+REPOS = [
+    "trixie",
+    "trixie_java",
+    "trixie_adoptium",
+    "trixie_python",
+    "bookworm",
+    "bookworm_python",
+]
+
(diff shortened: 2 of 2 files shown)


lib/linguist/languages.yml
```diff
@@ -7500,6 +7500,7 @@ Starlark:
   extensions:
   - ".bzl"
   - ".star"
+  - ".MODULE.bazel"
```
auscompgeek, line 7503: I believe both `.bzl` and `.star` should both be considered the "primary" extensions for Starlark (`.star` being the tool-agnostic extension, `.bzl` being widely used in Bazel), so I'm adding this after both per the contributing docs.

lildude: You can only have one primary and it should be first. The rest need to be in alphabetical order.

lildude:
I've just realised this is a dupe of https://github.com/github-linguist/linguist/pull/7624. Closing in favour of that PR as it doesn't have failures.

PR #7960, opened 2026-05-14
Mark Julia package manifests (Manifest.toml) as generated

ararslan:
## Description

Resolved package environments for the Julia language are generated by the package manager and captured in TOML files. They can be large and distract from more meaningful changes when reviewing PRs.

Julia's package manager produces a Manifest.toml file when the active project file is called Project.toml, and it produces a JuliaManifest.toml if the project file is called JuliaProject.toml. Manifests resolved with different Julia versions can live alongside one another, differentiated by a versioned suffix. Such files are not produced automatically; a generated manifest must be manually renamed to include a version suffix. In that case, it is still a generated file, despite the manual intervention.

The first line of these files is typically:
```toml
# This file is machine-generated - editing it directly is not advised
```
However, using the filename is much simpler than checking for this text. (Also note that these files would almost certainly never be written by hand.)

## Checklist:

### DIFF
--- a/lib/linguist/generated.rb
+++ b/lib/linguist/generated.rb
@@ -93,6 +93,7 @@ module Linguist
       gradle_wrapper? ||
       maven_wrapper? ||
       mise_lock? ||
+      julia_manifest? ||
       generated_go? ||
       generated_protocol_buffer_from_go? ||
       generated_protocol_buffer? ||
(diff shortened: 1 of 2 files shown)


lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #6565, opened 2023-10-05
Add language: liquidsoap

toots:
## Description

This PR adds support for the liquidsoap language. The language has been existing since ~2005 and is widely used to run media streaming applications. Although its original scope is specialized, the language itself if a general-purpose scripting language that is functional and statically typed with inferred typed.

## Checklist:
- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.liq
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - liquidsoap standard library: https://github.com/savonet/liquidsoap/tree/main/src/libs
    - Sample license(s): GPL 2.0 or later
  - [x] I have included a syntax highlighting grammar: https://github.com/savonet/vscode-liquidsoap
  - [x] I have added a color
    - Hex value: `#990066`
    - Rationale: Logo color. Logo has been used for probably 15 to 20 years.
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.
    - these extensions are unused by other languages at this time

PR #7966, opened 2026-05-17
Move fixtures added in #6565 to samples directory

Alhadis:
## Description
This pull-request moves the Liquidsoap (`*.liq`) files added in #6565 from Linguist's root directory to a named subdirectory under `samples/` like we require.

(Not sure how they managed to escape the confines of the `samples/` folder… 🤔)

_Template removed as it doesn't apply._


lildude (approved):
🤔 not sure how I missed this either. Thanks.

PR #7978, opened 2026-05-23
Allows custom language to be declared in the .gitattributes

andrey-moura:
# Allows custom language to be declared in the .gitattributes

## Description

Before the requested changes, adding a language in the .gitattributes which is not supported by linguist make it not appear.

.gitattributes:
`*.lv linguist-language=Lavi`

Running with `bundle exec bin/github-linguist lavi`:
```
96.96%  401776     C++
1.87%   7729       CMake
0.54%   2242       Shell
0.30%   1246       HTML
0.23%   960        Dockerfile
0.07%   303        C
0.01%   54         Ruby
0.01%   54         Python
```

Now with the proposed changes:
```
90.67%  401776     C++
6.49%   28749      Lavi
1.74%   7729       CMake
0.51%   2242       Shell
0.28%   1246       HTML
0.22%   960        Dockerfile
0.07%   303        C
0.01%   54         Ruby
0.01%   54         Python
```

I also had to add support to the `color` attribute to be override:
`*.lv linguist-color=ba5186`
This was needed because the language has a color to be displayed in the Languages section of the repository.

## Checklist:

I will need help with the tests.
Please, tell me if there's any more change needed.

### DIFF
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -4,7 +4,7 @@ Linguist supports a number of different custom override strategies for language 
 
 ## Using gitattributes
 
-Add [a `.gitattributes` file](https://git-scm.com/docs/gitattributes) to your project and use standard git-style path matchers for the files you want to override using the `linguist-documentation`, `linguist-language`, `linguist-vendored`, `linguist-generated`  and `linguist-detectable` attributes.
+Add [a `.gitattributes` file](https://git-scm.com/docs/gitattributes) to your project and use standard git-style path matchers for the files you want to override using the `linguist-documentation`, `linguist-language`, `linguist-vendored`, `linguist-generated` , `linguist-detectable` and `linguist-color` attributes.
 `.gitattributes` will be used to determine language statistics and will be used to syntax highlight files.
 You can also manually set syntax highlighting using [Vim or Emacs modelines](#using-emacs-or-vim-modelines).
 
--- a/lib/linguist/language.rb
+++ b/lib/linguist/language.rb
@@ -324,10 +324,10 @@ module Linguist
     # Returns a type Symbol or nil.
     attr_reader :type
 
-    # Public: Get color.
+    # Public: Get color or set color.
     #
     # Returns a hex color String.
-    attr_reader :color
+    attr_accessor :color
 
     # Public: Get aliases
     #
--- a/lib/linguist/lazy_blob.rb
+++ b/lib/linguist/lazy_blob.rb
@@ -9,7 +9,8 @@ module Linguist
                 'linguist-language',
                 'linguist-vendored',
                 'linguist-generated',
-                'linguist-detectable']
+                'linguist-detectable',
+                'linguist-color']
 
     # DEPRECATED: use Linguist::Source::RuggedRepository::GIT_ATTR_OPTS instead
     GIT_ATTR_OPTS = Linguist::Source::RuggedRepository::GIT_ATTR_OPTS
@@ -74,6 +75,10 @@ module Linguist
 
       @language = if lang = git_attributes['linguist-language']
         detected_language = Language.find_by_alias(lang) || Language.create(name: lang, type: :programming)
+
+        if color = git_attributes['linguist-color']
+          detected_language.color = color
+        end
 
         # If strategies are being tracked, get the original strategy that would have been used
         if detected_language && Linguist.instrumenter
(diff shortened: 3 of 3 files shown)


Alhadis:
GitHub doesn't support "custom" (user-defined) languages, and this is both a technical limitation and a matter of site policy. Linguist, by extension, has no means of supporting such a feature.

This feature has been requested in the past many times; see #4435 for one such discussion.

Closing this pull-request as there's nothing more to do here.

andrey-moura:
#4435

andrey-moura:
@Alhadis, please don't get me wrong, I'm not trying to start a discussion here if the feature I'm suggesting is good or not for the majority of users. I only want to confirm that my PR has been classified correctly and I will not complain any more after the next response. I read #4435 and I'm not sure if it's exactly the same problem, the problem I saw against such feature was about defining custom highlighting or replacing already defined languages (as mentioned the variations of Fortran). My PR does not deal with highlighting, It only allows unknown (not defined in linguist, if defined it does not override) languages to have their line of code counted.

The #4435 describe the problem as:

> Different implementations or specifications (specifically major language versions) of the same language may sometimes result in incompatibilities such as different syntax, new/deprecated language features, functions, builtin libraries, etc.
...
The repository could declare which exact versions or implementations of a language are being used in what files/directories to clarify and monitor that in the language statistics, therefore the problem is cosmetic in nature.

But the description of my problem is:
The linguist ignores lines of code written in unknown languages in the counting of lines of code, even when defined in the `.gitattributes`.

Today, when you have a language in your repo which linguist does not knows, it simply ignore it in the counting, even if you specify in the `.gitattributes` as shown here:

.gitattributes:
`*.lv linguist-language=Lavi`

Running with `bundle exec bin/github-linguist lavi`:
```
96.96%  401776     C++
1.87%   7729       CMake
0.54%   2242       Shell
0.30%   1246       HTML
0.23%   960        Dockerfile
0.07%   303        C
0.01%   54         Ruby
0.01%   54         Python
```
Note that Lavi has not been shown here, even if it's defined in the `.gitattributes`. It looks like a bug. In example, if the repo is full of Lavi code and some small C++ code, it will display C++ with 100%, let's say, 100 lines of code, when the the repo has thousands of lines of Lavi, making the repo looks smaller than it is.

After making the changes in the pull request, the result it:

```
90.67%  401776     C++
6.49%   28749      Lavi
1.74%   7729       CMake
0.51%   2242       Shell
0.28%   1246       HTML
0.22%   960        Dockerfile
0.07%   303        C
0.01%   54         Ruby
0.01%   54         Python
```

Note how:
- The total line count now is bigger
- All languages have decreased in % but not in line count

We:

- Do not override the label of already defined language
- Do not create custom highlight of language variations
- Do not modify the behavior of linguist across other repos, only the one with `.gitattributes` which asked to have this behavior

Please, confirm if "linguist ignores lines of code written in unknown languages in the counting of lines of code, even when defined in the `.gitattributes`" is the correct behavior or not.

If not, then my PR solves the problem.

If it is, then I have a new suggestion, to allow declaring languages in the .gitattributes and display all 'Other', showing the true line count:

.gitattributes:
`*.lv linguist-language=Lavi`

```
90.67%  401776     C++
6.49%   28749      Other
1.74%   7729       CMake
0.51%   2242       Shell
0.28%   1246       HTML
0.22%   960        Dockerfile
0.07%   303        C
0.01%   54         Ruby
0.01%   54         Python
```

Glad for your attention

lildude:
This has come up in several other issues over the years beyond the one @Alhadis has referenced (search our issues). Things have changed a lot in GitHub over the years so some of the previously explained limitations, explanations and rejections don't apply anymore, but a new one does.

Whilst this works in the isolated CLI environment, GitHub won't be able to take advantage of this. Over the years, GitHub has slowly split into micro-services, and pretty much every one of them relies purely on the language ID to name mapping once the initial (very expensive) analysis has been done on the git servers when code is pushed. This solution does not result in such a mapping so can't be used, and nothing other than the git servers, have access to the `.gitattributes` file (a deliberate separation of responsibilities design decision) so that can't be used as a fallback either.

lildude:
I should also add that it's a deliberate design decision that unknown languages are not considered in the LoC or percentages as without knowledge of the language, there's no way to differentiate between comments in that language and actual code. This would lead to more questions than answers. GitHub doesn't have a concept of a repo language, only an understanding of identified languages, and their prominence within that repo.

PR #7985, opened 2026-05-27
Add support for Valve Map Format (`.vmf`) files

meowcat767:
## Description
The Valve Map Format (VMF) is a data language used for Valve Source engine maps created in the Hammer Editor. It stores map geometry, brush data, and entity properties.

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*vmf
    - Sample source(s):
      - Internal sample created from standard Valve SDK map layout.
    - Sample license(s):
      - MIT / Public Domain
  - [x] I have included a syntax highlighting grammar: https://github.com/meowcat767/vmf-grammar
  - [x] I have added a color
    - Hex value: `#f79a10`
    - Rationale: This is the iconic orange brand color globally associated with the Source engine.


lildude (changes requested):
This PR is incomplete. You have not followed all of the instructions in the CONTRIBUTING.md file. 

We also can not accept this grammar either as it doesn't have a licence. We can only accept grammars with one of the licenses linked in the CONTRIBUTING.md file.

grammars.yml
```diff
@@ -1203,6 +1203,8 @@ vendor/grammars/typst-grammar:
 - source.typst
 vendor/grammars/verilog.tmbundle:
 - source.verilog
+vendor/grammars/vmf-grammar:
+- source.keyvalues
```
lildude: I see your grammar is very new. You should make this more specific as this is very generic and very likely to conflict with a grammar that really deals with legitimate key/value data which your samples don't seem to do.

meowcat767:
I have added a license (MIT) to the grammar, but please let me know what I have not followed in the CONTRIBUTING.MD as I am confused on that!

lildude (changes requested):
I've also just noticed your samples a huge! If they suppressed in the diff, they're too big. Please replace them with smaller, real world samples.

lildude:
> I have added a license (MIT) to the grammar, but please let me know what I have not followed in the CONTRIBUTING.MD as I am confused on that!

You've not generated the language ID as per step 4. You will also need to re-add your grammar to pull in the licence (see the instructions for replacing a grammar and replace your grammar with the same URL).

meowcat767:
> I've also just noticed your samples a huge

Yeah, VMF files are reasonably large. I can make smaller ones, but they won't reflect real maps in real projects.

I'm currently fixing the problems with the grammar and will also re-gen the IDs.

lildude:
I've just noticed we've already got a "Valve Data Format". How does this differ from a syntax perspective? Would this be better suited as adding this extension to the existing language? It's a data format so won't appear in the stats by default and if it's identical syntax, it saves you having to maintain a grammar.

meowcat767:
Valve Data Format is used to store game related metadata.  Valve Data Format files store information data for resources, installation scripts, configuration scripts, and visualization elements.

Valve Map Format files are the actual map source code used by Valve Hammer and VBSP, VVIS, and VRAD to turn the map into a BSP.

They have very different grammar, so using the VDF grammar would not be suitable for VMF.

meowcat767:
I have updated the IDs for VMF.

lildude (changes requested):
You've still not replaced the massive samples with smaller files nor re-added your grammar so that the license file can be cached.

meowcat767:
I have removed the large VMF and kept the small one (this still has the basics of a VMF). 

I have cached my grammar license as well.

meowcat767:
Any updates?

lildude:
Your sample is still too big. As I said before, if it’s suppressed in the diff, it’s too big.

meowcat767:
I have made a smaller VMF file that is just a box brush.

vendor/README.md
```diff
@@ -663,7 +663,8 @@ This is a list of grammars that Linguist selects to provide syntax highlighting
 - **VCL:** [fastly/vscode-fastly-vcl](https://github.com/fastly/vscode-fastly-vcl)
 - **VHDL:** [VHDL-LS/rust_hdl_vscode](https://github.com/VHDL-LS/rust_hdl_vscode)
 - **Vala:** [technosophos/Vala-TMBundle](https://github.com/technosophos/Vala-TMBundle)
-- **Valve Data Format:** [Nixinova/NovaGrammars](https://github.com/Nixinova/NovaGrammars)
+- **Valve Data Format:** [meowcat767/vmf-grammar](https://github.com/meowcat767/vmf-grammar)
```
lildude: Why are you changing this? 

You've previously stated:

> They have very different grammar, so using the VDF grammar would not be suitable for VMF.

The fact you're now replacing the VDF grammar with your own suggests this is not the case and that VMF and VDF files are really using the same underlying language and as such this PR should really be about adding this extention to the current entry.

meowcat767: For some reason it keeps changing? I haven't touched that file manually. Will fix now.

I'm sorry about how long this is taking...

meowcat767:
I have fixed the vendor README

lildude (changes requested):
You shouldn't be manually changing the files modified by the scripts provided by the repo as by doing so you've now made this PR invalid as we already have a language with the `source.keyvalues` TextMate scope… VDF. We can't have two languages with the same scope but different grammars. Using the scripts to add/update the grammars would have highlighted this.

This reminded me to look at the [PR](https://github.com/github-linguist/linguist/pull/5330) that added VDF. That PR included a link to https://developer.valvesoftware.com/wiki/KeyValues. The [VMF page](https://developer.valvesoftware.com/wiki/KeyValues) in the same wiki states:

> The Valve Map Format (VMF) is a plain-text file format that stores raw (pre-compile) map data, used by the [Valve Hammer Editor](https://developer.valvesoftware.com/wiki/Valve_Hammer_Editor) (since version 4.0) to save production-stage maps and [prefabs](https://developer.valvesoftware.com/wiki/Prefab). It contains information on all map brushes and entities in the [KeyValues](https://developer.valvesoftware.com/wiki/KeyValues) format stored with a ".vmf" extension.

So VDF and VMF do indeed use the same underlying language.

I'm also reminded of a previous attempt to add VMF in https://github.com/github-linguist/linguist/pull/5413 and I'm going to stick with my [decision](https://github.com/github-linguist/linguist/pull/5413#pullrequestreview-2101255556) in that PR here.

The VDF PR originally wanted to add the language as "KeyValues" as that's the language used for those files according to the wiki, but this was deemed too generic, hence "Valve Data Format". It think "Valve Data Format" is specific, yet generic enough, to cover these VMF files too.

Please update the PR to add `.vmf` as an extension to the current "Valve Data Format" entry. It's a data language so won't appear in the repo stats anyway and most users won't change this. They will however get the benefit of syntax highlighting.

Sorry to do this, but I didn't immediately remember we've effectively gone through this whole discussion before.

meowcat767:
I have added VMF as a alias to VDF and moved the sample accordingly.

lildude (changes requested):
You need to remove the grammar and related changes too.

meowcat767:
I have removed the grammar for VMF.

.gitmodules
```diff
@@ -1352,6 +1352,9 @@
 [submodule "vendor/grammars/verilog.tmbundle"]
 	path = vendor/grammars/verilog.tmbundle
 	url = https://github.com/textmate/verilog.tmbundle
+[submodule "vendor/grammars/vmf-grammar"]
+	path = vendor/grammars/vmf-grammar
+	url = https://github.com/meowcat767/vmf-grammar.git
```
lildude: You need to remove this too.

meowcat767: Done!

lildude (approved):

PR #7986, opened 2026-05-28
Add `.slnLaunch` to JSON extensions

brunt-toast:
.slnLaunch is a json-formatted file used by Visual Studio to create launch profiles for combinations of .NET projects with specific configurations and architectures.

As indicated by [Microsoft Learn - Set multiple startup projects](https://learn.microsoft.com/en-us/visualstudio/ide/how-to-set-multiple-startup-projects) and Visual Studio's own behaviour when creating a launch profile via GUI, the canonical format is ".slnLaunch", with the capital "L", however Visual Studio will also recognise the file with a lower-case L. In fact, Visual Studio matches case-insensitively, but `.*\.sln[Ll]aunch` covers most reasonable cases. 

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.slnLaunch
    - Sample source(s):
      - https://github.com/dotnet/samples/blob/c6eb3268ba912447e820ba78e3d35ee2c5d7e0b0/orleans/ChatRoom/ChatRoom.slnLaunch
    - Sample license(s):
      - CC-BY-4.0 

(Above: N/A, as far as I am aware there is no other use for .slnLaunch)


lib/linguist/languages.yml
```diff
@@ -3507,6 +3507,8 @@ JSON:
   - ".jsonl"
   - ".mcmeta"
   - ".sarif"
+  - ".slnlaunch"
+  - ".slnLaunch"
```
Alhadis: Linguist identifies file extensions case-insensitively, so there's no need for camelCased variant of `.slnLaunch`.

```suggestion
  - ".slnlaunch"
```

Alhadis (dismissed):

lildude (approved):

PR #8011, opened 2026-06-09
Test against Ruby 3.4 and 4.0

byroot:
The dependency on bundler in the gemspec is an old deprecated pattern that cause problem on newer rubies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ '3.3', '3.2', '3.1' ]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ '3.3', '3.2', '3.1' ]
+        ruby: [ '4.0', '3.4', '3.3', '3.2', '3.1' ]
```
lildude: Thanks for this. We may as well drop 3.2 and 3.1 whilst we're at it as they're now EoL.

byroot: Done.

lildude (approved):
Thanks for this.

PR #8032, opened 2026-06-21
Add Roblox `.rbxmx` as XML

### LINKED ISSUE #7928: Mark `.rbxmx` and `.rbxlx` as XML
`.rbxmx` and `.rbxlx` are [Roblox’s XML model format](https://dom.rojo.space/xml.html) used to save places and models from Roblox in a file format. `.rbxmx` is already used in a [significant amount](https://github.com/search?q=path%3A**.rbxmx) on GitHub and should be considered for addition to Linguist. `.rbxlx` should be considered as well for being `.rbxmx`'s brother, which uses the same XML syntax.

I'm checking if the above suggestion is possible and if there are any blockers to it.

gords2:
Add `.rbxmx` and `.rbxlx` to the XML language.

## Description

`.rbxmx` (Roblox **model**) is an XML-based serialization formats produced by Roblox Studio and the wider Roblox tooling ecosystem (Rojo, etc.). It is a well-formed XML document rooted in a `<roblox …>` element with an `xsi:noNamespaceSchemaLocation` pointing at `roblox.xsd`, so they classify naturally as XML.

This PR:
- Adds `.rbxmx` to the `XML` entry in `languages.yml`, kept in case-sensitive alphabetical order (between `.qhelp` and `.rdf`).
- Adds one real-world sample for each extension.

The extension is currently associated with any other language, so no heuristic is required.

Resolves #7928.

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.rbxmx+roblox
    - Sample source(s):
      - `samples/XML/roblox-model.rbxmx` — from [rojo-rbx/rbx-test-files `models/three-unique-parts/xml.rbxmx`](https://github.com/rojo-rbx/rbx-test-files/blob/master/models/three-unique-parts/xml.rbxmx)
    - Sample license(s): MIT (sample from [rojo-rbx/rbx-test-files](https://github.com/rojo-rbx/rbx-test-files), which is MIT-licensed)
  - N/A I have included a change to the heuristics to distinguish my language from others using the same extension. *(Not applicable — neither extension is shared with another language.)*


lib/linguist/languages.yml
```diff
@@ -8853,6 +8853,8 @@ XML:
   - ".pt"
   - ".pubxml"
   - ".qhelp"
+  - ".rbxlx"
+  - ".rbxmx"
```
Alhadis:
```suggestion
  - ".rbxmx"
```

samples/XML/roblox-baseplate.rbxlx
```diff
@@ -0,0 +1,609 @@
+<roblox xmlns:xmime="http://www.w3.org/2005/05/xmlmime" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.roblox.com/roblox.xsd" version="4">
```
Alhadis: Remove this sample, as explained in the review comments.

samples/XML/roblox-model.rbxmx
```diff
@@ -0,0 +1,246 @@
+<roblox xmlns:xmime="http://www.w3.org/2005/05/xmlmime" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.roblox.com/roblox.xsd" version="4">
```
Alhadis, line 1: Please shorten this file so it falls around ~100 lines (without invalidating the well-formedness of the XML source), again, as detailed in my review.

Alhadis (changes requested):
Usage for the `.rbxlx` extension is too thin to be eligible for addition; there's currently only [108 such files](https://github.com/search?type=code&q=path%3A**%2F*.rbxlx+NOT+is%3Afork+) indexed on GitHub. The `.rbxmx` extension, OTOH, clearly checks out ([~46.2k search results](https://github.com/search?type=code&q=path%3A**%2F*.rbxmx+NOT+is%3Afork) for that extension).

In addition, the size of the samples you've added are, uh, kind of huge (and they'll slow the classifier down needlessly). Could you cull the `.rbxmx` sample down to, say, ~100 lines or so? (Making sure the source remains well-formed XML, though).

You'll also need to remove the `.rbxlx` extension from this PR (and thus also its sample), owing to the scant in-the-wild usage.

> Both are well-formed XML documents

😉 Technically, a well-formed XML document includes an XML declaration (`<?xml version="1.0"?>`), which incidentally is the reason these files aren't being recognised by Linguist's [XML header] strategy.

lib/linguist/languages.yml
gords2: Done — `.rbxlx` removed; the XML `extensions` list now carries only `.rbxmx` (between `.qhelp` and `.rdf`).

samples/XML/roblox-baseplate.rbxlx
gords2: Removed this sample along with the `.rbxlx` extension.

samples/XML/roblox-model.rbxmx
gords2: Trimmed from 246 to 82 lines (kept a single `Part`). Still well-formed XML — validated with `xmllint --noout`.

gords2:
Pushed `71c75be` addressing the review:

- Dropped the `.rbxlx` extension and its sample (`roblox-baseplate.rbxlx`) given the thin in-the-wild usage.
- Cut `roblox-model.rbxmx` from 246 → 82 lines (one `Part`), still well-formed XML.

Good call on the XML-declaration point — and that is precisely why the extension mapping is needed here: Roblox Studio writes these files without an `<?xml ?>` prolog, so the XML-header strategy never fires. Left the sample prolog-less to stay representative of real-world `.rbxmx` output.

Alhadis (dismissed):
Excellent work!

lildude (changes requested):
You need to add the files identified in the failing tests to the test at https://github.com/github-linguist/linguist/blob/f39b13030cbbe5b2dd1ab8a98fc7f65bd239146e/test/test_strategies.rb#L177 to get CI to pass.

lildude (approved):

PR #8056, opened 2026-07-05
Add B programming language

aceinetx:
Added the [B programming language](https://en.wikipedia.org/wiki/B_(programming_language)) (C predecessor)

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.b+%2F%28%3F%3Aextrn%7Cauto%7Cgoto%29%2F
    - Sample source(s):
      - https://github.com/sergev/vak-opensource/blob/master/compilers/b/brtb.b
    - Sample license(s):
      - Apache-2.0
  - [ ] I have included a syntax highlighting grammar
  - [x] I have added a color
    - Hex value: `#da7666`
    - Rationale: The color is influenced by an unofficial logo in https://github.com/bext-lang/b (B doesn't have an official logo)


DecimalTurn:
Related issue: https://github.com/github-linguist/linguist/issues/6717

DecimalTurn:
For the search results, it might be better to use the same regex as used in the heuristic to illustrate that we have reached the 2k file threshold. Eg.: https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.b+%2F%28%3F%3Aextrn%7Cauto%7Cgoto%29%2F (7.2k files)

lib/linguist/heuristics.yml
```diff
@@ -107,6 +107,11 @@ disambiguations:
   - language: LTspice Symbol
     pattern: '^SymbolType[ \t]'
   - language: Asymptote
+- extensions: ['.b']
+  rules:
+  - language: B
+    pattern: '^\s*(?:extrn|auto|goto)\b'
+    negative_pattern: '^\s*(?:implement|include|load|ref)\b'
```
DecimalTurn, line 114: Ideally, we'd also include a rule for Limbo and Brainfuck (Bf).

There's already rules inside the heuristics.yml file that we can re-use:
```yml
  - language: Brainfuck
    pattern: '(>\+>|>\+<)'
```
and
```yml
  - language: Limbo
    pattern: '^\w+\s*:\s*module\s*\{'
```

DecimalTurn:
> B doesn't have an official logo

What about this one?

image

From here: https://github.com/bext-lang/b

The color of the logo is already close enough with what is suggested here (`#da7666`), so I'm not saying it has to be changed, but at least this logo could be used as a rationale for the color.

lib/linguist/languages.yml
```diff
@@ -529,6 +529,14 @@ Awk:
...
   ace_mode: text
   language_id: 28
+B:
+  type: programming
+  color: "#da7666"
+  extensions:
+  - ".b"
+  tm_scope: none
+  ace_mode: text
+  language_id: 700792152
```
DecimalTurn, line 539: For some of these fields, like for tm_scope, it's worth considering if using the same values as the C programming language would be beneficial. You can check how the C syntax highlighing of a B file would look like using https://novalightshow.netlify.app/

aceinetx: Apart from the extrn keyword it looks pretty much the same as C

lildude (changes requested):
The samples added do not match the list in the template. They are also mostly illustrations of using the language rather than real world examples of usage of the language. Please correct this.

aceinetx:
> The samples added do not match the list in the template. They are also mostly illustrations of using the language rather than real world examples of usage of the language. Please correct this.

Sorry for the wait. I've updated the samples and ran script/list-grammars because the tests were complaining.

lildude:
> Sorry for the wait. I've updated the samples and ran script/list-grammars because the tests were complaining.

Thanks, but you've now used samples we explicitly request _not_ be be used:

https://github.com/github-linguist/linguist/blob/021d61376df532faeb3d312b8067d44ea57e885f/CONTRIBUTING.md?plain=1#L155

Please replace these with _real_ examples of the language.

aceinetx:
> > Sorry for the wait. I've updated the samples and ran script/list-grammars because the tests were complaining.
> 
> Thanks, but you've now used samples we explicitly request _not_ be be used:
> 
> https://github.com/github-linguist/linguist/blob/021d61376df532faeb3d312b8067d44ea57e885f/CONTRIBUTING.md?plain=1#L155
> 
> Please replace these with _real_ examples of the language.

I updated the examples, some of which I have written myself, some from real world code.

lib/linguist/languages.yml
lildude: Oooof. Looks like https://github.com/github-linguist/linguist/pull/7651 generated the ID when the language entry was just B and then not updated it with the rename. That PR also pre-dates the test we added to prevent incorrect IDs.

I think as this language is quite rarely used, and was only added in the last release, we can safely do this:

```suggestion
  language_id: 993355937
```

It'll mean some current files shown as this language will report as B, but this will only be until the next push to the repo.

lildude (changes requested):
We need an update to the heuristics tests as shown in the failure and updating of the language ID suggested in the inline comment.

aceinetx:
> We need an update to the heuristics tests as shown in the failure and updating of the language ID suggested in the inline comment.

I've added heuristics tests and updated the language ID as you suggested. Limbo also didn't pass the heuristics tests the first time, so I strengthened it's heuristics

test/test_heuristics.rb
```diff
@@ -285,6 +285,13 @@ def test_asy_by_heuristics
     })
   end
 
+  def test_b_by_heuristics
+    assert_heuristics({
+      "B" => all_fixtures("B", "*.b"),
+      "Limbo" => all_fixtures("Limbo", "*.b")
```
lildude: This needs Brainfuck too.

aceinetx: > This needs Brainfuck too.

Thanks for pointing that out, I've added it right now

lildude (approved):

PR #8061, opened 2026-07-09
Security Patches

jacob-kraniak:
Many critical unpatched Dependency Vulnerabilities


lildude:
Thanks for this but we're don't need this.

We already have CodeQL (default config is good enough) and Dependabot configured and they already flag issues with the samples, but as the samples aren't actually used (we tokenise the files for the classifier), all such notifications are dismissed.

We also don't change samples unless they're blatantly wrong as the classifier needs to be trained on all forms of a language as it evolves.

Issue #8046, opened 2026-07-02
Add support for Civet language


AustinGrey:
## Language name
Civet

## URL of example repository
https://github.com/edemaine/texlish

## URL of syntax highlighting grammar
https://github.com/DanielXMoore/Civet/blob/df3ed3d6e36c2f031e90554df084ffc2f359c92e/lsp/tree-sitter/README.md

## Most popular extensions

- .civet

## Detected language

I think these are getting detected as javascript right now

hiSandog:
Because `.civet` files can resemble JavaScript, representative samples should exercise Civet-specific constructs rather than only translated-looking syntax. A generated or vendored example may also be useful if Civet commonly emits JavaScript beside its sources, so generated-code classification does not inflate language statistics. Does the linked tree-sitter repository expose a compatible TextMate grammar, or will a separate highlighting source be needed?

AustinGrey:
The repository linked does contain representative samples, as well as TemRevil's PR #8063 . I don't see how emitted JS or TS could inflate statistics. ".civet" files are immediately and obviously distinct from ".ts" or ".js" files. So there is no ambiguity. I haven't yet encountered a Civet repo that tries to emit JS by it's sources and also commits that. Generally the workflow is like typescript, where the results are built directly into a dist folder for distribution rather than transpiled beside for compilation.

> Does the linked tree-sitter repository expose a compatible TextMate grammar, or will a separate highlighting source be needed?

It does, which TemRevil's PR has made use of

PR #8063, opened 2026-07-10
Add Civet language support

TemRevil:
## Description

Adds support for [Civet](https://civet.dev), the TypeScript-superset language, per #8046 (labeled Add Language / Good First Issue). Includes the `languages.yml` entry with generated `language_id`, the official grammar as a submodule, three real-world samples, and the regenerated grammar list entry.

## Checklist:

      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.civet+civet
    - Sample source(s):
      - https://github.com/edemaine/texlish/blob/main/src/cli.civet (`cli.civet`)
      - https://github.com/DanielXMoore/Civet/blob/main/source/config.civet (`config.civet`)
      - https://github.com/DanielXMoore/Civet/blob/main/source/cache.civet (`cache.civet`)
    - Sample license(s): MIT (both source repositories)
  - [x] I have included a syntax highlighting grammar: https://github.com/DanielXMoore/Civet (MIT; TextMate grammar at `lsp/vscode/syntaxes/civet.json`, scope `source.civet`)
  - [x] I have added a color
    - Hex value: `#202c36`
    - Rationale: dominant color of the official Civet favicon/logo at civet.dev
    - Not needed: `.civet` is not associated with any other language in `languages.yml`.

Test results on this branch (Linux, `rake samples` regenerated): `test_language`, `test_blob` (10232 assertions) and `test_grammars` all pass with 0 failures.


lildude (changes requested):
When you ran `script/add-grammar`, the repo's license should have been cached into a yaml file. We need that file in this PR.

AustinGrey:
Using [this search](https://github.com/search?q=path%3A*.civet&type=code), it appears there are 1.9k civet files indexed so far. Which is close to the 2k threshold, but not yet there. The repos appear to be representative, coming from different authors, so that won't be a concern. Will have to keep an eye on the stats.

TemRevil:
@lildude good catch, thank you. I'd added the submodule by hand and missed the two steps `script/add-grammar` normally does for me. Pushed 58f9922 which sorts the `.gitmodules` entry into place and runs `licensed cache` for the Civet grammar, so `vendor/licenses/git_submodule/Civet.dep.yml` is now in the PR. Let me know if anything else is missing.

@AustinGrey appreciate you actually checking the numbers, that's the right way to look at it. Happy to sit on this until it clears 2k if that's the bar, will keep an eye on it too.

PR #8079, opened 2026-07-21
Add test validating language_id on new language additions

lildude:
Add a test that validates the `language_id` field is correctly generated for newly added languages.

## Description

New languages must have their `language_id` generated by `script/update-ids`, which derives the ID deterministically from the language name (`SHA-256(name) % (2**30 - 1)`). Contributors sometimes forget this step or hand-pick an ID, which is currently only caught during manual review.

This adds `test/test_language_id.rb`, which diffs `languages.yml` against the point the branch diverged from the default branch and asserts every **newly added** language uses the script-generated ID.

The check is deliberately scoped to new languages only:

- **Existing languages** predate the generator and keep their original (often sequential) IDs, so they are never re-validated.
- **Renames** intentionally reuse the original `language_id` so GitHub keeps tracking the language. An added entry whose ID already existed on the base revision is treated as a rename and skipped.

When a base revision can't be resolved (e.g. a shallow clone without the default branch available), the test skips rather than fails.

## Checklist:


Alhadis (approved):
Brilliant addition.

(I was under the impression the `language_id` was generated purely randomly (with a fixed number of digits), but yeah, a checksum of the human-readable language name makes more sense…)

PR #8096, opened 2026-07-31
Add JASS progamming language

Drake53:
Add JASS, the Warcraft III scripting language.

## Description

JASS is the original scripting language of Warcraft III.
It has been the World Editor's default output since 2002; Lua was added as an alternative in patch 1.31 (2019), but JASS remains what the editor emits unless a map opts into Lua, and it backs a large existing corpus of map scripts and hand-written libraries on GitHub.

Those `.j` files are currently either left unclassified or attributed to Jasmin or Objective-J, the two languages that share the `.j` extension.
This PR adds `JASS` with a syntax-highlighting grammar, four real-world samples, and a new `.j` heuristic that disambiguates all three languages.

### Scope: JASS, not vJASS

This adds **original JASS only**. vJASS (JassHelper) is usually described as a superset of JASS, but it is not strictly one, so folding the two together would be wrong.
The clearest example is the `debug` keyword: in JASS `debug` applies to the entire **statement** that follows it, so it carries through a multi-line construct such as an `if … endif` block, whereas in vJASS it applies only to the **line** it appears on.
The same source text can therefore mean different things in the two languages.
The grammar and the samples here are plain JASS; no vJASS, Zinc or Wurst constructs are included.

## Checklist:
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.j+endfunction — **14k files**
    - Sample source(s):
      - https://github.com/Drake53/War3Net/blob/master/tests/War3Net.TestTools.UnitTesting/TestData/MapFiles/TestGenerateUnitData/war3map.j
      - https://github.com/eGlint/Warcraft-III-Noise-Library/blob/master/JASS/Noise.j
      - https://github.com/eGlint/Warcraft-III-Noise-Library/blob/master/JASS/OctavePerlin.j
      - https://github.com/Son-Guhun/Titan-Land-Lands-of-Plenty/blob/master/release/trigger/41-Disable%20System%20-%20GDS/3-GDS%20Main.j
    - Sample license(s): MIT
  - [x] I have included a syntax highlighting grammar: https://github.com/Drake53/language-jass
  - [x] I have added a color
    - Hex value: `#FF0303`
    - Rationale: This is the color used by player 1 in Warcraft III. It's visually distinct from other languages that often share a repository with JASS (C#, lua, typescript).


lildude (approved):
LGTM. Thanks.

> [!IMPORTANT]

PR #8108, opened 2026-08-08
Fix .al heuristic: add profileextension, anchor to line start

SShadowS:
Two fixes to the `.al` heuristic, which disambiguates AL (Microsoft Dynamics 365 Business Central) from Perl's AutoLoader.

## Description

**1. `profileextension` is missing from the keyword list.**

`PROFILE` is present, but `\b` stops it matching inside `profileextension`, so a file whose only object is a profile extension falls through to the Perl fallback. Measured against 15358 `.al` files from Business Central application sources, this is the only AL object kind the heuristic misses; every other kind present in that corpus already matches.

**2. The heuristic matches its keywords anywhere in the file.**

`page` and `value` are among the most common variable names in Perl, so real AutoLoader chunks are classified as AL:

```perl
my ($page, $response, %headers) = $self->request(GET => $url);
my $value = $headers{'content-length'};
```

Of the 75 AutoLoader `.al` files that ship with Git for Windows' Perl, 8 are currently detected as AL. The trigger is a bare `$page` or `$value`, or the words "man page" in a comment.

The fix requires an object declaration to start a line, which is how AL sources are written. DotNet alias objects are a bare keyword on their own line, and `namespace`/`using` declarations are always semicolon terminated, so each gets its own pattern. Ruby anchors `^` to line start by default, so no flag is needed; the ALGOL rule directly below uses the same style.

`VALUE` is dropped from the list. It is an enum member declaration rather than an object kind, it is written `value(0; "Name")` with no space after the keyword so it cannot match the anchored pattern anyway, and it caused three of the eight Perl misdetections.

## Measurements

| Corpus | before | after |
| --- | --- | --- |
| 15358 `.al` files from Business Central application sources | 15358 (100%) | 15298 (99.609%) |
| 75 Perl AutoLoader `.al` files from Git for Windows | 67 of 75 (89.33%) | 75 of 75 (100%) |

The 60 AL files no longer detected have no object declaration at all: their entire contents are commented out behind a preprocessor directive, so there is nothing in them to match. Trading those for eight working Perl files seems right, since a file with no AL object in it is not recognisably AL.

## Testing

Both samples are picked up automatically by the existing `test_al_by_heuristics`, which globs `samples/AL/*.al` and `samples/Perl/*.al`. `BlankProfileExt.al` is detected as Perl before this change and AL after; `fetch_page.al` is detected as AL before and Perl after.

## Checklist:

    - Sample source(s):
      - `samples/AL/BlankProfileExt.al` and `samples/Perl/fetch_page.al` are original minimal samples written for this PR, not copied from a project, so there is no upstream URL or third party license to attribute.
      - The AL sample is modelled on profile extension objects in Microsoft's Business Central BaseApp (MIT licensed), for example `BlankExt.ProfileExt.al`.
      - The Perl sample is modelled on the AutoLoader output in Net::SSLeay, whose `do_httpx2.al`, `make_headers.al` and `ssl_read_until.al` are among the files currently misdetected.
    - Sample license(s): same license as Linguist, contributed as part of this PR.

## AI disclosure

This change was prepared with AI assistance (Claude). The measurements above were produced by running both the current and proposed patterns over the two corpora, and the sample classifications were verified against the patterns exactly as committed to `heuristics.yml`.


SShadowS:
Pushed a follow-up commit: the anchored patterns now tolerate a UTF-8 BOM.

While implementing the same disambiguation for cloc I found that anchoring introduced a blind spot I had missed here. AL tooling commonly writes files with a UTF-8 BOM, and a BOM is not whitespace, so an anchored pattern cannot match on line 1.

Measured against the same 15358 Business Central files: 1788 carry a BOM, and for 58 of them the only matching line is line 1. Those were being classified as Perl. So the measurement in the description above was too pessimistic, and my explanation of the misses was wrong for most of them.

Corrected numbers:

| | AL corpus (15358) | Perl corpus (75) |
| --- | --- | --- |
| current `master` | 15358 (100%) | 67 (89.33%) |
| this PR, before the BOM fix | 15298 (99.609%) | 75 (100%) |
| this PR now | 15356 (99.987%) | 75 (100%) |

The two files still not detected genuinely contain no object declaration: their entire contents are commented out behind a preprocessor directive.

One thing worth a maintainer's eye: I wrote the BOM as the byte escape `(?:\xEF\xBB\xBF)?` rather than `\uFEFF`, on the assumption that `blob.data` is raw bytes and a UTF-8 regex could raise an encoding incompatibility. I do not have Ruby on this machine, so I could not run `test_al_by_heuristics` to confirm. `samples/AL/BomProfile.al` is BOM-prefixed specifically so the existing test exercises it in CI; if the escape form is wrong, that test should fail and I will switch it.

lildude (changes requested):
> One thing worth a maintainer's eye: I wrote the BOM as the byte escape `(?:\xEF\xBB\xBF)?` rather than `\uFEFF`, on the assumption that `blob.data` is raw bytes and a UTF-8 regex could raise an encoding incompatibility. I do not have Ruby on this machine, so I could not run `test_al_by_heuristics` to confirm. `samples/AL/BomProfile.al` is BOM-prefixed specifically so the existing test exercises it in CI; if the escape form is wrong, that test should fail and I will switch it.

And the tests have failed 😁 . You can use a Codespace for free as detailed in the CONTRIBUTING.md file if you don't have a ruby environment.

SShadowS:
AL corpus 15356/15358, Perl 75/75. Installed Ruby this time rather than guessing.

lildude (approved):

PR #8157, opened 2026-08-31
Make Adblock Filter List heuristic regex portable

look:
Rewrite the `.txt` Adblock Filter List heuristic using syntax shared by Ruby, Go's `regexp` package, and Rust's `regex` crate. This allows Linguist ports using RE2-style, linear-time regex engines to compile and apply the heuristic.

The reusable version expression is inlined in place of Ruby's `\g<version>` subexpression call, possessive quantifiers and extended mode are removed, and version components use `[0-9]` so all three engines agree on ASCII digit semantics. This preserves canonical and combined headers while allowing ports to compile the pattern instead of skipping it.

The now-stale Ruby, Go, and Rust compatibility exceptions for this heuristic are removed.

Validated with:

- `bundle exec script/check-regex-compatibility`
- `bundle exec ruby -Itest test/test_heuristics.rb -n test_txt_by_heuristics`, covering all five Adblock samples
- Old/new Ruby regex comparison across 1,065 generated single and combined headers


look:
Closing for local review before publication.

lildude (approved):

PR #8159, opened 2026-09-01
Check regex portability across engines

look:
## Description

I am auditing the regexes for compatibility with GitHub's code search system, which uses the `regex` crate. We exclude certain regexes that use Ruby-only features. This test documents those regexes while introducing a barrier for adding more by checking for compatibility with the Go and Rust regex libraries.

This might seem unnecessary for a Ruby library, but regexes that don't use these features are often faster and it will help other projects that consume Linguist's data.

Prior examples of regex PRs to fix incompatabilities:

* https://github.com/github-linguist/linguist/pull/6897
* https://github.com/github-linguist/linguist/pull/7238
* https://github.com/github-linguist/linguist/pull/7240
* https://github.com/github-linguist/linguist/pull/7242
* https://github.com/github-linguist/linguist/pull/7259

## Checklist:

N/A, test only change.


lildude:
This is nice and has been on my list to do for a while. Any chance you could add ReDoS checking too? I think either checking in Go or Rust should be sufficient to cover all languages.

look:
@lildude great idea. I added a check for `linear_time?` on the regexes. There are a couple holdouts, but fortunately, they are the ones I'm already targeting for `re2`/`regex` compatibility.

lildude (approved):

look:
@lildude thanks for approving. I think you need to merge it, as I do not see the merge button.

PR #8182, opened 2026-09-09
Recognize M routine commands in .m heuristics

look:
## Description

Recognize M/MUMPS routines without standalone semicolon comments using dot-indented commands, `SET ^global` assignments, and `WRITE` newline controls. Require a formatting delimiter after `!` to distinguish abbreviated `WRITE` from expressions such as `w != value`. Existing `.m` rule ordering is unchanged.

Expand heuristic coverage to the existing M samples, recognizing seven that previously had no heuristic match: `arrays.m`, `fibonacci.m`, `forloop.m`, `helloworld.m`, `ifelse.m`, `indirectfunctions.m`, and `nesting.m`. Leave `mileage.m` to classification rather than matching generic `quit` expressions.

These forms also occur in [VPE routines](https://github.com/shabiel/VPE/blob/b1780afef7db73f8b696a36e0ec415031646d8a9/Routines/XVEMD.m#L40-L53) and [YottaDB routines](https://github.com/YottaDB/YDBTest/blob/bf8967c92b0ab3f24908555cd76363904f650826/com/fixtp.m#L14-L49). No sample files, language metadata, or grammars are added.

Linux heuristic, pedantic, strategy, and sample tests pass (1,861 tests, 7,489 assertions). Ruby/Go/Rust regex compatibility passes with only the existing allowlisted exceptions; full Linguist detection is unchanged across all 97 competing `.m` samples and fixtures.

## Checklist:

    - Sample source(s): N/A — reuses existing samples; no new sample code is copied.
    - Sample license(s): N/A — no new samples.

### DIFF
--- a/lib/linguist/heuristics.yml
+++ b/lib/linguist/heuristics.yml
@@ -561,7 +561,8 @@ disambiguations:
     # Dot-indented commands, global assignments, and WRITE newline controls.
     - '(?i)^[ \t]+(?:\.[ \t]+)+(?:d(?:o)?|f(?:or)?|i(?:f)?|n(?:ew)?|q(?:uit)?|s(?:et)?|w(?:rite)?)[ \t:]'
     - '(?i)^[ \t]+s(?:et)?[ \t]+\^[%a-z][a-z0-9]*(?:\(|=)'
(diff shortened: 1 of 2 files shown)


look:
Superseded by #8185, which now incorporates these M routine heuristic changes and their original commits alongside the MATLAB .m heuristics.

PR #8183, opened 2026-09-09
Recognize Apex .trigger files

look:
## Description

Recognize Apex `.trigger` files from trigger declarations and sharing-qualified class declarations without resorting to the classifier. Matching is case-insensitive and supports multiline declarations; unmatched files continue through existing detection, and Shell shebang precedence is unchanged.

Reuses all existing Apex/Shell `.trigger` samples and adds two test-only fixtures under their language directories for multiline syntax and Shell negative coverage. No classifier-training samples are added.

Syntax/usage evidence, not sources of the new fixture bytes: [Apex Recipes](https://github.com/trailheadapps/apex-recipes/blob/87c1c9b674de15870e546b862fa93fbd04a4e130/force-app/main/default/triggers/AccountTrigger.trigger), [NPSP](https://github.com/SalesforceFoundation/NPSP/blob/1e9e61908707e00faaef9f25982a85409d4b93e3/force-app/tdtm/triggers/TDTM_Lead.trigger), and [Alpine package triggers](https://github.com/alpinelinux/aports/blob/4242388adce973c9d66f1a2c8a36eda2bd6588f0/main/grub/grub.trigger). Further examples are discoverable through [non-fork usage search](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.trigger+%22trigger%22+%22on%22).

## Checklist:

    - Sample source(s):
      - `test/fixtures/Apex/audit.trigger`: AI-generated synthetic, source-informed regression snippet composed by GitHub Copilot for this PR after inspecting existing Apex samples and public syntax examples; not copied from an upstream file.
      - `test/fixtures/Shell/refresh.trigger`: AI-generated synthetic, source-informed negative regression snippet composed by GitHub Copilot for this PR after inspecting existing Shell samples and Alpine package triggers; not copied from an upstream file. The heuristic abstains, while full detection remains Shell.
      - Both additions are test-only fixtures, not classifier-training samples.
    - Sample license(s):
      - Both `test/fixtures/Apex/audit.trigger` and `test/fixtures/Shell/refresh.trigger` are contributed under [Linguist's MIT license](https://github.com/github-linguist/linguist/blob/30b32c7e3351d18dffe2698d913fa74b477b2bd0/LICENSE).


lildude (commented):
Why have you put the samples where you have rather than using the sample directories that already exist for these languages? "Generic" is going to cause confusion as people will start putting samples for extensions listed in `generic.yaml` there when they should be in the respective language directories.

We also need explicit mentioning of the licence of the samples in the PR template for all new extensions.

I'm only comment on this PR but the same applies to the others which are doing similar things.

look:
Thanks, I will fix the fixture location and license information in this and the other PRs. (I've got most of them draft because I want to make sure they are as streamlined as possible before requesting review. Directly fixing the mismatches in the code indexing pipeline results in..."creative"...regexes.)

test/test_heuristics.rb
```diff
@@ -1248,6 +1239,21 @@ def test_tpl_by_heuristics
     })
   end
 
+  def test_trigger_by_heuristics
```
lildude, line 1242: AI has over-engineered this test 😁 It should use the same approach used elsewhere in this file.

look: Agreed. I looked into disambiguating shell but left it as no heuristic.

lildude (approved):

PR #8002, opened 2026-06-04
Add Quarkdown language

iamgio:
## Description

This PR introduces the [Quarkdown](https://github.com/iamgio/quarkdown) language, an extension of GFM which includes support for Turing-complete function calls.

## Checklist:

- [ ] **I am adding a new extension to a language.**
  - [ ] The new extension is used in hundreds of repositories on GitHub.com
    - Search results for each extension:
      
      - https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.FOOBAR+KEYWORDS
  - [ ] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [x] **I am adding a new language.**
  - [x] The extension of the new language is used in hundreds of repositories on GitHub.com.
    - Search results for each extension:
      
      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.qd+.d
      - Note: search doesn't do it justice, there are many results cut off. Remove the keyword to find more results (although mixed up with some other language)
  - [x] I have included a real-world usage sample for all extensions added in this PR:
    - Sample source(s):
      - [example.qd](https://github.com/iamgio/linguist/blob/b5665abbfcb18e8a63d40c34600e6c1020931fa7/samples/Quarkdown/example.qd)
    - Sample license(s): can be under the MIT license that covers Linguist
  - [x] I have included a syntax highlighting grammar: https://github.com/quarkdown-labs/quarkdown-tm-grammar-lspless
      
  - [x] I have added a color
    - Hex value: `#376d6e`
    - Rationale: [quarkdown.com](https://quarkdown.com)'s accent color (the primary brand color, `#212d2d`, was too dark on GitHub's dark mode)
  - [ ] I have updated the heuristics to distinguish my language from others using the same extension.
    - No other language uses the `.qd` extension

- [ ] **I am fixing a misclassified language**
  - [ ] I have included a new sample for the misclassified language:
    - Sample source(s):
      - [URL to each sample source, if applicable]
    - Sample license(s):
  - [ ] I have included a change to the heuristics to distinguish my language from others using the same extension.

- [ ] **I am changing the source of a syntax highlighting grammar**
  - Old: [URL to grammar repo]
  - New: [URL to grammar repo]

- [ ] **I am updating a grammar submodule**
  

- [ ] **I am adding new or changing current functionality**
  
[...]

PR #8193, opened 2026-09-13
Add Quarkdown language

iamgio:
## Description

This PR introduces the [Quarkdown](https://github.com/iamgio/quarkdown) language, an extension of GFM which includes support for Turing-complete function calls.

This a follow up to #8002, which was closed due to not meeting adoption requirements.

## Checklist:

      -  https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.qd+.+%7B+%7D+NOT+user%3Aiamgio (already excluding myself)
          - Note that, as with my previous PR, this search doesn't do the language justice. Quarkdown doesn't come with a static set of keywords, which makes the result count far from reality.
    - Sample source(s):
      - [example.qd](https://github.com/iamgio/linguist/blob/b5665abbfcb18e8a63d40c34600e6c1020931fa7/samples/Quarkdown/example.qd)
    - Sample license(s): can be under the MIT license that covers Linguist
  - [x] I have included a syntax highlighting grammar: https://github.com/quarkdown-labs/quarkdown-tm-grammar-lspless
      
  - [x] I have added a color
    - Hex value: `#376d6e`
    - Rationale: [quarkdown.com](https://quarkdown.com)'s accent color (the primary brand color, `#212d2d`, was too dark on GitHub's dark mode)
    - No other language uses the `.qd` extension

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1097,6 +1097,9 @@
 [submodule "vendor/grammars/quakec-syntax"]
 	path = vendor/grammars/quakec-syntax
 	url = https://github.com/4LT/quakec-syntax.git
+[submodule "vendor/grammars/quarkdown-tm-grammar-lspless"]
+	path = vendor/grammars/quarkdown-tm-grammar-lspless
+	url = https://github.com/quarkdown-labs/quarkdown-tm-grammar-lspless.git
 [submodule "vendor/grammars/quint-grammars"]
 	path = vendor/grammars/quint-grammars
 	url = https://github.com/informalsystems/quint-grammars.git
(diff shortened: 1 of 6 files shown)


lildude:
This is still very premature. Many of the files returned in the search results are not Quarkdown. 

> Note that, as with my previous PR, this search doesn't do the language justice. Quarkdown doesn't come with a static set of keywords, which makes the result count far from reality.

That may be the case but they do have a common pattern of including a line that starts with a `.` and a word, eg `.docname`. You can use regex in the search query [like this](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.qd+%2F%5E%5C.%5Ba-z%5D%2B+%2F+NOT+user%3Aiamgio) which gives a much better and more realistic looking set of results.

PR #8200, opened 2026-09-16
Add Teyru

LangYa466:
## Description

Adds **Teyru**, a Java-shaped language whose compiler turns `.teyru` source
files into native binaries.

- Classes, interfaces, records, enums and annotation types; generics, text
  blocks, lambdas, pattern matching, `sealed` / `permits` / `non-sealed`,
  `when` guards, `_` unnamed variables.
- No semicolons anywhere: a `;` is `TY-SYN-0001`. A `for` header separates its
  three parts with colons, `for (init : cond : update)`.
- Native properties: `public int years { get { ... } set { ... } }`, with the
  `field` and `value` names inside the accessors.

| | |
|---|---|
| Language and compiler (Go) | https://github.com/teyru-lang/Teyru |
| End-to-end suite | https://github.com/teyru-lang/tests — 256 `.teyru` files: 220 programs, 23 diagnostics, 9 package fixtures, 3 module cases, 1 native case |
| Editor support and the grammar | https://github.com/teyru-lang/editors |
| Documentation | https://github.com/teyru-lang/docs |

The syntax-highlighting grammar here is the project's own:
`vscode/syntaxes/teyru.tmLanguage.json` in `teyru-lang/editors`, hand-written
for this language against the language as its own compiler accepts it. It is not
adapted from a Java grammar, and not derived from VS Code's bundled Java grammar
or any other third-party grammar. The same bytes ship in the project's JetBrains
plugin. `grammar-compiler add vendor/grammars/editors` compiles it and records
the one scope it declares, `source.teyru`.

## Checklist:

    - Search results for the extension:
      - https://github.com/search?type=code&q=NOT+is%3Afork+%22package+teyru%22
      - That search returns 49 files, **all of them under `teyru-lang/Teyru`**.
      - Not met, and I would rather say so than tick the box: `.teyru` exists in
        this project's repositories and, as far as I can find, nowhere else. I am
        opening this PR so the language and its grammar can be looked at rather
        than waiting for usage that does not exist yet. "Not yet" is a fine
        answer to this PR.
    - Sample source(s):
      - https://github.com/teyru-lang/Teyru/blob/main/tests/programs/t92_java25_sweep.teyru
      - https://github.com/teyru-lang/Teyru/blob/main/tests/programs/t07_properties.teyru
      - https://github.com/teyru-lang/Teyru/blob/main/lib/36_validation.teyru
    - Sample license(s): GPL-2.0-only with the Classpath exception — the
      project's `LICENSE` and `LICENSE-CLASSPATH-EXCEPTION-2.0`, in the same
      repositories. The three files are copied verbatim.
  - [x] I have included a syntax highlighting grammar: https://github.com/teyru-lang/editors
  - [x] I have added a color
    - Hex value: `#0e7c7b`
    - Rationale: picked by hand, not from official branding — the project has
      none, so there is nothing to match. A teal, chosen to be clearly
      distinguishable from Java's `#b07219` because the two languages look alike
      and would otherwise sit next to each other in a colour that says nothing.
      Happy to change it to whatever the maintainers prefer.
    - Not needed: no other language in `languages.yml` claims `.teyru`.

## Please look at the licence first

The grammar's repository is **GPL-2.0-only** (with the Classpath exception),
which is not one of the licences in `vendor/licenses/config.yml`, so
`bundle exec licensed status -c vendor/licenses/config.yml` will fail on
`vendor/grammars/editors`. I did not touch the licence: relicensing a file is
the project owner's decision, not mine. If a permissive licence on
`vscode/syntaxes/teyru.tmLanguage.json` alone is what it takes, say so and I will
ask for it; if a GPL grammar cannot be vendored at all, then this PR cannot be
merged as it stands and the answer is no.

## What I ran, and what I could not

`bundle install` does not complete in the environment I have:
`An error occurred while installing racc (1.8.1), and Bundler cannot continue.`
(a `nokogiri` dependency of `licensee`, itself a dependency of `licensed`), so
`bundle exec rake` did not run and I am not claiming it passes. Docker is not
available either, which is why the grammar compiler below is the native Go
build rather than `script/add-grammar`.

What did run, against this commit:

- **Linguist's own grammar compiler** (`tools/grammars`, built with the system
  Go and PCRE 8.45): `grammar-compiler add vendor/grammars/editors` →
  `OK! added grammar source 'vendor/grammars/editors'` / `new scope:
  source.teyru`, with no warning. Earlier revisions of the grammar did *not*
  compile: it used four variable-length lookbehinds, which Oniguruma allows and
  PCRE does not, and they are consuming rules now
  (https://github.com/teyru-lang/editors/commit/5a4dfad).
- **The assertions of `test_pedantic.rb`, `test_grammars.rb`,
  `test_language.rb` (the `tm_scope` rule) and `test_language_id.rb`**,
  re-implemented in plain Ruby — which needs no gems — because the bundle does
  not build: `languages.yml` and `grammars.yml` are sorted, no scope is listed
  twice, `.gitmodules` and `grammars.yml` agree and use an HTTPS URL, every
  `tm_scope` in `languages.yml` is a scope `grammars.yml` knows, the new
  `language_id` is the value `script/update-ids` generates for the name, and no
  two languages share an id.
- `script/sort-submodules` reports `.gitmodules` sorted.

Not run, so a maintainer should: `bundle exec rake samples`, which is what
writes `lib/linguist/samples.json` — the samples are in this PR, the generated
cache is not regenerated. `script/list-grammars` has the same problem, so the
line for Teyru in `vendor/README.md` is the line it generates rather than one it
wrote.

`language_id` is 614556994, the value `script/update-ids` derives from the
name — `test/test_language_id.rb` requires exactly that for a new entry. The id
is yours to reassign regardless.


Alhadis:
1. Read Linguist's [rules on adding new languages](https://github.com/github-linguist/linguist/blob/ac753dd778f3700debd0c43399d54511419844ce/CONTRIBUTING.md). We require new languages to be in-use by at least 200 unique user/repo combinations (not counting those from the language's own author/org); currently, there are only [57 search results](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A**%2F%3F*.teyru) for the `.teyru` file extension.
2. Fill out the pull-request template next time (it's a requirement).

